PR #1, opened 2020-07-30
per-stream record object store

darkdarkdragon:
  - add support for url-based object store
  - per-stream record object store in web hook

## -

-

**Screenshots (optional):**

**Checklist:**


packages/api/src/controllers/stream.js
```diff
@@ -322,6 +323,26 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
   res.end()
 })
 
+app.put('/:id/record', authMiddleware({}), async (req, res) => {
```
iameli, line 326: For API consistency, would kind of prefer this to be a `app.patch("/:id")` with a body of `{"record": true}`. Should validate that too or currently we'll allow any valid JSON value to be saved to `req.body.record`

packages/api/src/controllers/stream.js
```diff
@@ -322,6 +323,26 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
...
+  const { id } = req.params
+  const stream = await req.store.get(`stream/${id}`, false)
+  if (!stream || stream.deleted || !req.user.admin) {
```
iameli, line 329: Wait, won't this require that the user is an admin? `!req.user.admin` will always be false for regular users

packages/www/hooks/use-api.tsx
```diff
@@ -278,6 +278,22 @@ const makeContext = (state: ApiState, setState) => {
...
+      const [res, body] = await context.fetch(`/stream/${streamId}/record`, {
+        method: "PUT",
+        body: JSON.stringify({ record }),
+        headers: {
+          "content-type": "application/json"
+        }
+      });
+
+      if (res.status !== 204) {
+        return;
```
iameli, line 291:
```suggestion
        throw new Error(`http status ${res.status}`);
```

iameli (changes requested):
Really excellent work, just a few line-by-line things, small API adjustment, and this:

I don't think we should just pass `--recordObjectStore` in as an inline OS string. Very soon we're going to have multiple object stores for multiple regions and we're going to need a record of which stream is recorded where. My intention with `--recordObjectStore` was to have an admin create an OS for Backblaze, and then deploy the API servers with that ID: `--recordObjectSore=UUID`. I wasn't 100% sure what the end of that story was, but I think it's this:

When we get a webhook from go-livepeer, we can query the stream object. If `stream.recordStoreId` is undefined, we populate the stream with a recordStoreId and return that store's URL to the webhook. (The idea is that the API server that got the hook will know which OS is closest.) If `stream.recordStoreId` is already populated, we want to keep the stream's recordings in the same region, so we should just query and use that OS.

EDIT: And then we have a really easy answer to how we get the playback URLs: all we do is ask for `GET /stream/:id/recordUrl` and that will return a playback URL wi

darkdarkdragon:
@iameli fixed here: #7 
(created new PR because master branch was changed)

PR #23, opened 2020-08-10
initial api webhooks Docs

ya7ya:
Adds API Documentation for the webhooks

updates documentation on livpeer.com

## -

-

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/pages/docs/livepeer-api/webhooks.mdx
+++ b/packages/www/pages/docs/livepeer-api/webhooks.mdx
@@ -1,0 +1,94 @@
+import DocsLayout from "../../../components/DocsLayout";
+
+<DocsLayout>
+
+### How to Create a Webhook
+
+To create a webhook, We'll send a POST requst to `https://livepeer.com/api/webhook`.
+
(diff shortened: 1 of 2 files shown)


packages/www/pages/docs/livepeer-api/webhooks.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+<DocsLayout>
+
+## Wehbooks
+
+**`NOTE: This is still in development`**
+
+Livepeer offers a standard webhook to integrate with your platform's backend. Currently the following events are supported:
+
+- `streamStarted` : triggered when a stream is started.
```
ericxtang, line 11: We should explain what this event does - something like "this event will be triggered whenever a new stream is started associated with your account".

ericxtang:
Another question for the webhook - is there a way to list all the webhooks for a specific account?  (If I create one, and forget the ID, would it continue to be called forever?)

ya7ya:
> Another question for the webhook - is there a way to list all the webhooks for a specific account? (If I create one, and forget the ID, would it continue to be called forever?)

Yes, `GET /api/webhook` should return a list of webhooks defined by the user. I'll update the docs to reflect that

PR #25, opened 2020-08-10
add automated billing

### LINKED ISSUE #27: As a user, I can upgrade from the hobby plan to the pro plan so that I can take advantage of livepeer.com’s infinite usage and storage and streaming features.

### LINKED ISSUE #28: As a user, I can downgrade from the pro plan to the hobby plan if I no longer need livepeer.com’s services

### LINKED ISSUE #29: As a user, I can change my credit card so that I never have an expired or cancelled credit card on file.

### LINKED ISSUE #30: As a user, I can view my usage for the current billing cycle so that I have a good idea of what my total statement will be at the end of the cycle.

### LINKED ISSUE #31: As a user, I can view past invoices so that I have a record of how much I’ve paid over time.

### LINKED ISSUE #32: As a user, I can get in touch with the livepeer.com sales team if I’m interested in the Enterprise plan by filling out a form on the contact page.

### LINKED ISSUE #33: As a user, I can use livepeer.com’s hobby plan for up to 1000 ingested minutes so that I can try livepeer.com before committing to the pro plan.

### LINKED ISSUE #34: As a user, I can download my past invoices in pdf format so that I can keep records of or share my invoices.

### LINKED ISSUE #35: As a user, I can receive an email with my invoice at the end of each billing cycle so that I’m reminded of an upcoming payment.

### LINKED ISSUE #36: As a user, I'm prompted to upgrade to the pro plan from the navigation after signing up so that I'm always aware of the option to upgrade.

adamsoffer:
Adds self serve billing

- Introduces a pricing page
- Adds `usage` and `plans` views to the dashboard
- Stripe integration

closes #27, closes #28, closes #29, closes #30, closes #31, closes #32, closes #33, closes #34, closes #35, and closes #36


adamsoffer:
@iameli any idea how to get these preview urls working with the livepeer.com api? The preview url works but I can't login to the dashboard.

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
+import Router from 'express/lib/router'
+import { db } from '../store'
+const app = Router()
+const stripe = require('stripe')(process.env.LP_STRIPE_SECRET_KEY)
```
iameli: kinda like `import Stripe from 'stripe'` and `const stripe = new Stripe( ... )` here for consistency if that works

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+// Webhook handler for asynchronous events.
+app.post('/webhook', async (req, res) => {
+  const sig = req.headers['stripe-signature']
+
+  let event
+
+  try {
+    event = stripe.webhooks.constructEvent(req.body, sig, endpointSecret)
+  } catch (err) {
+    res.status(400).send(`Webhook Error: ${err.message}`)
```
iameli:
```suggestion
    return res.status(400).send(`Webhook Error: ${err.message}`)
```

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+  if (event.type === 'invoice.created') {
+    let invoice = event.data.object
+
+    if (invoice.status === 'draft') {
```
iameli: What if the invoice status is something other than draft? Is it correct for us to be just returning 200 in that case? If so, I think this is probably clearer as 

```javascript
if (invoice.status !== 'draft') {
  // we don't need to do anything
  return res.sendStatus(200)
}
```

And then we can remove the big `if` block

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+      const { data: userIds } = await req.store.query({
+        kind: 'user',
+        query: { stripeCustomerId: invoice.customer },
+      })
+      const user = await req.store.get(`user/${userIds[0]}`, false)
```
iameli: Cleaner and safer b/c no replica:
```suggestion
      const [user] = db.user.find({ stripeCustomerId: invoice.customer }, {useReplica: false})
```

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+        {
+          useReplica: false,
+        },
+      )
+      // Invoice items based on usage
+      await Promise.all(
+        products[user.stripeProductId].usage.map(async (product) => {
+          if (product.name === 'Transcoding') {
+            let quantity = Math.round(
+              (usageRes.sourceSegmentsDuration / 60).toFixed(2),
```
iameli: `toFixed` doesn't seem necessary if we're passing it to `Math.round` anyway

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+
+async function sleep(millis) {
+  return new Promise((resolve) => setTimeout(resolve, millis))
+}
+
+// Migrate existing users to stripe
+app.post('/migrate-users', async (req, res) => {
+  if (process.env.LP_STRIPE_SECRET_KEY != req.body.stripeSecretKey) {
+    res.status(403)
+    res.json({ errors: ['unauthorized'] })
```
iameli:
```suggestion
    return res.json({ errors: ['unauthorized'] })
```

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  const { data: userIds } = await req.store.query({
+    kind: 'user',
+    query: {},
+  })
```
iameli:
```suggestion
  const [users] = await db.user.find({}, {limit: 9999999999, useReplica: false})
```

iameli: Unless loading all the prod users at once uses too much RAM. In which case we'd need pagination anyway, because `req.store.query` loads all the records at once behind the scenes.

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+
+  for (let index = 0; index < userIds.length; index++) {
+    const id = userIds[index]
+    let user = await req.store.get(`user/${id}`, false)
+
+    const { data } = await stripe.customers.list({
+      email: user.email,
+    })
+
+    if (!data.length) {
```
iameli: Nervous that this might be true in some kind of error condition. Perhaps `if (data.length === 0)` at least?

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+        lookup_keys: products['prod_0'].lookupKeys,
+      })
+
+      // Subscribe the user to the free plan
+      const subscription = await stripe.subscriptions.create({
+        cancel_at_period_end: false,
+        customer: customer.id,
+        items: items.data.map((item) => ({ price: item.id })),
+      })
+      await sleep(500)
```
iameli: Perhaps should be at the end of the loop just so we don't `req.store.replace` a user that's 500ms out of date?

packages/api/src/controllers/stripe.js
```diff
@@ -0,0 +1,141 @@
...
+        stripeCustomerSubscriptionId: subscription.id,
+        stripeCustomerPaymentMethodId: null,
+      }
+      await req.store.replace(user)
+    }
+  }
+  res.json(userIds)
+})
+
+app.get('/retrieve-subscription', async (req, res) => {
```
iameli: No auth needed on these two? I get that the `stripeCustomerSubscriptionId` is probably private but still... is this endpoint hit by stripe, UI, API users, etc?

packages/api/src/controllers/user.js
```diff
@@ -83,6 +87,10 @@ app.post('/', validatePost('user'), async (req, res) => {
       admin: false,
       emailValidToken: emailValidToken,
       emailValid: validUser,
+      ...(firstName && { firstName }),
```
iameli: If `firstName` isn't present this will do `...undefined`, which surely is an error? Seems like we could just put these in unconditionally and if they're undefined we'll get the same results?

packages/api/src/controllers/user.js
```diff
@@ -403,4 +434,247 @@ app.post(
...
 )
 
+app.post(
+  '/create-customer',
+  validatePost('create-customer'),
+  async (req, res) => {
+    const { data: userIds } = await req.store.query({
+      kind: 'user',
+      query: { email: req.body.email },
+    })
```
iameli:
```suggestion
    const [users] = await db.user.find({ email: req.body.email }, {useReplica: false});
```

packages/api/src/controllers/user.js
```diff
@@ -403,4 +434,247 @@ app.post(
...
+      res.status(201)
+      res.json(customer)
+    } else {
+      res.status(403)
+      res.json({ errors: ['customer not created'] })
+    }
+  },
+)
+
+app.post(
```
iameli, line 462: Seems like we'd want authMiddleware for a specific user or admin here

packages/www/components/AdminUsageTable/index.tsx
```diff
@@ -115,10 +103,7 @@ export default ({ id }: { id: string }) => {
           <Box>Source segments</Box>
           <Box>Transcoded segments</Box>
         </TableRow>
-        <TableRow
-          key="just one row for now"
-          variant={TableRowVariant.Normal}
-        >
+        <TableRow key="just one row for now" variant={TableRowVariant.Normal}>
```
iameli, line 106: lmao i love it

packages/www/components/ChangePaymentForm/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+  const onSubmit = async (data, e) => {
+    e.preventDefault();
+
+    // Abort if form isn't valid
+    if (!e.target.reportValidity()) return;
+
+    setStatus("processing");
+
+    const cardElement = elements!.getElement(CardElement);
```
iameli, line 68: whoa cool `!.` is new to me

packages/www/components/HomeHero/index.tsx
```diff
@@ -200,9 +200,10 @@ const HomeHero = () => {
             <Text sx={{ fontSize: "18px", lineHeight: 1.7 }}>
               By leveraging decentralized infrastructure, Livepeer.com delivers
               a video platform that focuses on what user-generated content (UGC)
-              applications need. Affordable, scalable and reliable, the Livepeer
-              platform combines high-quality transcoding with powerful APIs and
-              features that are essential for a seamless creator experience.
+              applications need. Affordable, scalable and reliable, the
```
iameli: who do i need to fight to get an oxford comma added here (ง •̀_•́)ง

packages/www/components/Login/index.tsx
```diff
@@ -36,23 +43,95 @@ const Login = ({
           }
           const [hashedPassword] = await hash(password, FRONTEND_SALT);
           // hash password, then
-          onSubmit({ email, password: hashedPassword });
+          onSubmit({
+            email,
+            password: hashedPassword,
+            ...(firstName && { firstName }),
```
iameli: does this have the desired results if firstName is empty?

packages/api/src/schema/schema.yaml
```diff
@@ -425,6 +507,36 @@ components:
...
+          example: Livepeer
+        phone:
+          type: string
+          example: 2034212122
+        stripeProductId:
+          type: string
+          example: prod_0
+        stripeCustomerId:
+          type: string
+          example: cus_Jv6KvgT0DCH8HU
```
iameli, line 528: This should presumably have `unique: true`, if I understand the flow correctly things would break pretty hard if two users have the same stripe customer id

iameli: Not that it should ever happen... but it doesn't hurt to have the DB to enforce that too. Bonus: that unique index will make lookups by customer ID in faster than `O(every single user)` time

iameli (changes requested):
Reviewed the whole backend. I'll take another pass over the frontend later after I've had a chance to click around .monster.

Great work! I'm really happy this webhook pattern ended up working for us; the integration happens in a lot fewer lines of code than I thought.

Aside from the line-by-line stuff, here are the backend changes that would make me comfortable getting this to prod:

* `req.store` is deprecated. I'm not 100% opposed to its use in new code if `db.user` doesn't work but the payment code seems like exactly the kind of place where we'd want to make sure to use the new stuff. It's type-safe, more explicit, and:
* With the above, we can add `{useReplica: false}` everywhere in the payment workflow. I can't think of any _specific_ problems with using the read replicas but let's just go ahead and never ever find out.
* I really really think that we need to do https://github.com/livepeer/livepeer-com/issues/94 and change all instances of `db.user.replace` with `db.user.update` before this goes live. In something like the `/user/create-subscription` endpoint, the `req.store.replace` call could plausibly happen a few seconds after the `req.store.get` call if the Stripe API is slow or far away. In that case, any other changes to that user object will be clobbered. If someone's got two tabs open, and they switch subscriptions and switch payment information at the same time, one of those two isn't getting saved into the user record 😬

But yeah. Let's do that and try the migration against a clone of the prod database (I can walk you through this) and we should be good to go!

packages/api/src/controllers/stripe.js
adamsoffer: @iameli nice. Does `db.user.find()` return users inside an array?

adamsoffer: In other words do I need to access the user like so:
```
const [users] = db.user.find({ stripeCustomerId: invoice.customer }, {useReplica: false})
const user = user[0]
```

packages/api/src/controllers/stripe.js
adamsoffer: @iameli yeah good call - should I move these endpoints to the `user.js` controller and change to a post request?

adamsoffer: (It's hit by the UI in the user's `/usage` view)

packages/api/src/controllers/user.js
adamsoffer: This weird syntax will conditionally add the key value pair to the object. I used this syntax because I wasn't sure if the api would return `firstName: undefined` if I set it and it was undefined. If that's the the case and the api excludes undefined values then I can remove the conditional.

packages/api/src/controllers/user.js
adamsoffer: Does that middleware already exist? What would that look like?

packages/api/src/schema/schema.yaml
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/stripe.js
adamsoffer: Changed these two endpoints to `post` and moved them into the `user.js` controller in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/user.js
adamsoffer: Confirmed we can get rid of this weird syntax. Removed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/api/src/controllers/user.js
adamsoffer: Fixed in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/www/components/HomeHero/index.tsx
adamsoffer: I gotchu. Oxford comma added in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/www/components/Login/index.tsx
adamsoffer: Removed ugly conditional syntax in [04ff25](https://github.com/livepeer/livepeer-com/pull/25/commits/04ff2537a397629316e3ccef0ffba6a97665ad0a)

packages/www/hooks/use-api.tsx
```diff
@@ -99,9 +99,10 @@ const makeContext = (state: ApiState, setState) => {
         headers.set("authorization", `JWT ${state.token}`);
       }
       const endpoint =
+        window.location.hostname.includes("livepeer.monster") ||
         window.location.hostname.includes("livepeerorg.vercel.app") ||
         window.location.hostname.includes("livepeerorg.now.sh")
-          ? `https://livepeer.monster/api${url}`
```
iameli, line 104: This should get flipped back and we should make sure it works before shipping

adamsoffer: Done

PR #38, opened 2020-08-14
www: use ingest.base instead of ingest.playback

iameli:
fixes /hls/hls in a different way that doesn't break API compatibility


ya7ya (approved):
Looks good, I haven't tested it though (I can't get yarn to work properly right now :angry: )

PR #43, opened 2020-08-20
postgres refactor

iameli:
Pertinent changes:

1. Support for all stores except Postgres has been removed.
1. Instead of the hand-rolled indexing system I wrote before, we're now using proper Postgres indices.
1. `req.store` is deprecated. Use the `db` singleton instead, from `import { db } from '../store'`. `req.store` still exists, but it's just implemented using a compatibility layer on top of `db[table]`, so they shouldn't be used for anything new.
1. Standard CRUD operations are on the `db[table]` Table classes. We have `db.stream.get(id)`, `db.stream.create(document)`, `db.stream.replace(document)`, and `db.stream.delete(id)`.
1. There's also `db.stream.find()`, which can be invoked in two ways:
a.  Called with some fields, such as `await db.stream.find({userId: "ABC123"})`. This will work with an arbitrary number of fields.
b. Called with an array of SQL WHERE clauses using [sql-template-strings](https://github.com/felixfbecker/node-sql-template-strings), see as this example from the `getWebhook` helper:
```javascript
import { db } from '../store'
import sql from 'sql-template-strings'

[snip]

  const query = [sql`data->>'userId' = ${userId}`]
  if (event) {
    query.push(sql`data->>'event' = ${event}`)
  }
  if (!includeDeleted) {
    query.push(sql`data->>'deleted' IS NULL`)
  }
  const [webhooks, nextCursor] = await db.webhook.find(query, { limit, cursor })

  return { data: webhooks, cursor: nextCursor }
```
6. I'm also perfectly fine with adding more SQL directly to `Table` or subclassing it into individual model classes if we want to do more sophisticated queries per your feedback @darkdarkdragon. We can also kind of "just write SQL: using the `db.table.find([SQLStatement])` syntax that I think will work for many read queries. But for updating multiple documents and whatnot we can do it for sure.

There's a lot of tech debt to address on this; it doesn't need to block deploy but it should be handled within a reasonable time frame.

1. Lots and lots of code to clean up for Cloudflare KV, Firestore, LevelDB, test harness, etc...
2. Dev environment currently expects Postgres to be running locally; `cd packages/api && yarn run postgres:start` starts one
3. There should really be a `db[table].update({})` for partial updates, this will reduce the risk of different API servers clobbering the entire document with `replace()`
4. Now that we're Postgres-only we should lean into it more. Probably we shouldn't have all JSONB data everywhere (though it works surprisingly well — all queries and indexes I've tried have worked fine so far)
a. I think "deleted" should be a new first-class field on the documents next to `id` instead of being at `data->>'deleted'`. Right now all the deleted queries check for `data->>'deleted' IS NOT NULL` which would get messed up with a `{"deleted": false}` value.
b. We should probably have a way to do database migrations. I find the traditional way of numbering them to be kind of a pain, maybe we can think of something a little better.
5. No read replica support quite yet. We'll be operating out of a Postgres server in Europe. Users elsewhere will still see a big performance improvement, but we should add read replica support.
6. The worker is no longer capable of interacting with the database and should be split out into a separate package. On the other hand, it would probably be useful to have the API server still be capable of geolocation. So... I'm thinking that the worker compiles into a worker and also into something that can be included as an express middleware. This is pretty close to what's happening already.
7. Jest's `-i` in-band flag is presently required — parallel tests are broken because of the `db` singleton. Probably `db` should not be a singleton; we should instead turn our controllers into factory functions (like all other express controllers) and dependency inject `db`


ya7ya:
@iameli Nice work dude, Quick question, Is there some db initialization required before running the tests beside `yarn postgres:start` ?

iameli:
@ya7ya There shouldn't be — `ensureDatabase` and `ensureTable` handle the necessary creation. Let me know if you're seeing problems though, that could be an issue

packages/api/src/controllers/stream.js
```diff
@@ -46,27 +44,23 @@ const hackMistSettings = (req, profiles) => {
...
-  let output = resp.data
+  const [output, newCursor] = await db.stream.find(query, { cursor, limit })
+
   res.status(200)
 
-  if (output.length > 0) {
-    res.links({ next: makeNextHREF(req, resp.cursor) })
-  } // CF doesn't know what this means
-  output = output.map((o) => o[Object.keys(o)[0]])
+  if (newCursor.length > 0) {
```
darkdarkdragon: Error?

packages/api/src/controllers/stream.js
```diff
@@ -277,8 +270,7 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
 
   let objectStoreId
   if (req.body.objectStoreId) {
-    await req.store.get(`objectstores/${req.user.id}/${req.body.objectStoreId}`)
-    objectStoreId = req.body.objectStoreId
+    await req.store.get(`object-store/${req.body.objectStoreId}`)
```
darkdarkdragon: Result of this call is not used anywhere.

packages/api/src/controllers/stream.js
```diff
@@ -329,10 +321,9 @@ app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
 
     const { data: webhooksList } = await getWebhooks(
       req.store,
-      stream.userId,
+      req.user.id,
```
darkdarkdragon: this is bug in webhooks that was fixed by Yahya, need to revert this line

packages/api/src/controllers/webhook.js
```diff
@@ -19,17 +20,23 @@ app.get('/', authMiddleware({}), async (req, res) => {
       limit,
       filter,
     })
-    let output = resp.data
+    const [output, newCursor] = db.stream.find([`data->>'deleted' IS NULL`])
```
darkdarkdragon: This makes call six lines up unnecessary.

darkdarkdragon (changes requested):
Couple of changes, though not finished review yet.

packages/api/src/store/table.ts
```diff
@@ -0,0 +1,183 @@
...
+  [key: string]: any
+}
+
+interface FindOptions {
+  cursor?: string
+  limit?: number
+}
+
+export default class Table<T extends DBObject> {
+  db: DB
```
darkdarkdragon, line 24: `DB` here refers to the instance of DB class, not to a class itself

packages/api/src/store/table.ts
```diff
@@ -0,0 +1,183 @@
...
+    this.name = schema.table
+  }
+
+  // get a single document by id
+  async get(id: string): Promise<T> {
+    if (!id) {
+      throw new Error('missing id')
+    }
+    const res = await this.db.query(
+      sql`SELECT data FROM `.append(this.name).append(sql` WHERE id=${id}`),
```
darkdarkdragon:
```suggestion
      sql`SELECT data FROM `.append(this.name).append(sql` WHERE id=${id}`.setName(`${this.name}_by_id`)),
```
to use Postres prepared statement

darkdarkdragon (changes requested):
Two more small comments, plus made PR with some fixes https://github.com/livepeer/livepeer-com/pull/50

Otherwise looks good!

packages/api/src/controllers/stream.js
iameli: oh yeah. hm.

packages/api/src/store/table.ts
iameli: fixed by tweaking the import

PR #46, opened 2020-08-23
add api reference to docs

### LINKED ISSUE #40: As a user, I can view the API reference as part of the docs so I know what APIs I have available

adamsoffer:
This PR adds an API reference to the docs using generated markdown from the openapi spec.

Note: 
I noticed there are some endpoints that are part of the open api spec but that should be excluded from the public api like `/password-reset` and `/made-admin`. We'll need to figure out how to get `swagger-markdown` to ignore those.

Closes #40 

Screen Shot 2020-08-23 at 4 31 45 PM


adamsoffer:
@wohlner I've removed the non-public facing api endpoints. Good to merge?

Issue #48, opened 2020-08-24
api: implement postgres read replica support


iameli:
1. New flag, adjacent to `--postgres-url`: `--postgres-replica-url`. It can be pointed at a read replica of the primary database.
2. If set, initialize a second `Pool` object, `this.replicaPool`, alongside `this.pool` in `db.ts`.
3. In `db.ts`, add a new function adjacent to `query()` called `replicaQuery()` that can be called in exactly the same way. If `this.readPool` is unset, it does exactly the same thing as `query()`. If it is set, it performs the request against `this.replicaPool` instead of `this.pool`.
4. Update the `find()` and `get()` functions in `table.ts` with a new option `useReplica` that defaults to `true`. (`get()` will need to add a second parameter for options entirely.) This option determines whether they call `this.db.query()` or `this.db.replicaQuery()`. The idea here is that we want to hit the read replica almost all of the time but if you're in a situation where you absolutely must have the most up-to-date data you can set `{useReplica: false}` and sacrifice latency for correctness.

PR #60, opened 2020-08-27
Read Replica Pool

ya7ya:

implements #48 

## -

-

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/store/db.ts
+++ b/packages/api/src/store/db.ts
@@ -38,7 +38,7 @@ export class DB {
     // constructor logic has moved to start({}).
   }
 
-  async start({ postgresUrl, replicaUrl }) {
+  async start({ postgresUrl, postgresReplicaUrl }) {
     this.postgresUrl = postgresUrl
     if (!postgresUrl) {
       throw new Error('no postgres url provided')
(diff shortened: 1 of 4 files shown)


packages/api/src/store/db.ts
```diff
@@ -51,6 +53,14 @@ export class DB {
       connectionTimeoutMillis: CONNECT_TIMEOUT,
       connectionString: postgresUrl,
     })
+
+    if (replicaUrl) {
+      this.replicaPool = new Pool({
+        connectionTimeoutMillis: CONNECT_TIMEOUT,
+        connectionString: replicaUrl,
+      })
```
iameli, line 62: Let's do a `await this.replicaQuery('SELECT NOW()')` on the replica too

packages/api/src/store/table.ts
```diff
@@ -84,8 +99,8 @@ export default class Table<T extends DBObject> {
 
     q.append(' ORDER BY id ASC')
     q.append(sql` LIMIT ${limit}`)
-
-    const res = await this.db.query(q)
+    
+    const res = await this.db.replicaQuery(q)
```
iameli: Let's add a `useReplica` here too in case we want to query on the freshest data

iameli (changes requested):
minor changes, looks good

PR #67, opened 2020-09-03
gradient and video fixes

### LINKED ISSUE #70: serve one of three hero videos at random on page load
Here are the three different videos we'd like to feature on the home page hero. The idea is to serve a different one at random on page load.

[hero_videos.zip](https://github.com/livepeer/livepeer-com/files/5177300/hero_videos.zip)

julianbenegas:
This PR:
- Adds video swapper component to serve more than one video, one after the other.
- Adds a couple of videos, and serve them in random order each time. closes #70 
- Removes `display: flex` from `<Layout />`
- Makes hero input and CTA go full width on mobile
- Makes views number a random num between 100 and 600


julianbenegas:
@adamsoffer regarding #70 , I've improved the video component to be able to handle more than one vid, and serve them one after the other. Is this preferable to just looping over one video? They are served at random, as well

adamsoffer:
Nice! This looks awesome. I'll share with Mark and get his thoughts. One thing...each video should have a different view count so can we change the view count when the video changes?

adamsoffer:
Seems like the first video is always the singer. Is that due to vercel's cache? Possible to randomize the first video?

julianbenegas:
I see. Super weird, cause locally they shuffle correctly. Any ideas on how to solve this @adamsoffer ?

adamsoffer:
hm I'm able to reproduce locally when running in prod mode (`yarn build && yarn start`). It displays a random video after a rebuild but not on subsequent page refreshes.

adamsoffer:
Adding `revalidate: 1` seems to have done the trick 👍

julianbenegas:
Nice one!!

PR #69, opened 2020-09-04
add mailchimp to the footer

julianbenegas:


adamsoffer (changes requested):
Looks and works great thanks! Small request - can we just add a right arrow to the form input? Something along these lines? 
image

adamsoffer (approved):
Perfecto 💯

PR #89, opened 2020-09-14
enable cors

adamsoffer:
This PR enables CORs and teaches the frontend to use the `livepeer.monster/api` endpoint when on a vercel preview url.

### DIFF
--- a/packages/www/hooks/use-api.tsx
+++ b/packages/www/hooks/use-api.tsx
@@ -6,7 +6,7 @@ import {
   Error as ApiError,
   ApiToken,
   Stream,
-  Webhook,
+  Webhook
 } from "@livepeer.com/api";
 import qs from "qs";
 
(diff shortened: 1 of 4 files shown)


packages/www/hooks/use-api.tsx
```diff
@@ -76,9 +76,15 @@ const makeContext = (state: ApiState, setState) => {
       if (state.token && !headers.has("authorization")) {
         headers.set("authorization", `JWT ${state.token}`);
       }
-      const res = await fetch(`/api${url}`, {
+      const endpoint =
+        window.location.hostname !== "localhost" &&
```
iameli: Can we maybe do this the other way and check `livepeerorg.vercel.app` and `livepeerorg.now.sh`, using livepeer.monster if they're present? Otherwise we'll break deployment to all other domains

iameli (changes requested):
One change, looks good otherwise

packages/www/hooks/use-api.tsx
adamsoffer: Yeah good call. All set!

PR #90, opened 2020-09-14
Usage API endpoint and admin page

### LINKED ISSUE #78: sourceSecondsDuration endpoint
For any given account, we need an endpoint to query the `sourceSecondsDuration` for all streams created thus far in an active billing cycle and at the end of the billing cycle.

`sourceSecondsDuration` for the active billing cycle will be displayed on the usage tracking page in the livepeer.com admin and must be updated at least daily.  `sourceSecondsDuration` will also be used at the end of the billing cycle for billing via our Stripe integration. To get an accurate count, we will need to query for streams with a parent id for a given account.

The billing cycle is not the first of the calendar month to the end of the calendar month. The billing cycle should be able to be set arbitrarily.

darkdarkdragon:
/user/usage endpoint - returns usage data for the user in the specified time interval

## -

-

Fixes #78 

**Screenshots (optional):**
image

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/user.js
+++ b/packages/api/src/controllers/user.js
@@ -8,8 +8,26 @@ import validator from 'email-validator'
 import { makeNextHREF, sendgridEmail, trackAction } from './helpers'
 import hash from '../hash'
 import qs from 'qs'
+import { db } from '../store'
 
 const app = Router()
+
+app.get('/usage', authMiddleware({}), async (req, res) => {
+  let { userId, fromTime, toTime } = req.query
+  if (!fromTime || !toTime) {
+    res.status(400)
+    return res.json({ errors: ['should specify time range'] })
+  }
+  if (!userId || (req.user.admin !== true && req.user.id !== userId)) {
+    userId = req.user.id
+  }
+
+  const usageRes = await db.stream.usage(userId, fromTime, toTime, {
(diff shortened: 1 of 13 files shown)


wohlner:
The Usage page UI is helpful for reviewing what is available with the endpoint, but @adamsoffer will create a differently designed one based on the user's billing cycle dates. @adamsoffer your call if you want to keep the UI in the PR. 

This is super helpful, @darkdarkdragon, thanks!

adamsoffer:
Thanks @darkdarkdragon !

@wohlner I see the UI introduced in this PR is for the admin page; it doesn't interfere with the usage page I already implemented which is user facing and would be a nice internal tool for us to look up a user's usage for a given time period so I'm inclined to keep it.

adamsoffer (changes requested):
@darkdarkdragon good to merge once we fix this type error causing the build fail.

```
Type error: Argument of type 'Error \| User[]' is not assignable to parameter of type 'SetStateAction<any[]>'.
Type 'Error' is not assignable to type 'SetStateAction<any[]>'.
Type 'Error' is missing the following properties from type 'any[]': length, pop, push, concat, and 28 more.
```

darkdarkdragon:
@adamsoffer fixed

PR #93, opened 2020-09-15
Admin tool for getting various stream info

darkdarkdragon:
Adds 'tools' tab to admin's page.
Allows to enter playback ID, stream key or manfiest ID and will shows various information about stream

**Screenshots (optional):**
image

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -475,6 +475,49 @@ app.delete('/:id', authMiddleware({}), async (req, res) => {
   res.end()
 })
 
+app.get('/:id/info', authMiddleware({ anyAdmin: true }), async (req, res) => {
+  let { id } = req.params
+  let stream = await db.stream.getByStreamKey(id, { useReplica: true })
+  let session,
+    isPlaybackid = false,
+    isStreamKey = !!stream,
+    isSession = false
+  if (!stream) {
+    stream = await db.stream.getByPlaybackId(id, { useReplica: true })
+    isPlaybackid = !!stream
+  }
+  if (!stream) {
+    stream = await db.stream.get(id, { useReplica: true })
+  }
+  if (stream && stream.parentId) {
+    session = stream
+    isSession = true
+    stream = await db.stream.get(stream.parentId, { useReplica: true })
+  }
+  if (!stream) {
+    res.staus(404)
+    return res.json({
+      errors: ['not found'],
+    })
+  }
+  if (!session) {
+    // find last session
+    session = await db.stream.getLastSession(stream.id, { useReplica: true })
+  }
+  const user = await req.store.get(`user/${stream.userId}`)
+  const resp = {
+    stream,
+    session,
+    isPlaybackid,
+    isSession,
+    isStreamKey,
+    user
+  }
+
+  res.status(200)
+  res.json(resp)
+})
+
 app.post('/hook', async (req, res) => {
   if (!req.body || !req.body.url) {
     res.status(422)
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -480,22 +480,22 @@ app.delete('/:id', authMiddleware({}), async (req, res) => {
 
 app.get('/:id/info', authMiddleware({ anyAdmin: true }), async (req, res) => {
   let { id } = req.params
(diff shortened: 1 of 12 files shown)


darkdarkdragon:
@iameli @ya7ya 
I've also added 'Show active streams only' checkbox on admin's streams page:
image

wohlner:
First, this is great and super useful! One suggestion for how to make this workflow even better either in this PR or a future iteration.

It would be great in the tools tab to see a list of all active streams grouped by user with the playback ID, stream key or manfiest ID also in the table. And then when you click into a stream, you get all the info you have in your first screenshot.

That way instead of co-opting the streams tab that users use, we can have a separate admin tools tab. Also, then the person debugging wouldn't have to click into an active stream and copy and past an ID into the text field of new tool you created; it could be one step.

The tool with the text field shouldn't go away though, because it will still be helpful for non-active streams.

packages/api/src/controllers/stream.js
```diff
@@ -475,6 +478,49 @@ app.delete('/:id', authMiddleware({}), async (req, res) => {
   res.end()
 })
 
+app.get('/:id/info', authMiddleware({ anyAdmin: true }), async (req, res) => {
+  let { id } = req.params
+  let stream = await db.stream.getByStreamKey(id, { useReplica: true })
```
iameli: `{useReplica: true}` is the default; it's there so you can do an explicit `{useReplica: false}`

packages/www/components/AdminTools/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+      .catch((err) => {
+        console.log(err);
+        setMessage(`${err}`);
+      });
+  }, [id]);
+  useEffect(() => {
+    getUsers(10000)
+      .then((users) => {
+        if (Array.isArray(users)) {
+          users.sort((a, b) => a.email.localeCompare(b.email));
```
iameli, line 43: I'd never used `localeCompare`, why use it in this case?

darkdarkdragon:
@wohlner 
Let's merge this PR and then I'll improve on it in the next one
> That way instead of co-opting the streams tab that users use,

This tab is admin's only, users don't see it.

>It would be great in the tools tab to see a list of all active streams grouped by user

Now you can sort streams by user, that should do it for now, until we have much more streams

>  And then when you click into a stream, you get all the info you have in your first screenshot.

With the next PR I'll add all the info (under the 'admins only fields' section of course) to the regular streams page

wohlner:
No, i get that the users don't see the admin features. I just don't want that streams tab to get bloated with all sorts of features though. 

But I think that your suggestion "With the next PR I'll add all the info (under the 'admins only fields' section of course) to the regular streams page" is an excellent quick win. 

Nice work on this, Ivan. It's going to me immensely helpful

packages/api/src/store/db.ts
```diff
@@ -106,16 +108,32 @@ export class DB {
     )
   }
 
-  async query(query, ...params) {
+  queryWithOpts<T, I extends any[] = any[]>(
+    query: QueryConfig<I>,
+    opts: QueryOptions = {},
+  ): Promise<QueryResult<T>> {
+    if (opts.useReplica && this.replicaPool) {
```
iameli, line 115: This changes from a default of `true` to a default of `false` for useReplica. I like it `true` — we should almost always be using replica queries except for rare cases where we must have up-to-date data (probably billing-related)

iameli (changes requested):
Love nearly all of it, only change: I think `{"useReplica": true}` should be the default except for rare mission-critical cases where we might explicitly set `{"useReplica": false}`

packages/api/src/store/db.ts
darkdarkdragon: Fixed

packages/www/components/AdminTools/index.tsx
darkdarkdragon: Just because it is only function I can directly use in sort 😄

packages/api/src/controllers/stream.js
darkdarkdragon: Fixed

darkdarkdragon:
@iameli fixed

iameli (approved):
LGTM

PR #98, opened 2020-09-24
Add /api/version endpoint that returns hash of git commit

darkdarkdragon:
Add /api/version endpoint that returns hash of git commit
from which API server was built
Show this commit in the UI

**Screenshots (optional):**

image


adamsoffer (changes requested):
Nice idea. Can we just move this to the footer? 
image

darkdarkdragon:
@adamsoffer done in 62624a0ce7d20f04dddb780b73766a07dcf6689b

iameli (approved):
LGTM!

PR #99, opened 2020-09-26
Add filtering and pagination to the Admin Users page

darkdarkdragon:
Add filtering and pagination to the Admin Users page	

**Screenshots (optional):**

image

Update:
Added ability to filter users by their payment plan
image

### DIFF
--- a/packages/api/src/controllers/user.js
+++ b/packages/api/src/controllers/user.js
@@ -33,12 +33,15 @@ app.get('/usage', authMiddleware({}), async (req, res) => {
 })
 
 app.get('/', authMiddleware({ admin: true }), async (req, res) => {
-  let { limit, cursor, livepeer, filter } = req.query
+  let { limit, cursor, livepeer, filter, product } = req.query
   const query = []
   if (livepeer && livepeer !== 'false' && livepeer !== '0') {
     query.push(sql`data->>'email' LIKE '%livepeer%'`)
   } else if (filter) {
     query.push(sql`data->>'email' LIKE ${'%' + filter + '%'}`)
+  }
+  if (product) {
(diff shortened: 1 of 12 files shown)


wohlner:
I did not mean to close this. SORRY!

wohlner:
Awesome initiative on this! I am unable to test the previous and next buttons because there are less than 101 streams.

I tested the filter field, and when I type in "jen" i get way more results than expected, many that don't make sense. I would expect the results to be only `id`s and `emails` that contain `jen`. 

Screen Shot 2020-12-17 at 1 40 37 PM

wohlner:
Now I see that the preview text in the text field where I can filter says "email." But still I was getting back unexpected results besides emails that contain the string `jen`

darkdarkdragon:
@wohlner you won't be able to see it working correctly until it is deployed to staging - you're testing only frontend part, but filtering is done on backend which is not gets deployed to the Vercel's test deployment.

wohlner:
The "Show Livepeer's users" is confusing and takes up super valuable real estate above the table. 

Please change it to "Exclude users with a Livepeer email address" and this should only return users without @livepeer.com or @livepeer.org email addresses. This should be unchecked by default.

darkdarkdragon:
@wohlner 
> "Exclude users with a Livepeer email address"

This takes even more space 😃 
And what's the point - we have not that many Livepeer users? 
Initially I've added 'Show Livepeer's users only' when there was no filtering yet so it can be possible to find Livepeer user and make that user to be admin. But with filtering it's that important (though probably it's still good to have list of only Livepeer users - just to have overview).

wohlner:
Well, to get all LP users, we could just filter for "livepeer" and all the users with livepeer.com and .org email addresses would show up. How about you just remove that checkbox and not replace it with anything. 👍

PR #100, opened 2020-09-26
Allow to delete many streams at once in UI

darkdarkdragon:
Allows to delete many streams at once.
In my tests I'm creating streams using API, and so there are many unneeded stream objects left in my streams list, it would take too much time to delete them one-by-one using UI. And I saw one user in the Discord asking if it possible to delete many streams at once.

- Adds checkbox to select all streams
- Adds endpoint to delete many streams
- Slightly changes delete modal dialog's text if more than one stream selected

Manually

**Screenshots (optional):**
image

**Checklist:**

### DIFF
--- a/packages/api/src/store/table.ts
+++ b/packages/api/src/store/table.ts
@@ -49,6 +49,22 @@ export default class Table<T extends DBObject> {
     return res.rows[0].data as T
   }
 
+  async getMany(ids: Array<string>, opts: GetOptions = { useReplica: true }): Promise<Array<T>> {
+    if (!ids || !ids.length) {
+      throw new Error('missing ids')
+    }
+    let res: QueryResult<DBLegacyObject> = await this.db.queryWithOpts({
+      name: `${this.name}_by_ids`,
+      text: `SELECT data FROM ${this.name}  WHERE id IN (${ids.map((_, i) => '$' + (i + 1)).join(',')})`,
+      values: ids
+    }, opts)
+
+    if (res.rowCount < 1) {
+      return null
+    }
+    return res.rows.map(o => o.data as T)
+  }
+
   // returns [docs, cursor]
   async find(
     query: FindQuery | Array<SQLStatement> = {},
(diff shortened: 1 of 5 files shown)


packages/api/src/store/table.ts
```diff
@@ -173,6 +189,25 @@ export default class Table<T extends DBObject> {
...
+    if (res.rowCount < 1) {
+      throw new NotFoundError(`couldn't find ${this.name} id=${id}`)
+    }
+  }
+
+  async markDeletedMany(ids: Array<string>) {
+    const res = await this.db.query(`UPDATE ${this.name} SET data = jsonb_set(data, '{deleted}', 'true'::jsonb) WHERE id IN (${ids.map((_, i) => '$' + (i + 1)).join(',')})`, ids)
+
+    if (res.rowCount < 1) {
+      console.log(res)
```
adamsoffer: Can we remove this log?

packages/www/components/StreamsTable/index.tsx
```diff
@@ -242,15 +247,28 @@ export default ({ userId, id }: { userId: string; id: string }) => {
...
+          onClick={() => selectedStream.length && setDeleteModal(true)}
         >
           Delete
         </Button>
       </Box>
       <Table sx={{ gridTemplateColumns: "auto auto auto auto auto auto" }}>
         <TableRow variant={TableRowVariant.Header}>
-          <Box></Box>
+          <Flex
+            sx={{ alignItems: "baseline", pt: "8px" }}
```
adamsoffer: Can we remove the padding top so the checkbox is centered?
Screen Shot 2020-11-20 at 2 36 03 PM

adamsoffer (changes requested):
LGTM! Just a couple minor suggestions.

packages/www/components/StreamsTable/index.tsx
darkdarkdragon: fixed

packages/api/src/store/table.ts
darkdarkdragon: fixed

PR #123, opened 2020-10-16
Update segment based API docs

ericxtang:
Fix a few broken links, and update the language a bit to make things more clear.

## -

-

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/pages/docs/guides/segment-based-ingest/authentication.mdx
+++ b/packages/www/pages/docs/guides/segment-based-ingest/authentication.mdx
@@ -6,7 +6,7 @@ import DocsLayout from "../../../../components/DocsLayout";
 ## How to authenticate with your API key
 
 You need an API key to live transcode .ts segments via http push. Learn
-[how to create an API key](/docs/api-keys/create-an-api-key).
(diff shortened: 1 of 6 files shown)


wohlner:
Looking at /docs/guides/segment-based-ingest/create-a-stream-and-define-renditions. It doesn't include the actual `POST` request to `https://livepeer.com/api/stream`. This segment based transcoding docs page should be updated to be a near copy of https://livepeer.com/docs/guides/api/create-a-stream.

It seems like non-LP Inc people will be looking at these docs on livepeer.com, not just in the github UI. In that case, my only other comment is that since the menu for this section is purposely hidden from the main docs sidebar menu, it could be helpful to provide easy access back to the Segment Based Ingest Docs main page, /docs/guides/segment-based-ingest/segment-based-ingest. Two ideas: 1) open all links on /docs/guides/segment-based-ingest/segment-based-ingest in a new tab or 2) add a "Back to Segment Based Ingest docs home" link at the top of each doc page in that section.

ericxtang:
Thanks for the feedback @wohlner.  I updated the site based on your suggestions.

wohlner:
LGTM

PR #124, opened 2020-10-16
fix for sharding

darkdarkdragon:
updates stream-info-service for sharding

- updates /stream endpoint handlers to use `db.update` method
- add `db.add` method 
- uses `db.add` method in stream-info-service
- in stream-info-service look for object by id and by playbackId

manually

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/app/stream-info/stream-info-app.ts
+++ b/packages/api/src/app/stream-info/stream-info-app.ts
@@ -1,13 +1,12 @@
 import express, { Router } from 'express'
 import 'express-async-errors' // it monkeypatches, i guess
 import morgan from 'morgan'
-import createStore from '../../store'
+import { db } from '../../store'
+import { DB } from '../../store/db'
 import { healthCheck } from '../../middleware'
 import logger from '../../logger'
 import { Stream } from '../../schema/types'
 import fetch from 'isomorphic-fetch'
-
-import { IStore } from '../../types/common'
 
 import { StatusResponse, MasterPlaylist } from './livepeer-types'
 
@@ -17,9 +16,7 @@ const deleteTimeout = 5 * 60 * 1000
 const seenSegmentsTimeout = 60 * 1000
 
 async function makeRouter(params) {
-  // Storage init
   const bodyParser = require('body-parser')
-  let [db, store] = await createStore(params)
 
   // Logging, JSON parsing, store injection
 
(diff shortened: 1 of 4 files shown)


packages/api/src/app/stream-info/stream-info-app.ts
```diff
@@ -158,52 +151,25 @@ class statusPoller {
       const manifest = status.Manifests[mid]
       countSegments(si, manifest)
       if (needUpdate) {
-        const streamKey = `stream/${mid}`
-        const storedInfo: Stream = await this.store.get(streamKey, false)
+        let storedInfo: Stream = await db.stream.get(mid)
+        if (!storedInfo) {
```
iameli, line 155: Not required now but I've been thinking a function on the stream table that's like `db.stream.getByIdOrPlaybackIdOrStreamKey` would be a good idea

packages/api/src/app/stream-info/stream-info-app.ts
```diff
@@ -158,52 +151,25 @@ class statusPoller {
...
+          const [objs, _] = await db.stream.find({ playbackId: mid })
+          if (objs && Array.isArray(objs) && objs.length) {
```
iameli: This is a little defensive, `objs` will definitely always be an array

packages/api/src/app/stream-info/stream-info-app.ts
```diff
@@ -253,8 +219,9 @@ export default async function makeApp(params) {
     broadcaster,
   } = params
   // Storage init
+  await db.start({ postgresUrl, postgresReplicaUrl: undefined })
```
iameli, line 222: Did this cause problems somewhere?

packages/api/src/store/table.ts
```diff
@@ -143,6 +143,25 @@ export default class Table<T extends DBObject> {
     }
   }
 
+  async add(id: string, doc: T) {
```
iameli, line 147: Took me a long time to understand what was happening here. Looks good, but maybe let's add a comment:
```suggestion
  // Takes in an object of {"field": number} and increases all the fields by the specified amounts
  async add(id: string, doc: T) {
```

iameli (approved):
One documentation change you can just auto-apply but LGTM otherwise

packages/api/src/app/stream-info/stream-info-app.ts
darkdarkdragon: No, why are you asking?

PR #126, opened 2020-10-21
Fixed broken URL link

wohlner:
Changed /docs/guides/api-keys/when-do-you-need-an-API-key to /docs/guides/api-keys/when-do-you-need-an-api-key


darkdarkdragon (approved):
LGTM

PR #134, opened 2020-10-30
fix pricing display on mobile

adamsoffer:
This PR fixes the pricing display on mobile in order to make it clear which usage type each price point belongs to.

**Screenshots (optional):**
localhost_3008_


wohlner:
LGTM. Let's get this in prod 👍 
Thanks for the quick turnaround

PR #139, opened 2020-11-16
Add `publicUrl` field to the object-store object

darkdarkdragon:
Adds `publicUrl` field to the object-store object
`publicUrl` is a URL at which data stored to the object store can be publicly accessed.
For example, data stored to https://storage.googleapis.com/record-test/b64d415a-cc4b-465d-9faf-09a0132c2759/darkbook.lan/source/0.ts, but can be accessed from http://store.livepeer.live/record-test/b64d415a-cc4b-465d-9faf-09a0132c2759/nodeid/source/0.ts

Needed for stream recordings
https://github.com/livepeer/go-livepeer/pull/1582

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -588,7 +588,8 @@ app.post('/hook', async (req, res) => {
     return res.json({ errors: ['not found'] })
   }
   let objectStore,
-    recordObjectStore = undefined
+    recordObjectStore = undefined,
+    recordObjectStoreUrl
   if (stream.objectStoreId) {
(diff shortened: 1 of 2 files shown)


iameli (approved):
Code looks fine, one question: will this result in uploads being made to the OS's public URL? Is that what we want?

darkdarkdragon:
@iameli 
> will this result in uploads being made to the OS's public URL? Is that what we want?

Manifests will contain absolute URL to the segments, for that we need to know public facing, CDN backed URL at which they can be accessed.

PR #149, opened 2020-11-26
track session creation using user id of the stream

darkdarkdragon:
Fixes session creation tracking. 

Track session using owner of stream itself instead of using owner of the key (stream sessions usually created by Mist which uses own auth token). 

Manually

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -300,12 +300,24 @@ app.post(
 
     try {
       await req.store.create(doc)
-      trackAction(
-        req.user.id,
-        req.user.email,
-        { name: 'Stream Session Created' },
-        req.config.segmentApiKey,
-      )
+      setImmediate(async () => {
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/stream.js
```diff
@@ -300,12 +300,20 @@ app.post(
 
     try {
       await req.store.create(doc)
-      trackAction(
-        req.user.id,
-        req.user.email,
-        { name: 'Stream Session Created' },
-        req.config.segmentApiKey,
-      )
+      setImmediate(async () => {
```
iameli, line 303: I think because we're in a `setImmediate` here and we've exited the async loop, any unhandled rejection [will result in the server crashing entirely](https://nodejs.org/api/deprecations.html#DEP0018). No good. Easy enough to work around with a try/catch, feel free to merge after that's added.

iameli (changes requested):
I

PR #152, opened 2020-12-01
[READY] global Orchestrator list schema and endpoints

ya7ya:

## -

-

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -9,7 +9,11 @@ const getOrchestrators = async (req, res, next) => {
   return res.json(orchestrators.map(({ address }) => ({ address })))
 }
 
-app.get('/:region', authMiddleware({ admin: true }), getOrchestrators)
+app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
+  const region = await req.store.get(`region/${req.params.region}`)
+  return res.json(region)
+})
+
 app.get('/orchestrator', getOrchestrators)
 
 export default app
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -5,8 +5,8 @@ import { db } from '../store'
 const app = Router()
 
 app.get('/', authMiddleware({ admin: true }), async (req, res, next) => {
-  const [regions, cursor] = await db.region.find()
-  if (req.query.grouped) {
+  const [regions, cursor] = await db.region.find({}, { limit: 100 })
+  if (req.query.grouped === 'true') {
     return res.json(regions)
   }
 
@@ -40,9 +40,11 @@ app.put(
       orchestrators: req.body.orchestrators || [],
     }
 
-    const currentRegion = await db.region.get(req.params.region)
+    const currentRegion = await db.region.get(req.body.region)
+
     if (currentRegion) {
       const updateResp = await db.region.update(req.params.region, region)
+      console.log('update Resp: ', updateResp)
     } else {
       const resp = await db.region.create({
         id: region.region,
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -41,7 +41,7 @@ app.put('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
     orchestrators: req.body.orchestrators || [],
   }
 
-  const resp = await req.store.create({
+  const resp = await db.region.create({
     id: region.region,
     kind: 'region',
     region: region.region,
@@ -55,7 +55,7 @@ app.delete(
   '/:region',
   authMiddleware({ admin: true }),
   async (req, res, next) => {
-    const resp = await req.store.delete(`region/${req.params.region}`)
+    const resp = await db.region.delete(`region/${req.params.region}`)
     return res.json({ ok: 1 })
   },
 )
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -1,0 +1,15 @@
+import Router from 'express/lib/router'
+import { authMiddleware } from '../middleware'
+
+const app = Router()
+
+const getOrchestrators = async (req, res, next) => {
+  const orchestrators = await req.getOrchestrators(req)
+
+  return res.json(orchestrators.map(({ address }) => ({ address })))
+}
+
+app.get('/:region', authMiddleware({ admin: true }), getOrchestrators)
+app.get('/orchestrator', getOrchestrators)
+
+export default app
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -1,16 +1,28 @@
 import Router from 'express/lib/router'
 import { authMiddleware } from '../middleware'
+import { db } from '../store'
 
 const app = Router()
 
-const getOrchestrators = async (req, res, next) => {
-  const orchestrators = await req.getOrchestrators(req)
+app.get('/', authMiddleware({ admin: true }), async (req, res, next) => {
+  const [regions, cursor] = await db.region.find()
+  if (req.query.grouped) {
+    return res.json(regions)
+  }
 
-  return res.json(orchestrators.map(({ address }) => ({ address })))
-}
+  const flatOrchList = []
+  regions.forEach((region) => {
+    region.orchestrators.forEach((orch) => {
+      orch.region = region.region
+      flatOrchList.push(orch)
+    })
+  })
+
+  return res.json(flatOrchList)
+})
 
 app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
-  const region = await req.store.get(`region/${req.params.region}`)
+  const region = await db.region.get(req.params.region)
   return res.json(region)
 })
 
@@ -29,11 +41,11 @@ app.put('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
     orchestrators: req.body.orchestrators || [],
   }
 
-  console.log(`req.body ${JSON.stringify(req.body)}`)
   const resp = await req.store.create({
     id: region.region,
     kind: 'region',
-    ...region,
+    region: region.region,
+    orchestrators: region.orchestrators,
   })
 
   return res.json(region)
@@ -47,7 +59,5 @@ app.delete(
     return res.json({ ok: 1 })
   },
 )
-
-app.get('/orchestrator', getOrchestrators)
 
 export default app
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -16,4 +16,34 @@ app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
 
 app.get('/orchestrator', getOrchestrators)
 
+app.put('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
+  if (!req.body || !req.body.region) {
+    res.status(406)
+    return res.json({ error: 'region field is required' })
+  }
+
+  if (!req.body.orchestrators || !req.body.orchestrators.length) {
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,57 @@
...
+app.delete(
+  '/:region',
+  authMiddleware({ admin: true }),
+  async (req, res, next) => {
+    const resp = await req.store.delete(`region/${req.params.region}`)
+    return res.json({ ok: 1 })
+  },
+)
+
+app.get('/orchestrator', getOrchestrators)
```
iameli: This will need to go through every region and return a flat array in the same format as https://livepeer.com/api/orchestrator

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,57 @@
+import Router from 'express/lib/router'
+import { authMiddleware } from '../middleware'
+
+const app = Router()
+
+const getOrchestrators = async (req, res, next) => {
```
iameli: I don't think we need `req.getOrchestrators()` anywhere here. Maybe we will once we figure out how each region is individually updating the global O list, but this set of endpoints shouldn't need it for anything. It's all about just getting updated externally.

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return res.json(orchestrators.map(({ address }) => ({ address })))
+}
+
+app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
+  const region = await req.store.get(`region/${req.params.region}`)
+  return res.json(region)
+})
+
+app.post(
```
iameli: We should make a JSON-Schema validator for this and do a `validatePost()` middleware.

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,57 @@
...
+  '/:region',
+  authMiddleware({ admin: true }),
+  async (req, res, next) => {
+    if (!req.body || !req.body.region) {
+      res.status(406)
+      return res.json({ error: 'region field is required' })
+    }
+
+    if (!req.body.orchestrators || !req.body.orchestrators.length) {
```
iameli: If it's omitted, that can be an error. But I think an empty array is valid; there might not be an orchestrator in a given region.

packages/api/src/controllers/region.js
iameli: I also think this should be a PUT that creates the region if it doesn't exist, rather than a POST that makes a new region every time. The use case here is each region continually pushing its O list to the server here, we won't need to create a lot of new regions with new ids.

packages/api/src/schema/schema.yaml
```diff
@@ -469,6 +469,26 @@ components:
           example: 66E2161C-7670-4D05-B71D-DA2D6979556F
           index: true
 
+    region:
+      type: object
+      table: regions
+      required:
```
iameli, line 525: Let's require both

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,57 @@
...
+const app = Router()
+
+const getOrchestrators = async (req, res, next) => {
+  const orchestrators = await req.getOrchestrators(req)
+
+  return res.json(orchestrators.map(({ address }) => ({ address })))
+}
+
+app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
+  const region = await req.store.get(`region/${req.params.region}`)
```
iameli: `req.store` is deprecated, we should move to `db.region` for everything going forward.

packages/api/src/controllers/region.test.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+      adminUser = { ...user, admin: true, emailValid: true }
+      await server.store.replace(adminUser)
+    })
+
+    it('should be able to upsert a region as admin', async () => {
+      const region = {
+        region: 'BER',
+        orchestrators: [
+          { address: 'localhost:7935' },
```
iameli: It doesn't matter too much in the test case, but these will generally be formatted `https://orchestrator.example.com:443`

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,57 @@
...
+    if (!req.body.orchestrators || !req.body.orchestrators.length) {
+      console.log(`WARNING req.body.orchestrators is undefined`)
+    }
+
+    let region = {
+      region: req.body.region,
+      orchestrators: req.body.orchestrators || [],
+    }
+
+    console.log(`req.body ${JSON.stringify(req.body)}`)
```
iameli: can clean this up

packages/api/src/controllers/region.js
iameli: (But `validatePost()` should still work to validate even though it's a PUT)

iameli (changes requested):
Cool, good start. Main thing missing here is the `GET /region/orchestrators` endpoint that will serve a flat list of Os like https://livepeer.com/api/orchestrator

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+      orch.region = region.region
+      flatOrchList.push(orch)
+    })
+  })
+
+  return res.json(flatOrchList)
+})
+
+app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
+  const region = await db.region.get(req.params.region)
```
iameli, line 25: Need to account for the 404 case here, or else we'll return a 200 with `null` as the result. (This changed a while back - back in the day a missing response threw a 404 error automatically.)

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+  return res.json(region)
+})
+
+app.put('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
```
iameli: We should add a `handlePost` middleware here to validate the shape of the incoming object against the definition in `schema.yaml`

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+  if (!req.body || !req.body.region) {
```
iameli: The middleware will take care of this automatically :+1:

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+    res.status(406)
+    return res.json({ error: 'region field is required' })
+  }
+
+  if (!req.body.orchestrators || !req.body.orchestrators.length) {
```
iameli: An absent `req.body.orchestrators` will be treated as an error condition by the `handlePost` middleware. But an empty list for a region is totally legit, don't necessarily need to log it

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+
+  return res.json(region)
+})
+
+app.delete(
+  '/:region',
+  authMiddleware({ admin: true }),
+  async (req, res, next) => {
+    const resp = await db.region.delete(`region/${req.params.region}`)
+    return res.json({ ok: 1 })
```
iameli: Usually our DELETEs just do `res.sendStatus(204)`, no need for a response in this case.

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+  if (!req.body.orchestrators || !req.body.orchestrators.length) {
+    console.log(`WARNING req.body.orchestrators is undefined`)
+  }
+
+  let region = {
+    region: req.body.region,
+    orchestrators: req.body.orchestrators || [],
+  }
+
+  const resp = await db.region.create({
```
iameli: So we don't want to create a new region every time. The main use case here: we're going to run a daemon in each region that runs `PUT /regions/:regionName` every minute or so. So this will need to be an "upsert" operation — first try and get the region. If it already exists, do a `db.region.update({})` with the new list of Os. If it doesn't, `db.region.create({})` just like this.

packages/api/src/schema/schema.yaml
```diff
@@ -469,6 +469,27 @@ components:
...
+        region:
+          type: string
+          description: region name
+          example: ber
+          unique: true
+          index: true
+        orchestrators:
+          type: array
+          items:
+            type: object
```
iameli, line 538: Should add `additionalProperties: false` here or you can add whatever else you want. Fun! Also should do:
```
required:
- address
```
So that we don't accept empty objects.

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
+import Router from 'express/lib/router'
+import { authMiddleware } from '../middleware'
+import { db } from '../store'
+
+const app = Router()
+
+app.get('/', authMiddleware({ admin: true }), async (req, res, next) => {
+  const [regions, cursor] = await db.region.find()
```
iameli: Let's do this just to make sure we don't get bit by the default limit count:
```suggestion
  const [regions, cursor] = await db.region.find({limit: 1000})
```

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,63 @@
...
+  if (req.query.grouped) {
```
iameli: Hmm, this will show grouped results for `?grouped=true` and `?grouped=false` and `?grouped=potato`. Maybe 
```suggestion
  if (req.query.grouped === "true") {
```

packages/api/src/controllers/region.test.js
```diff
@@ -0,0 +1,133 @@
...
+        region: 'VNO',
+        orchestrators: [
+          { address: 'https://orchestrator.example.com:443' },
+          { address: 'https://orchestrator.example.com:443' },
+          { address: 'https://orchestrator.example.com:443' },
+          { address: 'https://orchestrator.example.com:443' },
+        ],
+      }
+
+      let res2 = await client.put(`/region/${region2.region}`, { ...region2 })
```
iameli, line 110: Let's do a test where we PUT to the same region twice and make sure the new results are overridden

iameli (changes requested):
Main things: let's add a `validatePost` middleware and handle the multiple `PUT` case.

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,70 @@
...
+    let region = {
+      region: req.body.region,
+      orchestrators: req.body.orchestrators || [],
+    }
+
+    const currentRegion = await db.region.get(req.body.region)
+
+    if (currentRegion) {
+      const updateResp = await db.region.update(req.params.region, region)
+      console.log('update Resp: ', updateResp)
```
iameli: cleanup

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,70 @@
...
+    })
+  })
+
+  return res.json(flatOrchList)
+})
+
+app.get('/:region', authMiddleware({ admin: true }), async (req, res, next) => {
+  const region = await db.region.get(req.params.region)
+  if (!region) {
+    return res.status(404)
```
iameli: This will actually hang forever because there's no `res.end()`. How 'bout
```suggestion
    res.status(404)
    return res.json({errors: ["not found"]})
```

packages/api/src/controllers/region.js
```diff
@@ -0,0 +1,70 @@
...
+
+    return res.json(region)
+  },
+)
+
+app.delete(
+  '/:region',
+  authMiddleware({ admin: true }),
+  async (req, res, next) => {
+    const resp = await db.region.delete(`region/${req.params.region}`)
```
iameli: Hmm, does this work? I don't think `region/` does anything here — we can just delete `req.params.region` directly

iameli (approved):
Sweet. Some tiny cleanup, feel free to merge after

PR #155, opened 2020-12-04
/admin/active-clean endpoint

### LINKED ISSUE #185: automatically clean active streams on the way out of the API
* On requests for GET /api/stream and GET /api/stream/{id}, add an additional check to the streams on the way out to users. Specifically we'll be looking at the active and lastSeen fields.
* If lastSeen is more than (for example) 3 minutes old, do two things:
1. On the way out to the user, tweak the field so the stream shows up as {active: false}. (Or if we're supposed to ONLY be listing active streams, don't include the stream at all.)
2. Asynchronously, dispatch a task to update the stream object to set {active: false} in the database. (But don't slow down the GET response to make this happen, just do it in the background.)

That way we can be confident that we are returning correct stream statuses to users even in between runs of the "active-clean" cron job.

darkdarkdragon:
Checks if there is streams marked 'active' but wasn't seen recently
and mark them as not active
Intended to be called periodically by cron job

- added `/admin/active-clean` endpoint

## -

-

Fixes #185 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/admin.js
+++ b/packages/api/src/controllers/admin.js
@@ -1,0 +1,45 @@
+import { authMiddleware } from '../middleware'
+import Router from 'express/lib/router'
+import { db } from '../store'
+import sql from 'sql-template-strings'
+
+const ACTIVE_TIMEOUT = 24 * 3600 * 1000 // 24h
+
+const app = Router()
+
+app.post('/active-clean', authMiddleware({ anyAdmin: true }), async (req, res) => {
+  // check if there is streams marked 'active' but wasn't seen recently
+
+  const query = []
+  query.push(sql`data->>'isActive' = 'true'`)
+
+  let cleaned = 0
+  let docs, cursor
+  const now = Date.now()
+  const toClean = []
+  do {
+    [docs, cursor] = await db.stream.find(query, { cursor, limit: 100 })
+    console.log(`got ${docs.length} active streams, cursor: ${cursor}`)
+    for (const stream of docs) {
+        const lastSeen = +new Date(stream.lastSeen)
+        const needClean = isNaN(lastSeen) ||(now - lastSeen) > ACTIVE_TIMEOUT
+        console.log(`id ${stream.id} seen ${new Date(stream.lastSeen)} need clean ${needClean} name ${stream.name}`)
+        if (needClean) {
+            toClean.push(stream)
+        }
+    }
+  } while (cursor && docs.length === 100)
+  for (const stream of toClean) {
+    await db.stream.update(stream.id, {
+        isActive: false,
+    })
+    console.log(`stream ${stream.id} (${stream.name}) active set to false`)
+    cleaned++
+  }
+  const msg = `cleaned ${cleaned} streams`
+
+  res.status(200)
+  res.json({ msg })
+})
+
+export default app
--- a/packages/api/src/controllers/admin.js
+++ b/packages/api/src/controllers/admin.js
@@ -4,8 +4,6 @@ import { db } from '../store'
 import sql from 'sql-template-strings'
 import consul from 'consul'
 
-// const consul = require('consul')()
-
 const ACTIVE_TIMEOUT = 1 * 3600 * 1000 // 1h
 
 const app = Router()
@@ -25,6 +23,9 @@ app.post(
       if (!baseUrl.endsWith('/v1')) {
         baseUrl += '/v1'
       }
+    } else {
+      res.status(501)
+      return res.end()
     }
     const cc = consul({ promisify: true, baseUrl })
     let keys
@@ -40,7 +41,7 @@ app.post(
       } else {
         console.log(`error getting keys:`, e)
         res.statusCode = 500
-        return res.json({ error: String(e) })
+        return res.json({ errors: [String(e)] })
       }
     }
     let i = 0,
@@ -86,6 +87,9 @@ app.post(
       ])
       if (keysResp && keysResp.Results) {
         // deleting keys
+        // 'delete-cas' will not delete key in case it was
+        // modified after we've read it
+        // (could happend if stream started after we've read key)
         const deleteTxn = keysResp.Results.map((rk) => {
           return {
             KV: {
@@ -147,12 +151,12 @@ app.post(
     let cleaned = 0
     for (const stream of toClean) {
       const upRes = await db.stream.update(
-        stream.id,
-        {
-          isActive: false,
-        },
-        ` (data->>'lastSeen')::bigint = ${stream.lastSeen}`,
-        true,
+        [
+          sql`id = ${stream.id}`,
+          sql`(data->>'lastSeen')::bigint = ${stream.lastSeen}`,
+        ],
+        { isActive: false },
+        { throwIfEmpty: false },
       )
       if (upRes.rowCount) {
         console.log(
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,45 @@
...
+
+  let cleaned = 0
+  let docs, cursor
+  const now = Date.now()
+  const toClean = []
+  do {
+    [docs, cursor] = await db.stream.find(query, { cursor, limit: 100 })
+    console.log(`got ${docs.length} active streams, cursor: ${cursor}`)
+    for (const stream of docs) {
+        const lastSeen = +new Date(stream.lastSeen)
```
iameli: Kind of prefer this compared to the magic `+`:
```suggestion
        const lastSeen = new Date(stream.lastSeen).getTime()
```

packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,45 @@
...
+        const needClean = isNaN(lastSeen) ||(now - lastSeen) > ACTIVE_TIMEOUT
+        console.log(`id ${stream.id} seen ${new Date(stream.lastSeen)} need clean ${needClean} name ${stream.name}`)
```
iameli: Trying to standardize this log formatting just a little
```suggestion
        console.log(`cleaning stream id=${stream.id} lastSeen=${new Date(stream.lastSeen)} name=${stream.name}`)
```
And maybe only log if we need to clean it? Otherwise we'll log every active stream every time

packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,45 @@
...
+        if (needClean) {
+            toClean.push(stream)
+        }
+    }
+  } while (cursor && docs.length === 100)
+  for (const stream of toClean) {
+    await db.stream.update(stream.id, {
+        isActive: false,
+    })
+    console.log(`stream ${stream.id} (${stream.name}) active set to false`)
```
iameli:
```suggestion
    console.log(`timed out stream id=${stream.id} name=${stream.name})`)
```

iameli (changes requested):
Logging changes, looks solid

packages/api/src/controllers/admin.js
darkdarkdragon: fixed

packages/api/src/controllers/admin.js
darkdarkdragon: Yeah, this line was here just for debugging, removed

packages/api/src/controllers/admin.js
darkdarkdragon: fixed

darkdarkdragon:
@iameli please take a look again - I've added check to prevent potential race

iameli:
Looking now. Love that we're interfacing with Consul from the API server, that feels correct.

packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,172 @@
+import { authMiddleware } from '../middleware'
+import Router from 'express/lib/router'
+import { db } from '../store'
+import sql from 'sql-template-strings'
+import consul from 'consul'
+
+// const consul = require('consul')()
```
iameli: Cleanup

packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,172 @@
...
+const traefikKeyPathMiddlewares = `traefik/http/middlewares/`
+
+app.post(
+  '/consul-routes-clean',
+  authMiddleware({ anyAdmin: true }),
+  async (req, res) => {
+    // check if there is routes in the consul that belongs to the streams that
+    // are not active anymore
+    let baseUrl
+    if (req.config.consul) {
```
iameli, line 21: Should presumably return an HTTP error here if this isn't set. Probably a 501 Not Implemented?

packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,172 @@
...
+        separator: '/',
+      })
+      keys = keys.map((k) => k.split('/')[3])
+    } catch (e) {
+      if (e.statusCode == 404) {
+        return res.json({ msg: 'no keys' })
+      } else {
+        console.log(`error getting keys:`, e)
+        res.statusCode = 500
+        return res.json({ error: String(e) })
```
iameli: I like that we're mostly consistent on returning errors as an array:
```suggestion
        return res.json({ errors: [String(e)] })
```

packages/api/src/controllers/admin.js
```diff
@@ -0,0 +1,172 @@
...
+      // sending progress should prevent request timing out
+      res.write('.')
+      for (const stream of docs) {
+        const lastSeen = new Date(stream.lastSeen).getTime()
+        const needClean = isNaN(lastSeen) || now - lastSeen > ACTIVE_TIMEOUT
+        if (needClean) {
+          toClean.push(stream)
+        }
+      }
+    } while (cursor && docs.length === 100)
```
iameli, line 148: No need for the other check; `cursor` comes back `null` if there aren't any more documents.
```suggestion
    } while (cursor)
```

packages/api/src/store/table.ts
```diff
@@ -146,17 +146,21 @@ export default class Table<T extends DBObject> {
     }
   }
 
-  async update(id: string, doc: T) {
+  async update(id: string, doc: T, where: string = '', doNotThrow = false) {
```
iameli: I don't love this interface. What about this one instead?
```typescript
async update(
    query: string | Array<SQLStatement>,
    doc: T,
    opts: UpdateOptions = {throwIfEmpty = true},
):
```
If `query` is a string it's the same as the current function, otherwise
Advantages:
* Very similar to the interface of find(), so it's consistent
* Usage of SQLStatement protects us from SQL injection down the line when someone does a WHERE with 
* `opts` object means we can add more options later without adding more arguments

Your call above could be rewritten as:
```typescript
      const upRes = await db.stream.update(
        [
          sql`id = ${stream.id}`,
          sql`(data->>'lastSeen')::bigint = ${stream.lastSeen}`,
        ],
        { isActive: false },
        { throwIfEmpty: false }
      )

**Review by iameli — CHANGES_REQUESTED** [review]
Main thing is a suggestion for how we could make `db.table.update()` a little cleaner

File: packages/api/src/controllers/admin.js (hunk 7 above)
Comment (darkdarkdragon) [reply packages/api/src/controllers/admin.js:148]: No, it isn't
```js
    return [docs, docs[docs.length - 1].id]
```
Even if there is only one document present, it will return id of the last document.

packages/api/src/store/table.ts
darkdarkdragon: fixed

packages/api/src/controllers/admin.js
darkdarkdragon: fixed

packages/api/src/controllers/admin.js
darkdarkdragon: fixed

packages/api/src/controllers/admin.js
iameli: Ah, boo. Should probably change that but alright.

darkdarkdragon: There is fix for that in my other PR )

iameli (approved):
LGTM. We'll need to add `LP_CONSUL` to `livepeer-infra.

PR #157, opened 2020-12-07
add stream debugger page

adamsoffer:
This PR introduces a "stream debugger" page per the product spec outlined [here](https://docs.google.com/document/d/1zZrZ7LC4Sqw7rvVlWqEhfJ2HbZMK5ZeyWgMl2uPRzRU/edit?usp=sharing).


wohlner:
Still figuring out how to test, but one design note on /app/user: The "stream debugger" link has to be WAY more prominent. Like "can't miss it" prominent. Want to try something else out, @adamsoffer? If not, let's put it in the tab menu between _api keys_ and _usage_

wohlner:
I am still struggling to load a playback URL into the debugger. I'll work in discord to figure that out. But here are some more UI comments: 

1. We need some sort of error messaging if the playback URL isn't going to load the rest of the page? Does shaka provide anything we could leverage?
2. Please change the example URL under the text field to something more neutral and with .com instead of .monster: `https://fra-cdn.livepeer.com/hls/123456abcdef7890/index.m3u8`

wohlner:
Never mind about this one . I see your stream not found message.
> 1. We need some sort of error messaging if the playback URL isn't going to load the rest of the page? Does shaka provide anything we could leverage?

wohlner:
Other stuff we talked about in zoom review:

1. Only show status and playback settings below the two players. Remove other metadata
2. If possibly, format the playback settings to be more human readable (new line for each profile)
3. Add text by the "source" header above the left player. Add this: `(Click on ⋮ in the lower right corner. Select the source resolution, usually the highest resolution.)`

wohlner:
@adamsoffer a couple more tweaks:

- [ ] Please change the URL slug to from /app/debugger to `/app/test-player`.
- [ ] Change the page title from Stream Debugger to `Test your Livepeer.com stream`
- [ ] Change the copy under the title to "Test and debug your Livepeer.com stream. Follow the step-by-step process in the [Livepeer.com debugging guide](url)."
- [ ] Remove the entire section that says "Does something not look right?" because that content will now be right below the page title.

adamsoffer:
Pushed up the above mentioned tweaks (with the exception of formatting profile settings json string...not sure how to accomplish that). I did manage to configure the playback controls and remove everything except for the resolution switcher. I also disabled ABR on the first player so it just shows the source video.

Also, if you're cool with it I changed the page title from "Test your Livepeer.com stream" to just "Test Player" since we have the same description right below and it felt redundant. It also reinforces the name of the page and will help users know what to refer to it as.

wohlner:
@adamsoffer great work, we are so close! The "Test Player" title for the page is great. Good call. And great work finding out how to remove the miscellaneous player controls.

Here are the small tweaks I'd like you to make:
- [ ] When the stream is not active, the stream status should be `Idle` instead of `Inactive`. This aligns with the naming conventions on the stream page in the .com dashboard.
- [ ] Change the left "Source" player heading to `Source (highest resolution only)`
- [ ] Change the right "ABR" player heading to `ABR (source + transcoded renditions`
- [ ] This may be a bug. I was testing this playback url, https://fra-cdn.livepeer.monster/hls/qw1ujho9rfqs5iu3/index.m3u8The right player. The stream is is `06896e99-e02a-4cf9-977a-18f6c9271840`. When I went to change the resolution in the lower right corner of the ABR player, I could only select from 720p and auto. However, I have 4 profile renditions (and resolutions) set for that stream. I was expecting to be able to switch among those resolutions on the ABR player.

Lastly, the test player link on /app/user is not prominent enough. Please move it to the tabbed menu between API Keys and Usage. 

I am going to keep testing to try and find any remaining bugs, but this is looking great!!

wohlner:
One last thing, the debugging guide link in the subtitle will be at /docs/guides/debugging-guide.

adamsoffer:
@wohlner just pushed up fixes for the above. Also, I realized I forgot to display the profile settings attached to the session instead of the stream - fixed that as well.

Regarding what you're seeing in the resolution switcher, that's odd I'm seeing a bunch of renditions. I'm streaming via OBS 🤷‍♂️
image

wohlner:
Yeah, I see all the renditions now too. Might have been user error, oops.

Final changes!
- [ ] Add a colon after playback settings to be consistent with status.
- [ ] Move "test player" to the third spot in the tab menu instead of the last spot.

adamsoffer:
@wohlner just pushed up final changes.

wohlner:
@adamsoffer a couple tweaks from others testing:

- [ ] Please add back volume/mute controls on both players 
- [ ] Please default both players to mute
- [ ] Make the playback settings a code block instead of regularly formatted text.

Thank you!

PR #159, opened 2020-12-09
Wildcard routing

darkdarkdragon:

Add `base-stream-name` command line argument
if set, calls to `/api/stream/baseStreamName/stream` should have name
to be 'baseStreamName+playbackId'. Then session object is created with
transcoding profiles taken from parent object.

## -

-

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -210,8 +210,25 @@ app.post(
         errors: ['missing name'],
       })
     }
+    let stream
+    let useParentProfiles = false
+    if (req.config.baseStreamId === req.params.streamId) {
+      if (!req.body.name.includes('+')) {
+        res.status(422)
+        return res.json({
+          errors: ['wrong name'],
+        })
+      }
+      const playbackId = req.body.name.split('+')[1]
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/stream.js
```diff
@@ -210,8 +210,25 @@ app.post(
         errors: ['missing name'],
       })
     }
+    let stream
+    let useParentProfiels = false
```
iameli: Typo 
```suggestion
    let useParentProfiles = false
```

packages/api/src/parse-cli.js
```diff
@@ -173,6 +173,12 @@ export default function parseCli(argv) {
             "id of the object store that should be used for `record: true` requests that don't otherwise have an os",
           type: 'string',
         },
+        'base-stream-name': {
+          describe:
+            'base stream name to be used in wildcard-based routing scheme.',
+          type: 'string',
+          default: '',
```
iameli: This is how Go would do it, but I think it would be better to just not have a default and it will be `undefined`.

iameli (changes requested):
Minor changes, feel free to merge after these two things

PR #160, opened 2020-12-09
api: add duration logging for db queries

iameli:
Changed SQL statement logging to be on a single line for easier parsing on Papertrail. Added duration logging. End result looks like this:

image


darkdarkdragon (approved):
LGTM

PR #163, opened 2020-12-09
api: fix typo - streamId isn't a thing but streamKey is

iameli:
Curious how this hasn't caused more problems before now...


darkdarkdragon (approved):
LGTM

PR #164, opened 2020-12-10
Do not use replica if /setactive sets active to true

darkdarkdragon:
If stream starts streaming right after creation sometimes stream object is not propagated to replica yet, so /setactive handler can't find it and rejects stream.

Do not use replica /setactive sets stream to active

## -

-

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -385,7 +385,8 @@ app.post('/', authMiddleware({}), validatePost('stream'), async (req, res) => {
 app.put('/:id/setactive', authMiddleware({}), async (req, res) => {
   const { id } = req.params
   // logger.info(`got /setactive/${id}: ${JSON.stringify(req.body)}`)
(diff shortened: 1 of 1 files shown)


iameli (approved):
This call is NOT in the blocking stream-rewrite trigger workflow, correct? If that's right, LGTM.

We should maybe consider implementing global consistency on stream creation, just waiting until the stream is populated on the replicas before we do anything else. I'll open a ticket.

darkdarkdragon:
@iameli 
It is IN blocking workflow - that's why this PR is needed. 
Currently REWRITE trigger does two API calls - first to just find stream, and second to /setactive (both of course blocking)
I'm thinking of creating separate /start_stream endpoint that will combine two, but that's the next step.

iameli:
Oh, I see, so this actually makes it have less latency — we'll be using the replica less often. That makes sense to me, then. :+1:

PR #167, opened 2020-12-11
test player updates

adamsoffer:
- Adds mute/volume control to player
- Sets player to muted by default
- Fixes issue where video won't play if the the browser has an ad blocker (brave and firefox)
- Formats profile settings inside a code block


adamsoffer:
@wohlner just pushed up a fix that address the issue where playback only worked for admin users.

PR #170, opened 2020-12-15
API and UI for 'user' sessions with recordings

darkdarkdragon:
UI for recordings

- adds /stream/:streamId/sessions endpoint that returns 'user' session with recording urls	
- adds sessions table on stream info page which shows status (only if recording url is not ready) and recording url

**Screenshots (optional):**

image
image


wohlner:
`/api/stream/:streamId/sessions` intuitively should return **all** streamer (already combined) sessions, which it does now. /api/stream/:streamId/sessions` with a filter for `record:true` should then return just the sessions with recordings, and `record:false` should return just the sessions without recording.

PR #209, opened 2021-01-14
api:discovery auth webhook implementation for /api/broadcaster/hook

### LINKED ISSUE #205: api: Webhook impl for orchestrators
At the very least will return a price. Remaining details to be filled in.

kyriediculous:

Implements a discovery webhook for broadcasters. The current implementation returns a 200 OK for all broadcasters. If provided as `LP_BROADCASTERS` will also return a response object with the set price info.

- returns 200 OK with an [optional response](https://github.com/livepeer/go-livepeer/blob/a5fa25aec93d01218cb36125d0690fb067d16b4c/server/rpc.go#L325) if a broadcaster is authenticated
- returns another status if broadcaster is not authenticated (e.g. blacklisted)

relies on `LP_BROADCASTERS` to be extended when this endpoint is used in our setup 

```
    LP_BROADCASTERS: |-
      [
        {
          "address": "https://mdw.livepeer.monster",
          "eth": {
            "address": "0xc3c7c4c8f7061b7d6a72766eee5359fe4f36e61e",
            "priceInfo": {"pricePerUnit: "0", pixelsPerUnit: "0"}
          }
        }
      ]
```

Fixes #205

### DIFF
--- a/packages/api/src/controllers/orchestrator.js
+++ b/packages/api/src/controllers/orchestrator.js
@@ -11,4 +11,23 @@ const getOrchestrators = async (req, res, next) => {
 app.get("/", getOrchestrators);
 app.get("/ext/:token", getOrchestrators);
 
+async function discoveryAuthWebhookHandler(req, res) {
+  const responseObj = {};
+  let prices = [];
+  try {
+    prices = await req.getPrices();
+  } catch (err) {
+    return res.status(400).send({ message: err.message });
+  }
+  for (let i = 0; i < prices.length; i++) {
+    if (prices[i].address == req.body.id) {
+      responseObj["priceInfo"] = prices[i].priceInfo;
+      break;
+    }
+  }
+  return res.status(200).json(responseObj);
+}
+
+app.get("/hook/auth", discoveryAuthWebhookHandler);
+
 export default app;
--- a/packages/api/src/app-router.js
+++ b/packages/api/src/app-router.js
@@ -53,6 +53,7 @@ export default async function makeApp(params) {
     orchestrators = "[]",
     broadcasters = "[]",
     ingest = "[]",
+    prices = "[]",
     insecureTestToken,
     firestoreCredentials,
     firestoreCollection,
@@ -111,7 +112,7 @@ export default async function makeApp(params) {
     );
   }
 
-  app.use(hardcodedNodes({ orchestrators, broadcasters, ingest }));
+  app.use(hardcodedNodes({ orchestrators, broadcasters, ingest, prices }))
 
   // Add a controller for each route at the /${httpPrefix} route
   const prefixRouter = Router(); // amalgamates our endpoints together and serves them out
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -27,6 +27,7 @@ export default async function makeApp(params) {
     fallbackProxy,
     orchestrators,
     broadcasters,
+    prices,
     s3Url,
     s3UrlExternal,
     s3Access,
--- a/packages/api/src/middleware/hardcoded-nodes.js
+++ b/packages/api/src/middleware/hardcoded-nodes.js
@@ -7,14 +7,16 @@ export default function hardcodedNodes({
   broadcasters,
   orchestrators,
   ingest,
+  prices,
 }) {
   try {
     broadcasters = JSON.parse(broadcasters);
     orchestrators = JSON.parse(orchestrators);
     ingest = JSON.parse(ingest);
+    prices = JSON.parse(prices);
   } catch (e) {
     console.error(
-      "Error parsing LP_BROADCASTERS and LP_ORCHESTRATORS and LP_INGEST"
+      "Error parsing LP_BROADCASTERS, LP_ORCHESTRATORS, LP_INGEST and AND LP_PRICES"
     );
     throw e;
   }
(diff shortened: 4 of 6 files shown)


packages/api/src/controllers/broadcaster.js
```diff
@@ -59,4 +59,23 @@ app.get('/status', authMiddleware({}), async (req, res, next) => {
...
+  let responseObj = {}
+  let broadcasters = []
+  try {
+    broadcasters = await req.getBroadcasters(req)
+  } catch (err) {
+    return res.status(400).send({message: err.message})
+  }
+  for (let i = 0; i < broadcasters.length; i++) {
+    if (broadcasters[i].eth.address == req.body.id) {
+      responseObj["priceInfo"] = broadcasters[i].eth?.priceInfo
```
yondonfu: priceInfo is used as the field name here, but the go-livepeer PR expects [price](https://github.com/livepeer/go-livepeer/pull/1736/files#diff-7157f6a8ee36dc4ada0f69ca25c9b79cbbae048dba0b4dde96270ad400d900eeR326).

packages/api/src/controllers/broadcaster.js
```diff
@@ -59,4 +59,23 @@ app.get('/status', authMiddleware({}), async (req, res, next) => {
   res.json(statuses)
 })
 
+async function discoveryAuthWebhookHandler(req, res) {
+  let responseObj = {}
+  let broadcasters = []
+  try {
+    broadcasters = await req.getBroadcasters(req)
```
yondonfu: The dependent on the broadcasters that the API is aware of right now results in a lot of duplication when the broadcasters use the same ETH address and price. I wonder if starting the server with `LP_PRICES` variable that accepts an array of objects of the form `{ethAddress: ..., price: }` would help.

packages/api/src/controllers/broadcaster.js
```diff
@@ -59,4 +59,23 @@ app.get('/status', authMiddleware({}), async (req, res, next) => {
...
+  for (let i = 0; i < broadcasters.length; i++) {
+    if (broadcasters[i].eth.address == req.body.id) {
+      responseObj["priceInfo"] = broadcasters[i].eth?.priceInfo
+      break
+    }
+  }
+  return res.status(200).json(responseObj)
+}
+
+app.get('/hook', discoveryAuthWebhookHandler)
```
yondonfu: The `/broadcaster` endpoints are for fetching and querying broadcasters. Should this instead be a part of the `/orchestrator` endpoints which are for fetching orchestrators? Maybe something like `/orchestrator/auth`?

kyriediculous: Fixed in 2baa3f4

packages/api/src/controllers/broadcaster.js
kyriediculous: Fixed in 2baa3f4 , also made the necessary changes in https://github.com/livepeer/livepeer-infra/pull/390

packages/api/src/controllers/broadcaster.js
kyriediculous: fixed in the go-livepeer PR

packages/api/src/controllers/orchestrator.js
```diff
@@ -11,4 +11,28 @@ const getOrchestrators = async (req, res, next) => {
...
+      pixelsPerUnit: ""
+    }
+  }
+  let prices = []
+  try {
+    prices = req.prices()
+  } catch (err) {
+    return res.status(400).send({message: err.message})
+  }
+  for (let i = 0; i < broadcasters.length; i++) {
```
yondonfu: `broadcasters` -> `prices`?

packages/api/src/controllers/orchestrator.js
```diff
@@ -11,4 +11,28 @@ const getOrchestrators = async (req, res, next) => {
 app.get("/", getOrchestrators);
 app.get("/ext/:token", getOrchestrators);
 
-export default app;
+async function discoveryAuthWebhookHandler(req, res) {
+  const responseObj = {
+    priceInfo: {
```
yondonfu: If the priceInfo response is always populated with empty strings, will the O treat this as meaning pricePerUnit = 0 and pixelsPerUnit = 0? I know that the O uses the omitempty label for the priceInfo field in the JSON response, but since we have an actual object here by default will the O accept it instead of treating it as nil?

yondonfu (changes requested):

packages/api/src/controllers/orchestrator.js
kyriediculous: You're right, omitting it from the object by default would be better. fixed in a37855d

packages/api/src/controllers/orchestrator.js
kyriediculous: fixed in a37855d

yondonfu (approved):
LGTM

yondonfu (changes requested):
Looks like tests are failing?

kyriediculous:
I fixed up some tests but the remaining failing ones also fail on master for me... investigating further.

yondonfu:
The latest commits on master in this repo pass - maybe need to pick up the latest changes on master?

kyriediculous:
Looks like I also still had a commit needed to be pushed up , tests pass now.

yondonfu (approved):

PR #219, opened 2021-01-19
fix hot reloading

adamsoffer:
Serving the frontend over a proxy (port 3004) while in dev mode currently breaks Next.js hot reloading which makes frontend development a pain. This PR fixes that issue by making API requests using an absolute url while in dev mode and accessing the frontend over the default next port (3000).


iameli (approved):
We discussed changing `0.0.0.0` to `localhost` but other than that LGTM

PR #230, opened 2021-01-21
add global usage history endpoint

### LINKED ISSUE #220: Add API endpoint that returns usage over time
The endpoint should accept a date range and return the total minutes transcoded per day over that range.

adamsoffer:
This PR introduces an endpoint for fetching global daily usage. It also introduces a cron job that updates a cache of daily usage history every hour.

Closes #220

### DIFF
--- a/packages/api/src/controllers/usage.js
+++ b/packages/api/src/controllers/usage.js
@@ -4,6 +4,18 @@ import { db } from "../store";
 const app = Router();
 
 app.get("/", async (req, res) => {
+  if (!req.headers.authorization) {
+    res.status(403);
+    return res.json({ errors: ["unauthorized"] });
+  }
+
+  let token = req.headers.authorization.split(" ")[1]; // Bearer <token>
+
+  if (process.env.LP_API_ADMIN_TOKEN != token) {
+    res.status(403);
+    return res.json({ errors: ["unauthorized"] });
+  }
+
   let { fromTime, toTime } = req.query;
 
   // if time range isn't specified return all usage
@@ -25,14 +37,14 @@ app.get("/", async (req, res) => {
 });
 
 app.post("/update", async (req, res) => {
-  let { apiToken } = req.body;
-
-  if (!apiToken) {
+  if (!req.headers.authorization) {
     res.status(403);
-    return res.json({ errors: ["missing api token"] });
+    return res.json({ errors: ["unauthorized"] });
   }
 
-  if (process.env.LP_API_TOKEN != req.body.apiToken) {
+  let token = req.headers.authorization.split(" ")[1]; // Bearer <token>
+
+  if (process.env.LP_API_ADMIN_TOKEN != token) {
     res.status(403);
     return res.json({ errors: ["unauthorized"] });
   }
--- a/packages/api/src/controllers/usage.js
+++ b/packages/api/src/controllers/usage.js
@@ -50,13 +50,11 @@ app.post("/update", async (req, res) => {
 
   // get last updated date from cache
   if (cachedUsageHistory.length) {
-    let lastUpdatedDate =
-      cachedUsageHistory[cachedUsageHistory.length - 1].date;
-    fromTime = lastUpdatedDate + 86400000; // the day after the last updated date
+    fromTime = cachedUsageHistory[cachedUsageHistory.length - 1].date;
   }
 
-  // get all usage up until midnight yesterday
-  toTime = new Date(new Date().setUTCHours(0, 0, 0, 0)).getTime(); // midnight yesterday
+  // get all usage up until now
+  toTime = new Date().getTime();
 
   let usageHistory = await db.stream.usageHistory(fromTime, toTime, {
     useReplica: true,
@@ -64,7 +62,12 @@ app.post("/update", async (req, res) => {
 
   // store each day of usage
   for (const row of usageHistory) {
-    await req.store.create({ kind: "usage", ...row });
+    // if row already exists in cache, update it, otherwise create it
+    if (cachedUsageHistory.find((c) => c.id === row.id)) {
+      await req.store.replace({ kind: "usage", ...row });
+    } else {
+      await req.store.create({ kind: "usage", ...row });
+    }
   }
 
   res.status(200);
--- a/packages/api/src/controllers/usage.js
+++ b/packages/api/src/controllers/usage.js
@@ -33,17 +33,13 @@ app.post(
     let fromTime = +new Date(2020, 0); // start at beginning
     let toTime = +new Date();
 
-    let cachedUsageHistory = await db.stream.cachedUsageHistory(
-      fromTime,
-      toTime,
-      {
-        useReplica: true,
-      }
-    );
+    let lastUpdatedRow = (
+      await db.usage.find({}, { limit: 1, order: "data->>'date' DESC" })
+    )[0];
 
     // get last updated date from cache
-    if (cachedUsageHistory.length) {
-      fromTime = cachedUsageHistory[cachedUsageHistory.length - 1].date;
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/usage.js
```diff
@@ -0,0 +1,89 @@
...
+
+const app = Router();
+
+app.get("/", async (req, res) => {
+  if (!req.headers.authorization) {
+    res.status(403);
+    return res.json({ errors: ["unauthorized"] });
+  }
+
+  let token = req.headers.authorization.split(" ")[1]; // Bearer <token>
```
yondonfu: Could we use [auth middleware](https://github.com/livepeer/livepeer-com/blob/master/packages/api/src/middleware/auth.js) here instead of implementing additional auth logic? [Example](https://github.com/livepeer/livepeer-com/pull/152/files#diff-d6816b6ed28204167b664833a39078434b25ba17d880f0c0ff91f643aca75df1R36).

packages/api/src/store/stream-table.ts
```diff
@@ -1,25 +1,158 @@
...
+    res = await this.db.queryWithOpts(q2, opts);
+
+    if (res.rowCount > 0) {
+      for (const row of res.rows) {
+        let existingRow = usage.filter((r) => r.day === row.day);
+        if (existingRow.length) {
+          let index = usage.findIndex((r) => r.day === row.day);
+          usage[index].sourceSegments += +row.sourcesegments;
+          usage[index].transcodedSegments += +row.transcodedsegments;
+          usage[index].sourceSegmentsDuration += row.sourcesegmentsduration;
```
yondonfu, line 138: IIUC the + operator converts the following var into a numeric expression. If so, why don't we need it for sourcesegmentsduration and transcodedsegmentsduration and streamcount?

packages/api/src/store/stream-table.ts
```diff
@@ -1,25 +1,158 @@
...
+      AND (data->>'createdAt')::bigint >= ${fromTime}
+      AND (data->>'createdAt')::bigint < ${toTime}
+      GROUP BY day
+      ORDER BY day
+    `;
+    res = await this.db.queryWithOpts(q2, opts);
+
+    if (res.rowCount > 0) {
+      for (const row of res.rows) {
+        let existingRow = usage.filter((r) => r.day === row.day);
```
yondonfu: This seems like it's going to end up being a pretty expensive operation. If there are N rows for query q1 and M rows for query q2, we'll then have M for loop iterations and in each of those loop iterations we'll have N iterations for the filter function (assuming that the filter function works something like that).

Could this be avoided by using a map to track known days when we create the usage array based on the results of query q1 and then have a single for loop here that checks if the ID of one of these rows is in the map with known IDs? Ex.

```javascript
res = await this.db.queryWithOpts(q1, opts)

knownDays = {}
if (res.rowCount > 0) {
    for (const row of res.rows) {
        ...
        knownDays[row.day] = ... // index of usage object in usage array
    }
}

...
res = await this.db.queryWithOpts(q2, opts)

if (res.rowCount > 0) {
    for (const row of res.rows) {
        if (knownDays[row.day]) {
            // Add usage data to existing usage object
        } else {
            // Create new usage object
        }
    }
}
```

packages/api/src/controllers/usage.js
```diff
@@ -0,0 +1,89 @@
...
+
+  if (process.env.LP_API_ADMIN_TOKEN != token) {
+    res.status(403);
+    return res.json({ errors: ["unauthorized"] });
+  }
+
+  let fromTime = +new Date(2020, 0); // start at beginning
+  let toTime = +new Date();
+
+  let cachedUsageHistory = await db.stream.cachedUsageHistory(
```
yondonfu: Instead of having to run a query that fetches all rows since the beginning for each update could it make sense to have another query + function that just returns the last update time?

packages/api/src/controllers/usage.js
```diff
@@ -0,0 +1,89 @@
...
+
+  // get all usage up until now
+  toTime = new Date().getTime();
+
+  let usageHistory = await db.stream.usageHistory(fromTime, toTime, {
+    useReplica: true,
+  });
+
+  // store each day of usage
+  for (const row of usageHistory) {
```
yondonfu: Do we need to loop through all rows?

Seems like the only case where a row might already exist in the cache is for the first row returned since there could be overlap between the last day in the cache and the first day in the new usage history rows?

packages/api/src/store/stream-table.ts
adamsoffer: For some reason [postgres returns sums as strings](https://stackoverflow.com/questions/18754007/sums-return-string-only-with-postgresql). I tried casting the result of the sum but it always returns a string 🤷‍♂️

packages/api/src/controllers/usage.js
adamsoffer: I think so. If the cache hasn't been updated in a few days for some reason (ie the cron job is down) then we'll want to create rows for each day beginning at the last updated date when the cron job resumes.

packages/api/src/controllers/usage.js
adamsoffer: I ended up using `authMiddleware({ anyAdmin: true })`. For some reason I can't get `authMiddleware({ admin: true })` to pass authentication when sending a test post request with the api token. I think `anyAdmin` is okay though.

packages/api/src/store/stream-table.ts
```diff
@@ -1,25 +1,161 @@
...
+      GROUP BY day
+      ORDER BY day
+    `;
+
+    res = await this.db.queryWithOpts(q2, opts);
+
+    if (res.rowCount > 0) {
+      for (const row of res.rows) {
+        if (knownDays[row.day]) {
+          let index = usage.findIndex((r) => r.day === row.day);
```
yondonfu: If we store the index of the usage object in the `usage` array as the value in the `knownDays` map I think we can avoid this `findIndex()` call.

```javascript
// Loop for q1
for (...) {
    usage.push({ ... })
    knownDays[row.day] = usage.length - 1
}

// Loop for q2
for (...) {
    if (knownDays[row.day]) {
        let index = knownDays[row.day]
        ...
    }
}
```

packages/api/src/controllers/usage.js
yondonfu: Ah I think I misread this before. We loop through all the rows in usageHistory because we need to create corresponding rows in the usage table. And if the fromTime matches the date of an existing row in the usage table then that is the one scenario where we replace the row with the new row instead of creating a row. Got it.

packages/api/src/controllers/usage.js
yondonfu: Hm looks like `admin: true` will require a JWT token and not just an API key.  But, in both cases the associated user needs to be an admin. I also see we use `anyAdmin: true` [elsewhere](https://github.com/livepeer/livepeer-com/blob/8d972764f7f0d9005a885600845ab569ab9b9450/packages/api/src/controllers/admin.js#L16) as well so using that here seems reasonable.

yondonfu (approved):
LGTM

adamsoffer:
@yondonfu great. I'm going to merge into master and monitor the usage table that gets created on staging to see if the cron job does its thing.

adamsoffer:
@yondonfu the cron job runs but postgres [throws the following error](https://github.com/livepeer/livepeer-com/runs/1798872227?check_suite_focus=true) after about 30 seconds while running the update method: `"error: canceling statement due to conflict with recovery`.

I think we need to increase the timeout for queries inside `postgresql.conf` according to [this](https://stackoverflow.com/a/47661313/352241).

PR #235, opened 2021-01-27
Add copy stream key button

### LINKED ISSUE #174: Add a quick copy icon next to the streamKey on a Dashboard stream page
livepeer.com/app/stream/`<stream-id>`

Right now we show a quick copy icon next to the rtmp ingest and playback URLs. 

Let's also show the quick copy icon next to the stream key once it is revealed. Only reveal the icon after the "Show secret stream key" button has been clicked.

Use the same icon that is used in the ingest and playback URLs list below.

julianbenegas:
closes #174


adamsoffer (approved):
Looks good 💯

wohlner:
Looks good to me as well!

wohlner (approved):

PR #236, opened 2021-01-27
Show recording UI for regular users on staging

darkdarkdragon:
Shows recording UI for regular users but only on staging

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/hooks/use-api.tsx
+++ b/packages/www/hooks/use-api.tsx
@@ -9,6 +9,7 @@ import {
   Webhook,
 } from "@livepeer.com/api";
 import qs from "qs";
+import { isStaging, isDevelopment } from "../lib/utils";
 
 /**
  * Primary React API client. Definitely a "first pass". Should be replaced with some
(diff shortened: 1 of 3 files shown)


packages/www/hooks/use-api.tsx
```diff
@@ -600,7 +598,7 @@ const makeContext = (state: ApiState, setState) => {
           order,
           limit,
           cursor,
-          filters: f
+          filters: f,
```
iameli, line 601: I think you might have autoformatted off of an old version of the repo, or Prettier - @adamsoffer and I reformatted everything to use their default settings while you were gone

iameli (approved):
LGTM. Tiny formatting thing.

packages/www/hooks/use-api.tsx
darkdarkdragon: @iameli this change done according to the new formatting rules - I don't see what's the problem?

PR #245, opened 2021-02-02
fix usage bug

adamsoffer:
This PR fixes a bug in the usage update endpoint causing the usage table to skip over the month of Novemeber.


yondonfu (approved):
LGTM

PR #271, opened 2021-02-12
Update stream endpoint to support a userId param

### LINKED ISSUE #256: Scope `/api/stream` to a specific user
Return all streams by a user. Should do the same thing that `/api/stream/user/<userId>` currently does.

The `200` response for the new endpoint and `/api/stream/user/<userId>` should be the same

This may be relevant https://github.com/livepeer/livepeer-com/blob/master/packages/api/src/schema/schema.yaml

### LINKED ISSUE #257: Update Livepeer.com dashboard stream list to use `/api/stream?userId=*`
https://livepeer.com/app/user

### LINKED ISSUE #258: Confirm Livepeer.com admin streams list still works
/app/admin/streams

julianbenegas:

- `/api/stream` now supports `userId` query param, that filters streams by that userId.
- User streams table now hits `/api/stream`. The old endpoint remains unchanged for backwards compatibility.

- closes #256 
- closes #257
- closes #258

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -116,8 +116,16 @@ app.get("/", authMiddleware({}), async (req, res) => {
   }
 
   const query = parseFilters(fieldsMap, filters);
-  if (userId || !all || all === "false") {
+  if (!all || all === "false" || !req.user.admin) {
     query.push(sql`stream.data->>'deleted' IS NULL`);
+  }
+  if (req.user.admin) {
+    if (nonLivepeerOnly && nonLivepeerOnly !== "false") {
+      query.push(sql`users.data->>'email' NOT LIKE '%livepeer%'`);
+    }
+  }
+  if (active && active !== "false") {
+    query.push(sql`stream.data->>'isActive' = 'true'`);
   }
   if (streamsonly && streamsonly !== "false") {
     query.push(sql`stream.data->>'parentId' IS NULL`);
@@ -126,13 +134,6 @@ app.get("/", authMiddleware({}), async (req, res) => {
   }
   if (userId) {
     query.push(sql`stream.data->>'userId' = ${userId}`);
-  } else {
-    if (active && active !== "false") {
-      query.push(sql`stream.data->>'isActive' = 'true'`);
-    }
-    if (nonLivepeerOnly && nonLivepeerOnly !== "false") {
-      query.push(sql`users.data->>'email' NOT LIKE '%livepeer%'`);
-    }
   }
 
   order = parseOrder(fieldsMap, order);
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/stream.js
```diff
@@ -101,11 +101,19 @@ app.get("/", authMiddleware({ admin: true }), async (req, res) => {
     nonLivepeerOnly,
     order,
     filters,
+    userId,
   } = req.query;
   if (isNaN(parseInt(limit))) {
     limit = undefined;
   }
 
+  if (userId && req.user.admin !== true && req.user.id !== userId) {
```
darkdarkdragon: This whole endpoint is accessible only for admins (`authMiddleware({ admin: true })`)
So you should remove `{admin: true}` to make it accessible to regular users.

darkdarkdragon (changes requested):

packages/api/src/controllers/stream.js
wohlner: Great! Glad we can easily open up this endpoint to regular users. I'm sure you are already doing this, but I want to make sure that a regular users only has access to their own streams, not everyones. Only admins should see all users' streams.

julianbenegas: Yup, that is handled.

darkdarkdragon (changes requested):
[unit test](https://github.com/livepeer/livepeer-com/blob/f2a421f962144babc46e8e944b6a3b50377439c6/packages/api/src/controllers/stream.test.js#L380) failing

julianbenegas:
> [unit test](https://github.com/livepeer/livepeer-com/blob/f2a421f962144babc46e8e944b6a3b50377439c6/packages/api/src/controllers/stream.test.js#L380) failing

thanks @darkdarkdragon, will fix

packages/api/src/controllers/stream.test.js
```diff
@@ -373,8 +373,8 @@ describe("controllers/stream", () => {
       expect(res.status).toBe(403);
     });
 
-    it("should not get all streams for admin user", async () => {
-      client.apiKey = adminApiKey;
+    it("should not get all streams for non admin user", async () => {
```
julianbenegas, line 376: @darkdarkdragon changed the test to test for `non admin user`.

It's a bit weird, the `authMiddleware({ admin: true })` creates the need for a JWT, but the endpoint as is doesn't require it. Do you think I should update the permissions for it to require a JWT? 🤔

darkdarkdragon: > Do you think I should update the permissions for it to require a JWT?

No, there is no need.

darkdarkdragon (changes requested):
`all` and `nonLivepeerOnly` arguments should be ignored for non-admin users

wohlner:
@julianbenegas This is now deployed on stage to livepeer.monster. I just tested a `GET` request to https://livepeer.monster/api/stream using a token for a non-admin .monster account and got this error `{"errors":["user can only request information on their own streams"]}`. 

And then when I made the same GET request using an admin .monster token, the request response is super slow/doesn't work. I got a 524 error, the response timed out. 

Both these `GET` requests to `/api/stream` should return a list of all the API key holder's `stream` objects.

cc:@darkdragon#2215

PR #275, opened 2021-02-18
/api/stream for non-admin return only own streams

darkdarkdragon:
/api/stream for non-admin return only own streams

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -107,12 +107,8 @@ app.get("/", authMiddleware({}), async (req, res) => {
     limit = undefined;
   }
 
-  if (req.user.admin !== true && (!userId || req.user.id !== userId)) {
-    // Not admin, and requesting all streams, or streams that don't belong to the user.
-    res.status(403);
-    return res.json({
-      errors: ["user can only request information on their own streams"],
-    });
+  if (!req.user.admin) {
+    userId = req.user.id;
   }
 
   const query = parseFilters(fieldsMap, filters);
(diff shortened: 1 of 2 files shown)


iameli (changes requested):
Minor change.

darkdarkdragon:
@iameli Github's interface doesn't show any suggested change 😞

wohlner:
Can I test this with a `curl` command or something in my terminal before it goes to stage? Or do I need to wait until then. 

Also, one implementation question. This change means that is you make the `api/stream` request with Livepeer.com admin's token and without a `userId` specified, you will get ALL streams by ALL users returned, correct?

darkdarkdragon:
@wohlner 
> Can I test this with a curl command or something in my terminal before it goes to stage? 

No. Only way is to start local dev env and test locally.

> This change means that is you make the api/stream request with Livepeer.com admin's token and without a userId specified, you will get ALL streams by ALL users returned, correct?

Yes. Same as it worked before.

wohlner:
@darkdarkdragon in that case, I'm looking forward to testing when it is on .monster.

packages/www/hooks/use-api.tsx
```diff
@@ -520,7 +520,7 @@ const makeContext = (state: ApiState, setState) => {
 
     async getStreams(userId): Promise<Array<Stream>> {
       const [res, streams] = await context.fetch(
-        `/stream?userId=${userId}&streamsonly=1`
+        `/stream?streamsonly=1`
```
iameli: I think this should still have `userId` — I'm an admin but I still want my "streams" page to only show mine.

darkdarkdragon: Sure! Missed that one 😢

PR #339, opened 2021-04-01
show URL to playback from broadcaster

darkdarkdragon:
Show URL for playing back stream from our broadcaster
Correctly set active status for streams that not passed through our Mist server
(will show active PlayDj streams)

## -

Manually

**Screenshots (optional):**
image

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -626,6 +626,7 @@ app.post("/", authMiddleware({}), validatePost("stream"), async (req, res) => {
     streamKey,
     playbackId,
     createdByTokenName: req.tokenName,
+    createdByTokenId: req.tokenId,
     isActive: false,
     lastSeen: 0,
   });
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -778,7 +778,7 @@ app.put("/:id/setactive", authMiddleware({}), async (req, res) => {
   });
 
   if (stream.parentId) {
-    const pStream = await db.stream.get(id);
+    const pStream = await db.stream.get(stream.parentId);
     if (pStream && !pStream.deleted) {
       await db.stream.update(pStream.id, {
         isActive: stream.isActive,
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/stream.js
```diff
@@ -62,7 +61,7 @@ function isActuallyNotActive(stream) {
 
 function activeCleanupOne(stream) {
   if (isActuallyNotActive(stream)) {
-    setActiveToFalse(stream);
+    db.stream.setActiveToFalse(stream);
```
iameli, line 64: `await` here?

packages/www/pages/app/stream/[id].tsx
```diff
@@ -193,6 +213,21 @@ const ID = () => {
...
+  if (!region && lastSession && lastSession.region) {
+    region = lastSession.region;
+  }
+  let broadcasterPlaybackUrl;
+  const domain = isStaging() ? "monster" : "com";
+  if (stream && stream.region && stream.broadcasterHost) {
+    broadcasterPlaybackUrl = `https://${stream.region}.livepeer.${domain}/stream/${stream.id}.m3u8`;
+  } else if (lastSession && lastSession.region) {
+    broadcasterPlaybackUrl = `https://${lastSession.region}.livepeer.${domain}/stream/${stream.playbackId}.m3u8`;
+  }
```
Comment (iameli) [inline packages/www/pages/app/stream/[id].tsx:230]: Hmmmmm. It seems like there should be a better way to do this... it would be pretty useful for us to return actual playback URLs from the API. Not something we have to do now but something to keep thinking about.

iameli (approved):
I like this a lot. One small change, just a place I'm not sure why there's not an `await`.

It seems like "get me the playback URL for stream `abc123` in region 'mdw'" and "stop all RTMP sessions for stream `abc123` in region `mdw`" could maybe be handled through the same mechanism - we want to teach these API nodes about other API nodes and how to contact them.

packages/api/src/controllers/stream.js
darkdarkdragon: We don't need result of this function and we don't want to wait for it to complete - it can execute in parallel.

PR #354, opened 2021-04-11
Suspend streams

### LINKED ISSUE #327: Suspend session for an Livepeer.com active stream and block new sessions
This new feature will allow a .com user to stop an ongoing stream. This should end any in progress session and prevent new sessions from being started with the same stream key. It will also allow a Livepeer admin to stop ANY active stream, and add it into the admin interface. 

### Requirements:
1. New boolean parameter on the parent stream object `suspend`. The default is `false`, and it should be hidden by default in the `stream` object response.

2. When `suspend` is set to `true`, end any in progress session and prevent new sessions from being started with the same stream key.

3. The `suspend` parameter does not need to exist on "child" `stream` objects (sessions).

4. In the Livepeer.com dashboard, on the stream page we will add a checkbox Recording with the label `Suspend and block`. 
   -  By default this will be unchecked. 
   - When a user checks the box, display a pop-up saying "Are you sure you want to suspend and block this stream? Any active stream sessions will immediately end. New sessions will be prevented from starting until unchecked." The user will need to click the "Confirm" button before the new setting takes effect.
   - When a user unchecks the box, display a pop-up saying "Are you sure you want to allow new stream sessions again?" The user will need to click the "Confirm" button before the new setting takes effect.

5. For Livepeer.com admins we will also need to display the `Terminate and block` checkbox on ALL stream pages.

darkdarkdragon:
Suspend streams
Terminate running live streams

@wohlner - I've named field `suspended` instead of `suspend` - it is more in line with other fields like `deleted`. But I can rename it to `suspend` if needed.

Manually

Fixes #327

**Screenshots (optional):**
Suspend checkbox:
image
---
Confirm dialog 1
image
---
Confirm dialog 2
image
---

Admin-only button that will just stop running stream without suspending
image

---
'Suspended' column on admin's streams page
image

### DIFF
--- a/packages/api/src/controllers/mist-api.js
+++ b/packages/api/src/controllers/mist-api.js
@@ -1,0 +1,110 @@
+import logger from "../logger";
+import fetch from "isomorphic-fetch";
+import crypto from "crypto";
+import querystring from "querystring";
+
+let challengeResponse;
+let challenge;
+
+function hashChallenge(authChallenge, password) {
+  const passHashed = crypto.createHash("md5").update(password).digest("hex");
+  challengeResponse = crypto
+    .createHash("md5")
+    .update(passHashed + authChallenge)
+    .digest("hex");
+  challenge = authChallenge;
+}
+
+export async function listActiveStreams(mistHost, mistPort, login, password) {
+  try {
+    const req = {
+      active_streams: 1,
+    };
+    if (challengeResponse) {
+      req.authorize = {
+        username: login,
+        password: challengeResponse,
+      };
+    }
+    const params = {
+      method: "POST",
+      headers: {
+        "content-type": "application/json",
+      },
+      body: querystring.encode({ command: JSON.stringify(req) }),
+    };
+    logger.info(
+      `requesting active streams list with request ${JSON.stringify(
+        req,
+        null,
+        2
+      )}`
+    );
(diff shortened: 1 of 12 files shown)


wohlner (changes requested):
Looks great! A couple comments

- [ ] On /app/admin/streams, the suspended column sort doesn't seem to work. Also you can change 
- [ ] Testing on the Livepeer.com dashboard preview link with one of my streams for the stream with `id a38b0743-29bf-48b0-b2f0-9c7db17ded38`: When I change the "suspend and block" from off to on, I am shown the confirmation pop up. When I click confirm, the pop-up goes away, but the radio button value is still "off".
- [ ] The "Terminate" button on the admin view for a single stream is a great addition. Good thinking! Please move it in line and directly to the left of the "delete" button.

darkdarkdragon:
moved 'terminate' button
image

darkdarkdragon:
@wohlner it is not working because we need to merge to master to deploy server-side changes. I'm merging this so we can test on staging.

wohlner:
Can you please add a tooltip next to Off with this copy: "When turned on, any active stream sessions will immediately end. New sessions will be prevented from starting until turned off."

darkdarkdragon:
> When turned on, any active stream sessions will immediately end. New sessions will be prevented from starting until turned off.

@wohlner deployed to staging

PR #374, opened 2021-04-25
new 'session' table

### LINKED ISSUE #361: Add new 'session' table
## Problem

Currently we have `sessions` which is basically same stream object but with `parentId` set. This `sessions` objects created by Mist when it starts transcoding. 
But recently we introduced concept of `user session` - which internally consists of number of `regular` `sessions`. Currently `user session` object gets constructed on the fly from `regular session` objects. This is inefficient and complicates code.

## Solution

Create new `session` table which will track properties of `user` session.

darkdarkdragon:
Adds new 'session' table. This table aggregates 'user' session into one object.
New endpoint /api/session introduced
Sessions list on stream page uses this new endpoint.

Manually

Fixes #361 

**Screenshots (optional):**

**Checklist:**


wohlner (approved):
I appreciate you tagging me to review. I think I need to wait to test until this is on stage. I don't want to be a blocker though.

Issue #378, opened 2021-04-28
Postgresql Message Queue adapter


ya7ya:
We'd like to use Postgresql as a Msg Queue for the webhooks
details: https://stackoverflow.com/questions/13005410/why-do-we-need-message-brokers-like-rabbitmq-over-a-database-like-postgresql/13005551#13005551

The adapter needs to mimic a Msg queue interface
- [ ] emit to channel 
- [ ] channel handler

darkdarkdragon:
@ya7ya 
1. Any chance we could use CockroachDb for this? It has PostgreSQL wire-compatible protocol, so from our side it will be almost the same. But that way we'll have a chance to get hands-on experience with CochroachDb - with the intention to replace centralized PostgresSQL with it eventually.
2. Also it would be good for queue to have 'postponed' messages - like 'that message should be processed after 5 min after creation'. That would help with processing 'recording status' of the session.

ya7ya:
@darkdarkdragon 
1. Yeah the next logical step is to use cockroachDB but when I discussed it with the team it seems this would be out of scope for this Epic and it'd mean there is some DevOps work needed to get it off the ground unlike Postgres which we already use in Prod. 

I do agree that CockroachDB would be an overall better match as a datastore for the livepeer-api 👍 

2. Yes this mechanism needs to be implemented either way to allow for a backOff strategy in case the client's server doesn't respond with a 2xx, one way of doing that is to have the wait time as a parameter when firing and it defaults to 0.

PR #390, opened 2021-05-11
webhook Cannon

ya7ya:

implements #378 

- create the queue object
- add the methods for `emitting` and `popping` an event

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -85,6 +85,46 @@ components:
           description: Timestamp (in milliseconds)
           example: 1587667174725
 
+    webhookResponse:
+      type: object
+      table: webhookResponse
+      required:
+        - webhookId
+        - eventId
+        - statusCode
+      properties:
+        id:
+          type: string
+          readOnly: true
+          example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+        kind:
+          type: string
+          readOnly: true
+          example: webhookResponse
+        webhookId:
+          type: string
+          readOnly: true
+          index: true
+        eventId:
+          type: string
+          readOnly: true
+          index: true
+        createdAt:
+          type: number
+          readOnly: true
+          description:
+            Timestamp (in milliseconds) at which stream object was created
+          example: 1587667174725
+        statusCode:
+          type: number
+          default: 0
+        responseObject:
+          type: object
+          additionalProperties: true
+          properties:
+            body:
+              type: string
+
     webhook:
       type: object
       table: webhook
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -24,6 +24,8 @@ tags:
     description: Webhooks endpoints
   - name: queue
     description: Webhooks msg queue
+  - name: webhookResponse
+    description: webhook client responses
 
 components:
   schemas:
@@ -87,7 +89,7 @@ components:
 
     webhookResponse:
       type: object
-      table: webhookResponse
+      table: webhook_response
       required:
         - webhookId
         - eventId
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -24,7 +24,7 @@ tags:
     description: Webhooks endpoints
   - name: queue
     description: Webhooks msg queue
-  - name: webhookResponse
+  - name: webhook-response
     description: webhook client responses
 
 components:
@@ -87,7 +87,7 @@ components:
           description: Timestamp (in milliseconds)
           example: 1587667174725
 
-    webhookResponse:
+    webhook-response:
       type: object
       table: webhook_response
       required:
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -35,6 +35,10 @@ export default class QueueTable extends Table<Queue> {
     }
   }
 
+  unsetMsgHandler() {
+    this.setMsgHandler(QueueTable.prototype.handleMsg);
+  }
+
   async listen() {
     this.listener = await this.db.pool.connect();
     this.listener.query(`LISTEN ${this.channel}`);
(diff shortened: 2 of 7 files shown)


packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,71 @@
...
+    super({ db, schema });
+    this.db = db;
+    this.schema = schema;
+    this.name = schema.queue;
+  }
+
+  // get next event in queue
+  async pop(opts: GetOptions = { useReplica: true }): Promise<DBObject> {
+    let res: QueryResult<DBLegacyObject>;
+    if (!opts.useReplica) {
```
iameli: Don't think we can probably `useReplica` here, cuz we're locking stuff

packages/api/src/schema/schema.yaml
```diff
@@ -24,6 +24,8 @@ tags:
     description: Webhooks endpoints
   - name: queue
     description: Webhooks msg queue
+  - name: webhookResponse
```
iameli: Probably `webhook-response`

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (newInterval > MAX_BACKOFF) {
+      return lastInterval;
+    }
+    return newInterval;
+  }
+
+  retry(event) {
+    event.lastInterval = this.calcBackoff(event.lastInterval);
+    event.status = "pending";
+    event.reteries = event.reteries ? event.reteries + 1 : 1;
```
iameli: `retries`

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.db.queue.updateMsg(event);
+  }
+
+  async storeResponse(webhook: Webhook, event: Queue, resp: Response) {
+    // TODO
+    // store response for each time a webhook fires.
+    await this.db.webhookResponse.create({
```
iameli, line 154: It would be awesome to have a `duration` field on here too

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+      webhookId: webhook.id,
+      eventId: event.id,
+      statusCode: resp.status,
+      response: resp,
+    });
+  }
+
+  async onTrigger(event: Queue) {
+    if (!event) {
+      return;
```
iameli, line 167: Could throw an error here

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+      event.userId,
+      event.event
+    );
+
+    let stream = await this.db.stream.get(event.streamId);
+    // basic sanitization.
+    let sanitized = { ...stream };
+    delete sanitized.streamKey;
+
+    let user = await this.store.get(`user/${event.userId}`);
```
iameli: `this.db.user.get`

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+            } catch (e) {
+              console.error("isLocal Error", isLocal, e);
+              throw e;
+            }
+          }
+          if (isLocal) {
+            // don't fire this webhook.
+            console.log(
+              `webhook ${webhook.id} resolved to a localIP, url: ${webhook.url}, resolved IP: ${ips}`
+            );
```
iameli: Can put a `throw` here and lose the `else` statement

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+                // 2xx requests are cool.
+                // all is good
+                logger.info(`webhook ${webhook.id} fired successfully`);
+                this.storeResponse(webhook, event, resp);
+                return true;
+              }
+              console.error(
+                `webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`
+              );
+              this.storeResponse(webhook, event, resp);
```
iameli: `await`?

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,184 @@
...
+              // retry
+              this.retry(event);
```
iameli: Can get rid of this when we move to a single hook per

iameli (changes requested):
Per discussion, let's have each webhook entry add its own entry to the queue

darkdarkdragon:
@ya7ya @iameli 
BTW, maybe we should tie webhook (message queue in general?) processing to the region?
Firing webhook from API server in Singapore to the receiver in Latin America is not the optimal thing to do...

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,204 @@
...
+  constructor({ db, store }) {
+    this.db = db;
+    this.store = store;
+    this.running = true;
+    this.start();
+  }
+
+  async start() {
+    if (this.running) {
+      this.db.queue.setMsgHandler(this.processEvent.bind(this));
```
darkdarkdragon, line 30: We should also drain pending events at start.

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,204 @@
...
+      event.userId,
+      event.event
+    );
+
+    let stream = await this.db.stream.get(event.streamId);
+    // basic sanitization.
+    let sanitized = { ...stream };
+    delete sanitized.streamKey;
+
+    let user = await this.db.user.get(`user/${event.userId}`);
```
darkdarkdragon, line 181:
```suggestion
    let user = await this.db.user.get(event.userId);
```
Also, what would happen if user is not found?

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,204 @@
...
+    await this.db.webhookResponse.create({
+      webhookId: webhook.id,
+      eventId: event.id,
+      statusCode: resp.status,
+      response: resp,
+    });
+  }
+
+  async onTrigger(event: Queue) {
+    if (!event) {
```
darkdarkdragon, line 163: We also need to check that event structure is valid (has all required fields at least).

darkdarkdragon (changes requested):

PR #414, opened 2021-05-19
Msg Queue

ya7ya:

implements #378 

- this implements 2 main methods 

- `queue.pop()` to get the next msg
- `queue.emit()` to push a msg into the queue

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -24,6 +24,8 @@ tags:
     description: Webhooks endpoints
   - name: queue
     description: Webhooks msg queue
+  - name: webhook-response
+    description: webhook client responses
 
 components:
   schemas:
@@ -130,6 +132,45 @@ components:
             If true, returning non 2xx value from webhook will prevent stream
             from starting
 
+    webhook-response:
+      type: object
+      table: webhook_response
+      required:
+        - webhookId
+        - eventId
+        - statusCode
+      properties:
+        id:
+          type: string
+          readOnly: true
+          example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+        kind:
+          type: string
+          readOnly: true
+          example: webhookResponse
+        webhookId:
+          type: string
+          readOnly: true
+          index: true
+        eventId:
+          type: string
+          readOnly: true
+          index: true
+        createdAt:
+          type: number
+          readOnly: true
+          description:
+            Timestamp (in milliseconds) at which stream object was created
+          example: 1587667174725
+        statusCode:
+          type: number
+          default: 0
+        responseObject:
+          type: object
+          additionalProperties: true
+          properties:
+            body:
+              type: string
     stream:
       type: object
       table: stream
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -24,8 +24,6 @@ tags:
     description: Webhooks endpoints
   - name: queue
     description: Webhooks msg queue
-  - name: webhook-response
-    description: webhook client responses
 
 components:
   schemas:
@@ -33,7 +31,7 @@ components:
       type: object
       table: queue
       required:
-        - time
+        - createdAt
         - channel
         - event
         - userId
@@ -47,10 +45,11 @@ components:
           type: string
           readOnly: true
           example: queue
-        time:
+        createdAt:
           type: number
           readOnly: true
-          example: usage
+          description: Timestamp (in milliseconds)
+          example: 1587667174725
         channel:
           type: string
           readOnly: true
@@ -86,46 +85,6 @@ components:
           readOnly: true
           description: Timestamp (in milliseconds)
           example: 1587667174725
-
-    webhook-response:
-      type: object
-      table: webhook_response
-      required:
-        - webhookId
-        - eventId
-        - statusCode
-      properties:
-        id:
-          type: string
-          readOnly: true
-          example: de7818e7-610a-4057-8f6f-b785dc1e6f88
-        kind:
-          type: string
-          readOnly: true
-          example: webhookResponse
-        webhookId:
-          type: string
-          readOnly: true
-          index: true
-        eventId:
-          type: string
-          readOnly: true
-          index: true
-        createdAt:
-          type: number
-          readOnly: true
-          description:
-            Timestamp (in milliseconds) at which stream object was created
-          example: 1587667174725
-        statusCode:
-          type: number
-          default: 0
-        responseObject:
-          type: object
-          additionalProperties: true
-          properties:
-            body:
-              type: string
 
     webhook:
       type: object
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -68,6 +68,7 @@ components:
         isConsumed:
           type: boolean
           default: false
+          index: true
         retries:
           type: number
           default: 0
@@ -77,6 +78,7 @@ components:
         status:
           type: string
           default: pending
+          index: true
         modifiedAt:
           type: number
           readOnly: true
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -56,7 +56,7 @@ components:
         event:
           type: string
           readOnly: true
-          example: stream_started
+          example: stream.started
         userId:
           type: string
           readOnly: true
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -24,6 +24,8 @@ tags:
     description: Webhooks endpoints
   - name: queue
     description: Webhooks msg queue
+  - name: webhook-response
+    description: webhook client responses
 
 components:
   schemas:
@@ -84,6 +86,46 @@ components:
           readOnly: true
           description: Timestamp (in milliseconds)
           example: 1587667174725
+
+    webhook-response:
+      type: object
+      table: webhook_response
+      required:
+        - webhookId
+        - eventId
+        - statusCode
+      properties:
+        id:
+          type: string
+          readOnly: true
+          example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+        kind:
+          type: string
+          readOnly: true
+          example: webhookResponse
+        webhookId:
+          type: string
+          readOnly: true
+          index: true
+        eventId:
+          type: string
+          readOnly: true
+          index: true
+        createdAt:
+          type: number
+          readOnly: true
+          description:
+            Timestamp (in milliseconds) at which stream object was created
+          example: 1587667174725
+        statusCode:
+          type: number
+          default: 0
+        responseObject:
+          type: object
+          additionalProperties: true
+          properties:
+            body:
+              type: string
 
     webhook:
       type: object
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -22,9 +22,53 @@ tags:
     description: Changes user admin status
   - name: webhook
     description: Webhooks endpoints
+  - name: queue
+    description: Webhooks msg queue
 
 components:
   schemas:
+    queue:
+      type: object
+      table: queue
+      required:
+        - time
+        - channel
+        - event
+        - userId
+        - streamId
+      properties:
+        id:
+          type: string
+          readOnly: true
+          example: 256d47d4-d59b-4fd5-a87d-d98a68ac27ef
+        kind:
+          type: string
+          readOnly: true
+          example: queue
+        time:
+          type: number
+          readOnly: true
+          example: usage
+        channel:
+          type: string
+          readOnly: true
+          example: webhook_recording_started
+        event:
+          type: string
+          readOnly: true
+          example: stream_started
+        userId:
+          type: string
+          readOnly: true
+          index: true
+        streamId:
+          type: string
+          readOnly: true
+          index: true
+        isConsumed:
+          type: boolean
+          default: false
+
     webhook:
       type: object
       table: webhook
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -68,6 +68,20 @@ components:
         isConsumed:
           type: boolean
           default: false
+        retries:
+          type: number
+          default: 0
+        lastInterval:
+          type: number
+          default: 1
+        status:
+          type: string
+          default: pending
+        modifiedAt:
+          type: number
+          readOnly: true
+          description: Timestamp (in milliseconds)
+          example: 1587667174725
 
     webhook:
       type: object
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -24,7 +24,11 @@ export default class QueueTable extends Table<Queue> {
   }
 
   async stop() {
-    await this.client.release();
+    try {
+      await this.client.release();
+    } catch (e) {
+      console.log("QueueTable: Stop: ", e);
+    }
   }
   // name: string;
   // constructor({ db, schema }) {
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -145,6 +145,8 @@ export default class QueueTable extends Table<Queue> {
     if (!doc.status) {
       doc.status = "pending";
     }
+
+    doc.createdAt = Date.now();
     this.client = await this.db.pool.connect();
     try {
       await this.client.query("BEGIN;");
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -29,7 +29,7 @@ export default class QueueTable extends Table<Queue> {
         sql`SELECT data FROM `
           .append(this.name)
           .append(
-            sql`LIMIT 1 FOR UPDATE SKIP LOCKED`.setName(
+            sql` LIMIT 1 FOR UPDATE SKIP LOCKED`.setName(
               `${this.name}_next_event`
             )
           )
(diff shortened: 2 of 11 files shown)


darkdarkdragon (changes requested):
This requires some improvements:
1. Message should be fully processed during transaction (between `BEGIN` and `COMMIT`) - there shouldn't be `processing` status. (And whole status field could be removed - `isConsumed` is enough). See this [article](https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/) for examples.
2. It would be good to use `LISTEN/NOTIFY` to notify workers about new messages instead of constantly pooling in the loop.

packages/api/src/schema/schema.yaml
```diff
@@ -22,9 +22,67 @@ tags:
...
+          example: stream_started
+        userId:
+          type: string
+          readOnly: true
+          index: true
+        streamId:
+          type: string
+          readOnly: true
+          index: true
+        isConsumed:
```
darkdarkdragon, line 71: This should be indexed

packages/api/src/schema/schema.yaml
```diff
@@ -22,9 +22,67 @@ tags:
...
+          type: boolean
+          default: false
+        retries:
+          type: number
+          default: 0
+        lastInterval:
+          type: number
+          default: 1
+        status:
```
darkdarkdragon, line 81: This should be indexed

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,150 @@
...
+  //   super({ db, schema });
+  //   this.db = db;
+  //   this.schema = schema;
+  //   this.name = schema.queue;
+  // }
+
+  // get next event in queue
+  async pop(
+    processFunc: Function,
+    opts: GetOptions = { useReplica: true }
```
darkdarkdragon: This is not used, can we remove this?

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,150 @@
...
+    );
+
+    if (res.rowCount < 1) {
+      return null;
+    }
+
+    let originalData = res.rows[0].data;
+    // this allows us to pass a function that will process the event within the lock
+    if (processFunc) {
+      await processFunc(originalData as Queue);
```
darkdarkdragon: There is not error handling. 
I think we can wrap this call in `try/catch` and explicitly rollback transaction in `catch`

packages/api/src/schema/schema.yaml
```diff
@@ -22,9 +22,67 @@ tags:
...
+          readOnly: true
+          example: usage
+        channel:
+          type: string
+          readOnly: true
+          example: webhook_recording_started
+        event:
+          type: string
+          readOnly: true
+          example: stream_started
```
iameli: I think `stream.started` maybe

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,150 @@
...
+    try {
+      await this.db.query(
+        `INSERT INTO ${this.name} VALUES ($1, $2)`, //p
+        [doc.id, JSON.stringify(doc)] //p
+      );
+
+      if (this.listener) {
+        // emit it via notifications
+        console.log("emitting NOTIFY");
+        await this.db.query(`NOTIFY ${this.channel}, '${doc.id}';`);
```
darkdarkdragon: Ideally we should emit `NOTIFY` inside one transaction with `INSERT`
(and it doesn't matter if there is `this.listener` - it can be that some other server listen for changes)

darkdarkdragon (changes requested):
I think we should remove ability to process events outside of `BEGIN/SELECT/COMMIT`
Also, we should add a way to schedule events to be processed at specified time in the future.
Webhook cannon's retry code could use it - say, for the first 5-10 seconds of retries it could keep transaction open, but if it keeps failing - it should mark current event as processed and schedule (inside same `BEGIN/COMMIT`) new event to be fired at some time in the future.

darkdarkdragon:
@iameli @ya7ya What do you think about moving processed events to the separate 'history' table? `SELECT` for events scans all the existing events (it will use index, but index will be growing and growing...)?

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+    logger.debug(`MsgQueue emitting ${doc.id}`);
+    return doc;
+  }
+
+  async updateMsg(doc: Queue): Promise<Queue> {
+    doc.modifiedAt = Date.now();
+    await this.client.query(
+      sql`UPDATE queue SET data = data || ${doc} `.append(
+        ` WHERE id = '${doc.id}'`
```
iameli: I think we'd want this to avoid injection? Want to make sure the id is parameterized
```suggestion
        sql` WHERE id = '${doc.id}'`
```

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,155 @@
...
+  //   super({ db, schema });
+  //   this.db = db;
+  //   this.schema = schema;
+  //   this.name = schema.queue;
+  // }
+
+  // get next event in queue
+  async pop(processFunc: Function): Promise<Queue> {
+    let res: QueryResult<DBLegacyObject>;
+    await this.client.query("BEGIN");
```
iameli, line 91: With all the async functions in here I'm a little concerned about what happens when this function gets called a lot. If we call `pop()` twice without waiting for the async, that could cause problems, yeah? Can we add a test something like this to make sure it works okay?

```javascript
for (let i = 0; i < 10; i++) {
  await db.queue.emit({ id: `queue-${i}` });
}

const promises = [];
for (let i = 0; i < 10; i++) {
  promises.push(db.queue.pop());
}

const results = await Promise.all(promises);

// expect results to contain `queue-0`, `queue-1`, etc...

```

My concern is that BEGIN; is going to be called 10 times before anything else happens, right? Because `db.queue.pop()` could run multiple times before any one of the transactions completes. So the transaction might get messed up. Is there a way for us to get a separate transaction object from the pg adapter? Or maybe run everything as part of one big statement?

iameli (changes requested):
Concerned about a race condition on pop() there, see notes

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    let originalData = res.rows[0].data;
+    // this allows us to pass a function that will process the event within the lock
+    if (processFunc) {
+      try {
+        await processFunc(originalData as Queue);
+      } catch (error) {
+        console.log("Msg queue process function error: ", error);
+        await this.client.query("ROLLBACK;");
+        throw error; // throwing the error till we can handle it better
```
darkdarkdragon: This will crash whole node.

packages/api/src/store/queue.ts
darkdarkdragon: >  Is there a way for us to get a separate transaction object from the pg adapter?

Transaction is per-connection, so every `pop` call should use own connection to the db.

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,155 @@
...
+  // name: string;
+  // constructor({ db, schema }) {
+  //   super({ db, schema });
+  //   this.db = db;
+  //   this.schema = schema;
+  //   this.name = schema.queue;
+  // }
+
+  // get next event in queue
+  async pop(processFunc: Function): Promise<Queue> {
```
darkdarkdragon: Let's remove ability to use `pop` function without `processFunc`.
Also, can we type `processFunc` more specifically (not just `Function`)?

packages/api/src/store/queue.ts
darkdarkdragon: `pop` function should take new connection from the pool and use it exclusively for processing one event.

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,155 @@
...
+      return null;
+    }
+
+    let originalData = res.rows[0].data;
+    // this allows us to pass a function that will process the event within the lock
+    if (processFunc) {
+      try {
+        await processFunc(originalData as Queue);
+      } catch (error) {
+        console.log("Msg queue process function error: ", error);
```
darkdarkdragon, line 113: If there is error processing this event - same event will be picked up in the next `pop` call - and probably will give same error - this will effectively stop all the queue. Probably we should change state of such event to `error` and try it anymore. Or give it a number of tries, but we should stop at some point.

darkdarkdragon (changes requested):
1. I think we should add `createdAt` field to the event, fill it in `emit`, and order by that field in the `pop`'s select statement - that way messages will be processed in order.
2. Right now, retries made by simply waiting while processing event. This will keep transaction open for whole time and will render that connection unavailable for processing other events. Couple of unresponsive webhooks - and system will run out of connections. Way to improve on this will be to add `fireAtTime` field to event - with timestamp set to time in the future when event should fire. `pop` will select event without this field or with `fireAtTime` that is less than current time. Also, we will send content of this field in the `NOTIFY` payload - that way listener will now that there will be event in the future, so it can `setTimeout` for that duration and call `pop` in `setTimeout` handler.

packages/api/src/schema/schema.yaml
```diff
@@ -22,9 +22,111 @@ tags:
...
+          readOnly: true
+          example: 256d47d4-d59b-4fd5-a87d-d98a68ac27ef
+        kind:
+          type: string
+          readOnly: true
+          example: queue
+        time:
+          type: number
+          readOnly: true
+          example: usage
```
iameli: Should be a number?

packages/api/src/schema/schema.yaml
```diff
@@ -22,9 +22,111 @@ tags:
...
+      properties:
+        id:
+          type: string
+          readOnly: true
+          example: 256d47d4-d59b-4fd5-a87d-d98a68ac27ef
+        kind:
+          type: string
+          readOnly: true
+          example: queue
+        time:
```
iameli: Probably should be named `createdAt` for consistency

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,210 @@
...
+        console.error(
+          `webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`
+        );
+        await this.storeResponse(webhook, event, resp);
+        // retry
+        this.retry(event);
+        return;
+        // return !webhook.blocking;
+      } catch (e) {
+        console.log("firing error", e);
```
iameli, line 160: Gotta store the response in the event of a timeout; gotta test this too

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  async storeResponse(
+    webhook: Webhook,
+    event: Queue,
+    resp: Response,
+    duration = 0
+  ) {
+    // TODO
```
iameli: Donezo, can be removed

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,210 @@
...
+      event.userId,
+      event.event
+    );
+
+    let stream = await this.db.stream.get(event.streamId);
+    // basic sanitization.
+    let sanitized = { ...stream };
+    delete sanitized.streamKey;
+
+    let user = await this.db.user.get(`user/${event.userId}`);
```
iameli:
```suggestion
    let user = await this.db.user.get(event.userId);
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,210 @@
...
+      );
+    }
+
+    try {
+      const responses = await Promise.all(
+        webhooksList.map(async (webhook, key) => {
+          try {
+            await this._fireHook(event, webhook, sanitized, user);
+          } catch (error) {
+            console.log("_fireHook Error", error);
```
iameli, line 238: Need to retry here

iameli (changes requested):

packages/api/src/store/queue.ts
```diff
@@ -0,0 +1,193 @@
...
+  //   super({ db, schema });
+  //   this.db = db;
+  //   this.schema = schema;
+  //   this.name = schema.queue;
+  // }
+
+  // get next event in queue
+  async pop(processFunc: (q: Queue) => any): Promise<Queue> {
+    let res: QueryResult<DBLegacyObject>;
+    this.client = await this.db.pool.connect();
```
darkdarkdragon, line 90: It shouldn't be assigned to the object field, `client` should be local variable.

darkdarkdragon (changes requested):
Cannon still uses `setTimeout` for retries - this will not work for any amount of load.

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,253 @@
...
+    const { data: webhooksList } = await this.getWebhooks(
+      event.userId,
+      event.event
+    );
+
+    console.log("webhooks : ", webhooksList);
+    let stream = await this.db.stream.get(event.streamId);
+    // basic sanitization.
+    let sanitized = { ...stream };
+    delete sanitized.streamKey;
```
victorges, line 222: Should we `db.stream.removePrivateFields` here as well? Probably only after grabbing the user below so we can send the `isAdmin` parameter as well and keep admin-only fields if desired.

packages/api/src/util.js
```diff
@@ -54,6 +54,13 @@ export const fetchWithTimeout = (url, options) =>
     );
   });
 
+export const fetchWithTimeoutAndSleep = (url, options, sleep) => {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      fetchWithTimeout(url, options).then(resolve).catch(reject);
+    }, sleep);
+  });
+};
```
victorges, line 63: **nit:** WDYT of just a `sleepMillis(ms): Promise<void>` helper instead, and callers can do
```ts
await sleepMillis(ms)
await fetchWithTimeout(url, options)
```
or even
```ts
await sleepMillis(ms).then(() => fetchWithTimeout(url, options))
```
instead of 
```ts 
await fetchWithTimeoutAndSleep(url, options, sleep)
```
?

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      id: "webhook_test_12",
+      time: Date.now(),
+      channel: "test.channel",
+      event: "streamStarted",
+      streamId: resJson.id,
+      userId: nonAdminUser.id,
+      isConsumed: false,
+    });
+
+    await sleep(3000);
```
victorges, line 217: I wonder if we could use a "semaphore-like" approach here instead of a sleep, but I think there's no built-in abstraction in node for that :/

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -0,0 +1,228 @@
...
+  });
+});
+
+async function sleep(duration) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      resolve(true);
+    }, duration);
+  });
+}
```
victorges, line 228: This is the exact helper I suggested to have in the `utils` file instead! 😆

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,253 @@
...
+const WEBHOOK_TIMEOUT = 5 * 1000;
+const MAX_BACKOFF = 10 * 60 * 1000;
+const BACKOFF_COEF = 1.2;
+
+export default class WebhookCannon {
+  db: DB;
+  store: Model;
+  running: boolean;
+  verifyUrls: boolean;
+  resolver: any;
```
victorges, line 25: **nit:** You can fix the typing of this one by changing the `dns` require above to an `import`. I played around with this on the stream typescript thing, the pattern I liked was:
```ts
import { promises as dns } from "dns";
...
  resolver: dns.Resolver;
...
    this.resolver = new dns.Resolver();
```
WDYT?

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,253 @@
...
+  stop() {
+    this.db.queue.unsetMsgHandler();
+    this.running = false;
+  }
+
+  disableUrlVerify() {
+    this.verifyUrls = false;
+  }
+
+  calcBackoff(lastInterval): number {
```
victorges, line 57: **typenit:** 
```suggestion
  calcBackoff(lastInterval?: number): number {
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (!lastInterval || lastInterval < 1000) {
+      lastInterval = 5000;
+    }
+    let newInterval = lastInterval * BACKOFF_COEF;
+    if (newInterval > MAX_BACKOFF) {
+      return lastInterval;
```
victorges, line 63: **nit:** Maybe to keep it at round 10 mins by the end
```suggestion
      return MAX_BACKOFF;
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -0,0 +1,253 @@
...
+      } catch (e) {
+        console.error("isLocal Error", isLocal, e);
+        throw e;
+      }
+    }
+    if (isLocal) {
+      // don't fire this webhook.
+      console.log(
+        `webhook ${webhook.id} resolved to a localIP, url: ${webhook.url}, resolved IP: ${ips}`
+      );
```
victorges, line 115: **nit:** Maybe `return` here so we reduce nesting without the need for the `else`?

victorges, line 115: **nit:** Maybe `return` here so we reduce nesting without the need for the `else`?

victorges, line 115: **nit:** Maybe `return` here so we reduce nesting without the need for the `else`?

victorges (commented):
Also ignore if the comments don't make sense 😬 
Mostly some nits though

PR #429, opened 2021-06-03
chore(api): Fix yarn test script

victorges:
### What does this pull request do?

This mainly fixes the `yarn test` command on my machine.

Took quite some while to find the real issue, but in the end it was basically [this same problem](https://github.com/facebook/jest/issues/8226): 
TL;DR: the argument file sent to `jest` is not actually a path but actually a **regex** matched
against the **absolute** file path.

Since I have all my cloned repositories under a folder like `~/workspace/src/`, basically all files 
inside the root folder where `jest` looks on were matching the filtered regex. This meant that all
test files on `dist/` folder as well as the `test.js` in the root of the `api` project were being
executed as if they were tests in the project.

### Specific updates
Fixed:
 - the `dist/` part by ignroring `.test.` files on babel which I think makes sense nevertheless;
 - and the `test.js` part by changing the `jest` command to consider only the local `src` folder 
 with `src` => `${PWD}/src`. This should also work on Windows as per [jest docs](https://jestjs.io/docs/cli#jest-regexfortestfiles).

Tried several other options for jest like using the `--roots`, `--rootDir` and `--runTestsByPath`,
but all of those make the tests run really slow for some weird reason, like the actual requests 
made to the test server took 100s of ms. I tried checking some configs like the DB endpoints
and fallback server but they were all fine, so I had not enough context to go further efficiently.

Also included in this change but unrelated to the main goal:
 - Created a `coverage` script in the `api` which calls the test command with the additional
`--coverage` flag, to make the root project `yarn coverage` to start working again.
 - Fixed what seemed like a typo in the swagger schema inverting descriptions of 2 tags.

##

### How did you test each of these updates

Ran the related yarn scripts like `install`, `prepare`, `test` and `coverage` on `api` and `root`
projects top ensure everything was fine.

### Does this pull request close any open issues?
No

### Checklist


victorges:
✅ Will investigate why the build broke on github pipeline.

packages/api/src/schema/schema.yaml
```diff
@@ -9,9 +9,9 @@ tags:
   - name: object-store
     description: Object Stores Endpoints
   - name: user
-    description: Api Token Endpoints
-  - name: api-token
     description: User Endpoints
+  - name: api-token
+    description: Api Token Endpoints
```
iameli, line 14: Hah, nice catch

iameli (approved):
LGTM!

PR #430, opened 2021-06-04
feat(api): Push target API

victorges:

This creates a new API object called a (RTMP(S) or SRT) Push Target, which is
meant for registering additional destinations that should receive a stream contents.
This has multiple meanings in the market like multi-cast, simulcast, re-streaming, etc.

 - Typing additions in some core helpers to be able to implement the controller
 fully on TS
 - New Push Target API under the `/push-target` root path segment
 - New `pushTargets` field inside the `stream` object to reference existing
 push targets or create a new one inlined in the stream creation payload.
 - Lots of tests

##

- Mostly `yarn test --silent` for now, which I believe are pretty comprehensive
- To do: run the full thing locally and call the APIs manually

Implements https://github.com/livepeer/livepeer-com/issues/431

**Checklist:**


packages/api/src/schema/schema.yaml
```diff
@@ -239,6 +241,45 @@ components:
...
+                description:
+                  Name of transcoding profile that should be sent. Use "source"
+                  for pushing source stream data
+                minLength: 1
+                maxLength: 500
+                example: aac
+              id:
+                type: string
+                description: ID of push target object where to push this stream
+              inline:
```
iameli: I like having this in a separate object (as opposed to dumping all of the pushTarget values directly into this linking object) but I think maybe it should be named `pushTarget`

packages/api/src/schema/schema.yaml
```diff
@@ -239,6 +241,45 @@ components:
...
+                type: object
+                description:
+                  Inline push target object. Allowed on stream creation, which
+                  will automate creating the push-targets on the same request.
+                required:
+                  - url
+                additionalProperties: false
+                properties:
```
iameli, line 275: I forget the exact syntax but we should be able to have this be a ref to the pushTarget definition already provided within this schema

packages/api/src/schema/schema.yaml
victorges: Don't you think it might be weird to have the `pushTargets{}.pushTarget` repetition?

Later yesterday I thought of also updating the `id` field to make it explicit on the field what it is the ID for. So I though of the combination `pushTarget{}.targetId` OR `pushTarget{}.targetSpec`, borrowing the `spec` from k8s objects. What do you think?

packages/api/src/schema/schema.yaml
victorges: I spent quite some time trying to make that work, which IMO would be the ideal way as well. Unfortunately, it seems like the way we use `json-schema-to-typescript` and `ajv` ([src](https://github.com/livepeer/livepeer-com/blob/b1402e06389c328cf2d311169f3f3a43d8c1fb4f/packages/api/src/compile-schemas.js#L46)) doesn't really support this, since we compile/validate each schema separately from one another. I tried digging a bit in the lib(s) options to see if there would be some easy way to make it work, but couldn't find one in the end. From what found, we only use `$ref` in our schema from the API paths referencing one of the existing components, but not in between the components themselves.

Then, after realizing that the object inside the `stream` would actually always be different from the full `push-target` object in its own API, since it would only be the "push-target creation object" here and never a full push-target object returned by the API, I thought it'd be good to have only the simpler schema here, with only the `url` and `name` which is all we need on its creation, and I could live in peace with repeating them 😆 

I just had some ideas that I could further explore to try to make the schema `$ref`s work though. I could try digging a bit further on that path if you think it's worth it, and it would probably allow to avoid some other repetitions in the schema as well (and maybe in the future even break it up in multiple files?). Let me know!

iameli: Ah yeah, that makes sense. We don't need to use Ajv in that way anymore - schema precompilation and whatnot doesn't buy us anything anymore. But yeah, out-of-scope for this project.

> I thought it'd be good to have only the simpler schema here, with only the url and name which is all we need on its creation, and I could live in peace with repeating them 😆

Yup, can totally live with it for now.

victorges: Cool. Maybe later we can give a try to this lib which is specialized on openapi->TS?
https://github.com/drwpow/openapi-typescript

packages/api/src/controllers/object-store.test.js
```diff
@@ -108,24 +108,23 @@ describe("controllers/object-stores", () => {
...
       }
 
-      const res = await client.get(`/object-store?userId=${adminUser.id}`);
+      const res = await client.get(`/object-store`);
       expect(res.status).toBe(200);
       const objStores = await res.json();
       expect(objStores.length).toEqual(4);
-
-      // making sure url is coming back as null
-      expect(objStores[0].url).toEqual(null);
```
victorges, line 128: Note: I removed this check here because the write-only fields are actually _not_ being cleared out when it is an admin making the list query. Is this an intended behavior or is it a mistake to not do that in the list API? 

I ask especially because on this new push-target API I actually included an explicit check _not_ to filter the write-only fields when the caller was an admin ([here](https://github.com/livepeer/livepeer-com/pull/430/files#diff-a53538e800c8499cf95fcfdae3a64b30091960541997bd3db13f28149a68d520R69)).

So should I keep that, or should we always clear these fields? And if the latter, a follow-up: How does it work to actually get these fields in production? Like how does `go-livepeer` get the object-store URLs so it can write the video records there, since in the other object-store APIs this field is always cleared out? (it seems like the only one letting it go is this listing one).

victorges: OK, for that last part I am understanding that `go-livepeer` calls [this `/stream/hook` API](https://github.com/livepeer/livepeer-com/blob/0bf3b3a12f34fe919d575d64507d7f76b6cc7983/packages/api/src/controllers/stream.js) from this [mediarserver.go code](https://github.com/livepeer/go-livepeer/blob/54ac51ba36d5dea7bcf0463823c0e90537b8880f/server/mediaserver.go) in which we return the object store URLs. I couldn't find any authorization in that `/hook` API, is that by design?

Would that be a place where we'd need to create some push-target logic to return the push-target URL as well? Or is there a separate hook that mist-server calls already?

packages/api/src/schema/schema.yaml
victorges: FTR: decided to go with `id` and `spec` as a naming pattern that we expect to follow on these "reference or inline object" API objects.

victorges:
FTR I believe this is ready for review now :D

iameli:
Right on, taking a look now.

packages/api/src/compile-schemas.js
```diff
@@ -32,15 +33,18 @@ const schemaStr = fs.readFileSync(
...
-const str = JSON.stringify(data, null, 2);
-write(path.resolve(schemaDir, "schema.json"), str);
-write(path.resolve(schemaDistDir, "schema.json"), str);
-const ajv = new Ajv({ sourceCode: true });
-
-const index = [];
-const types = [];
 
 (async () => {
+  await $RefParser.dereference({ components: data.components });
```
iameli, line 38: Sweet.

packages/api/src/controllers/stream.js
```diff
@@ -53,6 +54,51 @@ const hackMistSettings = (req, profiles) => {
...
 };
 
+async function validatePushTarget(userId, profileNames, pushTargetRef) {
+  const { profile, id, spec } = pushTargetRef;
+  if (!profileNames.has(profile) && profile !== "source") {
+    throw new BadRequestError(
+      `push target must reference existing profile. not found: "${profile}"`
+    );
+  }
+  if (!!spec === !!id) {
```
iameli, line 64: Maybe a smidge more explicit:
```suggestion
  if (!spec && !id) {
```

packages/api/src/controllers/stream.js
```diff
@@ -53,6 +54,51 @@ const hackMistSettings = (req, profiles) => {
...
+    throw new BadRequestError(
+      `push target must have either an "id" or a "spec"`
+    );
+  }
+  if (id) {
+    if (!(await db.pushTarget.hasAccess(id, userId))) {
+      throw new BadRequestError(`push target not found: "${id}"`);
```
iameli, line 71: Seems to me that this should be either a 404 "Push Target Not Found" or a 403 "You May Not Add This Push Target" kind of thing. If they already have the id of someone else's push target, we probably don't accomplish too much by hiding it with a 404, so I'm thinking the 403 might be appropriate.

packages/api/src/schema/schema.yaml
```diff
@@ -239,6 +241,42 @@ components:
           example: D8321C3E-B29C-45EB-A1BB-A623D8BE0F65
           description:
             ID of object store where to which this stream was recorded
+        pushTargets:
+          type: array
+          example: [{ "id": "PUSH123", "profile": "720p" }]
+          description:
+            References to targets where this stream will be simulatenously
```
iameli:
```suggestion
            References to targets where this stream will be simultaneously
```

packages/api/src/schema/schema.yaml
```diff
@@ -239,6 +241,42 @@ components:
...
+            additionalProperties: false
+            properties:
+              profile:
+                type: string
+                description:
+                  Name of transcoding profile that should be sent. Use "source"
+                  for pushing source stream data
+                minLength: 1
+                maxLength: 500
+                example: aac
```
iameli: "aac" would be a weird profile for this, implies an audio-only profile that we don't support. Alternative:
```suggestion
                example: 720p
```

packages/api/src/schema/schema.yaml
```diff
@@ -289,6 +327,17 @@ components:
           type: boolean
           description: If currently suspended
 
+    stream-patch-payload:
```
iameli, line 330: Siiiiiick

packages/api/src/store/push-target-table.ts
```diff
@@ -0,0 +1,62 @@
+import { PushTarget } from "../schema/types";
+import Table from "./table";
+import { GetOptions, WithId } from "./types";
+import * as uuid from "uuid";
+import { InternalServerError, UnprocessableEntityError } from "./errors";
+
+type DbPushTarget = WithId<PushTarget>;
```
iameli: Perhaps `DBPushTarget` for consistency with `DBLegacyObject` and whatnot

packages/api/src/controllers/object-store.test.js
iameli: > Note: I removed this check here because the write-only fields are actually not being cleared out when it is an admin making the list query. Is this an intended behavior or is it a mistake to not do that in the list API?

Admins should be allowed to retrieve the actual fields in all cases, I think. So I guess it's a mistake not to do that in the get API?

> OK, for that last part I am understanding that go-livepeer calls this /stream/hook API from this mediarserver.go code in which we return the object store URLs. I couldn't find any authorization in that /hook API, is that by design?

Yeah, right now anyone with a stream ID (not a playbackID, importantly) could query that for the data. We discussed some potential ways to improve that workflow a long time ago, discussion here: https://github.com/livepeer/go-livepeer/issues/1278

> Would that be a place where we'd need to create some push-target logic to return the push-target URL as well? Or is there a separate hook that mist-server calls already?

It _would_ be, if go-livepeer was the thing doing the RTMP pushing. But we're going to have Mist take care of that via mist-api-connector, so no changes necessary there quite yet.

iameli (changes requested):
Looks great! Mostly relatively minor changes; I'm going to boot this up locally and then give it one final go-through before we ship.

packages/api/src/controllers/stream.js
victorges: The difference with that one would be that we wouldn't catch when **both** are set, which IMO should also be an error here. I added a test case for that scenario which would break if we changed the `===` to a `&&` here: https://github.com/livepeer/livepeer-com/blob/174374a5e87fb0a280f0bbb176b2237e0a06f71f/packages/api/src/controllers/stream.test.js#L255

I can actually remove one of the `!` and have the same desired behavior though, WDYT?
```suggestion
  if (!spec === !id) {
```

Otherwise, do you think we should just ignore the spec in case both fields are set, only using the ID?

packages/api/src/controllers/stream.js
victorges: I guess there are 2 subjects here: 
 1. If we should separate the error cases when the stream doesn't exist from when it is from another user;
 2. On any case, what status code to use.

I'm ok with making this return a "real" `404` when the push target doesn't exist and `403` when it exists but is from another user, but I'll explain my reasoning below and you let me know if you do prefer it that way.

### 1
I think it ends up a [simpler code](https://github.com/livepeer/livepeer-com/blob/b8ac0cd01b7620bae5c94abe9b8a4ee5a0b79874/packages/api/src/controllers/stream.js#L956) overall to treat both non existence and lack of permission the same way, so that we don't need separate `if` checks in the code which becomes more verbose. That `hasAccess` (and the underlying `getAuthed`) function checks both of those and returns `true` iff push target exists and user has access. 

So maybe the question should be the opposite, of whether it's worth to break up the conditions to be explicit about whether the object exists and is forbidden or it just doesn't exist. IMO there's a small gain in hiding the existence fact, then since the code felt a bit cleaner to me I thought it was worth adotping the pattern of "404"s when trying to access a cross-user resource.

### 2
I thought it could be misleading to return a `404` on a `/stream/*` endpoint, when actually the stream in question does exist but something in the payload that had a bad reference, either of a push target that doesn't exist or a push target from another user. 

Especially on the `PATCH` scenario, a `404` from a request to `/stream/abc-1234-def` request would make me think that the stream in question doesn't exist instead. So then this is not necessarily an HTTP `404` error, referring to the resource being accessed*, but a problem with the payload, which is how I ended up on the `400`.

\* This may be just my understanding of the 404 though, that it should refer to the resource in the URL. Couldn't find a definite answer with a quick search.

packages/api/src/schema/schema.yaml
victorges: Oh good catch. I had copied it from the `profile.name` field in this schema, soo I'll update that one too.

packages/api/src/store/push-target-table.ts
victorges: Cool. Updated `WithId` to `WithID` as well, since I'm guessing we don't want our data types to develop [consciousness](https://en.wikipedia.org/wiki/Id,_ego_and_super-ego#Id) quite yet.

packages/api/src/controllers/object-store.test.js
victorges: Cool! Thanks for the discussion reference I'll take a look and will mark this as resolved.
FTR: I am leaving the code on the `push-target` controllers that hides the write-only fields _except_ for admin users.

victorges:
@iameli Ready for another round! Addressed most comments and replied to some others, just waiting on your position on them to make any related updates.

packages/api/src/controllers/stream.js
iameli: Ah, good catch, I didn't think of the "both" case. You can leave as-is

packages/api/src/controllers/stream.js
iameli: Ah sure, I didn't wrap my head around the fact this was in `/stream` rather than `/push-target` - sounds fine to me

iameli (approved):
LGTM! 🚢

Merge it and see if the API works on Livepeer.monster 😃

Issue #431, opened 2021-06-07
Implement RTMP push endpoints


iameli:
# RTMP Push API

### New API Object: /push-target

address: rtmp, rtmps, srt

```json
{
    "id": "PUSH123",
    "userId": "USER123",
    "name": "Eli's Facebook Live",
    "address": "rtmp://live.facebook.com/live/ABC123"
    
}
```

```json
{
    "id": "PUSH444",
    "userId": "USER123",
    "name": "Eli's Twitch",
    "address": "rtmp://rtmp.twitch.tv/live/SECRET"
}
```

### Add to Stream object

```json
{
    ...stream stuff
    id: "STREAM99",
    pushTargets: [
        { id :"PUSH123", profile: "720p" }
    ]
}
```

PATCH /stream/STREAM99
```json
{
    "pushTargets": [
        { "id": "PUSH123", "profile": "720p" }
        { "id": "PUSH444", "profile": "source" }
    ]
}
```

###  Inline creation of pushTargets

```json
x = {
  ...streamStuff,
  pushTargets: [
    {
      profile: "720p",
      pushTarget: {
        url: "rtmp://destination-url",
        ... other fields here
      }
    }
  ]
}
```

**Wowza workflow**
They should use the same syntax as above

Other considerations:
* Push target URLs should never be retrievable by users, only admins
* Should check to make sure that the push-target userId is the same as the stream userId

PR #449, opened 2021-06-15
fix(api): Fix some small issues in the Push Target API

victorges:
This fixes 2 issues in the just-merged push-target API, before it goes live to production.
Each of the fixes is on a separate commit.

First one is to avoid setting the default value of the `push-target` field to `null`, which feels
 weird when using the API and could lead to errors if a developer doesn't expect that to happen.
 It also wouldn't even be valid given the object JSON schema, so it makes a lot of sense to use
 just an empty array there. Now the possibilities are: 
 - streams created before the feature: there will be no `pushTarget` field in there
 - streams created after the feature but with no push target: an empty push targets array
 - streams created with push targets: the push targets array always with an id+profile (the
 spec is only for creation)

Second fix was actually a bug in the push-target listing API, which wasn't cleaning write-only
fields in the response so even non-admin fields were getting the `url`s in the payload. Added
some tests to check if that starts happening again.

 - Replace `null` with a `[]` for the default value of the `pushTargets` field
 - Properly filter write-only fields in the push-targets list API response

## -

 - Ran tests
 - Called the respective APIs (push-target list and stream create) and validated the created/response objects

Also related to https://github.com/livepeer/livepeer-com/issues/431

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/push-target.test.ts
+++ b/packages/api/src/controllers/push-target.test.ts
@@ -119,7 +119,7 @@ describe("controllers/push-target", () => {
       }
       const res = await client.get(`/push-target?userId=${nonAdminUser.id}`);
       expect(res.status).toBe(200);
-      expect(await res.json()).toEqual([userPushTarget]);
+      expect(await res.json()).toEqual([{ ...userPushTarget, url: undefined }]);
     });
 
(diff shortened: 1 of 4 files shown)


iameli (approved):
LGTM

PR #453, opened 2021-06-16
Proper Rabbit Queue, With custom delays

ya7ya:
I was not happy with using Postgresql as a msg queue, so I changed it to RabbitMQ

this solves the issue we had with the delayed emits


packages/api/src/store/rabbit-queue.ts
```diff
@@ -0,0 +1,106 @@
...
+export default class MessageQueue {
+  private channel: ChannelWrapper;
+  private connection: AmqpConnectionManager;
+
+  constructor() {}
+
+  public async connect(url: string): Promise<void> {
+    // Create a new connection manager
+    this.connection = await amqp.connect([
+      url || process.env.CLOUDAMQP_MQTT_URL,
```
iameli: Needs to get into parse-cli.js as `LP_CLOUDAMQP_MQTT_URL`

iameli: Or just `LP_AMQP_URL`

packages/api/src/store/rabbit-queue.ts
```diff
@@ -0,0 +1,104 @@
...
+  public async connect(url: string): Promise<void> {
+    // Create a new connection manager
+    this.connection = await amqp.connect([url]);
+    this.connection.on("connect", () => console.log("AMQP Connected!"));
+    this.connection.on("disconnect", (err: Error) =>
+      console.log("AMQP Disconnected.", err)
+    );
+    this.connection.on("error", (error: Error) => {
+      console.error(`Connection error: ${error}`);
+    });
```
victorges, line 23: Should we wait for either of these events before returning from `connect`?
Something like we have in [the server startup](https://github.com/livepeer/livepeer-com/blob/7eeb04d5bdd819a5d7022dd6455a2b51d15fbbfb/packages/api/src/index.js#L115), manually creating a `Promise` object and calling `resolve`/`reject` from these callbacks.

Edit: Although the `createChannel` below makes it seem like if `connect` returns we are indeed connected. Do we need these `connect` callbacks at all then?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  public ack(data: any): void {
+    this.channel.ack(data);
+  }
+
+  public async consume(func: (msg: ConsumeMessage) => void): Promise<void> {
+    if (!func) {
+      console.log("func is undefined");
+      func = this.handleMessage.bind(this);
+    }
```
victorges: From what I understood of the logic here, I am thinking that this default may be unnecessary as it may hide some coding error. WDYT of throwing an error here instead?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -0,0 +1,104 @@
...
+    await this.connection.removeAllListeners("disconnect");
+    await this.connection.removeAllListeners("error");
+    await this.connection.close();
+  }
+
+  public ack(data: any): void {
+    this.channel.ack(data);
+  }
+
+  public async consume(func: (msg: ConsumeMessage) => void): Promise<void> {
```
victorges, line 59: **nit:** Just a trip, might not be worth it if it's more complex than it seems (not much experience with RabbitMQ): Would be nice to provide an `async` version of this which auto acks/nacks the msg based on the success or failure of the async function. 

Say, receive an 
```ts
asyncFunc: (msg: ConsumeMessage) => Promise<void>
``` 
here and send 
```ts
(msg) => asyncFunc(msg).then(() => channel.ack()).catch(() => channel.nack())
```
to the `channel.consume` function.

packages/api/src/store/rabbit-queue.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      },
+      5000
+    );
+    emittedAt = Date.now();
+    await sleep(7000);
+    let duration = consumedAt - emittedAt;
+    console.log("duration: ", duration);
+    if (duration < 5000) {
+      throw new Error("the 5s delayed queue failed");
+    }
```
victorges: **nit:**
```suggestion
    expect(duration).toBeGreaterThanOrEqual(5000)
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -23,30 +24,37 @@ export default class WebhookCannon {
...
+    this.queue = queue;
     // this.start();
   }
 
   async start() {
     console.log("WEBHOOK CANNON STARTED");
-    if (this.running) {
-      await this.db.queue.setMsgHandler(this.processEvent.bind(this));
-    }
+    await this.queue.consume(this.handleQueueMsg.bind(this));
```
victorges, line 40: Does this `consume` function in the queue setup the consumption for all events from now on or it only consumes one? I thought it was the latter until now but if the first instead WDYT of renaming it to something like `setupConsume`, `startConsuming` or something like that?

packages/api/src/webhooks/cannon.ts
```diff
@@ -23,30 +24,37 @@ export default class WebhookCannon {
...
 
-  async processEvent(msg: Notification) {
-    console.log("EVENT TRIGGERED ON THE WEBHOOK");
-    let event = await this.db.queue.pop(this.onTrigger.bind(this));
-    console.log("event: ", event);
+  async handleQueueMsg(data: any) {
+    let message = JSON.parse(data.content.toString());
+    console.log("webhookCannon: got message", message);
+    await this.onTrigger(message);
+    this.queue.ack(data);
```
victorges: Should we `nack` in case there is some error with the `onTrigger` callback? Or is there an automatic timeout for that or something?

victorges (commented):
Pretty cool! Some comments, possibly out of context so feel free to disregard anything that doesn't make sense 😬 
Both because I'm new to the project and because I actually haven't checked the first PR yet.

packages/api/src/store/rabbit-queue.ts
```diff
@@ -0,0 +1,104 @@
...
+      return Promise.all([
+        channel.assertQueue(`delayedQueue_${delay / 1000}s`, {
+          messageTtl: delay,
+          deadLetterExchange: EXCHANGE_NAME,
+          expires: delay + 15000,
+        }),
+      ]);
+    });
+    console.log("emitting ", msg);
+    this.channel.sendToQueue(`delayedQueue_${delay / 1000}s`, msg);
```
victorges: Should we `await` for this one too?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  public handleMessage(data: any) {
+    var message = JSON.parse(data.content.toString());
+    console.log("subscriber: got message", message);
+    this.channel.ack(data);
+  }
+
+  public async emit(msg: Object): Promise<void> {
+    console.log("emitting ", msg);
+    this.channel.sendToQueue(QUEUE_NAME, msg);
```
victorges: Same regarding `await` or `return`ing this promise

packages/api/src/webhooks/cannon.ts
```diff
@@ -69,7 +77,7 @@ export default class WebhookCannon {
     event.lastInterval = this.calcBackoff(event.lastInterval);
     event.status = "pending";
     event.retries = event.retries ? event.retries + 1 : 1;
```
victorges: Since the `event` object is shared between different webhook deliveries, might be better to make a copy here instead
```suggestion
    event = {
      ...event,
      lastInterval: this.calcBackoff(event.lastInterval),
      status: "pending",
      retries: event.retries ? event.retries + 1 : 1,
    }
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -69,7 +77,7 @@ export default class WebhookCannon {
...
-    this.db.queue.updateMsg(event);
+    this.queue.delayedEmit(event, event.lastInterval);
   }
 
   async _fireHook(
```
victorges, line 91: (closest place I could comment where I actually meant this, on [line 135](https://github.com/livepeer/livepeer-com/pull/453/files#diff-8942d451dd2aa33dc5fdc9410a004c7011e39201e287b71a009216c88830ebf7L135))

There, we are doing a sleep before firing a webhook, but I understood that was a necessary logic only for the Postgres queue but not for RabbitMQ, since we handle retries in it with the message TTL+deadletter queue instead. 

So, do we still need that sleep there? Since the event will have been delayed already maybe it's just fire at the hook. I wonder what kind of tests could have caught this as well 🤔

packages/api/src/webhooks/cannon.ts
```diff
@@ -202,7 +210,7 @@ export default class WebhookCannon {
   }
 
   async onTrigger(event: Queue) {
```
victorges, line 216: Another thing that came to my mind about retries:

We currently use a single event for all webhooks, fetching the webhook list when consuming that event. I think that's nice, and it make the firing of the events/webhooks really cheap, only needing to send 1 event to rabbit and almost no other I/O.

But I think there might be some issues in case we have multiple webhooks registered for the same event. Say, one global webhook and one specific to a streamId, or maybe one that listen to all events (someday?) and one that listens to only that specific event. In those cases, if 1 of them fails, we will be retrying the same original `event` object which will trigger all of the webhooks again. Maybe worse, if all of them fail, we will multiply the retries sending 1 retry event for each failure and so the number of events multiplies by `N` on every retry round.

One solution that came to my mind would be adding a new field only for the retry events. A `webhookId` field to the event we send to rabbit for retry, and if the event we receive has that field we will fire only that specific hook. This keeps the simple&cheap publishing logic but makes the retry independent for each hook. WDYT?

packages/api/src/webhooks/cannon.ts
victorges: Additionally, do we/should we have a limit of retries? Like not retry for more than 24 hours the same event.

victorges (commented):
Some additional things that came to my mind later. Unless I understood something wrong (which is very likely 😆) I think some small changes might be necessary.

packages/api/src/webhooks/cannon.ts
ya7ya: it's for all the events from now (or if there are unconsumed msgs on the queue).
it's named `consume` following the `consume` function on the `amqplib` we use. i'm cool with changing it if the semantics makes a difference here though.

packages/api/src/controllers/stream.js
```diff
@@ -749,14 +749,30 @@ app.put("/:id/setactive", authMiddleware({}), async (req, res) => {
...
+    req.queue.emit({
+      id: uuid(),
+      createdAt: Date.now(),
+      channel: "webhooks",
+      event: "stream.started",
+      streamId: id,
+      userId: user.id,
+    });
+    res.status(204);
+    return res.end();
```
darkdarkdragon, line 763: Is this temporary?

packages/api/src/webhooks/cannon.ts
```diff
@@ -23,30 +24,42 @@ export default class WebhookCannon {
...
-    }
+    await this.queue.consume(this.handleQueueMsg.bind(this));
   }
 
-  async processEvent(msg: Notification) {
-    console.log("EVENT TRIGGERED ON THE WEBHOOK");
-    let event = await this.db.queue.pop(this.onTrigger.bind(this));
-    console.log("event: ", event);
+  async handleQueueMsg(data: any) {
+    let message = JSON.parse(data.content.toString());
```
darkdarkdragon, line 44: Should try/catch this.

darkdarkdragon (changes requested):
Can't comment needed parts of code, so commenting here:
1. function `calcBackoff` should return integer number. Right now it returns double, and RabbitMQ throws `unacceptable_type,double` error for that.
2. `onTrigger` function, after this line:
```js
    let stream = await this.db.stream.get(event.streamId);
    ``` should check if stream is found.
3. Retries should not be indefinite - need to stop retrying failing webhooks after some time or number of tries.

Issue #451, opened 2021-06-15
Add field for enabling scene classification on the Stream object


iameli:
I created a branch in livepeer-com https://github.com/livepeer/livepeer-com/tree/yf/scene-classification-example that contains:

- An example of the config that the stream auth webhook should pass to a B in order to configure scene classification https://github.com/livepeer/livepeer-com/commit/7b289bf79f566cb6ecb44c1643fecad8a8a06982
- An example of a detection webhook that accepts results posted by a B (currently just logs the results) https://github.com/livepeer/livepeer-com/commit/5e6156061953f914a377d488a0710be634f80080

iameli:
The shape of the incoming go-livepeer object is defined here https://github.com/livepeer/go-livepeer/issues/1917

victorges:
Cool, thanks for the examples!

Issue #450, opened 2021-06-15
Scene Classification Webhook


iameli:
Handle scene classification response from go-livepeer, passing it on to any users that have it registered.

The incoming go-livepeer webhook looks like this: https://github.com/livepeer/go-livepeer/issues/1917

Per discussion, our webhook object will look like:

```json5
{
    event: "scene.detected",
    url: "https://playdj.tv/api/livepeer-scene-detected",
    detection: {
        sceneClassification: [
            {name: "soccer"},
            {name: "adult"}
        ],
        
        // someday...
        objectDetection: [
            {name: "soccer ball"}
        ]
    }
}
```

victorges:
What about using something like `stream.detection` for the event name? 
Or sth similar like `stream.sceneDetection`.

The prefix would keep consistency with the other events generated from streams like `stream.(active|idle)`.
And the `detection` would use the same name as the field that configures that specific event, which feels nice and tight to me.

Additionally, a follow up, what should we do in case there are multiple webhooks registered? Should we merge configurations when configuring the stream auth webhook and fan-out the events to the appropriate specific webhooks when receiving?

victorges:
Just had another idea about this schema:

What if instead of having the detection configuration inside the webhook itself, we have it on the `stream` object themselves, but defaulting to some sane configuration in case the one on the stream doesn't exist but a detection webhook exists (global or specific)? 

I think that makes more sense to me than the webhook objects "understanding" the stream detection configurations, since the webhooks should rather be pretty simple instead with only information about type of event, URL to call and possibly scope of the event (currently only specific to a stream or global).

This would allow it to work for clients creating stream objects directly from Mist, since for now they will most likely just be using that default detection configuration anyway. In that default we will include every scene detection we currently have since they all run on the same model anyway.

In the (medium-term?) future, we will start having other scene classifications, but I believe they will likely remain on the same model so we can just include them in the "default" configuration, and by the time we actually have multiple models or classifications that clients want to configure in the low level, we will hopefully already have access to the mist plugin and be able to configure that payload as well, to make the config directly in the stream object.

To sum it up:
`webhook:`
```yaml
{
    event: "stream.detection",
    url: "https://playdj.tv/api/livepeer-scene-detected"
}
```

`stream:`
```yaml
{
    ... other stream stuff
    detection?: {
        sceneClassification: [
            {name: "soccer"},
            {name: "adult"}
        ],
        // someday...
        objectDetection: [
            {name: "soccer ball"}
        ]
[...]

PR #463, opened 2021-06-17
feat(api): Scene classification webhook

victorges:
This implements both sides of the interaction with `go-livepeer` for the
scene detection/classification feature we will provide. These are:
 - A `detection` configuration to be returned in the auth hook on `/stream/hook`
 which is called in the beginning of the stream by the broadcaster. This ensures 
 that the scene classification models are enabled in the transcoders so we get
 the results of those.
 - An additional detection hook endpoint that `go-livepeer` will call when there
 are some scene detection results, which will notify users by a `webhook`, which
 is why this is built on top of #453

The detection is configured on the `stream` object via a similar field with the
`detection` name, but a default configuration will be used in case the stream
doesn't configure it explicitly but the user has registered a webhook for detection.
This allows clients using Mist to create the streams to use this feature, since they
won't need to change anything on their stream objects only register a webhook in
their account.

 - new `detection` field on `/stream/hook` for detection configuration
 - new `/detection` API to be called with detection results (might be
 better to rename it to something else like `/webhook/detection/trigger`
 or something like that for consistency?)
 - new field in `stream` object with `detection` configuration

## -

 - `yarn test` for now. added some new tests for the auth hook part

implements https://github.com/livepeer/livepeer-com/issues/450 
implements https://github.com/livepeer/livepeer-com/issues/451

**Checklist:**


victorges:
Changed base to be on top of https://github.com/livepeer/livepeer-com/pull/453 so I can use some of the new abstractions already, specifically the webhook cannon.

victorges:
Merged into https://github.com/livepeer/livepeer-com/pull/466

PR #467, opened 2021-06-21
Further updates to scene classification/webhook

victorges:

## -

**Screenshots (optional):**

**Checklist:**


packages/api/src/store/rabbit-queue.test.ts
```diff
@@ -9,6 +10,20 @@ async function sleep(duration) {
   });
 }
 
+function semaphore() {
```
iameli: Neat. Want to put this in a shared helper someplace where we could import it in other tests?

iameli (approved):
LGTM

packages/api/src/store/rabbit-queue.test.ts
victorges: Sure! I'll even move/export the `sleep` one too and avoid repeating it on stream tests

PR #470, opened 2021-06-22
Tests for scene classification webhook

victorges:
**What does this pull request do? Explain your changes. (required)**
Add some tests to the detection/scene classification webhook.

**Specific updates (required)**
 - Add validation tests on webhook called by go-livepeer
 - Add tests for processing and receiving of actual detection event from a registered webhook
 - Translate some test utilities to typescript
 - Some other small fixes or changes to webhooks code

## -

- **How did you test each of these updates (required)**
 - `yarn test --silent`

**Does this pull request close any open issues?**
Closes https://github.com/livepeer/livepeer-infra/issues/573

**Checklist:**

- [x] I have read the **CONTRIBUTING** document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

PR #471, opened 2021-06-22
webhooks fixes

darkdarkdragon:
Various fixes

See commits

Fixes https://github.com/livepeer/livepeer-infra/issues/572
**Screenshots (optional):**

**Checklist:**


packages/api/src/webhooks/cannon.ts
```diff
@@ -41,8 +41,15 @@ export default class WebhookCannon {
...
   async handleQueueMsg(data: any) {
-    let message = JSON.parse(data.content.toString());
-    console.log("webhookCannon: got message", message);
+    let message;
+    try {
+      message = JSON.parse(data.content.toString());
+      console.log("webhookCannon: got message", message);
+    } catch (err) {
+      console.log("webhookCannon: error parsing message", err);
+      this.queue.ack(data);
```
victorges, line 61: I'm wondering if we should `ack` or `nack` in case of thrown errors. I had suggested to do a `nack` on the original PR (line 54 below), but when building the detection hooks tests on #470 I realized that a `nack` would cause RabbitMQ to keep immediately retrying a message that was getting an exception on processing. That is not exactly what I was thinking originally with the `nack`, which was rather something like a retry.

That said, WDYT of changing the `nack` below either to be a `this.retry` so the broken event doesn't keep flooding the consumers without any delays? I'd be fine with an `ack` too if you prefer.

packages/api/src/webhooks/cannon.ts
```diff
@@ -75,7 +82,8 @@ export default class WebhookCannon {
     if (newInterval > MAX_BACKOFF) {
       return lastInterval;
     }
-    return newInterval;
+    // RabbitMQ expects integer
+    return newInterval|0;
```
victorges: I think you're missing a lint here to add some spacing in the expression. Either way, WDYT of the more explicit
```suggestion
    return Math.floor(newInterval);
```
or even rounding
```suggestion
    return Math.round(newInterval);
```
?

victorges (commented):
LGTM, but I'll leave for someone with more context to officially approve

packages/api/src/webhooks/cannon.ts
darkdarkdragon: Changed to retry below. Here is we do `ack` because there is no way we could recover from `JSON.parse` error.

packages/api/src/webhooks/cannon.ts
darkdarkdragon: I prefer `|0` because it always returns integer number. `Math.round(Infinity)` returns `Infinity`.

packages/api/src/webhooks/cannon.ts
```diff
@@ -75,7 +82,8 @@ export default class WebhookCannon {
     if (newInterval > MAX_BACKOFF) {
       return lastInterval;
     }
-    return newInterval;
+    // RabbitMQ expects integer
+    return newInterval | 0;
```
iameli, line 97: A little more expressive:
```suggestion
    return Math.floor(newInterval);
```

iameli (approved):
Tiny suggestion but LGTM

packages/api/src/webhooks/cannon.ts
darkdarkdragon: But it is different - I like `| 0` more because it always returns integer. `Math.floor(1/0)` returns `Infinity`

PR #472, opened 2021-06-22
Implement basic auth and limit stream hooks to admins only

### LINKED ISSUE #1278: How should we handle authenticated webhook requests?
When #1263 lands, the API will be returning object store credentials to the broadcaster at `-authWebhookUrl`. We should be careful about exposing those credentials. Currently, anyone in possession of a stream ID could hit `https://livepeer.live/api/stream/hook` and retrieve the object store credentials. That's bad. Here are four ways to handle that which I posted in chat:

1. Very simple shared secret implementation. Run the API with (perhaps) `--webhook-key=SHARED_SECRE`T and require the broadcasters to hit the webhook at https://livepeer.live/api/stream/hook?key=SHARED_SECRET. We could do that without any changes to go-livepeer. But now SHARED_SECRET is in the URL, and URLs get logged a lot, and SHARED_SECRET plus USER_UUID equals access to ultra sensitive credentials, so that makes me nervous.
2. Maybe a slightly better option: basic auth? So in go-livepeer you'd specify -authWebhookUrl https://broadcaster-abcd:SHARED_SECRET@livepeer.live/api/stream/hook, and then that ends up in the authorization: Basic [basic encoded secrets go here] header, validated by the API. That's kind of neat. I like that it can be easily specified as part of `-authWebhookUrl`.
3. As #1, but use some kind of HMAC of SHARED_SECRET and the stream URL such that we can use a shared secret without ever sending that secret over the wire, which I'd be more comfortable with than option 1.
4. Or most complex/robust would probably be a JWT signed by the Ethereum identity associated with the B node. Which would require offchain B nodes to have an eth identity, but maybe that's a good thing.

victorges:
This is to add an authorization layer to the hooks called on the stream object by go-livepeer.

Solution has been discussed in the issue: https://github.com/livepeer/go-livepeer/issues/1278

 - Add support to HTTP Basic Auth in auth middleware
 - Make stream auth and detection hooks admin-only
 - TODO: Add an admin token to go-livepeer deployments for them to call the API

## -

 - For now mostly `yarn test` which already got some errors on the tests calling the hooks with a non admin token
 - TODO: Create some tests for the Basic auth logic

Fixes https://github.com/livepeer/go-livepeer/issues/1278

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.js
+++ b/packages/api/src/controllers/stream.js
@@ -1233,7 +1233,7 @@ async function terminateStreamReq(req, stream) {
 
 const streamDetectionEvent = "stream.detection";
 
-app.post("/hook", async (req, res) => {
+app.post("/hook", authMiddleware({ anyAdmin: true }), async (req, res) => {
   if (!req.body || !req.body.url) {
     res.status(422);
     return res.json({
@@ -1399,6 +1399,7 @@ app.post("/hook", async (req, res) => {
 
 app.post(
   "/hook/detection",
+  authMiddleware({ anyAdmin: true }),
   validatePost("detection-webhook-payload"),
   async (req, res) => {
     const { manifestID, seqNo, sceneClassification } = req.body;
--- a/packages/api/src/middleware/auth.js
+++ b/packages/api/src/middleware/auth.js
@@ -56,7 +56,7 @@ function authFactory(params) {
       throw new ForbiddenError(`unsupported authorization type ${tokenType}`);
     }
 
-    user = await req.store.get(`user/${userId}`);
+    user = await db.user.get(userId);
     if (!user) {
       throw new InternalServerError(`no user found for token ${authToken}`);
     }
--- a/packages/api/src/middleware/auth.js
+++ b/packages/api/src/middleware/auth.js
@@ -29,9 +29,7 @@ function authFactory(params) {
       throw new ForbiddenError(`no authorization header provided`);
     } else if (["Bearer", "Basic"].includes(tokenType)) {
       const isBasic = tokenType === "Basic";
-      tokenObject = await req.store.get(
-        `api-token/${isBasic ? basicUser.pass : req.token}`
-      );
+      tokenObject = await db.apiToken.get(isBasic ? basicUser.pass : req.token);
       const matchesBasicUser = tokenObject?.userId === basicUser?.name;
       if (!tokenObject || (isBasic && !matchesBasicUser)) {
         throw new ForbiddenError(`no token object ${tokenValue} found`);
--- a/packages/api/src/middleware/auth.js
+++ b/packages/api/src/middleware/auth.js
@@ -57,45 +57,31 @@ function authFactory(params) {
     }
 
     user = await req.store.get(`user/${userId}`);
-
     if (!user) {
       throw new InternalServerError(`no user found for token ${authToken}`);
     }
-
     if (user.suspended) {
       throw new ForbiddenError(`user is suspended`);
     }
-
     if (!params.allowUnverified && user.emailValid === false) {
       throw new ForbiddenError(
         `useremail ${user.email} has not been verified. Please check your inbox for verification email.`
       );
     }
+    // UI admins must have a JWT
+    const isUIAdmin = user.admin && tokenType === "JWT";
+    if ((params.admin && !isUIAdmin) || (params.anyAdmin && !user.admin)) {
+      throw new ForbiddenError(`user does not have admin priviledges`);
+    }
 
     req.user = user;
     req.authTokenType = tokenType;
-    req.isUIAdmin = req.user.admin && tokenType === "JWT";
+    req.isUIAdmin = isUIAdmin;
     if (tokenObject && tokenObject.name) {
       req.tokenName = tokenObject.name;
     }
     if (tokenObject && tokenObject.id) {
       req.tokenId = tokenObject.id;
-    }
-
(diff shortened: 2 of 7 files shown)


iameli (approved):
Love it. But let's wait to merge until the go-livepeer bits land, yes?

victorges:
Yep! We can only merge/deploy this on a region once go-livepeer is actually
configured with the credentials for the webhooks.

packages/api/src/controllers/stream.js
```diff
@@ -1233,7 +1233,10 @@ async function terminateStreamReq(req, stream) {
 
 const streamDetectionEvent = "stream.detection";
 
-app.post("/hook", async (req, res) => {
+app.post("/hook", authMiddleware({}), async (req, res) => {
+  if (!req.user.admin && !req.user.access?.broadcasterHooks) {
```
darkdarkdragon: Can we move this into `auth-middleware`, similar to `anyAdmin` flag?

packages/api/src/middleware/auth.js
```diff
@@ -11,31 +19,37 @@ function authFactory(params) {
...
-      throw new ForbiddenError(`no token object ${authToken} found`);
-    } else if (authToken.startsWith("Bearer")) {
-      tokenObject = await req.store.get(`api-token/${req.token}`);
-      if (!tokenObject) {
-        throw new ForbiddenError(`no token object ${authToken} found`);
+    if (!tokenType) {
+      throw new ForbiddenError(`no authorization header provided`);
+    } else if (["Bearer", "Basic"].includes(tokenType)) {
+      const isBasic = tokenType === "Basic";
+      tokenObject = await db.apiToken.get(isBasic ? basicUser.pass : req.token);
```
darkdarkdragon: `basicUser.pass ` - this assumes that `basicUser` was successfully parsed, which not necessary be the case.

darkdarkdragon (changes requested):

packages/api/src/controllers/stream.js
victorges: The problem with that would be how to represent the "admin OR broadcasterHooks" idea in the params object. For example something like `{ anyAdmin: true, access: { broadcasterHooks: true } }` could make sense for this specific case, but it could also look like that _both_ things need to be set instead, and not one or another. 

I could make a more generic interface in the auth middleware, like receiving a function that validates properties on the `User` object (and token type?) and returns a `boolean` instead. Then already provide some helper implementations like `isAdmin`, `isAnyAdmin` and `isAnyAdminOrBroadcaster` which can be set on the params objects. WDYT?

packages/api/src/middleware/auth.js
victorges: Good catch. I could swear I read somewhere it returned an empty object when the parsing failed, but apparently it's not the case.

darkdarkdragon: Empty object will not work too - then `db.apiToken.get` will throw error because it expect id to be provided.

victorges: I see. Alright so I can't even coalesce to an empty string here, I'll do an explicit check.

victorges: While we're at it, do you know where this `req.token` property comes from? I coudn't find it on express docs nor any code of ours setting it explicitly.

victorges: Nevermind just found it: https://github.com/livepeer/livepeer-com/blob/d903f4e5761b05bd7469c925c8a4201f558ca38e/packages/api/src/app-router.js#L120

Maybe I'll use the [`express-basic-auth`](https://www.npmjs.com/package/express-basic-auth) instead of this more manual `basic-auth` then, for consistency.

packages/api/src/controllers/stream.js
victorges: Implemented the idea we agreed offline :)

packages/api/src/middleware/auth.js
victorges: Fixed. I'll check again later if it's worth migrating to the `express-basic-auth`, it seems to do way more than just parsing the headers so it may be overkill.

darkdarkdragon (changes requested):
1. Can we remove last usage of `req.store` from `auth.js`?
2. Can we fix tests?

darkdarkdragon (approved):
LGTM

packages/api/src/middleware/auth.js
```diff
@@ -11,74 +19,70 @@ function authFactory(params) {
...
+    } else if (["Bearer", "Basic"].includes(tokenType)) {
+      const isBasic = tokenType === "Basic";
+      const tokenId = isBasic ? basicUser?.pass : req.token;
+      if (!tokenId) {
+        throw new ForbiddenError(`no authorization token provided`);
+      }
+      tokenObject = await db.apiToken.get(tokenId);
+      const matchesBasicUser = tokenObject?.userId === basicUser?.name;
+      if (!tokenObject || (isBasic && !matchesBasicUser)) {
+        throw new ForbiddenError(`no token object ${tokenValue} found`);
```
iameli: Nice, this will finally get rid of the really awkward `no token object undefined found`

iameli (approved):
LGTM also. Let's make the go-livepeer changes and merge at the start of the load test session?

victorges:
@iameli Yeah! Just created some tests for this one as well and will leave the go-livepeer-infra part ready for the load test now

PR #476, opened 2021-06-24
refact(api): Make streams/sessions controllers typescript

victorges:
This adds typing to a lot of stuff, using `/stream` API controller as starting point.

Starting point because, on the way to typing that controller as much as possible,
I also bumped into a couple helpers or dependencies that also weren't typed yet
so I needed to go deeper.

Also typed all of its tests, which was mostly some repetitive changes, but also  ended
up creating new branches of typing stuff like all the `index.ts` and `parse-cli.ts` so the
test server/clients were fully typed as well.

 - Add type checking to build pipeline (including `tsconfig.json` for the package)
 - Migrate to typescript:
   - `controllers/stream.ts` and its tests
   - `controllers/session.ts`
   - `middleware/` `hardcocded-nodes` and `kubetrnetes`, and their respective tests
   - `parse-cli.ts` (I recommend reviewing this one from this commit: https://github.com/livepeer/livepeer-com/pull/476/commits/0297730f626b4f3b0221d65120dbff8656639696?w=1)
   - `util.ts` (which involved some hacky typing of isomorphic fetch, but I could migrate to node-fetch completely if preferred)
 - Also some typing and sometimes actual code fixes on the way
 
## -

`yarn && yarn test --silent`

I bet the actual code transpiled in `dist/` has changed very little. 
I could make that safety check as well, let me know.

I don't think so. Just some housekeeping in the project.

**Checklist:**


packages/api/src/parse-cli.ts
```diff
@@ -0,0 +1,220 @@
+import yargs from "yargs";
+import { CamelKeys } from "./types/common";
+
+function coerceArr(arg) {
```
victorges, line 4: Ugh just noticed Github didn't catch the rename here. I believe it's because the `parsed` intermediate variable in the `parceCli` function (used for typing purposes) caused a too-big indentation change.

Reviewer, let me know if you want me to put this change on a separate commit or something so it's more reviewable.

victorges: OK, you can check the real changes made to this file, together with the related changes on this commit: 

https://github.com/livepeer/livepeer-com/pull/476/commits/0297730f626b4f3b0221d65120dbff8656639696?w=1

Link already ignoring whitespace changes which are basically prettier itself 👌

victorges:
~Will fix the build/tests in the CI in a sec. I think this is ready for review already anyway.~
Fixed!

packages/api/Dockerfile
```diff
@@ -6,19 +6,26 @@ RUN npm install -g --force yarn
...
 ENV NODE_ENV production
+
 ARG DOCKER_TAG
 ENV DOCKER_TAG ${DOCKER_TAG}
 ARG GITHUB_SHA
 ENV GITHUB_SHA ${GITHUB_SHA}
+
 ADD package.json .
-RUN yarn install
+RUN yarn install --frozen-lockfile
```
iameli, line 24: Hmm, I'm surprised this works absent the presence of a yarn.lock file.

packages/api/src/controllers/session.ts
```diff
@@ -40,9 +45,25 @@ const fieldsMap = {
...
+  for (const [key, value] of Object.entries(obj)) {
+    strObj[key] = value.toString();
+  }
+  return strObj;
+}
+
 app.get("/", authMiddleware({}), async (req, res, next) => {
-  let { limit, cursor, all, order, filters, userId, parentId, count } =
-    req.query;
+  let {
```
iameli, line 57: This just for typescript purposes? What would these things come in as if not for strings?

packages/api/src/middleware/hardcoded-nodes.ts
```diff
@@ -22,17 +39,17 @@ export default function hardcodedNodes({
...
-      req.getOrchestrators = async () => orchestrators;
+      req.getOrchestrators = async () => orchestrators as NodeAddress[];
     }
     if (!req.getIngest) {
-      req.getIngest = async () => ingest;
+      req.getIngest = async () => ingest as Ingest[];
     }
-    if (!req.prices) {
-      req.getPrices = async () => prices;
+    if (!req.getPrices) {
```
iameli, line 50: Hm, this check must have never worked with the wrong name there. I also don't really know what it's for. @yondonfu @kyriediculous?

packages/api/src/test-server.ts
```diff
@@ -1,14 +1,15 @@
...
-
-import argParser from "./parse-cli";
-import makeApp from "./index";
 import fs from "fs-extra";
-import uuid from "uuid/v4";
+import { v4 as uuid } from "uuid";
 import path from "path";
 import os from "os";
-import fetch from "isomorphic-fetch";
+import "isomorphic-fetch";
```
iameli: I know this works but dumping stuff on the global scope is a little frightening. Want to just use `node-fetch` instead?

iameli (approved):
Sweet. Only actual comment is about the awkwardness around `isomorphic-fetch`.

packages/api/Dockerfile
victorges: lol good catch. I'll add the yarn.lock here too, also expected it would have given me an error there.

darkdarkdragon (approved):
LGTM

packages/api/src/controllers/session.ts
victorges: This is actually not only a type issue, but the typing exposes a possible issue in code that only uses a `req.query.param` assuming it is always a `string`. The newer `@types/express` have a proper type for the `req.query` property which exposes the issue:
```
interface ParsedQs { [key: string]: undefined | string | string[] | ParsedQs | ParsedQs[] }
```

Which basically means that a `req.query` can contain a string, an array of strings or even recursive `ParsedQs` objects or array of those objects. That can happen on these cases:
 - `?cursor=abc&cursor=cde`
 - `?cursor[]=abc`
 - `?cursor[k1][k2]=value`
 - `?cursor[][k1][k2]=value` (?)

Long explanation here: https://evanhahn.com/gotchas-with-express-query-parsing-and-how-to-avoid-them/

Since I've found that issue, I've been using the `toString()` approach that I found somewhere, which seems the easier/safer to me. It doesn't explicitly give support to all possible corner cases that one might receive, but we probably don't want to give support to those anyway. It will work as expected if 1 value is sent or if an array is sent with only 1 element. In other cases it will generate a string that will cause an error somewhere else in the code (probably a 400, 422 or 404), but at least it won't explode a 500 due to possibly mistreating the object types. 

Anyway, if you feel like we should better handle all the `ParsedQs` possibilities let me know and I can improve that helper, maybe put it into a middleware which populates the `req` or at least in a common helper, idk. I didn't include that in this PR already as it was a bit massive already with the type changes.

packages/api/src/middleware/hardcoded-nodes.ts
victorges: I think it really has never worked indeed, always overwriting what was in the request. I think it wasn't really an issue though because no other middleware overwrites that property. References in repo: https://github.com/livepeer/livepeer-com/search?q=getPrices

The `kubernetes` one actually [only sets `getBroadcasters` and `getOrchestrators`](https://github.com/livepeer/livepeer-com/blob/bd7528e04a91310fe2baa833d4504df0f96dde70/packages/api/src/middleware/kubernetes.ts#L107) so I'm guessing we only use the "hardcoded" ones configured on `LP_PRICES` env [in livepeer-infra](https://github.com/livepeer/livepeer-infra/blob/f78b771069043f14bc6fa79145cfdb392aee46fd/values/prod/livepeer.values.yaml#L5-L10)

packages/api/src/test-server.ts
victorges: Switched to `node-fetch`!

Answering the comment though: It actually doesn't make a difference which way we import the library, as it always messes with the global scope when imported. So I had switched to not use the default export as I thought it was a bit misleading, using the value it sets on the global scope instead. I guess this import worked like a "blank identifier import" in go.

In fact:
 - `global.fetch` and `require("isomorphic-fetch")` are the exact same value as they do [`global.fetch = module.exports;` here](https://github.com/matthew-andrews/isomorphic-fetch/blob/master/fetch-npm-node.js#L12) (although we could argue that someone else might change the `global.fetch` after the lib has run)
 - Also learned that `isomorphic-fetch` basically exposes `node-fetch` on the global scope, doing like 1 `//`=>`https://` conversion before literally proxying to `node-fetch`. So the type hack I made doesn't even feel that much like a hack anymore 😆 (but it's much better to just use `node-fetch` directly anyway)

victorges: Btw, the whole typing problem with `isomorphic-fetch` is that [their DefinitelyTyped package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/isomorphic-fetch/index.d.ts#L8) assumes that it's going to run on a browser and have the `fetch` type in the global scope. So we then also had to transpile our node app with the `dom` module, which could lead to programming errors if you use something from the browser not available in node. Kinda non-sense if you ask me, given the purpose of this isomorphic package, but I'm also not a TS expert to know how to fix it easily on their repository. So let's just stick with `node-fetch` for now 😬

packages/api/Dockerfile
victorges: So I fixed this by running the `docker build` with the context of the repository root instead, so I could copy the `yarn.lock` from the root `package.json`. Felt a bit awkward to be honest, wish there was an easier way, but I think it works fine. Let me know if you have a better idea or rather keep it the way it was, without `yarn.lock` (seems risky tho). I didn't really need this change to get the `tsconfig.json` to work, so I can revert the other changes I included.

Here's the commit: https://github.com/livepeer/livepeer-com/pull/476/commits/52dd0aff7f57c0bebdf231597ab700343cc9ef4b

PR #480, opened 2021-06-24
Finish filters for streams and sessions table | Bring back API Keys table

### LINKED ISSUE #478: Complete dashboard streams + sessions table filters
Streams Table Remaining Filters
- Created (date range)
- Last seen (date range)
- Status (boolean)

Sessions Table Remaining Filters
- Name (containing string)
- Created (date range)
- Duration ("greater than" or "less than" some number)
- Status (boolean)

### LINKED ISSUE #481: [dashboard redesign] Add back API Keys table
https://github.com/livepeer/livepeer-com/blob/as/dashboard-redesign/packages/www/components/Dashboard/TokenTable/index.tsx#L194

julianbenegas:
closes #478 
closes #481


adamsoffer (changes requested):
image

Can we remove the empty row when "Is Active" is checked?

Or if it's easier, add two checkboxes like this:
image

packages/www/components/Dashboard/TokenTable/index.tsx
```diff
@@ -130,77 +129,153 @@ const TokenTable = ({
...
-        align="end"
-        justify="between"
-        css={{
-          mb: "$5",
-        }}>
-        <Heading size="2" css={{ fontWeight: 600 }}>
-          {title}
-        </Heading>
+    <>
+      {/* TODO fix this before merging! @julianbenegas */}
```
julianbenegas:
```suggestion
```

julianbenegas:
> Can we remove the empty row when "Is Active" is checked?
> 
> Or if it's easier, add two checkboxes like this:

@adamsoffer The second option is best, because the first option "Can we remove the empty row when "Is Active" is checked?" would not cover all the filter cases. (`undefined`, `true`, `false`).

But, on the stripe design, they look like checkboxes, but they are actually radio buttons 🤔

adamsoffer:
> But, on the stripe design, they look like checkboxes, but they are actually radio buttons 🤔

Oh yeah, radio buttons work too. Can look something like this maybe? Whatever UX you prefer.

image

adamsoffer:
Nice. Let's just convert those radios back to checkboxes. And by default neither "Active" nor "Idle" should be checked when  "Status" is checked.
image

adamsoffer:
Going to merge this so you can create a new branch for the remaining pieces (adding total results, initializing the cursor so pagination works, and adjusting the status dropdown).

adamsoffer (approved):

PR #483, opened 2021-06-25
feat(api): Allow restricting access for API tokens via resource/method rules

victorges:
**Note:** remember to toggle "hide whitespace changes" when reviewing cause the prettier changed indentation all around

This is to add support for creating API tokens that have limited access to the API.

It can be used for things like read-only API tokens or well-scoped tokens that only have access
to a specifc API or API group.

e.g.: A token that would only give access to `POST /api/stream/hook/*`, `POST /api/usage/update` 
and `PUT /region/:id`:
```ts
{
  access: {
    rules: [
      {
        resources: ["stream/hook/*", "usage/update"],
        methods: ["post"],
      },
      { resources: ["region/:id"], methods: ["put"] },
    ],
  }
}
```

 - New logic in auth middleware to authorize access rules on API tokens, if present
 - New logic in api-token API to validate access rules on tokens being created
 - Lots of tests on both of those
 - Some typescript changes and utilities that I ended finding helpful on the way

### How did you test each of these updates
 - `yarn test --silent`
 -  Ran locally and tested with 4 api tokens:
    - legacy: no access field configuration, should have all access
    - all-access: * * access, should have all access as the legacy above
    - read-only: access to GET methods only
    - infra: access to only the stream/hook, usage/update and region/:id apis mentioned above

All worked well 😌 

Implements: https://github.com/livepeer/livepeer-com/issues/491

**Checklist:**


victorges:
@livepeer/video-user This is now ready for review!

packages/api/src/middleware/auth.ts
```diff
@@ -73,6 +104,13 @@ function authFactory(params) {
     if ((params.admin && !isUIAdmin) || (params.anyAdmin && !user.admin)) {
       throw new ForbiddenError(`user does not have admin priviledges`);
     }
+    const accessRules = tokenObject?.access?.rules;
+    if (accessRules) {
+      const { httpPrefix } = req.config;
+      if (!isAuthorized(req.method, req.path, accessRules, httpPrefix)) {
```
darkdarkdragon: in `app.post("/hook/detection", ...` handler `req.path` is equal to `/hook/detection`
this means this whole system in the current state will not work at all.

darkdarkdragon (changes requested):

packages/api/src/middleware/auth.ts
victorges: Good catch, thanks. I'll try to make this work using the `req.baseUrl` or `req.originalUrl` instead and add some tests to catch that scenario.

victorges: Done!

victorges: Rebased now so Github might have f'ed up the inceremental changes for you. The only commits I added since you last looked are these:
 - https://github.com/livepeer/livepeer-com/pull/483/commits/1e12ecb1cc6c1a90057d6e23132cec38ca293cb5
 - https://github.com/livepeer/livepeer-com/pull/483/commits/315ba1c99c73c32768825d981d14d8d6f155d591

packages/api/src/store/errors.ts
```diff
@@ -3,6 +3,14 @@ class APIError extends Error {
   status: number;
 }
 
+export function isAPIError(err: any): err is APIError {
+  if (typeof err?.status !== "number") {
+    return false;
+  }
+  const st = err.status;
+  return st % 1 === 0 && st >= 400 && st < 600;
```
iameli, line 11: Hah, nice integer check in case someone does `error.status = 404.5` 😃

darkdarkdragon (approved):
LGTM

iameli (approved):
LGTM also. How would you feel about issuing new properly-scoped tokens for our prod webhooks and putting them in Vault?

victorges:
@iameli Sure, I can do that! Just not sure how I can test all the flows to make sure they work properly (especially the other ones apart from `/stream/hook/*`). 

Are you thinking of 1 api key for all those APIs, as I exemplified in the PR description? Or only the one specifically for `/stream/hook/*` for now?

victorges:
Just done on staging:
- Created the `infraservice@livepeer.org` account, made it an admin
- Created an API key with:
```
curl --location --request POST 'https://livepeer.monster/api/api-token' \
--header 'Authorization: JWT <infraservice JWT>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "name": "PRODUCTION KEY FOR STREAM HOOKS AND ORCH CRON DON'T DELETE",
    "access": {
        "rules": [
            {
                "resources": ["stream/hook/*"],
                "methods": ["post"]
            },
            {
                "resources": ["region/:id"],
                "methods": ["put"]
            }
        ]
    }
}'
```
- Checked token really only works for the specified APIs
- Updated it on staging vault, both for admin key used in orch cron and basic auth for webhooks
- Re-deployed staging stuff with `ansible-playbook ...deploy-staging.yaml`
- Checked token was working for `/region/:id` by orchs cronjob via `kubectl` and dashboard's `lastActive`
- Checked token was working for `/stream/hook` with successful stream through `stream-tester`
- Checked token was working for `/stream/hook/detection` by testing detection webhook from [webhook.site](https://webhook.site/#!/6009dc66-5302-4ab7-afde-722da8d887a8/e8ca4d9c-413f-4a66-9887-07ddcec6c702/1). Checked lastActive kept up to date as stream was live

I'm now confident about taking this to production, let's do it next week :D

PR #494, opened 2021-06-28
Convert swr to react query and pagination bug

### LINKED ISSUE #499: Make sure pagination works on tables
Initializing the cursor should fix this

emilianoyaryura:
This closes #499


adamsoffer (changes requested):
By default, `Active` and `Idle` shouldn't be checked when you check `Status` and you should be able to check both `Active` and `Idle`, not only one or the other.

1. If neither are checked, display all streams.
2. If both are checked, display all streams.
3. If only `Active` is checked, only display active streams.
4. If only `Idle` is checked, only display idle stream.s

adamsoffer:
@emilianoyaryura looks like [this line](https://github.com/livepeer/livepeer-com/blob/088ee1c26f45ebcdff6cba34e04ddf7bd9e9499e/packages/www/components/Dashboard/AllSessionsTable/index.tsx#L220) is breaking the build.

julianbenegas:
@emilianoyaryura based on our sprint check in:

- [ ] revert back to radio buttons
- [ ] default to Active state
- [ ] Remove swr lib (if we are not using it in other places)

adamsoffer:
@emilianoyaryura is this ready to merge once you fix the build?

adamsoffer:
@emilianoyaryura I noticed when you navigate from the home page to the streams page, the streams page is displaying the results from the home page (only five). Only when you refresh does it display the correct amount (20). Can you look into this?

adamsoffer (approved):
LGTM

PR #504, opened 2021-07-01
Make webhooks able to subscribe to multiple events at once

### LINKED ISSUE #502: webhooks: add `events` array field
We'd like folks to be able to sign up for more than one webhook event in the same hook:
```
{
  events: ['stream.started', 'stream.detection']
}
```
For backwards compatibility, `event` should still work when creating a webhook, but we should also update any pertinent documentation to reference `events` instead of `event`.

victorges:
This is to change the webhooks spec a bit by allowing them to subscribe to multiple
events at once with an `events` array instead of a single `event` subs.

We still want to keep retrocompatibility in the API since some documents have already 
been distributed where the `event` field is used instead of `events`, but we can
actually break the backward compatibility in the database for now, since the only
webhooks that exist are ones created by Livepeer employeed for testing. 

So we'll just deploy this change and then truncate the existing prod and staging
tables so we don't have webhooks in the legacy format saved anywhere (even though
they wouldn't cause much trouble anyway, would just never be used).

For simplicity, we also don't have any index in the `events` field, but it is possible to
be made in the future with [a GIN index](https://www.postgresql.org/docs/9.4/datatype-json.html) on the specific array field, which can be used
for this `@>` operator. It didn't "just work" when I experimented with it, so I kept the simple
version with no index for now.

## -

`yarn test`

Closes https://github.com/livepeer/livepeer-com/issues/502

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -34,8 +34,8 @@ components:
       table: webhook
       required:
         - name
-        - event
         - url
+      additionalProperties: false
       properties:
         id:
           type: string
@@ -58,12 +58,18 @@ components:
             Timestamp (in milliseconds) at which stream object was created
           example: 1587667174725
         event:
-          type: string
-          enum:
-            - stream.started
-            - stream.detection
-            - stream.idle
-            - recording.ready
+          index: true
+          writeOnly: true
+          $ref: "#/components/schemas/webhook/properties/events/items"
+        events:
+          type: array
+          items:
+            type: string
+            enum:
+              - stream.started
+              - stream.detection
+              - stream.idle
+              - recording.ready
         url:
           type: string
           format: uri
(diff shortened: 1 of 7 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -58,12 +58,20 @@ components:
...
           example: 1587667174725
         event:
-          type: string
-          enum:
-            - stream.started
-            - stream.detection
-            - stream.idle
-            - recording.ready
+          writeOnly: true
+          deprecated: true
```
iameli, line 62: Nice.

iameli (approved):
LGTM!

adamsoffer:
Nice!

victorges:
FTR, after deploying we need to erase the webhook table from postgres. Some step by step for safety:
 - Grab postgres IP and password from vault `LP_POSTGRES_URL`
 - Connect to postgres with`psql -h $POSTGRES_IP -U postgres`, input pwd from vault
 - Verify only internal users have webhooks created in the DB with: `SELECT DISTINCT users.data->>'email' FROM Users JOIN Webhook hook ON hook.data->>'userId' = users.id;`
 - If that's confirmed actually delete/truncate/erase the table with: `DELETE FROM Webhook;` (scary)
 - Profit

Issue #362, opened 2021-04-20
Verifiable webhook signatures


wohlner:
### Show Signing Secret in Livepeer.com Dashboard 
([related to this ticket](https://app.zenhub.com/workspace/o/livepeer/livepeer-com/issues/357))
![SIGNING SECRET hidden, Webhooks MVP in Livepeer.com Dashboard.jpg](https://images.zenhubusercontent.com/5d51866c4b419616253bda2e/ab427485-68cd-4408-9326-61c27f0e923b)![SIGNING SECRET revealed, Webhooks MVP in Livepeer.com Dashboard.jpg](https://images.zenhubusercontent.com/5d51866c4b419616253bda2e/242d2108-3d2a-47df-b17f-31278c0f34c5)

### How Signing Secrets work
Livepeer.com will include a signature in each event’s `Livepeer-Signature` header.  The timestamp is prefixed by `t=` and the signature is prefixed by a `scheme`. Schemes start with `v`, followed by an integer. Currently, the only valid signature scheme is `v1`. Livepeer generates signatures using `HMAC` with `SHA2-256`.

```bash
Livepeer-Signature: t=36285904404,v1=88f3ff0fds9sf8a98vb0b096e81507cfd5c932fc17cf63a4a55566fd38da3a2d3d2`
```

To validate the signature, take the following steps:

**Step 1: Extract the timestamp and signatures from the header**

Split the header, using the `,` character as the separator, to get a list of elements. Then split each element, using the `=` character as the separator, to get a prefix and value pair. The value for the prefix `t` corresponds to the timestamp, and `v1` corresponds to the signature (or signatures). You can discard all other elements.

**Step 2: Prepare the `signed_payload` string**
The `signed_payload` string is created by concatenating:

- The timestamp (as a string)
- The character `.`
- The actual JSON payload (i.e., the request body)

**Step 3: Determine the expected signature**
Compute an `HMAC` with the `SHA256` hash function. Use the endpoint’s signing secret as the key, and use the `signed_payload` string as the message.

**Step 4: Compare the signatures**
Compare the signature (or signatures) in the header to the expected signature. For an equality match, compute the difference between the current timestamp and the received timestamp, then decide if the difference is within your tolerance.

To protect against timing attacks, use a constant-time string comparison to compare the expected signature to each of the received signatures.

```php
/**
 * Verify the signature (laravel)
 *
 * @param Request $request
 * @return boolean
 */
protected function verifySignature(Request $request)
{
[...]

PR #508, opened 2021-07-02
[READY] hmac signature for webhook payload

ya7ya:

implements #362 

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -79,14 +79,9 @@ components:
         deleted:
           type: boolean
           default: false
-        blocking:
-          type: boolean
-          default: true
-          description:
-            If true, returning non 2xx value from webhook will prevent stream
-            from starting
         sharedSecret:
           type: string
+          writeOnly: true
           description: shared secret used to sign the webhook payload
 
     webhook-response:
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -85,6 +85,9 @@ components:
           description:
             If true, returning non 2xx value from webhook will prevent stream
             from starting
+        sharedSecret:
+          type: string
+          description: shared secret used to sign the webhook payload
 
     webhook-response:
       type: object
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -163,7 +163,6 @@ export default class WebhookCannon {
       console.log("preparing to fire webhook ", webhook.url);
       const timestamp = Date.now();
       // go ahead
-      let timestamp = Date.now();
       let params = {
         method: "POST",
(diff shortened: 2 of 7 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -77,6 +77,9 @@ components:
           description:
             If true, returning non 2xx value from webhook will prevent stream
```
victorges: Unrelated: Should we remove this `blocking` property now that we don't support it?

packages/api/src/schema/schema.yaml
```diff
@@ -77,6 +77,9 @@ components:
...
             from starting
+        sharedSecret:
+          type: string
```
victorges, line 83: Better make this field `writeOnly` to avoid returning it from the read APIs?

packages/api/src/webhooks/cannon.ts
```diff
@@ -173,6 +174,14 @@ export default class WebhookCannon {
         }),
       };
 
+      // sign payload if there is a webhook secret
+      if (webhook.sharedSecret) {
+        let signature = sign(event.payload, webhook.sharedSecret);
+        params.headers['Livepeer-Signature'] = `t=${Date.now()},v1=${signature}`
```
victorges: Some considerations:
 - Did you mean to send `params.body` here instead of `event.payload` (this last one is probably `undefined` for most events but `stream.detection`). The `params.body` is the one that is a `string` as well and that would work on the `Buffer.from` call.
 - Should we include the timestamp inside the signed payload? Otherwise, a replay attack could be possible here by re-using the signature but setting new timestamps. I guess the simplest way for that is to include a field in the `params.body` with the timestamp. I think it's a good approach in order to keep a simple interface for signature verification as well.

WDYT?

darkdarkdragon: Agree with both points.

darkdarkdragon (approved):
LGTM after rebasing

packages/api/src/controllers/stream.test.js
```diff
@@ -1004,6 +1004,7 @@ describe("controllers/stream", () => {
           expect(res.status).toBe(204);
 
           await hookSem.wait(1000);
+          delete hookPayload.time;
           expect(hookPayload).toEqual({
             id: expect.stringMatching(uuidRegex),
             event: "stream.detection",
```
victorges: **nit:** I think this could work too:
```suggestion
          expect(hookPayload).toEqual({
            id: expect.stringMatching(uuidRegex),
            event: "stream.detection",
            time: expect.any(Number),
```
(horrible diff from github, I only stopped deleting the `time` field and started checking it with `expect.any(Number)`)

victorges (approved):
LGTM!

PR #513, opened 2021-07-04
webhook payload improve

darkdarkdragon:
send event.id and event.createdAt in webhook's body

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -22,6 +22,7 @@ const MAX_RETRIES = 20;
 export interface WebhookMessage {
   id: string;
   event: DBWebhook["events"][0];
+  createdAt: number,
   userId: string;
   streamId: string;
   payload?: Object;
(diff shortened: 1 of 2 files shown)


packages/api/src/webhooks/cannon.ts
```diff
@@ -158,6 +159,7 @@ export default class WebhookCannon {
       );
     } else {
       console.log("preparing to fire webhook ", webhook.url);
+      const ts = Date.now();
```
victorges: Since this field is going on the public payload, WDYT of using the full name like `timestamp`? Or maybe `date` like the HTTP header, but I think I prefer `timestamp`

packages/api/src/webhooks/cannon.ts
```diff
@@ -167,7 +169,11 @@ export default class WebhookCannon {
         },
         timeout: WEBHOOK_TIMEOUT,
         body: JSON.stringify({
-          id: webhook.id,
+          id: event.id, // this will allow receiver to check if it is already received same hook
```
victorges, line 172: Nice

victorges (approved):
lgtm

packages/api/src/webhooks/cannon.ts
darkdarkdragon: Renamed to `timestamp`

PR #514, opened 2021-07-06
add webhook endpoint for deleting multiple more than 1

adamsoffer:
This PR adds another webhook endpoint for deleting more than one at a time.

### DIFF
--- a/packages/api/src/controllers/webhook.js
+++ b/packages/api/src/controllers/webhook.js
@@ -49,16 +49,8 @@ const fieldsMap = {
 };
 
 app.get("/", authMiddleware({}), async (req, res) => {
-  let {
-    limit,
-    cursor,
-    all,
-    event,
-    allUsers,
-    order,
-    filters,
-    count,
-  } = req.query;
+  let { limit, cursor, all, event, allUsers, order, filters, count } =
+    req.query;
   if (isNaN(parseInt(limit))) {
     limit = undefined;
   }
@@ -218,6 +210,31 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
     console.error(e);
     throw e;
   }
+  res.status(204);
+  res.end();
+});
+
+app.delete("/", authMiddleware({}), async (req, res) => {
+  if (!req.body || !req.body.ids || !req.body.ids.length) {
+    res.status(422);
+    return res.json({
+      errors: ["missing ids"],
+    });
+  }
+  const ids = req.body.ids;
+
+  if (!req.user.admin) {
+    const webhooks = await db.webhook.getMany(ids);
+    if (
+      webhooks.length !== ids.length ||
+      webhooks.some((s) => s.userId !== req.user.id)
+    ) {
+      res.status(404);
+      return res.json({ errors: ["not found"] });
+    }
+  }
+  await db.webhook.markDeletedMany(ids);
+
   res.status(204);
   res.end();
 });
--- a/packages/api/src/controllers/webhook.js
+++ b/packages/api/src/controllers/webhook.js
@@ -222,7 +222,6 @@ app.delete("/", authMiddleware({}), async (req, res) => {
       errors: ["missing ids"],
     });
   }
-  const ids = req.body.ids;
 
   if (!req.user.admin) {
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/webhook.js
```diff
@@ -49,16 +49,8 @@ const fieldsMap = {
...
-    limit,
-    cursor,
-    all,
-    event,
-    allUsers,
-    order,
-    filters,
-    count,
-  } = req.query;
+  let { limit, cursor, all, event, allUsers, order, filters, count } =
```
victorges, line 52: What prettier config are you using? When I save this file my prettier does the exact opposite change 😆 
https://github.com/livepeer/livepeer-com/pull/504/files#diff-d76010b6a95682958e6201f3d0e7b8991d28cc648f7f166baf44b783f6796974R52

I don't recall configuring anything explicitly tho, maybe we're not setting an "official" line limit in the `@livepeer/prettier-config` and then it's grabbing the user's default?

packages/api/src/controllers/webhook.js
```diff
@@ -222,4 +214,29 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
   res.end();
 });
 
+app.delete("/", authMiddleware({}), async (req, res) => {
+  if (!req.body || !req.body.ids || !req.body.ids.length) {
```
victorges: **nit:**
```suggestion
  const ids = req.body?.ids;
  if (!Array.isArray(ids)) {
```

Also what happens if not all IDs are strings? If it just leads to a 404 I guess it's fine, but idk if something will explode or be exposed in the getMany or markDeleted query 🤔 

Edit: Maybe we could do this instead just in case and to avoid the schema type overhead:
```suggestion
  const ids = req.body?.ids;
  if (!Array.isArray(ids) || !ids.every(id => typeof id === "string")) {
```

packages/api/src/controllers/webhook.js
```diff
@@ -222,4 +214,29 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
...
+      errors: ["missing ids"],
+    });
+  }
+  const ids = req.body.ids;
+
+  if (!req.user.admin) {
+    const webhooks = await db.webhook.getMany(ids);
+    if (
+      webhooks.length !== ids.length ||
+      webhooks.some((s) => s.userId !== req.user.id)
```
victorges, line 230: nice

packages/api/src/controllers/webhook.js
```diff
@@ -222,4 +214,29 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
...
+    ) {
+      res.status(404);
+      return res.json({ errors: ["not found"] });
+    }
+  }
+  await db.webhook.markDeletedMany(ids);
```
victorges, line 236: Mind updating the single-delete API above to use the `markDeleted` helper as well?

And btw, I think I have been moving around the API code without internalizing that we actually never delete objects from DB (had heard about it tho). So sometimes they may just come as a successful result but with a `deleted` field inside 😬 

In that sense, should we check for `deleted` on line 230 as well, to disallow "re-deleting" a webhook and giving a 404 instead? Similarly to the `get` behavior I guess

victorges (approved):
LGTM!

packages/api/src/controllers/webhook.js
adamsoffer: ah interesting. I'm not sure why that is. I'm using the repo's `.prettierrc`.

packages/api/src/controllers/webhook.js
adamsoffer: Nice one. Just pushed up this change.

packages/api/src/controllers/webhook.js
adamsoffer: Good call. I'll merge this PR as is and then do you want to create a separate follow up PR with these improvements?

victorges: Sure :)

adamsoffer: Thanks!

PR #519, opened 2021-07-07
Fix some sortings, and add fast redirects when not logged in

julianbenegas:


packages/www/hooks/use-api.tsx
```diff
@@ -87,6 +88,24 @@ const getStoredToken = () => {
   }
 };
 
+export const DashboardRedirect = () => {
+  return (
+    <Head>
+      <script
```
adamsoffer, line 94: Nice.

adamsoffer (approved):
Nice work!

PR #529, opened 2021-07-14
[READY] webhook sig verification tests

ya7ya:

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/webhook.js
+++ b/packages/api/src/controllers/webhook.js
@@ -31,7 +31,7 @@ function validateWebhookPayload(id, userId, createdAt, payload) {
     name: payload.name,
     events: payload.events ?? [payload.event],
     url: payload.url,
-    sharedSecret: payload.sharedSecret?? payload.sharedSecret,
+    sharedSecret: payload.sharedSecret,
   };
 }
 
--- a/packages/api/src/controllers/webhook.js
+++ b/packages/api/src/controllers/webhook.js
@@ -31,6 +31,7 @@ function validateWebhookPayload(id, userId, createdAt, payload) {
     name: payload.name,
     events: payload.events ?? [payload.event],
     url: payload.url,
+    sharedSecret: payload.sharedSecret?? payload.sharedSecret,
   };
 }
 
@@ -45,6 +46,7 @@ const fieldsMap = {
   createdAt: `webhook.data->'createdAt'`,
   userId: `webhook.data->>'userId'`,
   "user.email": `users.data->>'email'`,
+  sharedSecret: `webhook.data->>'sharedSecret'`,
 };
 
 app.get("/", authMiddleware({}), async (req, res) => {
@@ -154,7 +156,7 @@ app.get("/:id", authMiddleware({}), async (req, res) => {
   // get a specific webhook
   logger.info(`webhook params ${req.params.id}`);
 
-  const webhook = await req.store.get(`webhook/${req.params.id}`);
+  const webhook = await db.webhook.get(req.params.id);
   if (
     !webhook ||
     ((webhook.deleted || webhook.userId !== req.user.id) && !req.user.admin)
@@ -198,8 +200,7 @@ app.put(
 
 app.delete("/:id", authMiddleware({}), async (req, res) => {
   // delete a specific webhook
-  const webhook = await req.store.get(`webhook/${req.params.id}`);
-
+  const webhook = await db.webhook.get(req.params.id);
   if (
     !webhook ||
     ((webhook.deleted || webhook.userId !== req.user.id) && !req.isUIAdmin)
--- a/packages/api/src/controllers/webhook.test.js
+++ b/packages/api/src/controllers/webhook.test.js
@@ -108,10 +108,8 @@ describe("controllers/webhook", () => {
     });
 
     it("create a webhook", async () => {
-      // // console.log('mockWebhook: ', mockWebhook)
       let res = await client.post("/webhook", { ...mockWebhook });
       let resJson = await res.json();
-      // console.log("webhook body: ", resJson);
       expect(res.status).toBe(201);
       generatedWebhook = resJson;
       generatedWebhookIds.push(resJson.id);
@@ -120,7 +118,6 @@ describe("controllers/webhook", () => {
         name: "test 2",
       });
       resJson = await res.json();
-      // console.log("webhook body: ", resJson);
       expect(res.status).toBe(201);
       expect(resJson.name).toBe("test 2");
       generatedWebhookIds.push(resJson.id);
@@ -131,7 +128,6 @@ describe("controllers/webhook", () => {
         name: "test non admin",
       });
       resJson = await res.json();
-      // console.log("webhook body: ", resJson);
       expect(res.status).toBe(201);
       expect(resJson.name).toBe("test non admin");
       generatedWebhookIds.push(resJson.id);
@@ -146,7 +142,6 @@ describe("controllers/webhook", () => {
       });
       expect(res.status).toBe(201);
       const resJson = await res.json();
-      // console.log('jsonRes: ', resJson);
       expect(resJson).toMatchObject(mockWebhook);
       generatedWebhookIds.push(resJson.id);
 
@@ -318,11 +313,9 @@ describe("controllers/webhook", () => {
 
       let localWebhook = { ...mockWebhook };
       localWebhook.url = "192.168.1.1";
-      // // console.log("localwebhook: ", localWebhook);
       // create webhook
       const webhookRes = await client.post("/webhook", { ...localWebhook });
       let webhookResJson = await webhookRes.json();
-      // // console.log("webhook created: ", webhookResJson);
       expect(webhookRes.status).toBe(422);
     });
   });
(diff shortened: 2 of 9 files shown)


packages/www/components/Dashboard/WebhookDialog/index.tsx
```diff
@@ -151,6 +155,16 @@ const WebhookDialog = ({
               placeholder="https://"
               onChange={(e) => setWebhookUrl(e.target.value)}
             />
+
+            <Label htmlFor="sharedSecret">Secret</Label>
+            <TextField
+              size="2"
+              type="url"
```
adamsoffer: Input type should be "text"

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -102,6 +102,7 @@ describe("webhook cannon", () => {
       createdAt: Date.now(),
       events: ["stream.started"],
       url: "http://localhost:30000/webhook",
+      sharedSecret: "keyboardCat",
       // url: 'https://livepeer.com/'
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -316,11 +318,11 @@ describe("controllers/webhook", () => {
 
       let localWebhook = { ...mockWebhook };
       localWebhook.url = "192.168.1.1";
-      console.log("localwebhook: ", localWebhook);
+      // // console.log("localwebhook: ", localWebhook);
       // create webhook
       const webhookRes = await client.post("/webhook", { ...localWebhook });
       let webhookResJson = await webhookRes.json();
-      console.log("webhook created: ", webhookResJson);
+      // // console.log("webhook created: ", webhookResJson);
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -316,11 +318,11 @@ describe("controllers/webhook", () => {
 
       let localWebhook = { ...mockWebhook };
       localWebhook.url = "192.168.1.1";
-      console.log("localwebhook: ", localWebhook);
+      // // console.log("localwebhook: ", localWebhook);
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -145,6 +146,7 @@ describe("controllers/webhook", () => {
       });
       expect(res.status).toBe(201);
       const resJson = await res.json();
+      // console.log('jsonRes: ', resJson);
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -130,7 +131,7 @@ describe("controllers/webhook", () => {
         name: "test non admin",
       });
       resJson = await res.json();
-      console.log("webhook body: ", resJson);
+      // console.log("webhook body: ", resJson);
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -119,7 +120,7 @@ describe("controllers/webhook", () => {
         name: "test 2",
       });
       resJson = await res.json();
-      console.log("webhook body: ", resJson);
+      // console.log("webhook body: ", resJson);
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -107,10 +108,10 @@ describe("controllers/webhook", () => {
     });
 
     it("create a webhook", async () => {
-      // console.log('mockWebhook: ', mockWebhook)
+      // // console.log('mockWebhook: ', mockWebhook)
       let res = await client.post("/webhook", { ...mockWebhook });
       let resJson = await res.json();
-      console.log("webhook body: ", resJson);
+      // console.log("webhook body: ", resJson);
```
adamsoffer: Can remove commented out code

packages/api/src/controllers/webhook.test.js
```diff
@@ -107,10 +108,10 @@ describe("controllers/webhook", () => {
     });
 
     it("create a webhook", async () => {
-      // console.log('mockWebhook: ', mockWebhook)
+      // // console.log('mockWebhook: ', mockWebhook)
```
adamsoffer: Can remove commented out code

adamsoffer (changes requested):
Nice. Just left a couple comments.

packages/api/src/controllers/webhook.js
```diff
@@ -31,6 +31,7 @@ function validateWebhookPayload(id, userId, createdAt, payload) {
     name: payload.name,
     events: payload.events ?? [payload.event],
     url: payload.url,
+    sharedSecret: payload.sharedSecret?? payload.sharedSecret,
```
victorges: The coalesce here seems to be unnecessary. Does it have any effect?

```suggestion
    sharedSecret: payload.sharedSecret,
```

packages/api/src/controllers/webhook.js
```diff
@@ -45,6 +46,7 @@ const fieldsMap = {
   createdAt: `webhook.data->'createdAt'`,
   userId: `webhook.data->>'userId'`,
   "user.email": `users.data->>'email'`,
+  sharedSecret: `webhook.data->>'sharedSecret'`,
```
victorges, line 49: Should the shared secret even be query-able, if it is write-only? Does it get filtered out for non-admin users?

packages/api/src/controllers/webhook.test.js
```diff
@@ -47,6 +47,7 @@ beforeAll(async () => {
     kind: "webhook",
     events: ["stream.started"],
     url: "https://winter-darkness-88ea.livepeer.workers.dev/",
+    sharedSecret: "keyboardcat",
```
victorges, line 50: I'm wondering if you mean 
music cat
or
coding cat

packages/api/src/webhooks/cannon.ts
```diff
@@ -184,7 +184,7 @@ export default class WebhookCannon {
 
       // sign payload if there is a webhook secret
       if (webhook.sharedSecret) {
-        let signature = sign(params.body, webhook.sharedSecret);
+        let signature = await sign(params.body, webhook.sharedSecret);
```
victorges: I think the `sign` function can be made non-async instead, since it's code has no `async/await` logic

packages/www/components/Dashboard/WebhookDialog/index.tsx
```diff
@@ -151,6 +155,16 @@ const WebhookDialog = ({
...
+
+            <Label htmlFor="sharedSecret">Secret</Label>
+            <TextField
+              size="2"
+              type="text"
+              id="sharedSecret"
+              defaultValue={Action.Update ? webhook?.sharedSecret : ""}
+              placeholder="secret used to sign the webhook requests"
+              onChange={(e) => setWebhookSecret(e.target.value)}
+            />
```
victorges: Would be nice if we could "suggest" a random secret on webhook creation, say like generating a UUID here and pre-populating the shared secret field 🤔 
Would make it less easy for users to create some potentially insecure secrets

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -163,6 +163,7 @@ describe("webhook cannon", () => {
       webhookServer.app.use(bodyParser.json());
       webhookServer.app.post("/webhook", (req, res) => {
         console.log("WEBHOOK WORKS , body", req.body);
+        expect(req.headers['livepeer-signature']).not.toBe(undefined);
```
victorges: Would be nice to check if the secret actually matches what we would expect. WDYT of?
```suggestion
        const signature = req.headers['livepeer-signature']
        expect(signature).toEqual(sign(JSON.stringify(req.body), mockWebhook.sharedSecret));
```
Or maybe we should use `bodyParser.text()` instead of `json` above and avoid potential errors from mismatching `JSON.stringify` ordering here.
```suggestion
        const signature = req.headers['livepeer-signature']
        expect(signature).toEqual(sign(req.body, mockWebhook.sharedSecret));
```
(it parses the raw body bytes into a `string`)

packages/api/src/controllers/webhook.test.js
ya7ya: haha. second one, that first cat knows what it's doing. it has to be random :D

packages/api/src/webhooks/cannon.ts
ya7ya: need the await or the signature would be `[Promise Object]` , try it out 🤯

packages/www/components/Dashboard/WebhookDialog/index.tsx
ya7ya: yes 💯 , will do

packages/api/src/webhooks/cannon.ts
victorges: I think that's because of the `async` in the `sign` function declaration [here](https://github.com/livepeer/livepeer-com/blob/18bdffc0d95d1aa3ecd779b7e8a817c834eff14b/packages/api/src/controllers/helpers.js#L18), then it forces the result to be a `Promise`. Since the implementation has no `await` though, I think that `async` can be removed and the typing would be just `string` here even without the `await`. That because `crypto.Hmac.digest` also returns a `string`, or at least thats what [the docs](https://nodejs.org/api/crypto.html#crypto_hmac_digest_encoding) say.

victorges (approved):
LGTM! (build is failing tho)

PR #530, opened 2021-07-14
api: Fix .dockerignore file

victorges:
This PR fixes `.doickerignore` file by renaming it so Docker can find it and adding some 
new files/folders to the ignore list. 

Context:
After the recent migration of some code to typescript, in which I also included the `tsc`
in the build pipeline for type checking, I had to make some changes to the API docker
file so that the `tsc` tool would be present. In that change I also made sure we included
the `yarn.lock` though, to avoid version discrepancies.

The problem was that with that change, I changed the docker context and now the
dockerignore file was being ignored and our docker build context increased to a whopping
2.1GB with the entire `node_modules` being copied around. This probably increased the
docker build time by almost a minute, at least on my machine.

 - Rename `packages/api/.dockerignore` to `packages/api/Dockerfile.dockerignore`
 This fixes the issue of the ignore file not being used since docker first searches for
 a file with the name `${dockerfile-path}.dockerignore`.
 - Enabled BuildKit builds on CI since the new dockerignore path needs that
 - Some changes to dockerignore file to ignore even more stuff

## -

`yarn docker:build`

I don't think so

**Checklist:**


iameli (approved):
Cool.

Something to consider soon: @livepeer-com/www needs to be its own container for livepeer-in-a-box. Also, we're soon going to have another package — the Livepeer.com TypeScript SDK — that's going to need to get built into the other packages. I'm kind of wondering if we should do a single top-level Dockerfile with a few different build targets so we can do one big install at the start for the creation of the two child images. Just a thought.

victorges:
After this fix I actually played around a bit with our CI, and I found out that our builds are taking very long not because of the time to install dependencies, but the time to build the front-end projects. Especially design-system and www from what I could tell, which spend almost 7 minutes on their own mostly on the `next build` steps. 

I tried playing around with the CI config and separating the builds ([example](https://github.com/livepeer/livepeer-com/runs/3070339941?check_suite_focus=true)). The API could be building+testing in around 2 minutes if we parallelize the docker build step, while doing a `yarn install` from the root of the project takes 7 additional minutes than only for the API (not because of the additional packages tho).

So I was thinking more in the terms of how could we separate these projects a little further, since in the end there are not many code dependencies between them anyway... So it feels like this single global docker could easily get pretty bulky/unmanageable as well without much benefit in it being aggregated across all projects? Not sure, maybe we're doing something wrong for the `next build` as well? Tried searching around but didn't find any easy flag like "parallelize the generation of all those 130 pages".

PR #533, opened 2021-07-19
recording.ready and recording.started

ya7ya:

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -904,6 +904,17 @@ app.put(
       userId: user.id,
     });
 
+  if (req.body.active === false && stream.record === true) {
+    // emit recording.started
+    req.queue.delayedEmit({
+      id: uuid(),
+      createdAt: Date.now(),
+      channel: "webhooks",
+      event:  "recording.ready",
+      streamId: id,
+      userId: user.id,
+    }, USER_SESSION_TIMEOUT)
+  }
     stream.isActive = !!req.body.active;
     stream.lastSeen = +new Date();
     const { ownRegion: region } = req.config;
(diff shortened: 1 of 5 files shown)


victorges (commented):
I think the `setRecordingStatus` function is only called, and always called, on GET requests for the stream. I think that means that:
 - the recording started/ready events would only really be emitted when the users themselves actually query for the stream, in which case the event seems unnecessary (they'd already receive the status in the response)
 - the events might also be sent multiple times, every time a new GET request is made which contains the stream as a result. 

That said, I think what this function actually does is something like `addRecordingFields` instead? Not sure if I misunderstood something or missed some other place where this function is called.

And also following up on the design, is there any other place where this could be instead? Some place that is actively called by the broadcaster somehow after starting or saving the recording to S3, idk.

darkdarkdragon:
@victorges 
> That said, I think what this function actually does is something like addRecordingFields instead? Not sure if I misunderstood something or missed some other place where this function is called.

You right, `recordingStatus` is a computed field, and it is computed only when user gets stream object.
Way to do `recording.ready` webhook is:
1. In /setactive, when it's called with active == false, if recording is true - emit delayed `recording.ready` event with delay set to `USER_SESSION_TIMEOUT`
2. In webhook cannon pull this session object and ran `setRecordingStatus` on it - and if it is ready only then fire actual webhook (I think payload for this hook should contain session id and `recordingUrl` and `mp4Url` fields).


P.S.
`recording.started` - is just `stream.started` for streams with `recording` on

victorges (approved):
LGTM!

darkdarkdragon:
@ya7ya @victorges this will not work - recording __could__ be ready after stream ended, but it is not necessarily so.  In webhook cannon there should be check added (call to `setRecordingStatus`)

victorges:
@darkdarkdragon But doesn't the `setRecordingStatus` just check if the time has passed anyway? Do you mean that the stream/session could become active again and because of that the recording url wouldn't be injected in the stream object? (`lastActive` updated to sth later than `eventTime+5 minutes`)

darkdarkdragon:
@victorges 
>  Do you mean that the stream/session could become active again and because of that the recording url wouldn't be injected in the stream object? 

exactly

PR #535, opened 2021-07-23
Ya/single webhook per url queues

ya7ya:

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -1487,7 +1487,8 @@ app.post(
         sceneClassification,
       },
     };
-    await req.queue.publish("events.sceneDetection", msg);
+
+    await req.queue.publish("events.streams", msg);
     return res.status(204).end();
   }
 );
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -121,7 +121,7 @@ async function triggerManyIdleStreamsWebhook(ids, queue) {
     ids.map(async (id) => {
       const stream = await db.stream.get(id);
       const user = await db.user.get(stream.userId);
-      queue.publish("events.stream", {
+      queue.publish("events.streams", {
         id: uuid(),
         createdAt: Date.now(),
         channel: "webhooks",
@@ -883,6 +883,14 @@ app.put(
 
     // trigger the webhooks, reference https://github.com/livepeer/livepeerjs/issues/791#issuecomment-658424388
     // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
+    await req.queue.publish("events.streams", {
+      id: uuid(),
+      createdAt: Date.now(),
+      channel: "webhooks",
+      event: req.body.active === true ? "stream.started" : "stream.idle",
+      streamId: id,
+      userId: user.id,
+    });
 
     if (!req.body.active && stream.record === true) {
       // emit recording.ready
@@ -899,7 +907,7 @@ app.put(
           const ingest = ((await req.getIngest()) ?? [])[0]?.base;
           const recordingUrl = getRecordingUrl(ingest, session);
           const mp4Url = getRecordingUrl(ingest, session, true);
-          req.queue.delayedPublish(
+          await req.queue.delayedPublish(
             "events.recording",
             {
               id: uuid(),
@@ -933,7 +941,7 @@ app.put(
         }
       }
       if (shouldEmit) {
-        req.queue.publish("events.recording", {
+        await req.queue.publish("events.recording", {
           id: uuid(),
           createdAt: Date.now(),
           channel: "webhooks",
@@ -1480,7 +1488,7 @@ app.post(
       },
     };
 
-    await req.queue.publish("events.streams", msg);
+    await await req.queue.publish("events.streams", msg);
     return res.status(204).end();
   }
 );
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -307,16 +307,6 @@ function setRecordingStatus(
       session.recordingUrl = getRecordingUrl(ingest, session);
       session.mp4Url = getRecordingUrl(ingest, session, true);
     }
-
-    // fire recording.ready webhook
-    req.queue.emit({
-      id: uuid(),
-      createdAt: Date.now(),
-      channel: "webhooks",
-      event:  isReady ? "recording.ready" : "recording.started",
-      streamId: session.parentId,
-      userId: session.userId,
-    })
   }
 }
 
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -121,7 +121,7 @@ async function triggerManyIdleStreamsWebhook(ids, queue) {
     ids.map(async (id) => {
       const stream = await db.stream.get(id);
       const user = await db.user.get(stream.userId);
-      queue.publish("events.streams", {
+      await queue.publish("events.streams", {
         id: uuid(),
         createdAt: Date.now(),
         channel: "webhooks",
@@ -1488,7 +1488,7 @@ app.post(
       },
     };
 
-    await await req.queue.publish("events.streams", msg);
+    await req.queue.publish("events.streams", msg);
     return res.status(204).end();
   }
 );
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -116,21 +116,19 @@ function validatePushTargets(
   );
 }
 
-async function triggerManyIdleStreamsWebhook(ids, queue) {
-  return Promise.all(
-    ids.map(async (id) => {
-      const stream = await db.stream.get(id);
-      const user = await db.user.get(stream.userId);
-      queue.emit({
-        id: uuid(),
-        createdAt: Date.now(),
-        channel: "webhooks",
-        event: "stream.idle",
-        streamId: stream.id,
-        userId: user.id,
-      });
-    })
-  );
+async function triggerManyIdleStreamsWebhook (ids, queue) {
+  return Promise.all(ids.map(async (id) => {
+    const stream = await db.stream.get(id);
+    const user = await db.user.get(stream.userId);
+    queue.publish("events.stream",{
+      id: uuid(),
+      createdAt: Date.now(),
+      channel: "webhooks",
+      event: "stream.idle",
+      streamId: stream.id,
+      userId: user.id,
+    });
+  }));
 }
 
 export function getRecordingUrl(ingest, session, mp4 = false) {
@@ -881,19 +879,19 @@ app.put(
       return res.json({ errors: ["user is suspended"] });
     }
 
-    // trigger the webhooks, reference https://github.com/livepeer/livepeerjs/issues/791#issuecomment-658424388
-    // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
+  // trigger the webhooks, reference https://github.com/livepeer/livepeerjs/issues/791#issuecomment-658424388
+  // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
 
-    // -------------------------------
-    // new webhookCannon
-    req.queue.emit({
-      id: uuid(),
-      createdAt: Date.now(),
-      channel: "webhooks",
-      event: req.body.active === true ? "stream.started" : "stream.idle",
-      streamId: id,
-      userId: user.id,
-    });
+  // -------------------------------
+  // new webhookCannon
+  req.queue.publish("events.stream",{
+    id: uuid(),
+    createdAt: Date.now(),
+    channel: "webhooks",
+    event:  req.body.active === true ? "stream.started" : "stream.idle",
+    streamId: id,
+    userId: user.id,
+  });
 
     if (!req.body.active && stream.record === true) {
       // emit recording.ready
@@ -1489,7 +1487,7 @@ app.post(
         sceneClassification,
       },
     };
-    await req.queue.emit(msg);
+    await req.queue.publish("events.sceneDetection", msg);
     return res.status(204).end();
   }
 );
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -314,7 +314,7 @@ function setRecordingStatus(
       createdAt: Date.now(),
       channel: "webhooks",
       event:  isReady ? "recording.ready" : "recording.started",
-      streamId: session.streamId,
+      streamId: session.parentId,
       userId: session.userId,
     })
   }
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -116,19 +116,21 @@ function validatePushTargets(
   );
 }
 
-async function triggerManyIdleStreamsWebhook (ids, queue) {
-  return Promise.all(ids.map(async (id) => {
-    const stream = await db.stream.get(id);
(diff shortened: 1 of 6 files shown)


packages/api/src/store/rabbit-queue.ts
```diff
@@ -24,16 +33,18 @@ export default class MessageQueue {
 
     this.channel = await this.connection.createChannel({
       json: true,
-      setup: async function (channel: Channel) {
+      setup: async (channel: Channel) => {
         await Promise.all([
-          channel.assertQueue(QUEUE_NAME),
+          channel.assertQueue(this.queues.events),
```
victorges: I think the queues should be made `{ durable: true }` as well, WDYT?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -24,16 +33,18 @@ export default class MessageQueue {
...
           // channel.assertQueue("delayedQueue", {
           //   messageTtl: 5000,
           //   deadLetterExchange: EXCHANGE_NAME
           // }),
```
victorges: Can we remove this comment?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -81,12 +90,18 @@ export default class MessageQueue {
     await this.channel.sendToQueue(QUEUE_NAME, msg);
   }
 
+  public async publish (route: string, msg: object): Promise<void> {
```
victorges: **nit:** Could we either have only one of these `publish` and `emit`, or rename the `emit` to `sendToQueue` with the queue name as an argument? I think this would make the difference between them more explicit, and also keep it consistent with the `Channel`.

If we can have only the `publish` I think it's even better though, feels like an anti-pattern to send directly to a queue if we have a nice topic exchange available as well. (in the long term the publisher and consumer services could even be separate, and the publisher only knows the exchange)

packages/api/src/store/rabbit-queue.ts
```diff
@@ -81,12 +90,18 @@ export default class MessageQueue {
...
+    await this.channel.publish(EXCHANGE_NAME, route, msg);
+  }
+
   public async delayedEmit(msg: Object, delay: number): Promise<void> {
     await this.channel.addSetup((channel: Channel) => {
       return Promise.all([
         channel.assertQueue(`delayedQueue_${delay / 1000}s`, {
           messageTtl: delay,
           deadLetterExchange: EXCHANGE_NAME,
+          deadLetterRoutingKey: "webhooks.delayedEmits",
```
victorges: I think this should be received as an argument as well, otherwise the `recording.ready` event is going to the wrong queue.

Also 
**nit:** If we do receive the routing key as an argument here, WDYT of renaming this function to `delayedPublish` instead for consistency? Even though we actually do a `sendToQueue` internally, that's just an implementation detail that ends up being a delayed publish to the message exchange, so for the callers it's more similar to the `publish` one.

packages/api/src/webhooks/cannon.ts
```diff
@@ -52,7 +52,70 @@ export default class WebhookCannon {
...
+        })
+      );
+    } catch (err) {
+      console.log("handleEventQueue Error ", err);
+    }
+
+    this.queue.ack(data);
   }
 
   async handleQueueMsg(data: ConsumeMessage) {
```
victorges: **nit:** Rename to `handleWebhooksQueue`?

packages/api/src/webhooks/cannon.ts
```diff
@@ -66,7 +129,8 @@ export default class WebhookCannon {
       return;
     }
     try {
-      await this.onTrigger(message);
+      // TODO Activate URL Verification
```
victorges, line 157: What is this URL verification?

packages/api/src/webhooks/cannon.ts
```diff
@@ -52,7 +52,70 @@ export default class WebhookCannon {
...
+        );
+      }
+      // basic sanitization.
+      let sanitized = this.db.stream.addDefaultFields(
+        this.db.stream.removePrivateFields({ ...stream })
+      );
+      delete sanitized.streamKey;
+  
+      let user = await this.db.user.get(event.userId);
+      if (!user) {
```
victorges: Maybe check if user is suspended here well?
```suggestion
      if (!user || user.suspended) {
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -52,7 +52,70 @@ export default class WebhookCannon {
...
+              event: event,
+              stream: sanitized,
+              user,
+              webhook
+            })
+          } catch (error) {
+            console.log("Error firing single url webhook trigger", error);
+          }
+        })
+      );
```
victorges: I think we should avoid `ack`ing the event in case of an exception here. If we failed to send the event to rabbitmq, it most likely means that `rabbitmq` is going through some issues so we can just `nack` the message (after a `setTimeout` just in case?) to retry soon since none of the previous steps of fetching stream, user, etc had any issues (which would be the ones for us to drop the event on any error).

I think something like this would do it:
```ts
      try {
        const responses = await Promise.all(
          webhooksList.map((webhook) =>
              this.queue.publish("webhooks.triggers", {
                id: uuid(),
                event: event,
                stream: sanitized,
                user,
                webhook
              }))
        );
      } catch (err) {
         console.log("Error firing single webhook trigger events", err);
         setTimeout(10000, () => this.queue.nack(data));
         return
      }
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -52,7 +52,70 @@ export default class WebhookCannon {
...
+        // if user isn't found. don't fire the webhook, log an error
+        return console.error(
+          `webhook Cannon: onTrigger: User Not found , userId: ${event.userId}`
+        );
+      }
+
+      const responses = await Promise.all(
+        webhooksList.map(async (webhook, key) => {
+          try {
+            this.queue.publish("webhooks.triggers", {
```
victorges: I think you're missing an `await` here

packages/api/src/webhooks/cannon.ts
```diff
@@ -52,7 +52,70 @@ export default class WebhookCannon {
...
+        this.db.stream.removePrivateFields({ ...stream })
+      );
+      delete sanitized.streamKey;
+  
+      let user = await this.db.user.get(event.userId);
+      if (!user) {
+        // if user isn't found. don't fire the webhook, log an error
+        return console.error(
+          `webhook Cannon: onTrigger: User Not found , userId: ${event.userId}`
+        );
```
victorges: If we return here (or in the other return above) we won't `ack` the message in the end of this function, and it will come back (automatically nacked) when the connection to rabbitmq is closed.

Maybe you could just `throw new Error()` here with that same message, so it gets logged in the `catch` block before the `ack`? Otherwise you could have a separate helper function from which you can return from, like that previous `onTrigger`.

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -200,7 +200,7 @@ describe("webhook cannon", () => {
         sem.release();
       };
 
-      await server.queue.emit({
+      await server.queue.publish("events.stream",{
```
victorges: **nitnit:** Probably for the future, I think it'd be nice to have the event topics be composed with the inner `event` keys of our events. Like here `events.stream.started`. It could even be composed automatically by the `publish` function of the queues, like grabbing the `event` field from the event data and composing with the "route namespace" provided like:
```
publish(routeNs: "events" | "webhooks", msg: { event: string }): Promise<void> {
    const route = `${routeNs}.{msg.event}`
    ...
```
and callers would go like
```
await server.queue.publish("events", { ... event: "stream.started" ... }
```
with the final routing key becoming `events.stream.started`.

This would make it really easy to implement some "on-demand" subscription to events like via websockets or SSE in the future. But it can also be iterated later, I don't think we have to implement that now.

packages/api/src/store/rabbit-queue.ts
```diff
@@ -4,12 +4,21 @@ import { Channel, ConsumeMessage, Options } from "amqplib";
...
   private channel: ChannelWrapper;
   private connection: AmqpConnectionManager;
+  private queues: Record<string, string>;
 
-  constructor() {}
+  constructor() {
+    this.queues = {
+      events: "webhook_default_queue",
+      webhooks: "webhook_cannon_single_url",
+    }
```
victorges: **nit:** Since this is just a `const` in the end, WDYT of declaring it as a static const above and avoid repeating the queue names here? Say up there:
```
const Queues = {
      events: "webhook_default_queue",
      webhooks: "webhook_cannon_single_url",
    } as const
```

Then you could even type the `consume` function below more specifically, so it avoids typos in the consumer code as well:
```
  public async consume(queueName: keyof Queues, func: (msg: ConsumeMessage) => void): Promise<void> {
```

packages/api/src/controllers/stream.ts
```diff
@@ -892,6 +892,26 @@ app.put(
     userId: user.id,
   });
 
+  if (req.body.active === false && stream.record === true) {
```
victorges: **nit:** WDYT of having a separate helper that emits all these stream active events? `emitStreamActiveEvents`?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -81,12 +90,18 @@ export default class MessageQueue {
     await this.channel.sendToQueue(QUEUE_NAME, msg);
   }
 
+  public async publish (route: string, msg: object): Promise<void> {
+    console.log(`publishing to ${route} : ${JSON.stringify(msg)}`);
+    await this.channel.publish(EXCHANGE_NAME, route, msg);
```
victorges: I think we should also have a `{ persistent: true }` option on these events otherwise they will be lost on a rabbitmq server restart (even if the queues/exchanges themselves are `durable`). Both here and in `sendToQueue` above.

victorges (commented):
Really cool design with the same topic exchange and multiple queues :)

Some comments in the implementation, most of them are nits but I think some of the functional changes are necessary.

packages/api/src/webhooks/cannon.ts
```diff
@@ -52,10 +52,73 @@ export default class WebhookCannon {
...
+      const { data: webhooksList } = await this.db.webhook.listSubscribed(
+        event.userId,
+        event.event
+      );
+  
+      console.log("webhooks : ", webhooksList);
+      let stream = await this.db.stream.get(event.streamId);
+      if (!stream) {
+        // if stream isn't found. don't fire the webhook, log an error
+        return console.error(
```
victorges: We need to throw an error instead of returning from here as well

packages/api/src/webhooks/cannon.ts
ya7ya: it's to make sure the URLs are not pointing to some local IP.

packages/api/src/store/rabbit-queue.ts
```diff
@@ -4,12 +4,19 @@ import { Channel, ConsumeMessage, Options } from "amqplib";
...
 export default class MessageQueue {
   private channel: ChannelWrapper;
   private connection: AmqpConnectionManager;
+  private queues: Record<string, string>;
 
-  constructor() {}
+  constructor() {
+    this.queues = {
+      events: "webhook_default_queue",
+      webhooks: "webhook_cannon_single_url",
```
darkdarkdragon:
```suggestion
      webhooks: WEBHOOK_CANNON_SINGLE_URL,
```

packages/api/src/store/rabbit-queue.ts
```diff
@@ -4,12 +4,19 @@ import { Channel, ConsumeMessage, Options } from "amqplib";
...
 
 export default class MessageQueue {
   private channel: ChannelWrapper;
   private connection: AmqpConnectionManager;
+  private queues: Record<string, string>;
 
-  constructor() {}
+  constructor() {
+    this.queues = {
+      events: "webhook_default_queue",
```
darkdarkdragon:
```suggestion
      events: QUEUE_NAME,
```

packages/api/src/store/rabbit-queue.ts
```diff
@@ -56,17 +59,16 @@ export default class MessageQueue {
...
       throw new Error("RabbitMQ | consume | func is undefined");
     }
     console.log("adding consumer");
     await this.channel.addSetup((channel: Channel) => {
-      return Promise.all([
-        channel.assertQueue(QUEUE_NAME),
-        channel.bindQueue(QUEUE_NAME, EXCHANGE_NAME, "#"),
-        channel.consume(QUEUE_NAME, func),
-      ]);
+      return Promise.all([channel.consume(this.queues[queueName], func)]);
```
darkdarkdragon: Promise.all is not needed?

packages/api/src/store/rabbit-queue.ts
```diff
@@ -76,23 +78,34 @@ export default class MessageQueue {
...
         channel.assertQueue(`delayedQueue_${delay / 1000}s`, {
-          messageTtl: delay,
+          messageTtl: delay + 100,
           deadLetterExchange: EXCHANGE_NAME,
+          deadLetterRoutingKey: routingKey,
           expires: delay + 15000,
         }),
       ]);
     });
     console.log("emitting ", msg);
```
darkdarkdragon:
```suggestion
    console.log(`delayed emitting delay=${delay / 1000}s`, msg);
```

packages/api/src/controllers/stream.ts
```diff
@@ -884,17 +884,6 @@ app.put(
     // trigger the webhooks, reference https://github.com/livepeer/livepeerjs/issues/791#issuecomment-658424388
     // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
 
-    // -------------------------------
-    // new webhookCannon
-    req.queue.emit({
-      id: uuid(),
-      createdAt: Date.now(),
-      channel: "webhooks",
-      event: req.body.active === true ? "stream.started" : "stream.idle",
```
darkdarkdragon: So now `stream.started` and stream.idle` events are not emitted.

packages/api/src/controllers/stream.ts
```diff
@@ -910,7 +899,8 @@ app.put(
           const ingest = ((await req.getIngest()) ?? [])[0]?.base;
           const recordingUrl = getRecordingUrl(ingest, session);
           const mp4Url = getRecordingUrl(ingest, session, true);
-          req.queue.delayedEmit(
+          req.queue.delayedPublish(
```
darkdarkdragon: should add `await` to every call to `publish` - or else exception inside it will crash whole node with `unhandled promise rejection`

packages/api/src/webhooks/cannon.ts
```diff
@@ -69,7 +134,14 @@ export default class WebhookCannon {
       return;
     }
     try {
-      await this.onTrigger(message);
```
darkdarkdragon, line 72: `onTrigger` function is not used anymore - should be removed, and code handling `recording.ready` should be moved to `handleEventsQueue`

darkdarkdragon (changes requested):

packages/api/src/webhooks/cannon.ts
```diff
@@ -69,7 +134,14 @@ export default class WebhookCannon {
...
+      // TODO Activate URL Verification
+      await this._fireHook(
+        message.event,
```
darkdarkdragon: Can we check that `event`, `webhook`, `stream`, `user` fields exists in `message` object? Just to check that we got valid data.

darkdarkdragon (changes requested):

darkdarkdragon (approved):
LGTM

PR #423, opened 2021-06-01
add livepeerci/www docker container

iameli:
Allows the frontend to be used as part of livepeer-in-a-box.

PR #536, opened 2021-07-23
www: add dockerized build (attempt 2)

iameli:
Second attempt at adding a Dockerized version of `@livepeer.com/www` for livepeer-in-a-box. This one is less hacky than https://github.com/livepeer/livepeer-com/pull/423 but a bit heavier; it packages up the entire monorepo so it can run an actual Next.js server with `next start`.

Because it's a big build I added a separate Github Actions workflow that builds the WWW container in parallel; hopefully it won't slow down the overall time to completion of the GH actions that way.

### DIFF
--- a/.github/workflows/www.yml
+++ b/.github/workflows/www.yml
@@ -2,23 +2,7 @@ on: push
 jobs:
   build:
     runs-on: ubuntu-18.04
-    services:
-      postgres:
-        image: postgres
-        env:
-          POSTGRES_HOST_AUTH_METHOD: "trust"
-        # Set health checks to wait until postgres has started
-        options: >-
-          --health-cmd pg_isready --health-interval 10s --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 5432:5432
-      rabbitmq:
-        image: rabbitmq
-        env:
-          RABBITMQ_DEFAULT_VHOST: "livepeer"
-        ports:
-          - 5672:5672
+
     steps:
       - name: checkout
         uses: actions/checkout@v2
@@ -37,14 +21,6 @@ jobs:
           docker version
           docker login -u $DOCKER_USER -p $DOCKER_PASS
 
-      - name: yarn install
-        run: |
-          yarn install --frozen-lockfile
-
-      - name: yarn test
-        run: |
-          yarn test
-
       - name: docker build
         env:
           BRANCH: ${{ steps.extract_branch.outputs.branch }}
@@ -55,8 +31,9 @@ jobs:
           DOCKER_BUILDKIT: 1
         run: |
           export DOCKER_TAG=$(echo $BRANCH | tr -cd '[:alnum:]_')
-          yarn run lerna-run docker:build --scope=@livepeer.com/api
-          yarn run lerna-run docker:push --scope=@livepeer.com/api
+          cd packages/www
+          yarn run docker:build
+          yarn run docker:push
 
           # Livepeer internal build alerting
           curl -X POST https://holy-bread-207a.livepeer.workers.dev
--- a/packages/www/Dockerfile
+++ b/packages/www/Dockerfile
@@ -4,8 +4,9 @@ FROM node:12
 
 WORKDIR /app
 ADD package.json package.json
+ADD yarn.lock yarn.lock
 ADD lerna.json lerna.json
 ADD . .
-RUN yarn install
+RUN yarn install --frozen-lockfile
 WORKDIR /app/packages/www
 CMD yarn run start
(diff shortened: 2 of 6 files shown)


.github/workflows/www.yml
```diff
@@ -0,0 +1,39 @@
...
+          CF_API_KEY: ${{ secrets.CF_API_KEY }}
+          CF_EMAIL: ${{ secrets.CF_EMAIL }}
+          DOCKER_PASS: ${{ secrets.DOCKER_PASS }}
+          DOCKER_USER: ${{ secrets.DOCKER_USER }}
+          DOCKER_BUILDKIT: 1
+        run: |
+          export DOCKER_TAG=$(echo $BRANCH | tr -cd '[:alnum:]_')
+          cd packages/www
+          yarn run docker:build
+          yarn run docker:push
```
victorges, line 36: For consistency, is there any disadvantage with doing like this?
```suggestion
          yarn run lerna-run docker:build --scope=@livepeer.com/www
          yarn run lerna-run docker:push --scope=@livepeer.com/www
```

packages/www/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# To be run from the root of the repository so that we can pick up design-system too
+
+FROM node:12
+
+WORKDIR /app
+ADD package.json package.json
+ADD lerna.json lerna.json
```
victorges: Should we copy `yarn.lock` too?

packages/www/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# To be run from the root of the repository so that we can pick up design-system too
```
victorges, line 1: Maybe a note that this is a development image only? Or is it not?

I guess it could be explicited in the filename as well/instead? Like `Dockerfile.dev` or `dev.Dockerfile`, idk 🤔 
Not necessary anyway, just a nit.

packages/www/Dockerfile.dockerignore
```diff
@@ -0,0 +1,26 @@
...
+*/*/minio
+*/*/dist
+*/*/docs
+
+# WWW-specific
+*/*/.next
+*/*/out
+
+# So you can change this without forcing a rebuild
+Dockerfile
```
victorges: nice 😆 

Shouldn't this be `**/Dockerfile` though? Since it is evaluated from the root of the repository

packages/www/Dockerfile.dockerignore
```diff
@@ -0,0 +1,26 @@
...
+**/.next
+
+# API-specific
+*/*/minio
+*/*/dist
+*/*/docs
+
+# WWW-specific
+*/*/.next
+*/*/out
```
victorges, line 23: I'm wondering if we should have only 1 `.dockerignore` in the repo root after all 🤔 

Like feels there are some stuff copied from the API docker ignore, and maybe some stuff that could be copied back well. In the end might be simpler to just have them be the same.

If it is in the root, it can have the regular `.dockerignore` name and it'd be used for all docker builds since it would be in the context root (or at least that's what they say 😆 )

victorges (approved):
LGTM

packages/www/Dockerfile
iameli: Probably not terribly useful outside of development yet, but I'd like livepeer-in-a-box to be useful outside of dev, so that'd make this a non-development dependency I suppose.

packages/www/Dockerfile.dockerignore
iameli: Good call.

packages/www/Dockerfile.dockerignore
iameli: The main difference being that `api` can ignore all of the other packages, whereas www needs them as dependencies... kinda nice to have the api build a bit faster and leaner.

victorges: Oh that's true, forgot about that part

.github/workflows/www.yml
iameli: I forgot entirely why I did this until I committed this change - the `www` build skips the initial `yarn install` step entirely and jumps straight to the Docker parts. No need to download and link everything both inside and outside of the container; www doesn't have unit tests that could fail.

iameli: (So it doesn't have any idea what lerna is at this point)

packages/www/Dockerfile
iameli: done, added `--frozen-lockfile` too

.github/workflows/www.yml
victorges: Oh, I see. Yeah, it's nice to keep both a little more isolated for parallel speediness indeed. Maybe sometime we could even separate the "api" (docker?) build from the `main.yml` job as well, so it builds much faster. I bet we could get an isolated API build in ~4 minutes instead of the >~12 we have today.

PR #540, opened 2021-07-26
update stripe script to create customer accounts for users that don't have one

adamsoffer:
A user reported that the dashboard was broken for them. After looking into the issue it was discovered that this user didn't have stripe customer account created upon registration. This PR updates the stripe migration script to filter by user accounts that don't have a stripe account, so we can create one for them.

### DIFF
--- a/packages/api/src/controllers/stripe.js
+++ b/packages/api/src/controllers/stripe.js
@@ -89,6 +89,7 @@ app.post("/migrate-users", async (req, res) => {
   const [users] = await db.user.find(
     [sql`users.data->>'stripeCustomerId' IS NULL`],
     {
+      limit: 9999999999,
       useReplica: false,
     }
   );
--- a/packages/api/src/controllers/stripe.js
+++ b/packages/api/src/controllers/stripe.js
@@ -2,6 +2,7 @@ import Router from "express/lib/router";
 import { db } from "../store";
 import Stripe from "stripe";
 import { products } from "../config";
+import sql from "sql-template-strings";
 
 const app = Router();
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/stripe.js
```diff
@@ -86,8 +87,10 @@ app.post("/migrate-users", async (req, res) => {
   }
 
   const [users] = await db.user.find(
-    {},
-    { limit: 9999999999, useReplica: false }
+    [sql`users.data->>'stripeCustomerId' IS NULL`],
+    {
+      useReplica: false,
```
victorges, line 93: I think a high `limit` might be necessary here, or we will default to `100` [here](https://github.com/livepeer/livepeer-com/blob/049c0fa1442db71201f90ea5eed49e627e69973c/packages/api/src/store/table.ts#L98)

Unless 100 is definitely ok, I think it wouldn't hurt to have something like 1000 or 10000 here (we have 7.5k users total in the db, as of now)

Otherwise, if you explicitly want **no** limit, I think you can set `limit: ''` or an explicit `limit: NaN` here which will work as per the current code (there's another check before using it [here](https://github.com/livepeer/livepeer-com/blob/049c0fa1442db71201f90ea5eed49e627e69973c/packages/api/src/store/table.ts#L135)). Those will also work if we change the initial limit check to something like `opts.limit ?? 100`. (if we do `limit: null`, then it'd be coalesced to `100` in that updated version)

victorges (approved):
LGTM

packages/api/src/controllers/stripe.js
adamsoffer: ah yeah good call. I'll add back the limit.

PR #555, opened 2021-08-03
api: allow webhooks to be sent to emails

iameli:


iameli:
We went a different direction with this - archiving it here but not gonna try and merge

PR #568, opened 2021-08-16
api: promote first user to admin

iameli:
Useful for development and the box. Only happens if there are literally no users.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -12,8 +12,8 @@
   "scripts": {
     "cloud-dev": "npm run docker && kubectl delete pod -l app.kubernetes.io/name=livepeer-api-server --wait && sleep 1 && kubectl logs -f --pod-running-timeout=30s $(kubectl get pods -o name -l app.kubernetes.io/name=livepeer-api-server)",
     "cloud-dev:server": "nodemon -w src -x node -r esm src/cli.js --port 3040 --kubeBroadcasterService broadcaster --kubeOrchestratorService orchestrator --kubeNamespace default --kubeBroadcasterTemplate 'https://{{nodeName}}.livepeer-staging.live' --kubeOrchestratorTemplate 'https://{{nodeName}}.livepeer-staging.live:8935'",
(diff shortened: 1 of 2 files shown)


packages/api/package.json
```diff
@@ -12,8 +12,8 @@
...
     "cloud-dev:server": "nodemon -w src -x node -r esm src/cli.js --port 3040 --kubeBroadcasterService broadcaster --kubeOrchestratorService orchestrator --kubeNamespace default --kubeBroadcasterTemplate 'https://{{nodeName}}.livepeer-staging.live' --kubeOrchestratorTemplate 'https://{{nodeName}}.livepeer-staging.live:8935'",
-    "docker:build": "docker pull livepeerci/api:master && docker build --cache-from=livepeerci/api:master -t livepeerci/api:${DOCKER_TAG:-dev} --build-arg GITHUB_SHA --build-arg DOCKER_TAG -f Dockerfile ../.. && docker run livepeerci/api:${DOCKER_TAG:-dev} --help",
-    "docker:push": "docker push livepeerci/api:${DOCKER_TAG:-dev}",
+    "docker:build": "docker pull livepeerci/api:master && docker build --cache-from=livepeerci/api:master -t livepeerci/api:${DOCKER_TAG:-master} --build-arg GITHUB_SHA --build-arg DOCKER_TAG -f Dockerfile ../.. && docker run livepeerci/api:${DOCKER_TAG:-master} --help",
+    "docker:push": "docker push livepeerci/api:${DOCKER_TAG:-master}",
```
darkdarkdragon: Do we really want to have default for manual push to be master?

iameli: Hmm. I want the default _build_ to be master, so that you can build this image and then run `make dev` in livepeer-in-a-box and [have it use the new image](https://github.com/livepeer/livepeer-in-a-box/blob/master/Dockerfile#L1). But I suppose you could argue that push shouldn't have a default so we don't accidentally push local images to prod? 🤔

darkdarkdragon: Yes. I think we should just change default for `push` back to `dev`. If somebody really wants to manually push production build he should specify that explicitly.

iameli: Agreed, done in https://github.com/livepeer/livepeer-com/pull/568/commits/7011ddf6ec822526aeb3b9d560305fc42af1b293

darkdarkdragon (approved):
LGTM

PR #569, opened 2021-08-17
api: Implement common protocol for RabbitMQ messages

victorges:
This is to implement the common protocol proposed with the `data/analyzer` logic for 
events sent from our services into RabbitMQ. The main reason for this protocol is that
we need all events to be serialized on the **same** data stream (~a queue), while still being
able to process them correctly according to their corresponding type. In the low-level, this
meant:
 - a `type` field which specifies what is the type of the message. This field should be parsed
 first to determine how to parse the rest of the event (in [data/analyzer](https://github.com/livepeer/livepeer-data/blob/5b75a623cbb307520e42001663532c1572f6b46b/pkg/data/parse.go#L13))
 - a `timestamp` field instead of a `createdAt`, which is how I had followed for `data/analyzer`
 and I think it makes sense for events/messages. 
 - a required `id` field because why not, should be helpful
 - a required `manifestId`, since not all Livepeer services understand stream IDs so that's the global stream identifier

2 things here that I intend to update `livepeer-data` for consistency (and functionality):
 - Timestamps should be in milliseconds instead of nanoseconds. Nano required `int64` and that's just
 not portable. JSON is not defined that way, and node doesn't even know how to serialize `bigint`s. This
 is a backward-incompatible change on `analyzer`, but I intend to do anyway since we're only in `staging` yet.
 - Will rename "events" to "messages" instead, which is what they actually are. In the API we already
 had the higher-level user-facing "events", so we should better keep those separate. This is just nomenclature
 in the code, so lower-pri and should have no incompatibilities.

 - Improve data stored in `webhook_response` table, for more debuggability
 - Create separate types for the 2 cannon RabbitMQ messages (event and trigger)
 - Fix webhook retries (I think either the N*N bug was back or retries were broken)
 - Implement that common message protocol described above
 - Move `sessionId` field of `recording.ready` event out of the user-facing `payload`

## -

`yarn test` 
Mostly some structural changes on code very well exercised in the tests, so I believe
this is enough. Also added a test for webhook retries to make sure they keep working. 

This is part of the movement to get `multistream`s in prod, with connectivity data available
in stream health/`data/anlyzer`.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -138,7 +138,8 @@ async function triggerManyIdleStreamsWebhook(
       await queue.publish("events.stream.idle", {
         type: "webhook_event",
         id: uuid(),
-        createdAt: Date.now(),
+        timestamp: Date.now(),
+        manifestId: stream.playbackId,
         event: "stream.idle",
         streamId: stream.id,
         userId: user.id,
@@ -901,7 +902,8 @@ app.put(
     await req.queue.publish(`events.${event}`, {
       type: "webhook_event",
       id: uuid(),
-      createdAt: Date.now(),
+      timestamp: Date.now(),
+      manifestId: stream.playbackId,
       event: event,
       streamId: id,
       userId: user.id,
@@ -927,7 +929,8 @@ app.put(
             {
               type: "webhook_event",
               id: uuid(),
-              createdAt: Date.now(),
+              timestamp: Date.now(),
+              manifestId: session.playbackId,
               event: "recording.ready",
               streamId: id,
               userId: user.id,
@@ -947,9 +950,8 @@ app.put(
       let shouldEmit = true;
       const session = await db.stream.getLastSession(stream.id);
       if (session) {
-        const now = Date.now();
-        const since = now - (session.lastSeen ?? 0);
-        if (now - (session.lastSeen ?? 0) < USER_SESSION_TIMEOUT) {
+        const timeSinceSeen = Date.now() - (session.lastSeen ?? 0);
+        if (timeSinceSeen < USER_SESSION_TIMEOUT) {
           // there is recent session exits, so new one will be joined with last one
           // not emitting "recording.started" because it will be same recorded session
           shouldEmit = false;
@@ -959,7 +961,8 @@ app.put(
         await req.queue.publish("events.recording.started", {
           type: "webhook_event",
           id: uuid(),
-          createdAt: Date.now(),
+          timestamp: Date.now(),
+          manifestId: stream.playbackId,
           event: "recording.started",
           streamId: id,
           userId: user.id,
@@ -1494,8 +1497,9 @@ app.post(
     const msg: messages.WebhookEvent = {
       type: "webhook_event",
       id: uuid(),
+      timestamp: Date.now(),
+      manifestId: stream.playbackId,
       event: "stream.detection",
-      createdAt: Date.now(),
       streamId: stream.id,
       userId: stream.userId,
       payload: {
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -32,6 +32,7 @@ import {
 } from "./helpers";
 import { terminateStream, listActiveStreams } from "./mist-api";
 import wowzaHydrate from "./wowza-hydrate";
+import MessageQueue from "../store/rabbit-queue";
 
 type Profile = DBStream["profiles"][number];
 type MultistreamOptions = DBStream["multistream"];
@@ -126,15 +127,18 @@ async function validateMultistreamOpts(
   return { targets };
 }
 
-async function triggerManyIdleStreamsWebhook(ids, queue) {
+async function triggerManyIdleStreamsWebhook(
+  ids: string[],
+  queue: MessageQueue
+) {
   return Promise.all(
     ids.map(async (id) => {
       const stream = await db.stream.get(id);
       const user = await db.user.get(stream.userId);
-      await queue.publish("events.streams", {
+      await queue.publish("events.stream.idle", {
+        type: "webhook_event",
         id: uuid(),
         createdAt: Date.now(),
-        channel: "webhooks",
         event: "stream.idle",
         streamId: stream.id,
         userId: user.id,
@@ -893,11 +897,12 @@ app.put(
 
     // trigger the webhooks, reference https://github.com/livepeer/livepeerjs/issues/791#issuecomment-658424388
     // this could be used instead of /webhook/:id/trigger (althoughs /trigger requires admin access )
-    await req.queue.publish("events.streams", {
+    const event = req.body.active === true ? "stream.started" : "stream.idle";
+    await req.queue.publish(`events.${event}`, {
+      type: "webhook_event",
       id: uuid(),
       createdAt: Date.now(),
-      channel: "webhooks",
-      event: req.body.active === true ? "stream.started" : "stream.idle",
+      event: event,
       streamId: id,
       userId: user.id,
     });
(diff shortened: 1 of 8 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -888,11 +893,13 @@ app.put(
...
-    await req.queue.publish("events.streams", {
+    const event = req.body.active === true ? "stream.started" : "stream.idle";
+    await req.queue.publish(`events.${event}`, {
+      type: "webhook_event",
       id: uuid(),
-      createdAt: Date.now(),
-      channel: "webhooks",
-      event: req.body.active === true ? "stream.started" : "stream.idle",
+      timestamp: Date.now(),
+      manifestId: stream.playbackId,
```
darkdarkdragon: nowadays, in our codebase, `manifestId` can mean many different things. 
If we're passing `playbackId` - maybe it's better to call it `playbackId`? That (still 😄 ) has only one meaning 😄

packages/api/src/webhooks/cannon.ts
```diff
@@ -144,29 +144,21 @@ export default class WebhookCannon {
...
+      trigger = JSON.parse(data.content.toString());
+      console.log("webhookCannon: got trigger message", trigger);
     } catch (err) {
       console.log("webhookCannon: error parsing message", err);
       this.queue.ack(data);
       return;
     }
     try {
       // TODO Activate URL Verification
-      if (message.event && message.webhook && message.stream && message.user) {
```
darkdarkdragon, line 158: Why you've removed this check?

packages/api/src/controllers/stream.ts
victorges: Ideally, here we would use some kind of "global identifier" which is known by all Livepeer services and uniquely identifies each stream in our infrastructure. So that a service processing all the messages on RabbitMQ would be able to join all events from a given stream using only the ID value in that respective field.

Since `go-livepeer` apparently has no reference to `playbackId`s ([only one I found was a comment for a field that is actually `ManifestID`](https://github.com/livepeer/go-livepeer/blob/72047f1d2d95b89a308da1064f3f3be08c89d1ec/core/streamdata.go#L165)), I __think__ it makes more sense to stick with `manifestId` for the API-agnostic field here, unless you have some other idea. Some alternatives that came to my mind but are more like questions:
 - Could we assume on `go-livepeer` that every `manifestId = playbackId` and set a field called `playbackId` with the value known there as `manifestId` instead? So in the events we have `playbackId`, and on `go-livepeer` we set `event.playbackId = mid`, instead of the opposite here
 - Do we have a way to determine the "true" `manifestId` here? That since it can be set to different things, not only the `playbackId`, is that right? Or is it truly non-deterministic for the same stream, where different services use different values but call all of them `manifestId`?

I guess the more open questiion is: What's the best way to create this "global identifier" to be used by a service joining all events about a given stream?

packages/api/src/webhooks/cannon.ts
victorges: Those are already checked in the fist line of the `_fireHook` function: https://github.com/livepeer/livepeer-com/blob/ebb97569b815880295da59fb530bb8aa88e550ff/packages/api/src/webhooks/cannon.ts#L218-L223

It also throws an error, which I think is better since it will be logged in the catch below. Just realized that the message will be unwillingly retried on that case, so I can fix that if you prefer. But IMO this is more of a sanity check that should never happen in production, so we don't want to silently ignore those cases anyway. 

I guess that we only really needed it before because of that `retry` bug that was actually sending the original `event` to the `*_single_url` queue, instead of the actual "webhook trigger" as I'm calling it here. With that our retries were actually broken, only 1 attempt was made per webhook. This new test I created won't let that happen again :)

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -219,29 +228,34 @@ describe("webhook cannon", () => {
...
       res = await client.post("/webhook", {
         ...mockWebhook,
+        url: mockWebhook.url + "2",
         name: "test-2",
       });
       expect(res.status).toBe(201);
 
-      const sem = semaphore();
-      let callCount = 0;
+      const semaphs = [semaphore(), semaphore()];
```
iameli: `sems` not good enough for you? 😁

victorges: It actually sounded a bit cacophonic to me, but no biggie. Will change to `sems` 😁

victorges:
A couple minor issues related to this change, in case any of you can give some input. Some of them are actually breaking changes that we need to do now in this PR (changes to message schemas on rabbitmq) but others mostly code/organizational:

 - Do you think we should try to keep consistency between the messages that we send to the broker/RabbitMQ and the objects that we store in the DB? If so, WDYT of renaming the `type` field to `kind` and the `timestamp` field back to `createdAt` (including on `livepeer-data`)? This is the one that is better to do now rather than later.
 - On `livepeer-data`, WDYT of renaming all the "events" to "messages" instead? I implemented that distinction here and then reflecting about it felt like the things we send to RabbitMQ are messages, not events per se. This would include renaming the `data` package to `messages` and the `Event` struct/interfaces to something corresponding to that.

victorges:
Forgot to tag for notification: @darkdarkdragon @iameli

iameli:
> Do you think we should try to keep consistency between the messages that we send to the broker/RabbitMQ and the objects that we store in the DB? If so, WDYT of renaming the `type` field to `kind` and the `timestamp` field back to `createdAt` (including on `livepeer-data`)? This is the one that is better to do now rather than later.

No strong opinion. `createdAt` makes slightly more sense in the context of a database object which can also have an `updatedAt` field; immutable stream records perhaps just have a `timestamp`. `kind` and `type` are synonyms, I only used `kind` on the API because I rip off Kubernetes at every available opportunity.

victorges:
Cool then. I did think the `timestamp` made more sense in case of events/immutable messages, but wasn't sure if anyone could see a world where we would like to merge the objects/schemas somehow. I also hadn't remembered about the `kind` when I did the `type`, otherwise I'd have picked it since they're pretty much synonyms as you said. `type` can get a little tricky in `go` since it's a reserved word, so I'd wild guess that's why they started with `kind` themselves 😆 

Either way, I'll avoid the trouble of any of those refactors for now since we're running on a tight schedule for multistream.

packages/api/src/webhooks/cannon.ts
darkdarkdragon: > Just realized that the message will be unwillingly retried on that case, so I can fix that if you prefer

Yes, and that error will send message to retry loop. 
I've hit this problem when was developing and my old code sent message to the new one.

Everywhere we're getting data from external system we should check that is has format we expect. 
I think in this case we safely can just ignore such message.

victorges: Fixed: https://github.com/livepeer/livepeer-com/pull/569/commits/340eb2dc808943f69470d4ae839b4e6247aceaef

packages/api/src/controllers/stream.ts
victorges: Resolved by using the `streamId` as the global stream identifier.

 - livepeer-data: https://github.com/livepeer/livepeer-data/pull/7/commits/029683482b09a135f5b13a620213b43ec6dfb55c
 - go-livepeer: https://github.com/livepeer/go-livepeer/pull/2002/commits/b814dfb08208f1857b7bb6e01ac448246b006140
 - livepeer-com: https://github.com/livepeer/livepeer-com/pull/569/commits/74f1a4cb2a055be10fbf29b1db87bdb74ab7f02a
 - stream-tester: https://github.com/livepeer/stream-tester/pull/96/commits/b2ee197b342a7dc4499feeea8e1e9acbb59fa9a1

darkdarkdragon (approved):
LGTM

PR #579, opened 2021-08-24
www: Multistream frontend 

victorges:
It's happening!

This implements the frontend interfaces for interacting the multistream feature. 

 - New multistream targets table in stream details page
 - Creation/deletion/toggling of targets from that table
 - Display of each target status based on data from stream health (`livepeer-data/analyzer`)

## -

 - CRUD targets from table
 - Check calls being made to analyzer to grab state and display it accordingly per-target (faked it locally, couldn't get deploy preview/staging to work)
 - Rinse and repeat

Implements https://github.com/livepeer/livepeer-infra/issues/563

**Screenshots (optional):**
All around the PR discussion below. The one with the most overview:
image

**Checklist:**


victorges:
Getting there 😁 

Screen Shot 2021-08-25 at 21 22 10

adamsoffer:
Yesss

victorges:
Looking goood 😋

Screen Shot 2021-08-26 at 18 47 31

Thanks @adamsoffer!

victorges:
I believe we are now feature complete 🙌 

A few bugs here and there. The only big one is when closing the "delete target" dialog, focus gets lost on the window somewhere.

I am also currently not able to test on staging though, as the Vercel link that the bot sends on the PR here just never loads. How can I get one of those URLs pointing to my branch?

Screenshots:

Screen Shot 2021-08-27 at 04 36 24
(fake states, but should grab real data in staging/prod)

Screen Shot 2021-08-27 at 04 37 16

victorges:
> A few bugs here and there. The only big one is when closing the "delete target" dialog, focus gets lost on the window somewhere.

Fixed that one in this commit https://github.com/livepeer/livepeer-com/commit/b2e78a77d3356967c0fbb674eb6c4e97e93def2f!

The only pending thing in my mind now is to nicely display API errors, e.g. when trying to create target with a non-RTMP/SRT URL. Right now, we simply don't close the dialog without saying anything and let the user retry. I think this would be only a nice-to-have though as the main workflow is working great.

That said, I think it's pretty much done, we could even launch tomorrow if we want to 🙂

iameli:
Nice nice nice nice nice nice nice, taking this for a spin now

iameli:
> I am also currently not able to test on staging though, as the Vercel link that the bot sends on the PR here just never loads. How can I get one of those URLs pointing to my branch?

@victorges I'm using this one, seems to work for me? https://livepeer-com-h2pzhfxbt-livepeerorg.vercel.app/

adamsoffer:
So good ✨🚀

> I am also currently not able to test on staging though, as the Vercel link that the bot sends on the PR here just never loads. How can I get one of those URLs pointing to my branch?

This preview url points to the latest commit and is working for me. https://livepeer-com-git-vg-em-featmultistream-frontend-livepeerorg.vercel.app

packages/www/hooks/use-analyzer.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const makeUrl = (region: string, path: string) => {
+  if (isDevelopment() && useLocalAnalyzer) {
+    return `http://localhost:8080/data${path}`;
+  }
+  const tld = isStaging() || isDevelopment() ? "monster" : "com";
+  return `https://${region}.livepeer.${tld}/data${path}`;
+};
+
+class AnalyzerClient {
```
iameli, line 40: Nice. This is the thing I'd like to refactor out (along with `use-api`, natch) to become the TypeScript API client, but this is a good place to start

packages/www/layouts/streamDetail.tsx
```diff
@@ -276,7 +271,7 @@ const StreamDetail = ({
     <Layout id="streams" breadcrumbs={breadcrumbs}>
       <Box css={{ px: "$6", py: "$7" }}>
         {stream ? (
-          <>
+          <AnalyzerProvider region={stream.region}>
```
iameli: Will this work if the stream gains a region after the page loads? That'll be kind of common.

packages/www/hooks/use-analyzer.tsx
victorges: Yeah! I thought a class was a good choice here mainly because of eventually separating this to a standalone client.

packages/www/components/Dashboard/MultistreamTargetsTable/CreateTargetDialog.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const parsedUrl = useMemo(() => {
+    let parsed: url.UrlWithParsedQuery;
+    try {
+      parsed = url.parse(targetUrl, true);
+    } catch (err) {}
+    if (!streamKey || !parsed) {
+      return parsed;
+    }
+
+    switch (parsed.protocol) {
```
iameli, line 65: Nice. Better than the OBS frontend.

packages/www/layouts/streamDetail.tsx
victorges: Yeah! When the stream gains a region after the page loads, this whole component will be re-rendered with the updated `stream` object and thus the `Analyzer.Provider` will get the updated region. Btw when there is no `region` in the current `stream`, we default to `nyc` on the client just to get some proper API response, most likely a 404.

iameli (approved):
LGTM! We'll probably do another round of feedback from folks once it's in staging but I've used it successfully. Great work!

victorges:
Awesome! I think we can already share the URL that @adamsoffer mentioned here, which gets the latest version running from the branch: https://livepeer-com-git-vg-em-featmultistream-frontend-livepeerorg.vercel.app

(it's working for me now if I re-enter the URL in the browser after clicking the vercel preview link, weird)
Ah! Yesterday I was trying to convert the `/` to a `-` in the URL, that's why I couldn't access the one from the branch. The ones from each deploy preview are still weird, and I have to do the "re-insert" the URL in the browser for it to really load 🤷‍♂️

adamsoffer (approved):
LGTM! :shipit:

PR #573, opened 2021-08-18
api: Small tweaks on multistream and CLI parsing

victorges:
**Specific updates (required)**
 - Disallow additional properties on multistream options object in Streams
 - Disallow specifying a multistream target multiple times in the same stream 
 - Enable strict mode on CLI when not running tests

## -

- **How did you test each of these updates (required)**
`yarn && yarn test`

**Does this pull request close any open issues?**
No.
**Checklist:**

- [x] I have read the **CONTRIBUTING** document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

PR #580, opened 2021-08-25
Fix local run and production API command

victorges:
This fixes the `yarn dev` command when run from the root of the repository,
by using different ports for the `design-system` and `www` projects. Also
added a helper condition in the API client in order to create `stripe` stuff on
a pre-existing user that doesn't have those configured.

As for the API, fixed the command execution locally after this change: https://github.com/livepeer/livepeer-com/pull/573/commits/e832e9eeb2e383dd16dd6e817247123aec57de21
Both by removing inexistent args from the `.env` file (yarn complains about envs
on strict mode) and by fixing the Stripe client to use proper CLI args (also parsed
from envs) instead.

 - run design-system on a different port
 - add development helper on www API client to create stripe configs on an existing user
 - remove deprecated env vars from `.env`
 - create stripe client in the main app pipeline instead of statically, like the other existing clients

## -

`yarn && yarn test && yarn dev`

Fixes local dev and deploys.

**Checklist:**

### DIFF
--- a/packages/api/src/dotenv.js
+++ b/packages/api/src/dotenv.js
@@ -1,0 +1,5 @@
+// We need a separate file because of this issue: https://github.com/babel/babel/issues/2061
+// Otherwise, all our code would be evaluated without having the envs properly
+// set up, so things like a static stripe client creation wouldn't have access
+// to the env vars from the .env file.
+require("dotenv").config();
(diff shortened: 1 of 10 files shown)


packages/api/src/dotenv.js
```diff
@@ -0,0 +1,5 @@
+// We need a separate file because of this issue: https://github.com/babel/babel/issues/2061
```
iameli, line 1: Oh booooo. But cool.

victorges: It's apparently a babel only thing. If we eventually migrate to `tsc` we could avoid it 🌚 

(actually, now that we don't access env vars from static context anymore, I could even revert this change. lmk if you prefer that)

iameli (approved):
LGTM. Sneaky trick to get the Stripe credentials loaded at the right time.

PR #592, opened 2021-09-03
www/multistream: Add beta badge to multistream pages

victorges:
Mainly added `Beta` badge to multistream frontend components (only creation dialog in the end).

But then also added some improvements:
 - Sort profiles on create/update target dialogs, so it's always descending in quality starting from `Source`
 - Disallow profiles with the same name in the API. We assume uniqueness in several places and this bugs some stuff

**Screenshots:**
Screen Shot 2021-09-03 at 16 02 52

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -107,6 +107,8 @@ async function validateMultistreamOpts(
       continue;
     } else if (name === "source") {
       throw new BadRequestError(`profile cannot be named "source"`);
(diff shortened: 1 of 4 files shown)


victorges:
This one is DEPRECATED since we decided to keep the beta badge only on the creation dialog, but for history purposes I'll include the screenshot here:
Screen Shot 2021-09-03 at 16 02 43

adamsoffer (approved):
Looks good!

PR #596, opened 2021-09-08
www/docs: Finalize multistream target docs

victorges:


victorges:
Added API reference as well! I think we have all we need to deploy this, but would be nice to have a proof read :)

It should be live on Vercel preview: https://livepeer-com-git-vg-docsmultistream-livepeer.vercel.app/docs/

adamsoffer (approved):
LGTM!

PR #598, opened 2021-09-10
api: Make the --amqpUrl CLI flag optional

victorges:
This is to fix deploys of the API in the GPU regions, since we don't have a RabbitMQ running
there and currently cannot run the API without a queue. This fixes that latter.

 - Create interface for the current "public" usage of the queue, for publishing events
 - Create a fake queue implementation that only logs that events that were eaten
 - Remove default CLI flag value and initialize fake queue when not provided

## -

`yarn test` and `yarn dev` adding and removing the arg/env

Not that I know of.

**Checklist:**


darkdarkdragon (approved):
LGTM

yondonfu (approved):
LGTM! 🚢 

Confirmed that deploying this branch and omitting the `--amqpUrl` flag works - API pods came up successfully in a GPU region without RabbitMQ deployed.

PR #599, opened 2021-09-14
www/stream-health: update stream health view

adamsoffer:
TODOs:
- [ ] Add explainer tooltip copy for each health check item (Paige)
- [x] Remove the sort icon from health checks table


packages/www/components/Dashboard/HealthChecksTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        count: conditionTypes.length,
+        nextCursor: null,
+        rows: conditionTypes.map((condType) => {
+          const cond = conditionsMap?.[condType];
+          return {
+            id: condType,
+            name: {
+              children: (
+                <Box>
+                  {condType === "TranscodeRealTime" ? "Realtime" : condType}
```
victorges, line 88: WDYT of a more explicit:?
```suggestion
                  {condType === "TranscodeRealTime" ? "Real-time transcoding" : condType}
```
Or maybe just `Real-time`. Apparently both forms exist (even separated by a space as well), but I particularly prefer the hyphen. Not a strong opinion tho.

packages/www/components/Dashboard/HealthChecksTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+                  {!stream.isActive || !cond || cond.status === null ? (
+                    "-"
+                  ) : (
+                    <StatusBadge
+                      variant={
+                        cond.status
+                          ? StatusVariant.Healthy
+                          : StatusVariant.Unhealthy
+                      }
+                      timestamp={cond?.lastProbeTime}
```
victorges, line 104: Not something that we can fix right now, but just so you are aware as well:

There currently is a slight bug in these conditions, which is that we can't really tell whether they refer to the current stream or not. i.e. they might have been measured (`lastProbeTime`) on a previous session of the stream, but if they are not updated again for the current stream we have no way to tell whether the event was for the current or the previous session. This problem exists in the multistream table as well, which is why I added the `moment` tooltips so at least the user can have a tip in case the state refers to a previous session.

An easy way to reproduce this is to start a stream with multistream targets enabled, stop that stream, disable all the multistream targets and then restart the stream. We will get a permanent "unhealthy" for the `Multistreaming` condition even though the `lastProbeTime` is from before the current session ever started! (I couldn't find any existing API to tell when a stream was started).

Some ways to fix this:
 - Use the more detailed `multistream` field in the `streamHealth` object, and disconsider states about targets that are currently disabled, in the end re-computing the `multistreaming` condition in the frontend considering additional information from the API. Although this would work right now, it might be too much trouble for a workaround. Maybe not that common of a scenario anyway to be worth it. WDYT?
 - Provide some kind of `activeSince` data from the API so that in the frontend we can do some logic of disconsidering states/health conditions from previous sessions.
 - Handle the `stream.active` event from the `analyzer` as well (stream health API) and consider that when calculating all these conditions. This seems like the right option for me, but not something we could do before tomorrow.

packages/www/pages/dashboard/streams/[id]/health.tsx
```diff
@@ -1,23 +1,27 @@
...
+import Logger from "@components/Dashboard/Logger";
 
 const interval = 10000;
 const maxItems = 6;
 
 const Health = () => {
-  const [dataChart, setDataChart] = useState<{ name: number; kbps: number }[]>([
-    { name: 0, kbps: 0 },
+  const [dataChart, setDataChart] = useState<{ name: number; kpbs: number }[]>([
+    { name: 0, kpbs: 0 },
```
Comment (victorges) [inline packages/www/pages/dashboard/streams/[id]/health.tsx]: Wasn't it correct like `kbps` "kilobits per second"?

victorges (approved):
Awesome! :shipit:

packages/www/components/Dashboard/HealthChecksTable/index.tsx
adamsoffer: Good to know. Let's address in a follow up PR!

packages/www/pages/dashboard/streams/[id]/health.tsx
Comment (adamsoffer) [reply packages/www/pages/dashboard/streams/[id]/health.tsx]: oops yeah I was experimenting with something and changed it back with a typo.

Issue #575, opened 2021-08-20
Support combining orchestrator lists from multiple sources


yondonfu:
**What's the problem? (required)** Describe the problem or need in exactly 1
sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

At the moment, the API server can be configured with an orchestrator list to return for the orchestrator discovery webhook using a variety of sources i.e. hardcoded list, k8s, etc. After https://github.com/livepeer/livepeer-com/issues/574, the API server will also be able to fetch all active orchestrators on the public network. However, there is no support for combining orchestrator lists from multiple sources. An example where this feature would be desirable is to combine a hardcoded list of orchestrators that are trusted with the list of all public network orchestrators from the subgraph.

**What's the current behavior? (required)** A clear and concise description of
the current behavior of the DApp or package (or lack thereof).

The API server is configured with a single data source and returns the list of orchestrators from that source for the orchestrator discovery webhook.

**Describe a possible solution you've considered (optional)** A clear and
concise description of what you want to happen.

One option could be to combine the orchestrator lists from each source if multiple sources are configured into a single list to return. For example, if a hardcoded list is provided and a subgraph URL is provided, a single list of orchestrators with members from each of the sources would be returned.

yondonfu:
Closed by #603

Issue #574, opened 2021-08-20
Add subgraph orchestrator discovery


yondonfu:
**What's the problem? (required)** Describe the problem or need in exactly 1
sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

The [Livepeer subgraph](https://thegraph.com/legacy-explorer/subgraph/livepeer/livepeer) can be used to fetch a list of active orchestrators on the public network to be returned via the `/api/orchestrator` endpoint that broadcasters typically use for the orchestrator discovery webhook feature.

**What's the current behavior? (required)** A clear and concise description of
the current behavior of the DApp or package (or lack thereof).

The API server currently is not able to return a list of active orchestrators on the public network for the `/api/orchestrator` endpoint.

**Describe a possible solution you've considered (optional)** A clear and
concise description of what you want to happen.

This feature is mostly implemented on [this branch](https://github.com/livepeer/livepeer-com/commits/subgraph-discovery). The branch only contains a hardcoded "block list" that filters out orchestrators that we don't want to include in the webhook response based on ETH address. We may want a configurable block list for the API server to exclude certain orchestrators returned by the subgraph in the future, but for this issue I think it's fine to leave the hardcoded block list as is.

yondonfu:
Closed by #603

PR #603, opened 2021-09-15
Return 'score' field in /api/orchestrator endpoint

### LINKED ISSUE #576: Support returning scores in orchestrator discovery webhook response
**What's the problem? (required)** Describe the problem or need in exactly 1
sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

We want to indicate to a broadcaster a score for each orchestrator. Later on, the values for this score could be computed based on some to be determined formula. Initially, the values for this score can be just 0 or 1 with 0 indicating that the broadcaster does not trust the orchestrator and 1 indicating that the broadcaster should trust the orchestrator.

**What's the current behavior? (required)** A clear and concise description of
the current behavior of the DApp or package (or lack thereof).

The webhook response just includes a list of URIs.

**Describe a possible solution you've considered (optional)** A clear and
concise description of what you want to happen.

Add an additional `score` field to the JSON object for each orchestrator. Ex.

```
[
    {
        "address": ...
        "score": 0
    },
    {
        "address": ...
        "score": 1
    }
]
```

The response format decided on here should match the response format expected in https://github.com/livepeer/livepeer-com/issues/576 when the broadcaster reads the webhook response.

darkdarkdragon:
- Adds `verificationFreq` to `/stream/hook` response (hardcoded for now)
- Adds `Score` field to `/orchestrator` response
- Merges 'subgraph' branch
- Allows to combine different source in one `/orchestrator` response
- Allows to return data from `region` table in `/orchestrator` response

See commits

## -

Fixes #576, #575, #574

**Screenshots (optional):**

**Checklist:**


packages/api/src/middleware/subgraph.ts
```diff
@@ -0,0 +1,83 @@
...
+}: {
+  subgraphUrl: string;
+}) {
+  const CACHE_REFRESH_INTERVAL = 12 * 60 * 60 * 1000; // 2 hours
+
+  const blockList = {
+    "0x3e2b450c0c499d8301146367680e067cd009db93": true,
+    "0x23ca66656701ea524c92c5040cf022bd1bf8a4f8": true,
+    "0x2559ae126336207c93060ed626f8bdefd998b66f": true,
+    // TODO: Remove these addresses after some time for upgrading to >= v0.5.18
```
yondonfu: FYI this TODO was addressed in https://github.com/livepeer/livepeer-com/commit/f672b591ce818fcde1a06dae7b1030eca53c0a34 so should be addressed by picking up that commit.

packages/api/.env
```diff
@@ -1,4 +1,4 @@
-LP_ORCHESTRATORS=[{"address":"https://127.0.0.1:3086","cliAddress":"http://127.0.0.1:3076"}]
+#LP_ORCHESTRATORS=[{"address":"https://127.0.0.1:3086","cliAddress":"http://127.0.0.1:3076","score":1}]
```
victorges: Does this need to be commented out? If so, WDYT of moving it to the `.env.local.example` file instead?

packages/api/src/app-router.ts
```diff
@@ -18,7 +19,14 @@ import { getBroadcasterHandler } from "./controllers/broadcaster";
 import WebhookCannon from "./webhooks/cannon";
 import Queue, { NoopQueue, RabbitQueue } from "./store/queue";
 import Stripe from "stripe";
+import { CliArgs } from "./parse-cli";
+import { regionsGetter } from "./controllers/region";
 
+enum OrchestratorSource {
+  hardcoded = "hardcoded",
+  subgraph = "subgraph",
```
victorges, line 27: Nice! 😁

packages/api/src/controllers/orchestrator.ts
```diff
@@ -1,11 +1,18 @@
...
+const getOrchestrators = async (req: Request, res, next) => {
+  const orchestrators = [];
+  for (const getter of req.orchestratorsGetters) {
+    const orchs = await getter();
+    orchestrators.push(...orchs);
+  }
 
-  return res.json(orchestrators.map(({ address }) => ({ address })));
+  return res.json(
+    orchestrators.map(({ address, score }) => ({ address, score }))
```
victorges: Should the `defaultScore` be here instead of in the `regionOrchestratorsGetter`? So that any getter might return a non-scored orch and we always give it a default here.
```suggestion
    orchestrators.map(({ address, score = defaultScore }) => ({ address, score }))
```

packages/api/src/controllers/region.js
```diff
@@ -14,7 +29,7 @@ app.get("/", async (req, res, next) => {
   regions.forEach((region) => {
     region.orchestrators.forEach((orch) => {
       orch.region = region.region;
-      flatOrchList.push(orch);
+      flatOrchList.push({ score: defaultScore, ...orch });
```
victorges: Do we need to mutate the `orch` in-place? How about adding the `region` on this obj literal too?
`flatOrchList.push({ score: defaultScore, ...orch, region: region.region });`

packages/api/src/middleware/hardcoded-nodes.test.ts
```diff
@@ -76,7 +76,7 @@ describe("kubernetes middleware", () => {
   });
 
   it("should return orchestrators from getOrchestrators()", async () => {
-    const response = await req.getOrchestrators();
+    const response = await req.orchestratorsGetters[0]();
```
victorges, line 79: Could we keep calling the `getOrchestrators` function that iterates over the getters list?

packages/api/src/middleware/hardcoded-nodes.ts
```diff
@@ -41,8 +41,10 @@ export default function hardcodedNodes({
     if (!req.getBroadcasters) {
       req.getBroadcasters = async () => broadcasters as NodeAddress[];
     }
-    if (!req.getOrchestrators) {
-      req.getOrchestrators = async () => orchestrators as NodeAddress[];
+    if (orchestrators.length) {
```
victorges: If the `orchestrators` might ever be `null/undefined`, might be good to use a coalesce `?.` here
```suggestion
    if (orchestrators?.length) {
```

packages/api/src/middleware/kubernetes.test.ts
```diff
@@ -48,19 +48,22 @@ describe("kubernetes middleware", () => {
     });
 
     it("should return orchestrators from getOrchestrators()", async () => {
-      const response = await req.getOrchestrators();
+      const response = await req.orchestratorsGetters[0]();
```
victorges, line 51: Same, could we keep using `getOrchestrators`? I think there are some other occurrences too.

packages/api/src/middleware/subgraph.ts
```diff
@@ -0,0 +1,83 @@
...
+    "0xbaec9a6508b8270f4ff61049502d19cfe6ba428b": true,
+    "0xd84781e1a9b74d71ea76cda8bb9f30893bfd00d1": true,
+    "0xfb9849b0b53f66b747bfa47396964a3fa22400a0": true,
+    "0xc08dbaf4fe0cbb1d04a14b13edef38526976f2fb": true,
+    "0x2a15f9e60968451780bd2d09ca4e8c4498f168c3": true,
+  };
```
victorges: Should these be a configuration/CLI-arg instead?

packages/api/src/middleware/subgraph.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (lastCachedRespUpdate + CACHE_REFRESH_INTERVAL < Date.now()) {
+      try {
+        const res = await fetch(subgraphUrl, {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+            Accept: "application/json",
+          },
+          body: JSON.stringify({ query }),
+        });
```
victorges: Can this request take some relevant time to be resolved? If so, I think it might be better to initialize it on the construction of this middleware (make it `async`), and then have a `setInterval` function re-fetching it and updating some shared variable, WDYT?

If wanna be  extra safe regarding clean-up and all, could  even have some way  of stopping the fetching with a  `clearInterval` maybe, but I think it's unnecessary it will shutdown with the process anyway.

packages/api/src/middleware/subgraph.ts
```diff
@@ -0,0 +1,83 @@
...
+        for (const tr of transcoders) {
+          if (tr.id.toLowerCase() in blockList) {
+            continue;
+          }
+
+          cacheUpdate.push({
+            address: tr.serviceURI,
+            score: defaultScore,
+          });
+        }
```
victorges: **nit:** Could be a little functional (certainly not necessary tho)
```suggestion
        const cacheUpdate: OrchestratorNodeAddress[] = transcoders
            .filter(tr => !(tr.id.toLowerCase() in blockList))
            .map(tr => ({ address: tr.serviceURI, score: defaultScore }));
```

packages/api/src/types/common.d.ts
```diff
@@ -1,10 +1,18 @@
 import { Ingest, Price } from "../middleware/hardcoded-nodes";
-import { NodeAddress } from "../middleware/kubernetes";
 import { Stream, User, ApiToken } from "../schema/types";
 import Queue from "../store/queue";
 import { CliArgs } from "../parse-cli";
 import Stripe from "stripe";
 
+export interface NodeAddress {
```
victorges, line 7: Nice! thanks for all the typing 😁

packages/api/src/types/common.d.ts
```diff
@@ -28,7 +36,7 @@ declare global {
       tokenId?: string;
 
       getBroadcasters?: () => Promise<NodeAddress[]>;
-      getOrchestrators?: () => Promise<NodeAddress[]>;
+      orchestratorsGetters?: Array<() => Promise<OrchestratorNodeAddress[]>>;
```
victorges, line 39: Wait now I'm confused. I thought that the `getOrchestrators` function still existed on the request, doesn't it?

Checked now, and it's actually a helper in some specific file. Could we move that helper, that foreaches on the getters and concatenates the results, to the request, so we can re-use it anywhere we previously needed the orchs? 

Seems like we're doing `orchestratorGetters[0]` in a couple places (mostly tests for now?) that could be avoided if we still exposed the `getOrchestrators` helper as before, but now iterating over the `orchestratorGetters`. WDYT?

victorges (approved):
LGTM with only a couple nits!

packages/api/src/middleware/subgraph.ts
darkdarkdragon: fixed

packages/api/.env
darkdarkdragon: fixed

packages/api/src/controllers/orchestrator.ts
darkdarkdragon: fixed

packages/api/src/controllers/region.js
darkdarkdragon: fixed

packages/api/src/middleware/hardcoded-nodes.ts
darkdarkdragon: fixed

packages/api/src/middleware/subgraph.ts
darkdarkdragon: I've used `fetchWithTimeout`. Don't want to use `setInterval` - it will put unnecessary load on thegrap servers.

packages/api/src/middleware/subgraph.ts
darkdarkdragon: fixed

victorges (approved):
LGTM!

PR #617, opened 2021-09-23
www: Fix sorting of streams and webhooks table

victorges:
Fix sorting on Streams table.

Also found that Webhooks table had a similar issue. It's sortBy function was always
returning `NaN` and I could only guess that it still works because it falls back to some
default "text value" sorting? 

 - Fix sortBy function in streams and Webhooks tables

## -

Tried sorting those tables

Fixes https://discord.com/channels/423160867534929930/730147691698585660/890321994002870292

**Checklist:**


adamsoffer (approved):
LGTM 🛳

PR #625, opened 2021-09-28
CI: Publish version tags with all images

victorges:
This is to start using the new https://github.com/livepeer/action-gh-release-tags to publish
docker images with both version and branches tags.

Related: https://github.com/livepeer/livepeer-infra/pull/686


victorges:
This was reviewed in conjunction with https://github.com/livepeer/livepeer-com/pull/626#pullrequestreview-767245306 because I had forgotten to change the base branch. So considering this one LGTM'ed as well 😄

packages/api/package.json
```diff
@@ -12,8 +12,8 @@
...
     "cloud-dev:server": "nodemon -w src -x node -r esm src/cli.js --port 3040 --kubeBroadcasterService broadcaster --kubeOrchestratorService orchestrator --kubeNamespace default --kubeBroadcasterTemplate 'https://{{nodeName}}.livepeer-staging.live' --kubeOrchestratorTemplate 'https://{{nodeName}}.livepeer-staging.live:8935'",
-    "docker:build": "docker pull livepeerci/api:master && docker build --cache-from=livepeerci/api:master -t livepeerci/api:${DOCKER_TAG:-master} --build-arg GITHUB_SHA --build-arg DOCKER_TAG -f Dockerfile ../.. && docker run livepeerci/api:${DOCKER_TAG:-master} --help",
-    "docker:push": "docker push livepeerci/api:${DOCKER_TAG:-dev}",
+    "docker:build": "TAGS=\"${DOCKER_TAGS:-dev}\"; docker pull livepeerci/api:master && docker build --cache-from=livepeerci/api:master $(printf ' -t livepeerci/api:%s' $TAGS) --build-arg GITHUB_SHA --build-arg VERSION -f Dockerfile ../.. && docker run livepeerci/api:$(printf $TAGS)  --help",
```
iameli: I use the `master` build here for building new things for livepeer-in-a-box. Any idea the best way to handle that workflow?

iameli (commented):
Nice, just one question about `in-a-box` workflow.

packages/api/package.json
victorges: Humm, not really sure. Would it be possible to use the `dev` tag on `livepeer-in-a-box` instead? If not easily, I can just revert that specific commit that did this: https://github.com/livepeer/livepeer-com/pull/625/commits/0c881f725b79c9db7ee69b7c056c560a77f249bb
I don't think it is a must.

victorges: Reverted but kept the default `push` tag different than the build so we don't accidentally push to the production image tag.

PR #626, opened 2021-09-28
Allow video-only multistream sessions

victorges:

This is to enable multistreaming only a stream video, to allow streaming to a platform
where there might be strong copyright enforcement. Needed by PlayDJ.

 - Add proper field and handling in the API
 - Add UI components for toggling mutestream for a given target
 - Update the docs

## -

Create, view and update multistream targets video-only fields.

Implements https://github.com/livepeer/livepeer-com/issues/595

**Screenshots (optional):**
Screen Shot 2021-09-28 at 19 33 34
Screen Shot 2021-09-28 at 19 33 53

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -93,7 +93,8 @@ async function validateMultistreamTarget(
     url: spec.url,
     userId,
   });
-  return { profile, id: created.id };
+  const { spec: _, ...specless } = target;
+  return { ...specless, id: created.id };
 }
 
 async function validateMultistreamOpts(
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -424,7 +424,13 @@ describe("controllers/stream", () => {
         const res = await client.post("/stream", {
           ...postMockStream,
           multistream: {
-            targets: [{ profile: "test_stream_360p", spec: mockTarget }],
+            targets: [
+              {
+                profile: "test_stream_360p",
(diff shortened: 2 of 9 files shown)


adamsoffer (approved):
Looks great!

packages/api/src/controllers/stream.ts
```diff
@@ -93,7 +93,8 @@ async function validateMultistreamTarget(
     url: spec.url,
     userId,
   });
-  return { profile, id: created.id };
+  const { spec: _, ...specless } = target;
```
iameli, line 96: Neat trick!

iameli (approved):
LGTM!

I tried to convince myself that `muteAudio` made more sense than `videoOnly` in case we ever wanted to push things other than video. Subtitles? I don't think you can push those over RTMP. `videoOnly` is fine

victorges:
> @iameli I tried to convince myself that muteAudio made more sense than videoOnly in case we ever wanted to push things other than video. Subtitles? I don't think you can push those over RTMP. videoOnly is fine

Cool! The idea came actually from `youtube-dl -F <url>` command, which fetches the available tracks for a YouTube video, and some of them are audio-only or video-only. What I liked the most about it is that we'll be able to create an `audioOnly` flag in the future with a very consistent language 👌

Could even create it now already if you think it's worth, should be really easy 👀. I didn't do it only 'cause I couldn't find a service to test it with. (Also wasn't sure if it'd really be useful in the livestreaming space, but maybe?)

PR #633, opened 2021-10-05
replace Shaka with Video.js

### LINKED ISSUE #632: Replace Shaka with Video.js
**What's the problem? (required)** 
The Shaka player we use in the dashboard buffers every 10 seconds. 

**Do you have any personal experiences dealing with this problem? (optional)**
I've confirmed that the buffering issue is unique to Shaka player. Playback with other plays such as JWPlayer and Video.js is much smoother.

**Describe a possible solution you've considered (optional)**
Replace Shaka player with Video.js.

adamsoffer:
This PR replaces the Shaka player in the Dashboard with Video.js 

Tested locally. Confirmed Video.js fixes the issues described in #632 

Fixes #632


packages/www/components/Dashboard/Player/index.tsx
```diff
@@ -1,63 +1,50 @@
...
-
-    // Store Shaka's API in order to expose it as a handle.
-    controller.current = {
-      player,
-      ui,
-      videoElement: video.current,
-      config: {},
-    };
+    // make sure Video.js player is only initialized once
+    if (!playerRef.current) {
```
victorges, line 14: What if `options` or `onReady` changes? Do we really need this check if the `useEffect` memo below supposedly only calls this when one of the memos change? Or is this that one that might have some exceptions? not sure

packages/www/components/Dashboard/Player/index.tsx
```diff
@@ -1,63 +1,50 @@
...
+      const videoElement = videoRef.current;
+      if (!videoElement) return;
+
+      const player = (playerRef.current = videojs(videoElement, options, () => {
+        console.log("player is ready");
```
victorges: Apparently forgotten `console.log` here
```suggestion
```

packages/www/components/Dashboard/Player/index.tsx
```diff
@@ -1,63 +1,50 @@
...
+  useEffect(() => {
     return () => {
-      player.destroy();
-      ui.destroy();
+      if (playerRef.current) {
+        playerRef.current.dispose();
+        playerRef.current = null;
+      }
     };
   }, []);
```
victorges: I think this callback should be returned from the above `useEffect` instead, so that any time that `options` or `onReady` changes the previous player would be disposed as well.

If that `if (!playerRef.current)` check in the previous effect is correct though, then I guess it should not depend on any variables from the local scope like `[options, onReady]`, and then this `dispose` callback could still be moved there as well. Am I missing something?

packages/www/layouts/streamDetail.tsx
```diff
@@ -268,6 +268,33 @@ const StreamDetail = ({
...
+
+    player.hlsQualitySelector();
+
+    player.on("waiting", () => {
+      console.log("player is waiting");
+    });
+
+    player.on("dispose", () => {
+      console.log("player will dispose");
+    });
```
victorges: Do we need this logs?

packages/www/layouts/streamDetail.tsx
```diff
@@ -268,6 +268,33 @@ const StreamDetail = ({
...
+        src: globalPlaybackUrl,
+        type: "application/x-mpegURL",
+      },
+    ],
+  };
+
+  const handlePlayerReady = (player) => {
+    playerRef.current = player;
+
+    player.hlsQualitySelector();
```
victorges: Maybe this could be default, always done from the player itself 🤔 
Will we ever want not to have this?

victorges (approved):
Awesome!

packages/www/layouts/streamDetail.tsx
adamsoffer: Yeah good call. I made it a default.

packages/www/layouts/streamDetail.tsx
adamsoffer: Removed 👍

packages/www/components/Dashboard/Player/index.tsx
adamsoffer: I tried moving the `dispose` callback to the above effect, but for some reason it caused an error. 🤷 Kept as is.

packages/www/components/Dashboard/Player/index.tsx
adamsoffer: Removed

packages/www/components/Dashboard/Player/index.tsx
adamsoffer: Actually that works :) I combined the two.

Issue #634, opened 2021-10-07
Bug: Stream Health dashboard reports much lower kpbs than OBS


pglowacky:
@dob  mentioned this in Discord and I was able to reproduce. 

For whatever reason, my stream looked like this on our Stream Health dashboard:

![image.png](https://images.zenhubusercontent.com/602ea479bcf4ef10a9aace61/dafd4141-43dd-4776-915d-f1cda2be1160)

But OBS was reporting that the kpbs was much higher. 

![image.png](https://images.zenhubusercontent.com/602ea479bcf4ef10a9aace61/6352c592-6613-4619-bcf2-4ef74d047609)

Is this information getting displayed incorrectly? Is there a big slow down when we ingest?

iameli:
This looks to be a pretty simple bits/bytes discrepancy, with OBS showing 2400 kilobits and our graph showing `2400 / 8 = ~300` kilobytes. We can fix by just multiplying the number from stream-info-service by 8.

pglowacky:
@iameli Is this easy to fix? I've heard from a few customers how confusing this is, so if it's easy enough, I'd love to see this done ASAP

victorges:
@pglowacky This should have been fixed yesterday! Are you still observing the issue?

It had been fixed on https://github.com/livepeer/livepeer-com/pull/635 already but wasn't deployed explicitly. So it only got deployed yesterday with Adam's fix to the dashboard player.

pglowacky:
Look at that! It's fixed. Thanks all!

PR #635, opened 2021-10-07
www: show kilobits rather than kilobytes per second

iameli:
All credit to @victorges for finding the right line, I just typed the `* 8` into GitHub 🤷


iameli:
Fixes https://github.com/livepeer/livepeer-com/issues/634

victorges (approved):
LGTM

PR #649, opened 2021-10-19
cdn stats

### LINKED ISSUE #601: api: Implement CDN usage
* New API object: `/usage` or `/cdn-usage` or something
* Scheduled task that makes use of https://github.com/livepeer/cdn-log-analytics to push StackPath usage data into `PUT /cdn-usage/${hash of StackPath log file}` endpoints
* Some sort of something we'll work out for exposing data to users

darkdarkdragon:
Adds CDN usage data

- new table `cdn_usage_reg`
- new endpoint /cdn-data to receive CDN usage data from cdn-puller

Fixes #601 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/cdn-data.ts
+++ b/packages/api/src/controllers/cdn-data.ts
@@ -65,9 +65,9 @@ async function addMany(
         [
           date,
           region,
+          tdoc.playback_id,
           tdoc.user_id,
           tdoc.user_email,
-          tdoc.playback_id,
           tdoc.unique_users,
           tdoc.total_filesize,
           tdoc.total_cs_bytes,
@@ -218,7 +218,12 @@ app.post(
       }
       // insert data into db
       // @ts-ignore
-      const rows = data.data.filter((obj) => !obj.stream_id);
+      const rows = data.data.filter((obj) => obj.playback_id);
+      const badRows = data.data.filter((obj) => !obj.playback_id);
+      console.log(`==> bad rows:`, JSON.stringify(badRows, null, 2))
+      const badRows2 = data.data.filter((obj) => !obj.user_id);
+      console.log(`==> bad rows 2:`, JSON.stringify(badRows2, null, 2))
+      console.log(`==> good rows:`, JSON.stringify(rows, null, 2))
       const err = await addMany(data.date, data.region, data.file_name, rows);
       if (err) {
         logger.error(`Error saving row to db hour=${hour} err=${err}`);
--- a/packages/api/src/controllers/cdn-data.ts
+++ b/packages/api/src/controllers/cdn-data.ts
@@ -167,10 +167,12 @@ app.post(
       // aggregate stream_ids into playback_ids
       for (const row of data.data) {
         // console.log(`===> checking row `, row)
-        const user = await getUser(row.playback_id)
-        if (user) {
-          row.user_id = user.id
-          row.user_email = user.email
+        if (row.playback_id) {
+          const user = await getUser(row.playback_id)
+          if (user) {
+            row.user_id = user.id
+            row.user_email = user.email
+          }
         }
         if (row.stream_id) {
           // find playbackId by streamId
@@ -206,6 +208,11 @@ app.post(
           } else {
             row.playback_id = stream.playbackId;
             row.stream_id = null;
+            const user = await getUser(row.playback_id)
+            if (user) {
+              row.user_id = user.id
+              row.user_email = user.email
+            }
           }
         }
       }
--- a/packages/api/src/controllers/cdn-data.ts
+++ b/packages/api/src/controllers/cdn-data.ts
@@ -1,100 +1,19 @@
-import { SQLStatement } from "sql-template-strings";
-
 import { authMiddleware } from "../middleware";
 import { validatePost } from "../middleware";
 import { Response, Router } from "express";
 import { db } from "../store";
-import { FindOptions, FindQuery } from "../store/types";
 import logger from "../logger";
-import uuid from "uuid/v4";
-import { User } from "../schema/types";
+import { User, CdnDataPayload } from "../schema/types";
+import { CdnUsageRowReq } from "../store/cdn-usage-table";
 
 const app = Router();
 
-interface CdnUsageRowReq {
-  stream_id: string;
-  playback_id: string;
-  unique_users: number;
-  total_filesize: number;
-  total_cs_bytes: number;
-  total_sc_bytes: number;
-  count: number;
-  // filled in /api/cdn-data handler
-  user_id: string;
-  user_email: string;
-}
-
-interface SendData {
-  date: number;
-  region: string;
-  file_name: string;
-  data: Array<CdnUsageRowReq>;
-}
-
-async function addMany(
-  date: number,
-  region,
-  fileName: string,
-  rows: Array<CdnUsageRowReq>
-): Promise<Error> {
-  const name = "cdn_usage_reg";
-  const client = await db.pool.connect();
-  const newId = uuid();
-  console.log(
-    `-----> add many date=${date} region=${region} fileName=${fileName}`
-  );
-  try {
-    await client.query("BEGIN");
-    for (const tdoc of rows) {
-      await client.query(
-        `INSERT INTO ${name} VALUES (to_timestamp($1), $2, $3, $4, $5, $6, $7, $8, $9, $10)
-          ON CONFLICT (date, region, playback_id)
-          DO UPDATE SET 
-              unique_users = ${name}.unique_users + EXCLUDED.unique_users,
-              total_filesize = ${name}.total_filesize + EXCLUDED.total_filesize,
-              total_cs_bytes = ${name}.total_cs_bytes + EXCLUDED.total_cs_bytes,
-              total_sc_bytes = ${name}.total_sc_bytes + EXCLUDED.total_sc_bytes,
-              count = ${name}.count + EXCLUDED.count;
-          `,
-        [
-          date,
-          region,
-          tdoc.playback_id,
-          tdoc.user_id,
-          tdoc.user_email,
-          tdoc.unique_users,
-          tdoc.total_filesize,
-          tdoc.total_cs_bytes,
-          tdoc.total_sc_bytes,
-          tdoc.count,
-        ]
-      );
-    }
-    await client.query(
-      `INSERT INTO cdn_usage_last VALUES ($1, $2)
-          ON CONFLICT ((data ->> 'region'::text) )
-          DO UPDATE SET 
-            data = EXCLUDED.data
-      `,
-      [newId, JSON.stringify({ region, fileName })]
-    );
-    await client.query("COMMIT");
-  } catch (e) {
-    logger.warn(`--> error, rolling back`);
-    logger.info("--");
-    await client.query("ROLLBACK");
-    logger.warn(`--> error: ${e}`);
-    if (e.message.includes("duplicate key value")) {
-      // throw new BadRequestError(e.detail);
-      // return new BadRequestError(e.detail);
-    }
-    // throw e;
-    return e;
-  } finally {
-    client.release();
-  }
-  return null;
-}
+// interface SendData {
+//   date: number;
+//   region: string;
+//   file_name: string;
+//   data: Array<CdnUsageRowReq>;
+// }
 
 app.get("/region/:region", authMiddleware({}), async (req, res) => {
   const { region } = req.params;
@@ -136,7 +55,7 @@ app.post(
     };
 
     const start = Date.now();
-    const dataAr = req.body as Array<SendData>;
+    const dataAr = req.body as CdnDataPayload;
     console.log("Got data: ", JSON.stringify(dataAr, null, 2));
     if (!Array.isArray(dataAr)) {
       res.status(400);
@@ -154,8 +73,9 @@ app.post(
         if (row.playback_id) {
           const user = await getUser(row.playback_id);
           if (user) {
-            row.user_id = user.id;
-            row.user_email = user.email;
+            const urow = row as CdnUsageRowReq;
+            urow.user_id = user.id;
+            urow.user_email = user.email;
           }
         }
         if (row.stream_id) {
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+
+const app = Router();
+
+interface CdnUsageRowReq {
+  stream_id: string;
+  playback_id: string;
+  unique_users: number;
+  total_filesize: number;
+  total_cs_bytes: number;
+  total_sc_bytes: number;
```
victorges: What does `cs` and `sc` mean? Would it be possible to be more explicit on what these mean? If it'd be too long, maybe just a comment on the fields or the schema perhaps.

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+      const badRows2 = data.data.filter((obj) => !obj.user_id);
+      console.log(`==> bad rows 2:`, JSON.stringify(badRows2, null, 2));
+      console.log(`==> good rows:`, JSON.stringify(rows, null, 2));
+      const err = await addMany(data.date, data.region, data.file_name, rows);
+      if (err) {
+        logger.error(`Error saving row to db hour=${hour} err=${err}`);
+        process.exit(1);
+        throw "fuckk";
+      }
+      /*
```
victorges: Can this commented code be removed?

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+    res.json(docs[0]);
+  }
+);
+
+app.post(
+  "/",
+  //   authMiddleware({}),
+  // validatePost("stream"),
+  async (req, res) => {
+    // if (!req.body || !req.body.name) {
```
victorges: This also doesn't seem necessary, can it be removed?

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+  file_name: string;
+  data: Array<CdnUsageRowReq>;
+}
+
+async function addMany(
+  date: number,
+  region,
+  fileName: string,
+  rows: Array<CdnUsageRowReq>
+): Promise<Error> {
```
victorges: I think this should be moved to the `cdnUsage` table instead, not to have a lot SQL in the API controller.

Edit: Noticed that the `db.cdnUsage` is actually unused as of now. I think we should have a separate table for this in the `store` folder, even if it is just implemented from scratch (without inheriting from `Table`). Then any SQL logic should be kept there (including creating tables and indexes perhaps).

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+import { Response, Router } from "express";
+import { db } from "../store";
+import { FindOptions, FindQuery } from "../store/types";
+import logger from "../logger";
+import uuid from "uuid/v4";
+import { User } from "../schema/types";
+
+const app = Router();
+
+interface CdnUsageRowReq {
```
victorges: Could this be defined in `schema.yaml` instead, so that we also have some schema validators to use here? You can even create a separate `DbCdnUsage` or `CdnUsageRow` type inheriting from the request type to add the additional fields you want in the DB but not in the request.

e.g. [`DBStream` type](https://github.com/livepeer/livepeer-com/blob/a21b96ad54399ce8694e32e013dc2414b6b16bce/packages/api/src/store/stream-table.ts#L50), tho CDN usage would be much simpler as of now, something like
```
type DBCdnUsage = CdnUsageReq & {
    user_id: string;
    user_email: string;
}
```

Edit: Just saw that there **is** a schema type. Is it unused? Could we use it instead? We can remove the `table:` directive in the type to avoid the table creation and just use the generated types/validators here.

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+    `-----> add many date=${date} region=${region} fileName=${fileName}`
+  );
+  try {
+    await client.query("BEGIN");
+    for (const tdoc of rows) {
+      await client.query(
+        `INSERT INTO ${name} VALUES (to_timestamp($1), $2, $3, $4, $5, $6, $7, $8, $9, $10)
+          ON CONFLICT (date, region, playback_id)
+          DO UPDATE SET 
+              unique_users = ${name}.unique_users + EXCLUDED.unique_users,
```
victorges: **nit:** "unique users" seems confusing to me here since this entry is for a specific user already. If it's not too much trouble to rename the fields/APIs, WDYT of `unique viewers` or even a more explicit `unique_client_ips` so there's no question about what these mean/how accurate it is?

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+  const client = await db.pool.connect();
+  const newId = uuid();
+  console.log(
+    `-----> add many date=${date} region=${region} fileName=${fileName}`
+  );
+  try {
+    await client.query("BEGIN");
+    for (const tdoc of rows) {
+      await client.query(
+        `INSERT INTO ${name} VALUES (to_timestamp($1), $2, $3, $4, $5, $6, $7, $8, $9, $10)
```
victorges: Any reason not to use those `SQLStatement` literals like 
```
sql`SELECT ... ${var} ...`
```
?

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+        })
+        const err = await db.cdnUsageRegular.add({
+          date: data.date,
+          region: data.region,
+          ...row,
+        })
+        if (err) {
+          logger.error(`Error saving row to db hour=${hour} err=${err}`)
+          process.exit(1)
+          throw ('fuckk')
```
victorges: been there done that 😆

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+      console.log(`==> bad rows:`, JSON.stringify(badRows, null, 2));
+      const badRows2 = data.data.filter((obj) => !obj.user_id);
+      console.log(`==> bad rows 2:`, JSON.stringify(badRows2, null, 2));
+      console.log(`==> good rows:`, JSON.stringify(rows, null, 2));
+      const err = await addMany(data.date, data.region, data.file_name, rows);
+      if (err) {
+        logger.error(`Error saving row to db hour=${hour} err=${err}`);
+        process.exit(1);
+        throw "fuckk";
+      }
```
victorges: Reminder to remove this `exit(1)` from here as well

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+            logger.error(
+              `For stream_id=${stream.parentId} in hour=${hour} can't find playbackId.`
+            );
+            continue;
+          }
+          const playbackRow = data.data.find(
+            (row) => row.playback_id === stream.playbackId
+          );
+          if (playbackRow) {
+            playbackRow.count += row.count;
```
victorges, line 99: Curious, what are the cases when there is a row coming from CDN data that only has a `stream_id` and the `playback_id` is on a separate value? Would this catch like API calls for the stream info?

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+        }
+      }
+      // insert data into db
+      // @ts-ignore
+      const rows = data.data.filter((obj) => obj.playback_id && obj.user_id);
+      const badRows = data.data.filter((obj) => !obj.playback_id);
+      console.log(`==> bad rows:`, JSON.stringify(badRows, null, 2));
+      const badRows2 = data.data.filter((obj) => !obj.user_id);
+      console.log(`==> bad rows 2:`, JSON.stringify(badRows2, null, 2));
+      console.log(`==> good rows:`, JSON.stringify(rows, null, 2));
```
victorges: Probably better to remove these logs in production (at least the "good rows" one, which should be a lot of data I guess)

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+    logger.warn(`--> error, rolling back`);
+    logger.info("--");
+    await client.query("ROLLBACK");
+    logger.warn(`--> error: ${e}`);
+    if (e.message.includes("duplicate key value")) {
+      // throw new BadRequestError(e.detail);
+      // return new BadRequestError(e.detail);
+    }
+    // throw e;
+    return e;
```
victorges: Would it be possible to throw the error instead of returning? Returning seems very uncommon for js/ts

packages/api/src/store/db.ts
```diff
@@ -243,4 +257,23 @@ async function ensureDatabase(postgresUrl) {
...
+            region character varying(128) COLLATE pg_catalog."default" NOT NULL,
+            playback_id character varying(128) COLLATE pg_catalog."default" NOT NULL,
+            user_id character varying(128) COLLATE pg_catalog."default" NOT NULL,
+            user_email character varying(512) COLLATE pg_catalog."default" NOT NULL,
+            unique_users integer NOT NULL,
+            total_filesize bigint NOT NULL,
+            total_cs_bytes bigint NOT NULL,
+            total_sc_bytes bigint NOT NULL,
+            count integer NOT NULL,
+            CONSTRAINT prim PRIMARY KEY (date, region, playback_id)
```
victorges: I think it _might_ be better to have a single separate column for the `PRIMARY KEY`, like an auto-inremented `id` on the simplest case or a manually generated UUID like we have for our other tables (bonus of consistency on that).

Not sure if that is the case for Postgres, but I heard that at least for MySQL it's much more performant to keep the primary key as simple as possible for low-level data access in the storage system, and create separate more complex indexes for any other helpful queries we may want to run.

So here maybe have an `id VARCHAR(128)` (UUID) column as the primary key and then create helpful indexes for querying. This current might be one of them, and I believe we might want one with the `user_id` as well. Maybe just add the `user_id` before the `playback_id`, since any query for a `playback_id` could/should always filter by the `user_id` as well? `(date, region, user_id, playback_id)`

packages/api/src/controllers/cdn-data.ts
```diff
@@ -0,0 +1,272 @@
...
+      const stream = await db.stream.getByPlaybackId(playbackId);
+      if (!stream) {
+        logger.error(`Can't find stream for playbackId=${playbackId}`);
+        return null;
+      }
+      const user = await db.user.get(stream.userId);
+      if (!user) {
+        logger.error(`Can't find user for playbackId=${playbackId}`);
+        return null;
+      }
```
victorges, line 45: Maybe worth saving in the cache when it's not found as well.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: client-server and sever-client

victorges: Oh I see. WDYT of using "upstream" to cs and "downstream" to sc?

darkdarkdragon: Personally for, with words `upstream`/`downstream` I'm always struggle to understand where is _up_ and where is _down_ in relevant to context 😄 
I would prefer to have `client` and `server` in names 😃

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: In CDN data we have either `playback_id` or `stream_id`. We have `stream_id` in case of recording playback - recording URL contains UUID of the session.

packages/api/src/controllers/cdn-data.ts
victorges: Fair enough! I personally had trouble with the up/downstream terms as well until I read some explanation somewhere (after googling several times) that the analogy is basically a "river of data". Data generally flows from the server to the client, so that's downstream/download. Sending data from the client to the server is going against the flow of the data river, so it's upstream/upload.

But it's definitely not the most self-explanatory term. Maybe just some comments explaining what `cs` and `sc` mean here/in the schema should be enough.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: removed

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: I've removed `cdnUsage` table, moved SQL into own class.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: Used schema.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: renamed

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: No particular reason, this was just simple.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: removed

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: removed

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: Changed to throw.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: It really __shouldn't__ happen, so not optimising for that.

packages/api/src/controllers/cdn-data.ts
darkdarkdragon: I've added descriptions to schema. Took from logs format description.

PR #682, opened 2021-10-28
api: Create authorization API to be used by other services

victorges:
This is to create a new `/api/auth` endpoint which can be used as an API for authorizing
requests in other services in our infrastructure. The design was strongly based on 
[NGINX'es sub-request authorization](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication), so that the authorization request is as much as a "proxy"
as possible, adding only additional headers to account for additional information that may
be useful for the authorization process.

In other words, the authorization request has:
 - The same method as the original request
 - No body (authorization should be based on the accessed resource only, not the contents of the request)
 - All the original headers from the request (but we only really need the `Authorization` one, so services may optimize)
 - An `X-Original-Uri` header with... the original URI of the request being authorized
 - An `X-Livepeer-Stream-Id` header with the ID of the stream which is being accessed. This is used to 
 detect cross-account accesses which should not be authorized.

And the repsonse of the `/api/auth` request can also be directly proxied to the client, as it will contain
 whatever verbatim authorization error should be given to the client.

 - Add support in the `auth` middleware for an `originalUriHeader`, which should be used instead of
 the URI for authorizing a given request.
    - The middleware is responsible mostly for authentication, and authorization with the built-in 
    API key auth access rules (`X-Original-Uri` header).
 - Create a new `auth` controller and register its API on our router.
    - The controller is responsible for the authorization of access to the given resource. This currently only
    means that the stream ID belongs to the calling user (or the user is an admin).

## -

 - Start API and analyzer configured to authorize with the API
 - Try making calls without an auth header and check it fails
   - Same with invalid auth headers
 - Check that a user can access their own stream
 - Check that a user cannot access other users streams
 - Check that an admin can access any stream
All good ✅ 

Implements the first part of https://github.com/livepeer/livepeer-data/issues/4

**Checklist:**


packages/www/hooks/use-analyzer.tsx
```diff
@@ -1,6 +1,8 @@
-import { useContext, createContext, ReactNode } from "react";
+import EventSource from "eventsource";
```
iameli, line 1: Oh neat. So is this a totally custom fetch-based `eventsource` implementation now? I can't imagine that's a problem but cool.

iameli (approved):
LGTM

packages/api/src/controllers/auth.ts
```diff
@@ -0,0 +1,30 @@
...
+import { db } from "../store";
+
+const app = Router();
+
+app.all(
+  "/",
+  authMiddleware({ originalUriHeader: "x-original-uri" }),
+  async (req, res) => {
+    const streamId = req.headers["x-livepeer-stream-id"];
+    const stream = await db.stream.get(streamId?.toString() ?? "");
```
darkdarkdragon, line 16: If `x-livepeer-stream-id` is absent this will throw error, and request will return 500 and some stack trace.
Wouldn't it be better to return 400 with clear message?

packages/api/src/middleware/auth.ts
```diff
@@ -106,8 +108,14 @@ function authFactory(params: AuthParams): RequestHandler {
     }
     const accessRules = tokenObject?.access?.rules;
     if (accessRules) {
-      const fullPath = pathJoin2(req.baseUrl, req.path);
-      const { httpPrefix } = req.config;
+      let fullPath = pathJoin2(req.baseUrl, req.path);
+      let { httpPrefix } = req.config;
+      if (params.originalUriHeader) {
+        const header = req.headers[params.originalUriHeader];
+        const originalUri = new URL(header?.toString() ?? "");
```
darkdarkdragon, line 115: If `header` is empty this will throw error and return 500. 
Can we at least return 400?

PR #707, opened 2021-11-10
api: remove unnecessary dependency

iameli:
This has been in Node.js since 15.0.0. We run on 16. Unnecessary.

### DIFF
--- a/packages/api/src/middleware/geolocate.js
+++ b/packages/api/src/middleware/geolocate.js
@@ -1,8 +1,6 @@
 "use strict";
 
 import fetch from "node-fetch";
-// This is a spec but it doesn't look like it's very well supported, so
-import promiseAny from "promise.any";
 
 export function addProtocol(url, protocol = "https") {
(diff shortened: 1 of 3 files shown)


packages/api/src/middleware/geolocate.js
```diff
@@ -58,7 +56,7 @@ function geoLocateFactory({ first = true, region = "region" }) {
     let data;
     try {
       if (first) {
-        data = await promiseAny(promises);
+        data = await Promise.any(promises);
```
victorges, line 59: Do we actually use this on the API as well? Thought it was a worker thing only

victorges (approved):
Noice

packages/api/src/middleware/geolocate.js
iameli: The functionality is still there, maybe we'd want to use it to geolocate the second-nearest entrypoint or something. But yeah.

PR #709, opened 2021-11-10
www: Fix health progression on beginning of stream

### LINKED ISSUE #770: Stream health shows as "unhealthy" while stream is loading
## Current state

When beginning a stream, the stream health dashboard's global health indicator will display as "Unhealthy" while the stream is loading into the player. Once the video loads, the global stream indicator displays as "Healthy." 

[Video](https://drive.google.com/file/d/1a9zCxB9XFRz1xowHX8-YRlnCw-DnZdrj/view?usp=drivesdk)

This is a potential source of confusion for customers who may believe that there is something wrong with their streams when their stream simply hasn't loaded completely. 

## Desired state

How we display Multistream health solves this confusion by displaying a "Pending" health indicator until the multistream destination has connected. 

A "Pending" indicator for the global stream health indicator would solve this confusion

victorges:
This fixes a couple of glitches that happen with the current stream health UI, in which we
might both display health information about previous stream sessions, as well as display
invalid health information when the stream config has changed after the stream had already
been started (e.g. displaying "pending" for a multistream target only created after the stream
was active).

 - Start processing and using the "stream active since" information, which can now be obtained from
 the stream health payload via the `Active` condition!
 - Ignore conditions and multistream status from before the stream was active
 - Also ignore multistream targets created after the stream was active

## -

TODO:
1
 - [ ] Start a stream 
 - [ ] Make sure stream health statuses still works
 - [ ] Change multistream configuration
 - [ ] Make sure any target created after stream started is displayed as "Idle" instead of "Pending"
2
 - [ ] Stop the stream, restart it
 - [ ] Make sure all stream health stuff restarts as "-" instead of "Healthy" right from the start
 - [ ] Make sure multistream status still works (from the target created above)
3
 - [ ] Stop the stream, restart it
 - [ ] Make sure multistream target status starts as "Pending" instead of "Offine" (from previous stream event)

Now yeah! Fixes https://github.com/livepeer/livepeer-com/issues/770

**Screenshots (optional):**
<pending>

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/HealthChecksTable/index.tsx
+++ b/packages/www/components/Dashboard/HealthChecksTable/index.tsx
@@ -65,11 +65,18 @@ const HealthChecksTable = ({
 
   const conditionsMap = useMemo(
     () =>
-      streamHealth?.conditions.reduce(function (map, condition) {
+      (streamHealth?.conditions ?? []).reduce(function (map, condition) {
         map[condition.type] = condition;
         return map;
-      }, {} as Record<string, Condition>),
+      }, {} as Record<Condition["type"], Condition>),
     [streamHealth?.conditions]
+  );
+  const streamActiveSince = useMemo(
+    () =>
+      conditionsMap.Active?.status
+        ? conditionsMap.Active.lastTransitionTime
+        : null,
+    [conditionsMap.Active]
   );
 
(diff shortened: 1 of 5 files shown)


iameli (approved):
LGTM though I don't have quite all the context here

PR #745, opened 2021-11-25
api/auth: Allow access to deleted streams

victorges:
We currently can't get health or events about
streams that have been deleted. This prohibits
from investigating a bad stream after it has been
deleted, like from record-tester, and could create
similar problems to end-users as well. Don't see why
we shouldn't allow that access anyway, so skipping the
check for the stream.deleted in the auth controller logic.

Stream not found error grabbing stream health from a deleted stream.

**Checklist:**


iameli (approved):
Cool yeah, fair enough.

PR #746, opened 2021-11-29
handle test for the Recaptcha v3

FullStackTiger:

- handle test for the Recaptcha v3

**Screenshots (optional):**

**Checklist:**


packages/api/src/parse-cli.ts
```diff
@@ -237,6 +237,11 @@ export default function parseCli(argv?: string | readonly string[]) {
         describe: "google recaptcha secret key",
         type: "string",
       },
+      recaptchaMode: {
+        describe: "google recaptcha mode",
```
victorges: WDYT of
```suggestion
      enableRecaptcha: {
        describe: "whether to validate user registrations with google recaptcha",
```
Would also need to update places that use it with the new var name.

Alternatively, just a nit: maybe invert the configuration as well and call it `disableRecaptcha` instead, just so the default value is more natural (`false` for a boolean).

victorges (approved):
LGTM for fixing the tests.

Even better than this would be actually updating the test code to account for the recaptcha logic, and actually validate that it is doing what we expect somehow.

PR #768, opened 2021-12-03
Blank page on Dashboard

FullStackTiger:

- When it runs through the flow on Livepeer.monster, it gets a blank page after loading. 

- Added email verification function to every page on frontend side
- Remove allowUnverified param function on the auth middleware

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/pages/app/admin/keys.tsx
+++ b/packages/www/pages/app/admin/keys.tsx
@@ -1,21 +1,16 @@
 /** @jsx jsx */
 import { jsx } from "theme-ui";
 import useApi from "hooks/use-api";
-import Layout from "layouts/admin";
+import { withEmailVerifyMode } from "layouts/withEmailVerifyMode";
 import useLoggedIn from "hooks/use-logged-in";
 import TabbedLayout from "@components/Admin/TabbedLayout";
 import AdminTokenTable from "@components/Admin/AdminTokenTable";
 import { getTabs } from "../admin";
 
-const emailVerificationMode =
-  process.env.NEXT_PUBLIC_EMAIL_VERIFICATION_MODE === "true";
-
 const Keys = () => {
   useLoggedIn();
   const { user, logout } = useApi();
-  if (!user || (emailVerificationMode && user.emailValid === false)) {
-    return <Layout />;
-  }
+
   const tabs = getTabs(1);
 
   return (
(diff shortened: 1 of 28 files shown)


adamsoffer (changes requested):
Instead of including this `emailVerificationMode` conditional on every page, I think we should only include once in a higher order component. Cleaner.

adamsoffer:
@FullStackTiger  hmm on second thought, possible to just add one conditional to the layout component rather than every page? This solution still requires we wrap each page with the HOC which would be good to avoid if possible.

adamsoffer (approved):
LGTM. @iameli want to give this a quick look over as well since it touches the API?

adamsoffer:
@FullStackTiger I just noticed it's failing a prettier check.

FullStackTiger:
> @FullStackTiger I just noticed it's failing a prettier check.

I fixed this prettier issue with another PR, already merged it to the master

packages/www/pages/app/admin/keys.tsx
```diff
@@ -10,7 +10,7 @@ import { getTabs } from "../admin";
 const Keys = () => {
   useLoggedIn();
   const { user, logout } = useApi();
-  if (!user || user.emailValid === false) {
+  if (!user) {
```
iameli, line 13: Example of the parameter I was discussing:

```suggestion
  if (!user || user.emailValid === req.config.requireEmailVerification) {
```

iameli (changes requested):
Let's do the same thing we did on the front-end and add a new configuration parameter in `src/parse-cli.js` — `--require-email-verification`. It can default to false.

I left an example on one of the routes for what a check could look like. I'd also be okay with adding a new middleware helper that implements that check so we don't have to copy the `req.config.requireEmailVerification` everywhere.

packages/www/pages/app/admin/keys.tsx
victorges: Maybe we need to check both separately? Not sure what we want to happen on a `emailValid = true` and `req.config.requireEmailVerification = false` case.

```suggestion
  if (!user || (!user.emailValid && req.config.requireEmailVerification)) {
```

pglowacky (approved):

PR #784, opened 2021-12-13
(testing) return half of the orchestrators as untrusted

darkdarkdragon:

PR for fast verification testing in staging

## -

**Screenshots (optional):**

**Checklist:**


packages/api/src/parse-cli.ts
```diff
@@ -242,6 +242,12 @@ export default function parseCli(argv?: string | readonly string[]) {
         default: false,
         type: "boolean",
       },
+      "half-region-orchestrators-untrusted": {
+        describe:
+          "mark half of the orchestrators returned by /api/region as untrusted. For use in staging!",
+        default: false,
+        type: "boolean",
+      },
```
victorges: Maybe we don't even need to merge this and thus don't need the flag. We can just change the image used for staging deployments on `livepeer-infra` to be the one coming from this branch.

darkdarkdragon: I did it initially, but then it was overwritten in today's deploy 😢 
But yah, let's try it again

victorges: Oh I see. We could merge to `master` the change on `livepeer-infra` values/image for the staging deploy. The flag/env setting would also be overwritten if we didn't merge it anyway

victorges: So... I wasn't aware this would be such a long running test. I think we should actually merge this and add the CLI arg only to staging instead as you were planning on doing originally. Otherwise, no new changes are being picked up by staging when we merge to `master`. WDYT?

victorges: For now, I rebased this on top of `master` if you don't mind, which is OK as well for me. Just keep in mind that you might need a `reset` or `pull --rebase` to update to the `origin` version of the branch now

thomshutt: Hey @victorges, I'm picking this up again - is reinstating the flag and passing that in for Staging still the way you want to go?

victorges: Sure! I think in the end this was a long-running test that we wanna keep going apart from other potential tests we want to make. In hindsight it would've been better to just merge this to master but only configure the flag in staging. 

So it makes more sense to pick this up, merge it to `master`, and set the flag only on staging! Let me know when you rebase this and it's ready for a review/merging.

thomshutt:
Messed up the rebase. I'm going to just close this and move to https://github.com/livepeer/livepeer-com/pull/1064

PR #785, opened 2021-12-13
api: Make auth header parsing insensitive to case and whitespace

victorges:
This is to improve the developer experience by not requiring that they specify their authorization
headers with perfect casing. [The "specs"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) also state that the auth scheme should be case-insensitive.

Also made it a little more whitespace insensitive while on it, since it should not affect our ability to
process those anyway.

 - whitespace-trim any value read from the auth header
 - toLowerCase the auth scheme and compare with lower case constants
 - Rename some vars and helpers for consistency with spec

## -

Make some requests and ensure authorization still work.

No. Just an invisible one that's been annoying me in my head.

**Checklist:**

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -68,7 +68,6 @@
     "email-validator": "^2.0.4",
     "express": "^4.16.4",
     "express-async-errors": "^3.1.1",
-    "express-bearer-token": "^2.4.0",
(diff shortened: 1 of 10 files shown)


iameli (approved):
LGTM! This'll be nice.

PR #792, opened 2021-12-14
api: Set helpful application name in all Postgres connections

victorges:

This improves connection metadata by specifying a helpful application name on
all PostgreSQL connections. Also includes the own region name on that name in case
it is available.

 - Set `application_name` property on replica connections as well
 - Include own region name on `application_name`

## -

Just run tests and make sure they still work. If so, will just check in prod if the metadata updated.

FTR, this is the query to get all connected clients and metadata:
```
select pid as process_id, 
       usename as username, 
       datname as database_name, 
       client_addr as client_address, 
       application_name,
       backend_start,
       state,
       state_change
from pg_stat_activity;
```

No.

**Checklist:**


iameli (approved):
LGTM! Nice to see all that cleanup there

PR #793, opened 2021-12-15
api: Fix admin queries table scans

victorges:
This fixes some queries that have been performing table scans on the streams due to a
`LIKE %x%` filter on a field with only a B-TREE index.

Fixed it 2-fold:
 - Started using an `=` filter for most of the fields we allow user-defined filters on;
 - Specify a `userId` on that specific query, so on the worst case we still don't do a full table scan.

## -

Open `pages/app/stream/[id]` page (old dashboard) and make sure query still works (and is cheaper)

Fixes [this issue](https://eu-metrics-monitoring.livepeer.live/grafana/explore?orgId=1&left=%5B%22now-7d%22,%22now%22,%22Loki%22,%7B%22exemplar%22:true,%22expr%22:%22%7Bregion%3D~%5C%22.%2B%5C%22,%20app%3D%5C%22prod-livepeer-api%5C%22%7D%20%7C%3D%20%5C%22runQuery%20phase%3Dsuccess%5C%22%20%7C%3D%20%5C%22SELECT%5C%22%20%7C%20pattern%20%5C%22%3C_%3Eelapsed%3D%3Celapsed%3E%20%3C_%3E%5C%22%20%7C%20elapsed%20%3E%2015s%22%7D%5D)

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/api-token.js
+++ b/packages/api/src/controllers/api-token.js
@@ -45,10 +45,10 @@ app.delete("/:id", async (req, res) => {
 
 const fieldsMap = {
   id: `api_token.ID`,
-  name: `api_token.data->>'name'`,
+  name: { val: `api_token.data->>'name'`, type: "full-text" },
   lastSeen: `api_token.data->'lastSeen'`,
   userId: `api_token.data->>'userId'`,
-  "user.email": `users.data->>'email'`,
+  "user.email": { val: `users.data->>'email'`, type: "full-text" },
 };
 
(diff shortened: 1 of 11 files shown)


darkdarkdragon (approved):
can you please also add userId here? https://github.com/livepeer/livepeer-com/blob/b8a176bd603dc03ad5a888b2822313fb019a1faa/packages/www/layouts/streamDetail.tsx#L220
otherwise looks good

PR #800, opened 2021-12-18
api: Simpler AMQP logic for delayed events

victorges:
This is to simplify the logic for delaying messages on AMQP, also avoiding the
creation and destruction of multiple queues in runtime, as well as the dynamic
management of setup functions on the AMQP connection manager. 

Not the biggest priority, but I was bored while monitoring some incidents over
the weekend and decided to hack this out, seems to have worked nicely.
Sponsored by all the RabbitMQ learnings for stream health :D

 - Create a separate exchange E for publishing delayed messages (necessary for separate routing keys for each message)
 - Create a delayed queue Q, bound to everything on that exchange, with a dead-letter exchange set to the (other) main webhooks exchange
( - Publish messages on the exchange E, with the desired routing key and with an `expiration` set to desired delay. Message expires on the queue and is sent to the webhook's main exchange with the same routing key.)

## -

`yarn test`, we have plenty of tests on that part of the code :)
(at least for the basic retry/delay logic)

Not really, was an on-call bonus.
Edit: Created one. Now this implements #812

**Checklist:**

### DIFF
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -19,7 +19,7 @@ type RoutingKey = `events.${EventKey}` | `webhooks.${string}`;
 
 export default interface Queue {
   publishWebhook(key: RoutingKey, msg: messages.Webhooks): Promise<void>;
-  delayedPublish(
+  delayedPublishWebhook(
     key: RoutingKey,
     msg: messages.Any,
     delay: number
@@ -40,7 +40,11 @@ export class NoopQueue implements Queue {
     );
   }
 
-  async delayedPublish(key: RoutingKey, msg: messages.Any, delay: number) {
+  async delayedPublishWebhook(
+    key: RoutingKey,
+    msg: messages.Any,
+    delay: number
+  ) {
     console.warn(
       `WARN: Delayed publish event to noop queue. key=${key} delay=${delay} message=`,
       msg
@@ -165,7 +169,7 @@ export class RabbitQueue implements Queue {
     });
   }
 
-  public async delayedPublish(
+  public async delayedPublishWebhook(
     routingKey: RoutingKey,
     msg: messages.Any,
     delay: number
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -91,6 +91,7 @@ export class RabbitQueue implements Queue {
           channel.bindQueue(QUEUES.webhooks, EXCHANGES.webhooks, "webhooks.#"),
           channel
             .assertQueue(QUEUES.delayed, {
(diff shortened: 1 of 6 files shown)


gioelecerati (approved):
LGTM!

packages/api/src/store/queue.ts
```diff
@@ -137,40 +159,25 @@ export class RabbitQueue implements Queue {
...
-        persistent: true,
-      });
-    } finally {
-      // avoid accumulating duplicate setup funcs on the channel manager
-      await this.channel.removeSetup(setupFunc, () => {});
-    }
+    console.log(`emitting delayed message: delay=${delay / 1000}s msg=`, msg);
+    await this.channel.publish(EXCHANGES.delayed, routingKey, msg, {
+      persistent: true,
+      expiration: delay,
```
hjpotter92, line 184: just confused here, `expiration` implies that the event/message would expire after the defined value, whereas `delay` signifies the message would be published after the `value` time has passed. which is it? should it be reworded?

hjpotter92: to be clear, i get that the channel will delete message after `delay` time has passed and move it to the `deadLetterExchange` declared, my question is more in terms of terminology

victorges: So it is actually both! One refers to what it means for the delayed publishing "interface", while the other refers to how it is actually implemented in RabbitMQ and thus is passed on the lower-level AMQP publish here.

The pipeline works as this:
 - Message is published to the `delayed` exchange
 - The only queue subscribed to that is the `delayed` queue
 - No one is actually subscribed to that queue, but that queue has a `deadLetterExchange` configured to the regular `webhooks` exchange
 - Once the message reaches its **`expiration`** deadline, it is removed from that queue and re-published on the `webhooks` exchange

So in the low-level RabbitMQ, it really means the **expiration** of the message as the time that the message will spend on the `delayed` queue before being re-published on the main `webhooks` exchange where we bind queues and consume the events.

I think it would be weird to name the public argument as expiration though, as for the high level interface for publishing a delayed message it is just as if the message was delayed by some time before actually reaching the `webhooks` exchange. 

WDYT? Maybe I could add a comment here just above this `publish` or above this `expiration: delay` with some TL;DR of this workflow, to try to avoid confusion.

hjpotter92: yes, I think a small-ish comment in the function definition should help. When going through the PR, I dived directly through function definition and got confused until i reading PR description

victorges: Done!

hjpotter92 (approved):

PR #802, opened 2021-12-20
added redirection function

### LINKED ISSUE #801: Livepeer.com Verify Page is not redirect correctly

FullStackTiger:

 - added redirection function

- Fixes #801 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/pages/verify.jsx
+++ b/packages/www/pages/verify.jsx
@@ -14,8 +14,27 @@ import Guides from "components/Marketing/Guides";
 const Verify = () => {
   useLoggedIn();
   const router = useRouter();
-  const { user } = useApi();
+  const { user, verify } = useApi();
   const { email, emailValidToken } = router.query;
+
+  useEffect(() => {
+    if (email && emailValidToken) {
+      verify(email, emailValidToken).then(() => {
+        if (selectedPlan === "1") {
+          router.replace("/dashboard/billing/plans?promptUpgrade=true");
+        } else {
+          router.replace("/dashboard");
+        }
+      });
+    }
+  }, [email, emailValidToken]);
+
+  // If they've already validated their email, get 'em out of here
+  useEffect(() => {
+    if (user && user.emailValid !== false) {
+      router.replace("/dashboard");
+    }
+  }, [user]);
 
   return (
     <Layout>
--- a/packages/www/pages/verify.jsx
+++ b/packages/www/pages/verify.jsx
@@ -31,7 +31,7 @@ const Verify = () => {
 
   // If they've already validated their email, get 'em out of here
   useEffect(() => {
-    if (user && user.emailValid !== false) {
+    if (user?.emailValid === true) {
       router.replace("/dashboard");
     }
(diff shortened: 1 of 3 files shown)


packages/www/pages/verify.jsx
```diff
@@ -14,9 +14,28 @@ import Guides from "components/Marketing/Guides";
 const Verify = () => {
   useLoggedIn();
   const router = useRouter();
-  const { user } = useApi();
+  const { user, verify } = useApi();
   const { email, emailValidToken } = router.query;
 
+  useEffect(() => {
+    if (email && emailValidToken) {
+      verify(email, emailValidToken).then(() => {
```
victorges: Does `verify` throw an exception in case verification fails? (e.g. invalid token)
What would happen on that case? Any easy way to display an error, maybe with a toast or sth?

No problem if there's currently no pattern for that in the dashboard and it's harder than it should be (I know we have a current tech debt for not showing errors to the user on most ops).

packages/www/pages/verify.jsx
```diff
@@ -14,9 +14,28 @@ import Guides from "components/Marketing/Guides";
...
+        } else {
+          router.replace("/dashboard");
+        }
+      });
+    }
+  }, [email, emailValidToken]);
+
+  // If they've already validated their email, get 'em out of here
+  useEffect(() => {
+    if (user && user.emailValid !== false) {
```
victorges: **nit:**
```suggestion
    if (user?.emailValid === true) {
```
since `user.emailValid: undefined` would also pass the current test and we probably don't want undefined/null (or anything not a boolean) to be considered a validated user.

packages/www/pages/verify.jsx
```diff
@@ -14,9 +14,28 @@ import Guides from "components/Marketing/Guides";
 const Verify = () => {
   useLoggedIn();
   const router = useRouter();
-  const { user } = useApi();
+  const { user, verify } = useApi();
   const { email, emailValidToken } = router.query;
 
+  useEffect(() => {
```
victorges, line 20: So, in practice, we weren't validating the user's email anymore even when they did click on the link received on the email? Good thing we got this back.

victorges (approved):
LGTM!

packages/www/pages/verify.jsx
FullStackTiger: if they did click on the link received on the email, we will have the user's email validating, so emailValid will be true.

packages/www/pages/verify.jsx
FullStackTiger: in case verification fails, does the user need to signup again?

FullStackTiger: so we need to display the text like "We got an invalid token, you need to [signup](https://livepeer.com/) again"?

FullStackTiger: in case verification fails, added the text like "Email verification failed: ${err}. Please access the link you received by email again."

PR #805, opened 2021-12-22
api: Remove Segment API integration

victorges:

It is adding seconds of latency to our APIs further away from the US.

 - Make the `trackAction` function not even return a promise so it's not possible to await for it
 - Fix usage anyway just in case
 - You know what, just remove it completely. We found out no one uses it.

## -

`yarn test`

No. Only some increased API latency I noticed on EU regions (which don't make much sense since our DB is in EU)
With the new metric that @hjpotter92 implemented, also confirmed that indeed adds seconds of latency to the EU regions.

Edit: Now it does, implements #813

**Checklist:**

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+packages/api      @livepeer/video-user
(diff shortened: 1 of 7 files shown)


victorges:
FTR @iameli found out that we don't even use segment anymore. Will pivot this to completely remove segment API integration from the code.

iameli (approved):
LGTM!

PR #806, opened 2021-12-22
api: Modernize users controller

victorges:
Translate it to TypeScript (no implicit any) and remove usages of `req.store`.

 - Translate it to TypeScript (no implicit any)
 - Remove usages of `req.store` 

## -

`yarn test`

No. It's tech debt end of year!
Edit: Implements #815 

**Checklist:**


packages/api/src/controllers/multistream.ts
```diff
@@ -29,11 +30,17 @@ function adminListQuery(
   orderStr: string,
   filters: string
 ): [SQLStatement[], FindOptions] {
+  type ResultRow = {
```
hjpotter92, line 33: Shouldn't the type declaration be outside function?

victorges: It could be declared externally as well, but since it's a strict local type that never leaves the function I thought it would make sense to keep it local as well, just like a local var (it's allowed by `ts`). I also followed the pattern (that I had started myself 🤣) in the `stream.ts` controller which declares a similar type when calling the generic `db.table.find`.

I don't have a strong opinion about it though and could easily move it outside the function instead. WDYT?

hjpotter92: resolved in discord conversation.

---

I was only confused about performance benefits. But in practice, it doesn't matter.

hjpotter92 (approved):

Issue #772, opened 2021-12-06
api: add initial prometheus metrics exporter


iameli:
Some basic ones to add, by endpoint:

* HTTP Response codes
* Postgres request count
* HTTP latency
* Postgres latency

PR #809, opened 2021-12-23
Add prometheus metrics aggregator for APIs.

hjpotter92:

Adds minimal pprometheus metrics reporting for the APIs.

- adds a prometheus client dependency
- middleware to track response latency
- hook into db class and record pgsql call timings and counts

booted up the service, and tested against `/metrics` route. The results look similar to:

<details>

```
# HELP pgsql_query_duration_seconds duration histogram of pgsql queries
# TYPE pgsql_query_duration_seconds histogram
pgsql_query_duration_seconds_bucket{le="0.003"} 52
pgsql_query_duration_seconds_bucket{le="0.03"} 127
pgsql_query_duration_seconds_bucket{le="0.1"} 127
pgsql_query_duration_seconds_bucket{le="0.3"} 127
pgsql_query_duration_seconds_bucket{le="1.5"} 127
pgsql_query_duration_seconds_bucket{le="10"} 127
pgsql_query_duration_seconds_bucket{le="+Inf"} 127
pgsql_query_duration_seconds_sum 0.7180695500000001
pgsql_query_duration_seconds_count 127

# HELP http_request_duration_seconds duration histogram of http responses labeled with: status_code, method, path
# TYPE http_request_duration_seconds histogram
http_request_duration_seconds_bucket{le="0.003",status_code="304",method="GET",path="/api/stream"} 2
http_request_duration_seconds_bucket{le="0.03",status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_bucket{le="0.1",status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_bucket{le="0.3",status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_bucket{le="1.5",status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_bucket{le="10",status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_bucket{le="+Inf",status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_sum{status_code="304",method="GET",path="/api/stream"} 0.22272602700000008
http_request_duration_seconds_count{status_code="304",method="GET",path="/api/stream"} 25
http_request_duration_seconds_bucket{le="0.003",status_code="200",method="GET",path="/api/user/#val"} 0
http_request_duration_seconds_bucket{le="0.03",status_code="200",method="GET",path="/api/user/#val"} 1
http_request_duration_seconds_bucket{le="0.1",status_code="200",method="GET",path="/api/user/#val"} 1
http_request_duration_seconds_bucket{le="0.3",status_code="200",method="GET",path="/api/user/#val"} 1
http_request_duration_seconds_bucket{le="1.5",status_code="200",method="GET",path="/api/user/#val"} 1
http_request_duration_seconds_bucket{le="10",status_code="200",method="GET",path="/api/user/#val"} 1
http_request_duration_seconds_bucket{le="+Inf",status_code="200",method="GET",path="/api/user/#val"} 1
http_request_duration_seconds_sum{status_code="200",method="GET",path="/api/user/#val"} 0.006025716
http_request_duration_seconds_count{status_code="200",method="GET",path="/api/user/#val"} 1
```

</details>

Related to #772 

**Checklist:**


packages/api/src/store/db.ts
```diff
@@ -103,6 +106,13 @@ export class DB {
       console.log("no replica url found, not using read replica");
     }
 
+    this.metricHistogram = new Histogram({
+      name: "pgsql_query_duration_seconds",
```
victorges: Can we namespace the metric with `livepeer_api`? Following the pattern from the `livepeer-data` metrics (not merged yet tho, still on review), we could set up a `namespace: "livepeer"` and a `subsystem: "api"`, WDYT? 

If the `prom-client` node lib does not have the concepts of namespace and subsystem, you can just prefix the metric name with `livepeer_api_` instead, so the fully qualified name is 
```suggestion
      name: "livepeer_api_pgsql_query_duration_sec",
```
(also abbreviated `seconds`, just a nit feel free to keep it full)

packages/api/src/store/db.ts
```diff
@@ -195,17 +205,19 @@ export class DB {
...
     if (typeof query === "string") {
       queryLog = JSON.stringify({ query: query.trim(), values });
     } else {
       queryLog = JSON.stringify(query);
     }
-    let result;
+    let result: QueryResult;
     logger.info(`runQuery phase=start query=${queryLog}`);
     const start = Date.now();
+    const queryTimer = this.metricHistogram.startTimer();
```
victorges: I think the most useful part of this metric would be to have it broke down by the actual query being executed. 

In the ideal version, we would specify the table + the function that is being executed, for example `StreamTable#getByStreamKey` or something like that. There **might** be a way to grab the "caller" information automatically here, but I'm not so sure myself how to do it. On C# there is an attribute called [`[CallerMemberName]`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-6.0) which was filled by the compiler and could be used for that, but I don't think there's such a thing in TypeScript.

So I see a couple of options now, which we can discuss further on the most appropriate (or come up/decide for some other):
 1. Stick to trying to determine the caller automatically somehow. One idea is to use the stack trace for that, but the implementation could get quite messy and hard to maintain, especially with all the `async` stuff going on and the common `Table` base class that we have for all tables (stack trace might show that instead of the "real" table?). May also get less useful when different queries can be made by the same function, like for example from [`filter`](https://github.com/livepeer/livepeer-com/blob/049c0fa1442db71201f90ea5eed49e627e69973c/packages/api/src/store/table.ts#L114) function, which accepts a list of filters and builds a dynamic query from them, so they can vary in performance **a lot** (we just had one set of filters there [which wasn't using any index at all](https://github.com/livepeer/livepeer-com/pull/793)).
 2. Set a label on the metric with the full `query` but without the `values` (what to do when `typeof query !== "string"`?). I think that will be the easiest to implement, but might have issues with the length of the prometheus labels? Not sure, but maybe it can take some labels with 100s of characters with no issues. If you could make sure that is the case, I'd be down with just doing this then. Just gotta figure out when does that `typeof query !== "string"` happen and what do do about it.
 3. Have some new parameter(s) here like `queryName` or `queryId` (or separate `table` and `queryName`), which is passed explicitly by every caller. This seems to be the most bulletproof, so I'm growing more fond of it, it might just be a little cumbersome to implement initially. With it, we could also circumvent all issues mentioned above, by sometimes building the `queryId` dynamically in the case of the dynamic `find` query for example (tho on those cases it might as well grow very similar to just labeling with the whole `query`).

packages/api/src/store/model.js
```diff
@@ -188,7 +188,7 @@ export default class Model {
     return responses;
   }
 
-  close() {
+  async close() {
     this.db.close();
```
victorges: Might want to add an `await` or `return` here?

victorges (commented):
Looking food!

While at it, would you mind already adding some metrics to the segment API [[here]](https://github.com/livepeer/livepeer-com/blob/1c9390d3490f6f7473d770d967c672599b3e7d51/packages/api/src/controllers/helpers.js#L187)? I'm doing some work for improving our integration to it [(PR)](https://github.com/livepeer/livepeer-com/pull/805) and it would be really helpful to know if my guess there was correct (that it's costing **seconds** on API calls from outside the US).

packages/api/src/store/db.ts
```diff
@@ -66,6 +67,8 @@ export class DB {
   pool: Pool;
   replicaPool: Pool;
 
+  metricHistogram: Histogram;
```
victorges: I think build is also failing here
```
src/store/db.ts:70:20 - error TS2314: Generic type 'Histogram<T>' requires 1 type argument(s).
--
14:23:07.387 |  
14:23:07.388 | 70   metricHistogram: Histogram;
14:23:07.388 | ~~~~~~~~~
14:23:07.388 |  
14:23:07.389 |  
14:23:07.389 | Found 1 error.
```

hjpotter92: fixing

packages/api/src/store/model.js
hjpotter92: yes, added return statement.

packages/api/src/store/db.ts
hjpotter92: I was thinking for splitting the `queryLog` string and selecting the first element (to identify CRUD operation), but for fetching the table information, either the function signature would need change or we introduce a sql parser (not recommended imo).

setting a label with the full query would split the metric across multiple dimensions and might make it more difficult to aggregate results (although, i'm unfamiliar with prometheus in this regard).

I'm also in favour of providing the table name as a parameter. Labels can be dynamically updated for the histogram once the `startTimer()` call has been issued. The implementation might pose a bigger challenge as you stated, since all `.query` and `.replicaQuery` calls would need to be updated (or we introduce overloaded signatures for these functions?).

packages/api/src/store/db.ts
hjpotter92: It doesn't have a concept for namespace or prefixes, and when initialising a metric type; you specify the full metric `name` value (as done here).

I'll update the name prefix to reflect the service name (same would be applicable for http response duration metric I guess).

For the suffix `sec`, [prometheus recommends][1] having full, pluralised unit in the metric name. If we do have a consensus for naming, it'd be quicker to follow; or we can start writing/discussing one for later usage.

  [1]: https://prometheus.io/docs/practices/naming/

packages/api/src/controllers/helpers.js
```diff
@@ -185,18 +193,22 @@ export async function trackAction(userId, email, event, apiKey) {
...
     Authorization: "Basic " + Buffer.from(`${apiKey}:`).toString("base64"),
   };
 
-  await fetch(`${segmentApiUrl}/${endpoint}`, {
+  let response = await fetch(`${segmentApiUrl}/${endpoint}`, {
     method: "POST",
     body: JSON.stringify(body),
     headers: headers,
   });
+  labels.status_code = response.status.toString();
```
victorges: Is there a way to send the labels only when we finish the timer, instead of modifying an object that has already been sent as an argument? I find that API kind of weird, encouraging mutable objects, surprising that it works.

packages/api/src/store/db.ts
```diff
@@ -103,6 +106,13 @@ export class DB {
       console.log("no replica url found, not using read replica");
     }
 
+    this.metricHistogram = new Histogram({
+      name: "livepeer_api_pgsql_query_duration_seconds",
+      help: "duration histogram of pgsql queries",
+      buckets: [0.003, 0.03, 0.1, 0.3, 1.5, 10],
+      registers: [register],
```
victorges: Why does this need a register and the segment one doesn't? Also don't we need to specify `labelNames` here as well?

packages/api/src/store/db.ts
```diff
@@ -195,36 +205,44 @@ export class DB {
     query: string | QueryConfig<I>,
     values?: I
   ): Promise<QueryResult<T>> {
-    let queryLog;
+    let queryLog: string;
+    const labels: any = {};
```
victorges: **nit:** Perhaps a more accurate type.

```suggestion
    const labels: Record<string, string> = {};
```

We should avoid `any` at all costs since that is just regressing to ~~savagery~~ JavaScript :laughing:

packages/api/src/store/db.ts
```diff
@@ -195,36 +205,44 @@ export class DB {
...
       );
+      queryTimer();
       throw e;
     }
     logger.info(
       `runQuery phase=success elapsed=${Date.now() - start}ms rows=${
         result?.rowCount
       } query=${queryLog}`
     );
+    queryTimer();
```
victorges: WDYT of having this in a `finally` block instead of calling it twice?

packages/api/src/store/db.ts
```diff
@@ -195,36 +205,44 @@ export class DB {
...
       queryLog = JSON.stringify(query);
     }
-    let result;
+    let result: QueryResult;
     logger.info(`runQuery phase=start query=${queryLog}`);
     const start = Date.now();
+    const queryTimer = this.metricHistogram.startTimer(labels);
     try {
       result = await pool.query(query, values);
+      labels.phase = "success";
```
victorges: I think the term phase made sense when we also had a `start` phase. Since we are only splitting here whether the query was a success or error, WDYT of calling it `result` instead?

packages/api/src/store/db.ts
hjpotter92: its the default register. if we're going to have multiple registries, we would need this. Can be removed here. and yes, I forgot to stage the line with `labelNames` field here.

packages/api/src/controllers/helpers.js
hjpotter92: labels can be set anywhere. i was also surprised by mutable object observers, but it felt more js/ts. can be removed from here and be sent as param in the end

packages/api/src/store/db.ts
hjpotter92: my first name chosen was `success = true/false`, but just kept it as `phase` for consistency. otoh, `result` label felt like it should record the count of rows from query result/equivalent and not the `status` (?) of the query itself.

packages/api/src/controllers/helpers.js
victorges: Yeah I think that would be a better pattern IMO. Just a reminder that you changed on the prometheus client but not here.

packages/api/src/store/db.ts
```diff
@@ -60,12 +66,14 @@ export class DB {
...
 
-  postgresUrl: String;
-  replicaUrl: String;
+  postgresUrl: string;
+  replicaUrl: string;
   ready: Promise<void>;
   pool: Pool;
   replicaPool: Pool;
 
+  metricHistogram: Histogram<string>;
```
victorges: **nit:** For the extra typescript checks
```suggestion
  metricHistogram: Histogram<keyof QueryHistogramLabels>;
```

packages/api/src/store/db.ts
```diff
@@ -41,6 +42,11 @@ export type DBSession = WithID<Session> &
 
 type Table<T> = BaseTable<WithID<T>>;
 
+type QueryHistogramLabels = {
+  query: string;
+  result?: string;
```
victorges: This one doesn't even need to be optional anymore since we're initializing the labels with it ✨ 
```suggestion
  result: string;
```

victorges (approved):
LGTM!

PR #811, opened 2021-12-23
api: A couple functional improvements to the `/user` API

### LINKED ISSUE #317: Add proper authentication to methods used in Stripe integration
**What's the problem? (required)** 
Bunch of API methods is not properly authenticated. 
Methods:
`create-customer`, `update-customer-payment-method`, `create-subscription`, `update-subscription`,
`retrieve-subscription`, `retrieve-invoices`

### LINKED ISSUE #544: Investigate why some users were not assigned Stripe accounts
We recently heard from a customer that they had trouble using parts of the dashboard. After an investigation by @adamsoffer we determined that it was because they did not have an associated Stripe account. It was discovered that there were several other users who similarly did not have Stripe accounts. Users who did not have Stripe accounts had created their accounts at various times, including just recently, so this issue was not related to the Stripe migration. 

We'd like to determine the root cause of this issue and determine the steps to fix it.

victorges:
This is to propose some improvements to the `/user` API, specifically regarding
the integration with the Stripe API.

Made a couple changes, but I'd say that the main were:
 - Make the signup/registration API a single-request flow, no need for the client to make 3 separate
 calls. With that, also no risk that users will be created without a proper Stripe customer&subs configured.
 - Add authentication/torization to most of the APIs, since only a few of them need to be called without 
 proper user credentials.
 - Make sure that we never re-created the Stripe customer/subscription for a given user. On this [Google Sheet](https://docs.google.com/spreadsheets/d/13a_Ms3sST6KAL35e5K4ueVt7O5WzbY0hCUEjqEhFSwE/edit#gid=1397264974&fvid=2112356961),
 exported from the Stripe dashboard we can see that there are a couple users with duplicate customers in Stripe,
 so it is something that really happens in production.

## -

 - `yarn test`
 - tested all the user flows work on dashboard ✅ 

No. Tech debt christmas!
Edit: Implements #816  and closes #544 and fixes #317

**Checklist:**

### DIFF
--- a/packages/api/src/app-router.ts
+++ b/packages/api/src/app-router.ts
@@ -105,7 +105,9 @@ export default async function makeApp(params: CliArgs) {
       "Warning: Missing Stripe API key. In development, make sure to configure one in .env.local file."
     );
   }
-  const stripe = new Stripe(stripeSecretKey, { apiVersion: "2020-08-27" });
+  const stripe = stripeSecretKey
+    ? new Stripe(stripeSecretKey, { apiVersion: "2020-08-27" })
+    : null;
   // Logging, JSON parsing, store injection
 
   const app = Router();
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -1,5 +1,5 @@
 import validator from "email-validator";
-import { Router } from "express";
+import { RequestHandler, Router } from "express";
 import jwt from "jsonwebtoken";
 import ms from "ms";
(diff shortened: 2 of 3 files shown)


adamsoffer:
This is great! I think this PR also closes #544.

victorges:
Ohh, also realized this fixes this one too https://github.com/livepeer/livepeer-com/issues/317

hjpotter92 (approved):

PR #835, opened 2022-01-06
www: Prevent recording and multistream changes while stream is live

### LINKED ISSUE #797: Dashboard: Prevent users from turning recording on and off while a stream is active, remove warning for disabling recording
## Current Behavior

Video:
https://drive.google.com/file/d/1L4itstS4Q8KM2C8sqTv0gy38C9SOrbUl/view?usp=drivesdk

Currently, when a stream is active, you can turn recording on and off. When you turn it off, you get an alert describing the consequences of turning recording off. When you turn it on during a stream, you get a notification saying that recording has been turned on. 

## Desired Behavior

Customers cannot change their recording preferences while a session is active in a given stream. Therefore, this should be reflected in the dashboard. 

### When a stream is active...

-   and a user attempts to change the recording preferences on the dashboard, no visual changes are made and a notification pops up:

### When a stream is inactive...

- users are able to freely toggle between recording preferences with a notification only at the bottom of the screen (this is currently in place today). There is no need for a pop-up confirming the change in recording preferences.

FullStackTiger:

###  Dashboard: Prevent users from turning recording on and off while a stream is active, remove warning for disabling recording

- Handling Switch button for Record sessions on the stream details
- Apply to changing multistream targets
- Remove AlertDialog for disabling recording

Fixes #797 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/ErrorRecordDialog/index.tsx
+++ b/packages/www/components/Dashboard/ErrorRecordDialog/index.tsx
@@ -1,0 +1,44 @@
+import {
+  Button,
+  Flex,
+  AlertDialog,
+  AlertDialogTitle,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogCancel,
+  Heading,
+  Text,
+} from "@livepeer.com/design-system";
+
+const ErrorRecordDialog = ({
+  isOpen,
+  onOpenChange,
+}: {
+  isOpen: boolean;
+  onOpenChange: (isOpen: boolean) => void;
+}) => {
+  return (
+    <AlertDialog open={isOpen} onOpenChange={onOpenChange}>
+      <AlertDialogContent css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4" }}>
+        <AlertDialogTitle as={Heading} size="1">
+          Error
+        </AlertDialogTitle>
+        <AlertDialogDescription
+          as={Text}
+          size="3"
+          variant="gray"
+          css={{ mt: "$2", lineHeight: "22px" }}>
+          You cannot change recording preferences while a session is active
+        </AlertDialogDescription>
+
+        <Flex css={{ jc: "flex-end", gap: "$3", mt: "$5" }}>
+          <AlertDialogCancel size="2" as={Button} ghost>
+            Ok
+          </AlertDialogCancel>
+        </Flex>
+      </AlertDialogContent>
+    </AlertDialog>
+  );
+};
+
+export default ErrorRecordDialog;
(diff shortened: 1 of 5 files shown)


packages/www/components/Dashboard/ErrorRecordDialog/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <AlertDialogContent css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4" }}>
+        <AlertDialogTitle as={Heading} size="1">
+          Error
+        </AlertDialogTitle>
+        <AlertDialogDescription
+          as={Text}
+          size="3"
+          variant="gray"
+          css={{ mt: "$2", lineHeight: "22px" }}>
+          You cannot change recording preferences while a session is active
```
victorges: This text needs to be parametrized as well, since we are showing the same thing for when trying to configured multistream and recording.

Can just change to the same thing but replacing `recording` with `multistream` like
```
You cannot change multistream preferences while a session is active
```

Then I guess the component can just be renamed to `ErrorDialog` since it will just receive a generic text to show as the error.

packages/www/components/Dashboard/MultistreamTargetsTable/Toolbox.tsx
```diff
@@ -208,10 +211,14 @@ const Toolbox = ({
         checked={!target?.disabled}
         value={`${!target?.disabled}`}
         onCheckedChange={useCallback(async () => {
-          if (target?.disabled) {
-            await setTargetDisabled(false);
+          if (stream.isActive) {
+            errorRecordDialogState.onOn();
           } else {
-            setDisableDialogOpen(true);
+            if (target?.disabled) {
```
victorges: No need to nest further here, can just have a `else if`

packages/www/components/Dashboard/StreamDetails/Record.tsx
```diff
@@ -1,113 +1,55 @@
...
-  const [open, setOpen] = useState(false);
   const [openSnackbar] = useSnackbar();
+  const errorRecordDialogState = useToggleState();
+
+  const onCheckedChange = async (e) => {
+    e.preventDefault();
+    if (stream.isActive) {
+      errorRecordDialogState.onOn();
+    } else {
+      if (!stream.record) {
```
victorges: Same
```suggestion
    } else if (!stream.record) {
```

packages/www/components/Dashboard/StreamDetails/Record.tsx
```diff
@@ -1,113 +1,55 @@
...
-            e.preventDefault();
-            if (!stream.record) {
-              await patchStream(stream.id, { record: true });
-              await invalidate();
-              openSnackbar("Recording has been turned on.");
-            } else {
-              setOpen(true);
-            }
-          }}>
+        <Box as={DropdownMenuItem} onSelect={(e) => onCheckedChange(e)}>
```
victorges, line 41: Nice, this def deserved a helper

victorges (approved):
LGTM! Just need to fix the error message for multistream.

PR #838, opened 2022-01-07
api: Improve Webhook Retry Logic and Notification

### LINKED ISSUE #818: Improve Webhook Retry Logic and Notification
_Open to tweaks and suggestions_

## Current State

From Victor:

>  - backoffs start at 5s and grow 20% on every error up to a maximum of 10 minutes
> - but with a plot twist, we only retry up to 20 times, so the max delay we actually have is 5*1.2^20=191s~3 minutes, total cumulative retry period of ~15.5min
 > * (all the above is easily tune-able for better retry behavior (just checked those numbers, seems like we can improve them a lot))
> - after those 15.5 minutes of retrying, we don't really send an email. we just log it for now, should be easy to send an email as well (not so easy to avoid spamming the user tho)

Breakdown of the existing (and proposed) exponential backoff schedule can be seen [here.](https://docs.google.com/spreadsheets/d/13yftN2gUV8t1I43SIlSlWdrUarqlw1ZvmPIXZKgusM4/edit#gid=0)

##  Improved State

**1) Notify customers when webhooks fail**
Right now, we don't tell customers when webhooks are failing. I propose that we notify customers at ~3min of retries via email. The email should include:

- Which webhook and URL is failing 
- When the endpoint will be automatically disabled

**2) Improve existing retry policy**
Our current retry policy only lasts for ~15.5 min, which is too short for customers to react to their webhooks failing. I propose that we alter the existing retry policy to  5s initial backoff, doubling backoffs exponentially until a maximum of 1 hr (3600s) and then retrying every hour for ~2 days. Seen here:

gioelecerati:

Changes the webhook retry policy as specified at #818 and add a notification email to the user if max retries are reached

- Modifies the backoff coefficient to 2
- Sets the max retries attempt to 33 (24h)
- Send an email to the user before the webhook is disabled

CI

Fixes #818 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -270,7 +270,7 @@ export default class WebhookCannon {
 
     let signature_header = "";
     if (params.headers[SIGNATURE_HEADER]) {
-      signature_header = `-H  ${SIGNATURE_HEADER}: ${params.headers["Livepeer-Signature"]}`;
+      signature_header = `-H  "${SIGNATURE_HEADER}: ${params.headers["Livepeer-Signature"]}"`;
     }
 
     let payload = params.body;
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -9,14 +9,16 @@ import messages from "../store/messages";
 import Queue from "../store/queue";
 import Model from "../store/model";
 import { DBWebhook } from "../store/webhook-table";
-import { fetchWithTimeout } from "../util";
+import { fetchWithTimeout, RequestInitWithTimeout } from "../util";
 import logger from "../logger";
 import { sign, sendgridEmail } from "../controllers/helpers";
 
 const WEBHOOK_TIMEOUT = 5 * 1000;
-const MAX_BACKOFF = 10 * 60 * 1000 * 6;
+const MAX_BACKOFF = 60 * 60 * 1000;
 const BACKOFF_COEF = 2;
 const MAX_RETRIES = 33;
+
+const SIGNATURE_HEADER = "Livepeer-Signature";
 
 export default class WebhookCannon {
   db: DB;
@@ -221,8 +223,14 @@ export default class WebhookCannon {
       console.log(
         `Webhook Cannon| Max Retries Reached, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
       );
-      this.notifyDisabledWebhook(trigger, webhookPayload, err);
-      return;
+      try {
+        this.notifyFailedWebhook(trigger, webhookPayload, err);
+        return;
+      } catch (err) {
+        console.error(
+          `Webhook Cannon| Error sending notification email to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
+        );
+      }
     }
 
     trigger = {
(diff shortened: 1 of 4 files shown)


gioelecerati:
Needs edits on the email template, as the body text contains the full webhook payload and an example curl to replicate the webhook error, both not properly displayed

packages/api/src/webhooks/cannon.ts
```diff
@@ -4,33 +4,49 @@ import isLocalIP from "is-local-ip";
...
 import { fetchWithTimeout } from "../util";
 import logger from "../logger";
-import { sign } from "../controllers/helpers";
+import { sign, sendgridEmail } from "../controllers/helpers";
 
 const WEBHOOK_TIMEOUT = 5 * 1000;
-const MAX_BACKOFF = 10 * 60 * 1000;
-const BACKOFF_COEF = 1.2;
-const MAX_RETRIES = 20;
+const MAX_BACKOFF = 10 * 60 * 1000 * 6;
```
victorges: **nit:**
```suggestion
const MAX_BACKOFF = 60 * 60 * 1000;
```
or even
```suggestion
const MAX_BACKOFF = 3600_000; // 1h in ms
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +239,69 @@ export default class WebhookCannon {
     );
   }
 
+  async notifyDisabledWebhook(
```
victorges: **nit:** Since we're not disabling the webhook anymore
```suggestion
  async notifyFailedWebhook(
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +239,69 @@ export default class WebhookCannon {
...
+    trigger: messages.WebhookTrigger,
+    params?: any,
```
victorges: The type here should be like this one
https://github.com/livepeer/livepeer-com/blob/a21b96ad54399ce8694e32e013dc2414b6b16bce/packages/api/src/util.ts#L57

(if you're feeling typescripty, you can create and export a new `type` on that `util` file like `RequestInitWithTimeout` or sth like that and use it in all the places we have that `&` 😋 )

Also in the `retry` function that calls this.

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +239,69 @@ export default class WebhookCannon {
...
+    err?: any
+  ) {
+    try {
+      if (trigger.user.emailValid) {
```
victorges: **nit:** You can invert this if condition since you return from the `else`, then you can reduce 1 level of indentation on the whole thing (just like on the if below)

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +239,69 @@ export default class WebhookCannon {
...
+          console.error(
+            `Sending emails requires supportAddr, sendgridTemplateId, and sendgridApiKey`
+          );
+
+          return;
+        }
+
+        let signature_header = "";
+        if (trigger.webhook.sharedSecret) {
+          signature_header = "-H " + params.headers["Livepeer-Signature"];
```
victorges: Nice! I think you're missing the header name before the value tho, can use a template string as well
```suggestion
        if (params.headers["Livepeer-Signature"]) {
          signature_header = `-H  Livepeer-Signature: ${params.headers["Livepeer-Signature"]}`;
```

Also, we could have a constant for that `"Livepeer-Signature"` not to repeat it everywhere.

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +239,69 @@ export default class WebhookCannon {
...
+      } else {
+        console.error(
+          `Webhook Cannon| User email is not valid, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
+        );
+        return;
+      }
+      console.log(
+        `Webhook Cannon| Email sent to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
+      );
+    } catch (err) {
```
victorges: WDYT of moving the try/catch to the calling function just so we don't have the huge block here? 🥴

victorges (commented):
looking good!

packages/api/src/webhooks/cannon.ts
gioelecerati: My bad! Thanks for noticing

packages/api/src/webhooks/cannon.ts
```diff
@@ -167,7 +185,7 @@ export default class WebhookCannon {
       await this._fireHook(trigger, false);
     } catch (err) {
       console.log("_fireHook error", err);
-      await this.retry(trigger);
+      await this.retry(trigger, err);
```
victorges: I think you're sending `err` for a `webhookPayload` param. Shouldn't this be?
```suggestion
      await this.retry(trigger, null, err);
```

This should also be detected by TypeScript automatically when you set the right types to the inputs of the `retry` function. I'll make another suggestion there.

packages/api/src/webhooks/cannon.ts
```diff
@@ -200,12 +218,19 @@ export default class WebhookCannon {
     return newInterval | 0;
   };
 
-  retry(trigger: messages.WebhookTrigger) {
+  retry(trigger: messages.WebhookTrigger, webhookPayload?: any, err?: any) {
```
victorges:
```suggestion
  retry(trigger: messages.WebhookTrigger, webhookPayload?: RequestInitWithTimeout, err?: Error) {
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -200,12 +218,19 @@ export default class WebhookCannon {
...
       );
-      return;
+      try {
+        this.notifyFailedWebhook(trigger, webhookPayload, err);
+        return;
+      } catch (err) {
+        console.error(
+          `Webhook Cannon| Error sending notification email to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
+        );
+      }
```
victorges: The return should be after the `catch` block otherwise when an exception is thrown sending the email we are going to retry the webhook past the maximum limit we want to allow. 

```suggestion
      try {
        this.notifyFailedWebhook(trigger, webhookPayload, err);
      } catch (err) {
        console.error(
          `Webhook Cannon| Error sending notification email to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
        );
      }
      return;
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +247,60 @@ export default class WebhookCannon {
...
+      subject: "Your webhook has been disabled",
+      preheader: "Failure notification",
+      buttonText: "Manage your webhooks",
+      buttonUrl: `https://${this.frontendDomain}/dashboard/developers/webhooks`,
+      unsubscribe: `https://${this.frontendDomain}/contact`,
+      text: [
+        `Your webhook ${trigger.webhook.url} failed to receive our payload in the last 24 hours`,
+        //`<code>${payload}</code>`,
+        //`This is the error we are receiving:`,
+        //`${err}`,
```
victorges: We should uncomment at least this part with the error message, even in the simpler email.

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +247,60 @@ export default class WebhookCannon {
...
+      sendgridTemplateId: this.sendgridTemplateId,
+      sendgridApiKey: this.sendgridApiKey,
+      subject: "Your webhook has been disabled",
+      preheader: "Failure notification",
+      buttonText: "Manage your webhooks",
+      buttonUrl: `https://${this.frontendDomain}/dashboard/developers/webhooks`,
+      unsubscribe: `https://${this.frontendDomain}/contact`,
+      text: [
+        `Your webhook ${trigger.webhook.url} failed to receive our payload in the last 24 hours`,
+        //`<code>${payload}</code>`,
```
victorges, line 294: Mind adding a `// TODO:` here to change the template, uncommenting the lines including the payload and the example `curl`? e.g.
```suggestion
        // TODO: Uncomment the additional information here once we get access to Sendgrid to change the tempalte
        //`<code>${payload}</code>`,
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +247,60 @@ export default class WebhookCannon {
...
+      );
+      console.error(
+        `Sending emails requires supportAddr, sendgridTemplateId, and sendgridApiKey`
+      );
+      return;
+    }
+
+    let signature_header = "";
+    if (params.headers[SIGNATURE_HEADER]) {
+      signature_header = `-H  "${SIGNATURE_HEADER}: ${params.headers["Livepeer-Signature"]}"`;
```
victorges: **nit:** Just for naming consistency, local variables showed be in camelCase
```suggestion
    let signatureHeader = "";
    if (params.headers[SIGNATURE_HEADER]) {
      signatureHeader = `-H  "${SIGNATURE_HEADER}: ${params.headers["Livepeer-Signature"]}"`;
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +247,60 @@ export default class WebhookCannon {
...
+      buttonUrl: `https://${this.frontendDomain}/dashboard/developers/webhooks`,
+      unsubscribe: `https://${this.frontendDomain}/contact`,
+      text: [
+        `Your webhook ${trigger.webhook.url} failed to receive our payload in the last 24 hours`,
+        //`<code>${payload}</code>`,
+        //`This is the error we are receiving:`,
+        //`${err}`,
+        //`We disabled your webhook, please check your configuration and try again.`,
+        //`If you want to try yourself the call we are making, here is a curl command for that:`,
+        //`<code>curl -X POST -H "Content-Type: application/json" -H "user-agent: livepeer.com" ${signature_header} -d '${payload}' ${trigger.webhook.url}</code>`,
```
victorges: followup from other comment
```suggestion
        //`<code>curl -X POST -H "Content-Type: application/json" -H "user-agent: livepeer.com" ${signatureHeader} -d '${payload}' ${trigger.webhook.url}</code>`,
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -305,7 +384,7 @@ export default class WebhookCannon {
         }
 
         if (resp.status >= 500) {
-          await this.retry(trigger);
+          await this.retry(trigger, params, "Status code: " + resp.status);
```
victorges: Just for type consistency, send an `Error` here instead of a `string`:
```suggestion
          await this.retry(trigger, params, new Error("Status code: " + resp.status));
```

victorges (approved):
LGTM! Just some small fixes

packages/api/src/webhooks/cannon.ts
gioelecerati: Nice catch!

PR #849, opened 2022-01-11
api/queue: Create webhook queues as quorum type

victorges:

This is to guarantee some higher availability for the webhook queues by making them of quorum type instead.

 - Create new queues, appending a `_v1` to their previous names
 - Create logic to unbind the old queues from the exchanges and delete them once they are empty+unused

## -

 - Ran api locally and ensure it creates the new queues and delete the old ones.
 - Ran it again to ensure that neither the `unbind` nor the `delete` explode when the queue is not 
 - `yarn test`

Implements #823 

**Checklist:**

- [x] I have run tests to cover my changes.

### DIFF
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -81,11 +81,9 @@ export class RabbitQueue implements Queue {
         await Promise.all([
           channel.assertQueue(QUEUES.events, {
             arguments: { "x-queue-type": "quorum" },
-            durable: true,
           }),
           channel.assertQueue(QUEUES.webhooks, {
             arguments: { "x-queue-type": "quorum" },
-            durable: true,
           }),
           channel.assertExchange(EXCHANGES.webhooks, "topic", {
             durable: true,
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -9,8 +9,10 @@ const EXCHANGES = {
   delayed: "webhook_delayed_exchange",
 } as const;
 const QUEUES = {
-  events: "webhook_events_queue",
-  webhooks: "webhook_cannon_single_url",
+  events: "webhook_events_queue_v1",
+  webhooks: "webhook_cannon_single_url_v1",
+  events_old: "webhook_events_queue",
+  webhooks_old: "webhook_cannon_single_url",
   delayed: "webhook_delayed_queue",
 } as const;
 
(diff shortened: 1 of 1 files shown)


packages/api/src/store/queue.ts
```diff
@@ -91,6 +99,8 @@ export class RabbitQueue implements Queue {
           channel.bindQueue(QUEUES.webhooks, EXCHANGES.webhooks, "webhooks.#"),
           channel
             .assertQueue(QUEUES.delayed, {
+              // Quorum queues do not support message expiration, so this has to
```
gioelecerati, line 100: Just a thing I am missing, does this affect also the events and webhooks queues? What happens if the consumers on these two queues fails to consume the messages and they haven't a TTL?

packages/api/src/store/queue.ts
```diff
@@ -77,8 +79,14 @@ export class RabbitQueue implements Queue {
       json: true,
       setup: async (channel: Channel) => {
         await Promise.all([
-          channel.assertQueue(QUEUES.events, { durable: true }),
-          channel.assertQueue(QUEUES.webhooks, { durable: true }),
+          channel.assertQueue(QUEUES.events, {
+            arguments: { "x-queue-type": "quorum" },
+            durable: true,
```
gioelecerati: Little thing: Do these quorum queues need to be specified as durable? As far as I can understand quorum queues are durable by default and do not support non-durability

gioelecerati (approved):
LGTM, just a couple of comments for clarification

packages/api/src/store/queue.ts
victorges: This means that we won't be able to send messages with an expiration to the events/webhooks queues. Which at least currently is the case, since we don't want messages to disappear directly within RabbitMQ, but want to explicitly handle the messages that we can't deliver instead, after the specified number of retries and emailing the user, in the case of webhooks.

If we ever want to have messages that have some expiration and thus are not worth processing after some period, we could either set that expiration inside the payload and just immediately ignore/ack those messages when processing, or create a separate queue which is not of `quorum` type to be able to set an `expiration` for them within RabbitMQ.

In the case of the delayed messages, the expiration is actually only the delay that the message spends in the delayed queue before being dead-letter published to the regular queues. You can check this comment for a slightly longer explanation: https://github.com/livepeer/livepeer-com/blob/d2c7b91ca7dfb5fb9eeea201638066d53843018c/packages/api/src/store/queue.ts#L172

packages/api/src/store/queue.ts
victorges: You are right, quorum queues cannot be non-durable. I ended up keeping that here to have it set explicitly here, but if it might create confusion I'll just remove it, which makes sense and is cleaner.

PR #850, opened 2022-01-11
api: Call Sendgrid email validation API on signup

victorges:

This is to test Sendgrid email validation API to see if it would be interesting for us.
Runs totally in background and only logs the result, so not to pollute our DB with any
data nor affect the API performance. With loki we should be able to do some smart 
queries to know how well it is performing.

Let's keep this testing new signups and check how accurate it would be to detect
malicious users like soccer streamers.

The idea is that we would eventually **require** email verification only when the repsonse
from this API is something like `Risky` or `Invalid`.

Add a function that calls sengrid verification API and logs the result.

## -

Will test on staging where we deploy the API with a sendgrid key. 
Might need to change the key to have access to email verification API though.

Not really. But works on improving the email verification flow which now is totally open.

**Checklist:**


packages/api/src/controllers/helpers.js
```diff
@@ -163,6 +164,32 @@ export async function sendgridEmail({
...
 
+export function sendgridValidateEmail(email, sendgridApiKey) {
+  SendgridClient.setApiKey(sendgridApiKey);
+
+  const request = {
+    url: `/v3/validations/email`,
+    method: "POST",
+    body: { email, source: "signup" },
+  };
+  client
```
hjpotter92: typo here? should it be `SendgridClient`?

victorges: Ugh yes, thanks for spotting that. I hate JS, doesn't warn even on a typo 🙄

victorges: Couldn't hold myself. Converted these helpers to TypeScript 😡

hjpotter92: :rofl: I was going to suggest maybe make it a module? `helpers/sendgrid.ts` etc?

victorges: I think we can do that if we end up keeping this email validation logic. Otherwise, it might be just 1 function in there for sending emails.

hjpotter92 (approved):

PR #852, opened 2022-01-13
add pkg capabilities

iameli:
This adds static binaries of `api` and `www` via https://github.com/vercel/pkg. These will go into livepeer-in-a-box. Someday we'll have a Golang API server that will embed a statically-exported WWW, but this will get us to v1 of the box expediently.

Remaining work:
* On semver releases, create a new GitHub Release and upload all the binaries from `packages/api/bin` and `packages/www/bin` into there. That's where the box will pull them down from.
* I think we're going to need to do some kind of Mac code signing or the Mac binaries won't be usable? Testing after I do the first thing.
* `api` is ~100MB, which is big but okay. `www` is currently ~500MB, which is absurd. It's happening because the entrypoint for `www` does a `require("next")`, which then proceeds to package up Babel and every other piece of build-time nonsense we don't need.


.github/workflows/pkg.yml
```diff
@@ -0,0 +1,121 @@
...
+          yarn run lerna-run pkg
+
+      - name: move pkgs
+        run: |
+          mkdir -p ./pkg-releases
+          for package in api www; do
+            for arch in linux-arm64 linux-x64 macos-arm64 macos-x64 win-x64; do
+              inName="$package-$arch"
+              outName="livepeer-$package"
+              archiveName="livepeer-$(echo $inName | sed s/x64/amd64/ | sed s/win/windows/ | sed s/macos/darwin/).tar.gz"
```
hjpotter92: can be shortened to:

    archiveName="livepeer-$(echo $inName | sed -e 's/x64/amd64/;s/win/windows/;s/macos/darwin/').tar.gz"

iameli: Nice! Applied

iameli:
D'oh, `lerna version` pushes to `origin` even if your local `master` is elsewhere. TIL.

PR #853, opened 2022-01-13
api: add status to Webhooks

gioelecerati:

Extends [#838](https://github.com/livepeer/livepeer-com/pull/838) to add a status object on the api webhooks.
Now the lastTriggerTime and the lastFailure timestamps are added on _fireHook, making possible a notification of unhealthy or healthy state on the dashboard in a future step, as discussed [here](https://github.com/livepeer/livepeer-com/issues/818#issuecomment-1007765675).

* Added webhook status schema  
* Store failure and trigger time on webhook status  
  

CI

Extends [#838](https://github.com/livepeer/livepeer-com/pull/838) on issue [#818](https://github.com/livepeer/livepeer-com/compare/gioele/webhooks-lastfailure)

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -170,37 +170,31 @@ components:
         status:
           type: object
           description: status of webhook
-          default: null
           readOnly: true
           properties:
             lastFailure:
               type: object
               description: failure timestamp and error message with status code
-              default: null
               readOnly: true
               properties:
                 timestamp:
                   type: number
                   readOnly: true
-                  default: null
                   example: 1587667174725
                   description:
                     Timestamp (in milliseconds) at which the webhook last failed
                 error:
                   type: string
                   readOnly: true
-                  default: null
                   description: Webhook failure error message
                   example: "Error message"
                 statusCode:
                   type: number
                   readOnly: true
-                  default: null
                   description: Webhook failure status code
                   example: 500
             lastTriggerTime:
               type: number
-              default: null
               readyOnly: true
               description:
                 Timestamp (in milliseconds) at which the webhook last was
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -193,7 +193,7 @@ components:
                   readOnly: true
                   description: Webhook failure status code
                   example: 500
-            lastTriggerTime:
+            lastTriggeredAt:
               type: number
               readyOnly: true
               description:
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -188,6 +188,11 @@ components:
                   readOnly: true
                   description: Webhook failure error message
                   example: "Error message"
+                response:
+                  type: string
+                  readOnly: true
+                  description: Webhook failure response
+                  example: "Response body"
                 statusCode:
                   type: number
                   readOnly: true
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -167,6 +167,45 @@ components:
           type: string
           writeOnly: true
           description: shared secret used to sign the webhook payload
+        status:
+          type: object
+          description: status of webhook
+          default: null
+          readOnly: true
+          properties:
+            lastFailure:
+              type: object
+              description: failure timestamp and error message with status code
+              default: null
+              readOnly: true
+              properties:
+                timestamp:
+                  type: number
+                  readOnly: true
+                  default: null
+                  example: 1587667174725
+                  description:
+                    Timestamp (in milliseconds) at which the webhook last failed
+                error:
+                  type: string
+                  readOnly: true
+                  default: null
+                  description: Webhook failure error message
+                  example: "Error message"
+                statusCode:
+                  type: number
+                  readOnly: true
+                  default: null
+                  description: Webhook failure status code
+                  example: 500
+            lastTriggerTime:
+              type: number
+              default: null
+              readyOnly: true
+              description:
+                Timestamp (in milliseconds) at which the webhook last was
+                triggered
+              example: 1587667174725
 
     webhook-response:
       type: object
--- a/packages/api/src/store/webhook-table.ts
+++ b/packages/api/src/store/webhook-table.ts
@@ -1,6 +1,7 @@
 import sql from "sql-template-strings";
 
 import { Webhook } from "../schema/types";
+import { NotFoundError } from "./errors";
 import Table from "./table";
 
 export type EventKey = Webhook["events"][0];
@@ -37,5 +38,19 @@ export default class WebhookTable extends Table<DBWebhook> {
       cursor,
     });
     return { data: webhooks, cursor: nextCursor };
+  }
+
+  async updateStatus(id: string, status: DBWebhook["status"]) {
+    const res = await this.db.query(
+      `UPDATE ${
+        this.name
+      } SET data = jsonb_set(data, '{status}', case when data->'status' is null then '{}' else data->'status' end || '${JSON.stringify(
+        status.status
+      )}') WHERE id = '${id}'`
+    );
+
+    if (res.rowCount < 1) {
+      throw new NotFoundError(`couldn't find ${this.name} id=${id}`);
+    }
   }
 }
--- a/packages/api/src/store/webhook-table.ts
+++ b/packages/api/src/store/webhook-table.ts
@@ -45,7 +45,7 @@ export default class WebhookTable extends Table<DBWebhook> {
       `UPDATE ${
         this.name
       } SET data = jsonb_set(data, '{status}', case when data->'status' is null then '{}' else data->'status' end || '${JSON.stringify(
-        status.status
+        status
       )}') WHERE id = '${id}'`
     );
 
--- a/packages/api/src/util.ts
+++ b/packages/api/src/util.ts
@@ -1,6 +1,4 @@
 import fetch, { RequestInit, Response } from "node-fetch";
-
-// extend requestinit with timeout parameter
 export interface RequestInitWithTimeout extends RequestInit {
   timeout?: number;
 }
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -387,8 +387,7 @@ export default class WebhookCannon {
         logger.info(`webhook ${webhook.id} firing`);
         const startTime = process.hrtime();
         resp = await fetchWithTimeout(webhook.url, params);
-        // getting the response text here to avoid the overhead of reading the body twice
-        let responseBody = await resp.text();
+        const responseBody = await resp.text();
         await this.storeResponse(webhook, event, resp, startTime, responseBody);
         statusCode = resp.status;
 
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -303,7 +303,7 @@ export default class WebhookCannon {
     });
 
     console.log(
-      `Webhook Cannon| Email sent to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
+      `Webhook Cannon| Email sent to user ${trigger.user.email}, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
     );
   }
 
@@ -379,13 +379,13 @@ export default class WebhookCannon {
       }
       const triggerTime = Date.now();
       let resp: Response;
-      let errorMessage;
-      let statusCode;
+      let errorMessage: string;
+      let statusCode: number;
       try {
         logger.info(`webhook ${webhook.id} firing`);
         const startTime = process.hrtime();
         resp = await fetchWithTimeout(webhook.url, params);
-        const responseBody = await resp.text();
+        const responseBody = (errorMessage = await resp.text());
         await this.storeResponse(webhook, event, resp, startTime, responseBody);
         statusCode = resp.status;
 
@@ -394,7 +394,6 @@ export default class WebhookCannon {
           logger.info(`webhook ${webhook.id} fired successfully`);
           return true;
         }
-        errorMessage = responseBody;
         if (resp.status >= 500) {
           await this.retry(
             trigger,
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -274,7 +274,7 @@ export default class WebhookCannon {
 
     let signatureHeader = "";
     if (params.headers[SIGNATURE_HEADER]) {
-      signatureHeader = `-H  "${SIGNATURE_HEADER}: ${params.headers["Livepeer-Signature"]}"`;
+      signatureHeader = `-H  "${SIGNATURE_HEADER}: ${params.headers[SIGNATURE_HEADER]}"`;
     }
 
     let payload = params.body;
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -442,7 +442,7 @@ export default class WebhookCannon {
   async storeTriggerTime(webhook: DBWebhook) {
     await this.db.webhook.update(webhook.id, {
       status: {
-        lastTriggerTime: Date.now(),
+        lastTriggeredAt: Date.now(),
         lastFailure: webhook.status?.lastFailure,
       },
     });
@@ -455,7 +455,7 @@ export default class WebhookCannon {
   ) {
     await this.db.webhook.update(webhook.id, {
       status: {
-        lastTriggerTime: webhook.status?.lastTriggerTime,
+        lastTriggeredAt: webhook.status?.lastTriggeredAt,
         lastFailure: {
           timestamp: Date.now(),
           statusCode: statusCode,
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -378,15 +378,15 @@ export default class WebhookCannon {
         params.headers[SIGNATURE_HEADER] = `t=${timestamp},v1=${signature}`;
       }
       const triggerTime = Date.now();
+      const startTime = process.hrtime();
       let resp: Response;
       let errorMessage: string;
+      let responseBody: string;
       let statusCode: number;
       try {
         logger.info(`webhook ${webhook.id} firing`);
-        const startTime = process.hrtime();
         resp = await fetchWithTimeout(webhook.url, params);
-        const responseBody = (errorMessage = await resp.text());
-        await this.storeResponse(webhook, event, resp, startTime, responseBody);
+        responseBody = errorMessage = await resp.text();
         statusCode = resp.status;
 
         if (resp.status >= 200 && resp.status < 300) {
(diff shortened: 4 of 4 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -167,6 +167,45 @@ components:
           type: string
           writeOnly: true
           description: shared secret used to sign the webhook payload
+        status:
+          type: object
+          description: status of webhook
+          default: null
```
victorges: I think we should not have this default so we don't need to initialize the field to `null` when we create a new webhook. It should be handled basically the same way by someone using the API anyway, as checking whether the field is present or not null should be almost the same on every language, and we don't create the expectation that the field will be always there.

Edit: Just saw that there are a bunch of other `default: null`s below as well. I don't think we should have any of those, we should never create the fields just to store them as `null`, and since they are not required fields anyway it is implicit that the field might not be present (which IMO is cleaner than `null`, especially if it is a non-nullable like a `number` (in other languages)).

packages/api/src/schema/schema.yaml
```diff
@@ -167,6 +167,45 @@ components:
...
+                  default: null
+                  example: 1587667174725
+                  description:
+                    Timestamp (in milliseconds) at which the webhook last failed
+                error:
+                  type: string
+                  readOnly: true
+                  default: null
+                  description: Webhook failure error message
+                  example: "Error message"
```
victorges: WDYT of naming this `response` or `payload` instead just to be more explicit where the error comes from?

Edit: Hmm I see, sometimes this might be an actual exception-like error, right? WDYT of even having 2 fields instead? One is `error` the other is `response`, and we fill one or the other. If an "exception" occurred, there will be no `statusCode` nor `response`, otherwise there will be no `error`

packages/api/src/schema/schema.yaml
```diff
@@ -167,6 +167,45 @@ components:
...
+                statusCode:
+                  type: number
+                  readOnly: true
+                  default: null
+                  description: Webhook failure status code
+                  example: 500
+            lastTriggerTime:
```
victorges: Hmm, I didn't remember whether we hat a naming pattern for timestamps when I originally suggested this one. Tho taking a look at our schema now, WDYT of naming this `lastTriggeredAt` instead to be consistent with our other timestamps in the API objects? Like `createdAt`, `verifiedAt`, `lastStreamedAt`...

packages/api/src/util.ts
```diff
@@ -1,5 +1,10 @@
 import fetch, { RequestInit, Response } from "node-fetch";
 
+// extend requestinit with timeout parameter
```
victorges: I think this comment is unnecessary as the type itself is pretty simple/self-explanatory once you get used to the TypeScript syntax.

packages/api/src/webhooks/cannon.ts
```diff
@@ -200,11 +218,22 @@ export default class WebhookCannon {
     return newInterval | 0;
   };
 
-  retry(trigger: messages.WebhookTrigger) {
+  retry(
+    trigger: messages.WebhookTrigger,
```
victorges: Ops, you might need to do a manual `rebase` of your branch with master, cause the changes of the first PR are showing here. LMK if you need help with it. This command should do magic: `git rebase --onto master <last commit on this PR history that is not from this PR> <this PR branch>`

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +251,62 @@ export default class WebhookCannon {
...
+        //`We disabled your webhook, please check your configuration and try again.`,
+        //`If you want to try yourself the call we are making, here is a curl command for that:`,
+        //`<code>curl -X POST -H "Content-Type: application/json" -H "user-agent: livepeer.com" ${signatureHeader} -d '${payload}' ${trigger.webhook.url}</code>`,
+
+        // TODO: Uncomment the additional information here once we get access to Sendgrid to change the tempalte
+      ].join("\n\n"),
+    });
+
+    console.log(
+      `Webhook Cannon| Email sent to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
```
victorges: Would be helpful to have the user ID here instead/additionally of the trigger and stream ID. Or rather, the user email which is easier to understand and is also unique as the ID.

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
+        await this.storeTriggerTime(webhook);
+      } catch (e) {
+        console.log(
+          `Unable to store trigger time of webhook ${webhook.id} url: ${webhook.url}`
+        );
       }
 
+      let resp: Response;
+      let errorMessage;
+      let statusCode;
```
victorges: **nit:** Should add types to these
```
      let errorMessage: string;
      let statusCode: number;
```

Edit: Even better, instead of declaring individual fields, WDYT of declaring the "last failure" field itself?
```suggestion
      let lastFailure: Webhook["status"]["lastFailure"] = { timestamp: Date.now() };
```
(yes, indexing types works!)

Then when using it you could just set the fields on the object like `lastFailure.statusCode = resp.status` etc

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
+
       try {
         logger.info(`webhook ${webhook.id} firing`);
         const startTime = process.hrtime();
-        let resp = await fetchWithTimeout(webhook.url, params);
-        await this.storeResponse(webhook, event, resp, startTime);
+        resp = await fetchWithTimeout(webhook.url, params);
+        const responseBody = await resp.text();
+        await this.storeResponse(webhook, event, resp, startTime, responseBody);
```
victorges: Hmm, not a regression but I wonder if we should surround this with a separate `try/catch` as well

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
+        statusCode = resp.status;
+
         if (resp.status >= 200 && resp.status < 300) {
           // 2xx requests are cool. all is good
           logger.info(`webhook ${webhook.id} fired successfully`);
           return true;
         }
-
+        errorMessage = responseBody;
```
victorges: Since we already check the `statusCode` below and not the `errorMessage`, we can do this on the same line where we set the `responseBody` here like
```
const responseBody = errorMessage = await resp.text();
```
(if we don't rename it to just `response(Body)`/`payload` directly, in which case we don't even need to declare a new `responseBody` variable here.

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
         // this.retry(event);
-        return;
       } catch (e) {
         console.log("firing error", e);
-        await this.retry(trigger);
+        errorMessage = e.message;
+        await this.retry(trigger, params, e);
+      } finally {
+        if (statusCode >= 300 || !statusCode) {
+          try {
```
victorges: Hmmm, on this case, since the `storeTriggerTime` and `storeWebhookFailure` functions are so small, I think it would be cleaner if the `try/catch` was inside of them an we have less code branches on this main logic here. WDYT? 

This would apply here and around line 380, to move the `try/catch` to inside the helper `store*` functions.

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
         return;
       }
     }
   }
 
+  async storeTriggerTime(webhook: DBWebhook) {
+    await this.db.webhook.update(webhook.id, {
+      status: {
+        lastTriggerTime: Date.now(),
+        lastFailure: webhook.status?.lastFailure,
```
victorges: You don't need to include the same `lastFailure` value, since I think the PostgreSQL `||` operator that we use inside `update` does a **deep** merge of the JSONs ✨ 

Or have you checked that it does not do that? If it doesn't, you should instead make sure that we keep all the same fields from the previous status object like this (js trick):
```suggestion
        ...webhook.status,
        lastTriggerTime: Date.now(),
```
```suggestion
        lastTriggerTime: Date.now(),
        lastFailure: webhook.status?.lastFailure,
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
+  ) {
+    await this.db.webhook.update(webhook.id, {
+      status: {
+        lastTriggerTime: webhook.status?.lastTriggerTime,
+        lastFailure: {
+          timestamp: Date.now(),
+          statusCode: statusCode,
+          errorMessage: errorMessage,
+        },
+      },
```
victorges: Same, either remove the `lastTriggerTime` or use a spread (`...webhook.status`) operator instead.

If you do change to declare the whole `lastFailure` object on the main logic above, this might also get cleaner cause you only need to set the whole `lastFailure` obj 😁

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
       // sign payload if there is a webhook secret
       if (webhook.sharedSecret) {
         let signature = sign(params.body, webhook.sharedSecret);
-        params.headers["Livepeer-Signature"] = `t=${timestamp},v1=${signature}`;
+        params.headers[SIGNATURE_HEADER] = `t=${timestamp},v1=${signature}`;
+      }
+      try {
+        await this.storeTriggerTime(webhook);
```
victorges: Hmm, I think we should make sure to use the same timestamps for both the `lastTriggerTime` and the `lastFailure` fields, otherwise we would never be able to tell whether the last failure was actually the last trigger or not. Maybe declare the field locally here and send it as an argument to both functions.

packages/api/src/webhooks/cannon.ts
```diff
@@ -290,43 +375,102 @@ export default class WebhookCannon {
...
         // we don't retry on non 400 responses. only on timeouts
         // this.retry(event);
-        return;
       } catch (e) {
         console.log("firing error", e);
-        await this.retry(trigger);
+        errorMessage = e.message;
+        await this.retry(trigger, params, e);
+      } finally {
+        if (statusCode >= 300 || !statusCode) {
```
victorges: Hmm, since we execute this in a `finally` anyway, which will always get executed, I think it might be cleaner to update both the `lastTriggerTime` and the `lastFailure` on a single go, WDYT? Then we can have a single function that receives the timestamp (either from before making the request or after), and whether there was any errors and its status code etc, then we always update the status code and maybe update the `lastFailure` depending on there being an error or not. Doesn't seem like it would be a big refactor here in the end.

victorges (commented):
Looking good! What you need now is to update the dashboard/frontend to use these new fields from the API. I think you should start looking around from the [WebhooksTable](https://github.com/livepeer/livepeer-com/blob/0ebf5c4b02a2df9c2fbf3fb2bbcb4e7a2e4bf7eb/packages/www/components/Dashboard/WebhooksTable/index.tsx#L37), where you should add those buttons/interactions that we mentioned. 

Let me know if you need any help setting up the www (frontend project, in this same repo) development environment, or some getting started with React. I'm not an expert myself but hit my head against the wall a few times and I think I relatively get it now 😆 

Also remember to rebase this on top of `master` so it's easier to review with only the new changes, after having merged the base PR (github is not that smart yet). It's a healthy practice to rebase your PR every now and then anyway to avoid too many conflicts with `master` at once.

victorges:
Here is a recording of @adamsoffer giving an overview of the code architecture on the `www` project! 😁 

https://discord.com/channels/423160867534929930/445623681251672064/877630072767533056

packages/api/src/schema/schema.yaml
gioelecerati: Yeah, that was the thing, when the call fails is an error message. Anyway, it makes sense, I make two separate fields!

victorges: Hmmm, not to have the `error` empty when there was in fact an (response) error, WDYT of setting it always to a pre-formatted error message when the webhook returns a status error?

e.g. When some exception happens:
`{ error: err.Message }`
When request fails:
```js
{ 
  error: `Request failed with status code ${res.status}`,
  status: res.status, 
  response: await res.text()
}
```

Just so the schema is a little more "stable", with the error always being present, only status/response that can be omitted.

packages/api/src/webhooks/cannon.ts
```diff
@@ -222,6 +251,62 @@ export default class WebhookCannon {
...
+      );
+      console.error(
+        `Sending emails requires supportAddr, sendgridTemplateId, and sendgridApiKey`
+      );
+      return;
+    }
+
+    let signatureHeader = "";
+    if (params.headers[SIGNATURE_HEADER]) {
+      signatureHeader = `-H  "${SIGNATURE_HEADER}: ${params.headers["Livepeer-Signature"]}"`;
```
hjpotter92: is this used anywhere? also, the reference in assignment can be updated to the variable:

```ts
signatureHeader = `-H  "${SIGNATURE_HEADER}: ${params.headers[SIGNATURE_HEADER]}"`;
```

gioelecerati: Thanks for the catch!

Anyway, this is used in case the webhook is configured with a sharedSecret, so we could send a correct curl example via email to the user to test the call itself. 
However right now the part of the email where we provide the example curl is commented until I make a different template on sendGrid to support HTML, or it would come out as a non-readable blob of text

packages/api/src/webhooks/cannon.ts
```diff
@@ -303,7 +303,7 @@ export default class WebhookCannon {
     });
 
     console.log(
-      `Webhook Cannon| Email sent to user, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
+      `Webhook Cannon| Email sent to user ${trigger.user.email}, id: ${trigger.id}, streamId: ${trigger.stream?.id}`
```
victorges: For proper `logfmt` (https://brandur.org/logfmt) formatting, which is the main format Loki supports, you should actually use the `key="value"` format for the fields (quotes are optional in case your fields have no whitespace). A comma would also be interpreted as part of the value, so rather avoid it.

```suggestion
      `Webhook Cannon| Email sent to user="${trigger.user.email}" id=${trigger.id} streamId=${trigger.stream?.id}`
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -377,18 +377,23 @@ export default class WebhookCannon {
         let signature = sign(params.body, webhook.sharedSecret);
         params.headers[SIGNATURE_HEADER] = `t=${timestamp},v1=${signature}`;
       }
-
+      const triggerTime = Date.now();
+      const startTime = process.hrtime();
+      let resp: Response;
+      let errorMessage: string;
+      let responseBody: string;
+      let statusCode: number;
```
victorges, line 385: I think we can avoid declaring at least the `statusCode` here, since we can grab directly from `resp` in the `finally` block. I had thought the same about `responseBody`, but I think we might want to keep that inside the `try` block just in case, to catch any exceptions.

packages/api/src/webhooks/cannon.ts
```diff
@@ -402,36 +407,84 @@ export default class WebhookCannon {
...
+              timestamp: triggerTime,
+              statusCode: statusCode,
+              errorMessage: errorMessage,
+            },
+          },
+        });
+      } else {
+        await this.db.webhook.updateStatus(webhook.id, {
+          status: {
+            ...webhook.status,
```
victorges: We dont need this anymore 😁 
```suggestion
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -402,36 +407,84 @@ export default class WebhookCannon {
...
+            lastTriggeredAt: triggerTime,
+          },
+        });
```
victorges: **nit:** A more linear implementation calling `updateStatus` only in 1 place, wdyt?
```suggestion
      let status: DBWebhook["status"] = { lastTriggeredAt: triggerTime };
      if (statusCode >= 300 || !statusCode) {
        status = {
          ...status, 
          lastFailure: {
            timestamp: triggerTime,
            statusCode,
            errorMessage,
          }
        };
      }
      await this.db.webhook.updateStatus(webhook.id, status);
```

packages/api/src/store/webhook-table.ts
```diff
@@ -38,4 +39,18 @@ export default class WebhookTable extends Table<DBWebhook> {
     });
     return { data: webhooks, cursor: nextCursor };
   }
+
+  async updateStatus(id: string, status: DBWebhook["status"]) {
+    const res = await this.db.query(
+      `UPDATE ${
+        this.name
+      } SET data = jsonb_set(data, '{status}', case when data->'status' is null then '{}' else data->'status' end || '${JSON.stringify(
+        status.status
```
victorges: Since this function only updates the `status`, I don't think it should receive a full webhook object and then use only the `status` field of it. WDYT of removing the `.status` here and changing the callers to only send the `status` payload, and not an object that contains a `status` field?

I am also not sure how `TypeScript` compiler is allowing this, since `DBWebhook["status"]` object has no `status` field. Isn't it complaining about this line here? I guess on the call-site it does not complain because the `status` schema allows `additionalProperties` (the default is `true`), so it won't complain if you set properties that actually aren't there. 
Ah! And by explaining that I also understood how it is not complaining about anything right now. It is basically allowing you to call `status.status` here because of the additional properties, and for the same reason it is allowing you to send an object that doesn't actually match the `status` schema as the arguments here. To avoid that, WDYT of setting `additionalProperties: false` in the schema so it avoids these issues?

Edit: Also realized that, since typescript is not checking anything, the object you are sending for the `status` is actually not following the schema you defined on `schema.json`. You are setting an `errorMessage` field instead of an `error`, as it is in the schema right now. Just in case, let's add `additionalProperties: false` to all objects there unless we do need to allow additional properties. That means to both the `status` and the `lastFailure` objects.

packages/api/src/webhooks/cannon.ts
```diff
@@ -402,36 +407,84 @@ export default class WebhookCannon {
...
+  ) {
+    try {
+      if (statusCode >= 300 || !statusCode) {
+        await this.db.webhook.updateStatus(webhook.id, {
+          status: {
+            lastTriggeredAt: triggerTime,
+            lastFailure: {
+              timestamp: triggerTime,
+              statusCode: statusCode,
+              errorMessage: errorMessage,
```
victorges: I think we're missing the separate `responseBody` vs `errorMessage` here. Do you think it will be simpler with only a single field? I'm ok with that as well.

victorges (approved):
LGTM as is, just gotta fix a TypeScript/schema issue! All other comments are nits/optional 😁

packages/api/src/store/webhook-table.ts
gioelecerati: Yeah sounds good, I removed the status and send that only in the payload. I had some kind of trouble working on this query as it was nesting the objects and I took this path, but I think now that everything works properly I can send only that. Thanks for pointing it out & also noticing about the errorMessage and the error inconsistency, fixed it

packages/api/src/webhooks/cannon.ts
gioelecerati: I think it's better to have separate fields as well, now there're error and response instead of only errorMessage!

PR #865, opened 2022-01-19
api: Asset and Task resource types

### LINKED ISSUE #843: New API Object: Task
Brainstorming so far

### LINKED ISSUE #861: New API Object: Asset

### LINKED ISSUE #880: Implement upload of files to signed S3 URLs

gioelecerati:

Create the new resource types for Assets and Tasks

Fixes [#861](https://github.com/livepeer/livepeer-com/issues/861)
Fixes #880 
Fixes #843 

**Screenshots (optional):**

**Checklist:**


packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+          example: 2022-01-01T02:20:00Z
+        updatedAt:
+          type: string
+          description: Timestamp (in milliseconds) at which task was updated
+          example: 2022-01-01T02:20:00Z
+        params:
+          type: object
+          description: Parameters of the task
+          properties:
+            import:
```
gioelecerati, line 1099: Should we have the params further broken down into `Import`, `Export` and `Transcode` groups, or the `kind` parameter is enough and the params could be on the same level in the object?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+                      example: ABC123
+            transcode:
+              type: object
+              description: Parameters for the transcode task
+              properties:
+                assetId:
+                  type: string
+                  description: ID of the asset to transcode
+                  example: 123123123123
+                preset:
```
gioelecerati: Should we include transcoding preset adding a Preset resource type in a second step? 

- If yes, should the video and audio transcoding parameters be only in the transcoding preset object and a transcode task takes all the informations from that?
- If no, maybe we should decide which transcoding params are available in a first release

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+          example: https://cdn.livepeer.com/ABC123/playlist.m3u8
+        status:
+          type: string
+          description: Status of the asset
+          example: ready
+        originTaskId:
+          type: string
+          description: Task ID of the task that created the asset
+          example: 1-23-123-123-1123-123
+        relatedTasks:
```
gioelecerati: I don't know if it's better to have the `relatedTasks` here in the Asset object to have a view of the task related to it, or if we can display them just with an `assetId` param in the task object, wdyt?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+                  example: 123123123123
+                preset:
+                  type: string
+                  description: Preset to use for the transcode
+                  example: 1080p
+                container:
+                  type: string
+                  description: Container to use for the transcode
+                  example: mp4
+                audio:
```
gioelecerati: Is a transcoding task with audio specification a muted output or a `acodec copy`? Should we specifiy a `muted` param in the audio parameters?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+          example: filename.mp4
+        duration:
+          type: number
+          description: Duration of the asset in seconds (float)
+          example: 23.8328
+        aspectRatio:
+          type: string
+          description: Aspect ratio of the asset
+          example: 16:9
+        meta:
```
gioelecerati, line 906: I think this needs a `MAX_META_SIZE` in the post validation

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+              example: variant_1080.m3u8
+            width:
+              type: number
+              description: width of the profile
+              example: 1920
+            height:
+              type: number
+              description: height of the profile
+              example: 1080
+            frameRate:
```
iameli: I think we call this `fps` elsewhere

packages/api/src/schema/schema.yaml
hjpotter92: agreed! might need another value to check max size of key-value as well maybe? for eg. total meta size = 1MB, fields inside can be at most 100KB?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+            {
+              "title": "My awesome video",
+              "description": "This is a video of my awesome life",
+              "tags": ["awesome", "life", "video"],
+            }
+        updatedAt:
+          type: string
+          description:
+            Timestamp (in milliseconds) at which the asset was last updated
+          example: 2022-01-01T02:20:00Z
```
hjpotter92: nit: example is not the same as description :sweat_smile:

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+        createdAt:
+          type: string
+          description: Timestamp (in milliseconds) at which asset was created
+          example: 2022-01-01T02:20:00Z
+        format:
+          type: string
+          description: Format of the asset
+          example: mp4
+        hash:
```
hjpotter92, line 930: should it be string or object with key = hash used?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+                example: 123
+
+    task:
+      type: object
+      table: task
+      properties:
+        id:
+          type: string
+          description: Task ID
+          example: 1-23-123-123-1123-123
```
hjpotter92: should be uuid?

packages/api/src/schema/schema.yaml
```diff
@@ -1795,6 +2145,42 @@ paths:
...
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
+    post:
+      description: creates a new vod asset
+      responses:
+        200:
```
hjpotter92: POST should respond with `201 Created` status ^_^

gioelecerati: Thanks for the catch!

packages/api/src/schema/schema.yaml
```diff
@@ -24,6 +24,8 @@ tags:
     description: Changes user admin status
   - name: webhook
     description: Webhooks endpoints
+  - name: vod-asset
```
victorges: I think we're missing the tasks endpoint as well

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
             created
           example: 1587667174725
 
+    vod-asset:
+      type: object
+      table: vod_asset
+      required:
+        - name
```
victorges: Should also have `additionalProperties: false`. Please add to **every object** on the schema unless there is need for it.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+      properties:
+        id:
+          type: string
+          readOnly: true
+          example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+          index: true
+        userId:
```
victorges, line 886: This should also be readonly. Check the other schemas in this files to copy these default fields.

packages/api/src/schema/schema.yaml
victorges: I think there's a JSON schema way to enforce max number of elements in an array. Maybe we can enforce on the keys somehow as well.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+            custom name or title
+          example: filename.mp4
+        duration:
+          type: number
+          description: Duration of the asset in seconds (float)
+          example: 23.8328
+        aspectRatio:
+          type: string
+          description: Aspect ratio of the asset
+          example: 16:9
```
victorges: I wonder if we should keep the video-specific fields under a separate subobject here, WDYT? That so this same object could be later used for other file types, like for example images/thumbnails etc.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+        meta:
+          type: object
+          description: User input metadata associated with the asset
+          example:
+            {
+              "title": "My awesome video",
+              "description": "This is a video of my awesome life",
+              "tags": ["awesome", "life", "video"],
+            }
```
victorges, line 914: Good vibes example 😆

packages/api/src/schema/schema.yaml
victorges: All timestamps should be a `number` representing a millisecond-precision Unix timestamp since the epoch, which is also what that `Date.now()` returns by default in node.js. That's our current format in the other objects and we should keep the same pattern here. Mind copying the format/example/etc from the other objects in the schema here?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+        updatedAt:
+          type: string
+          description:
+            Timestamp (in milliseconds) at which the asset was last updated
+          example: 2022-01-01T02:20:00Z
+        createdAt:
+          type: string
+          description: Timestamp (in milliseconds) at which asset was created
+          example: 2022-01-01T02:20:00Z
+        format:
```
victorges: This might fit in the `videoParams` or `videoSpec` or something like that.

WDYT of also making this an`enum` with only the formats that we support?

packages/api/src/schema/schema.yaml
victorges: I also think we should be explicit about what is this hash. For example `md5hash` or `sha256hash`. WDYT
Then we can add other hashes in the future without breaking the contract.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+          type: string
+          description: Format of the asset
+          example: mp4
+        hash:
+          type: string
+          description: Hash of the asset
+          example: sha256:hu31498hf14398
+        tracks:
```
victorges: Another one for `videoSpec`

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+          type: array
+          description:
+            List of tracks associated with the asset when the format contemplate
```
victorges: **nit:** typo

```suggestion
            List of tracks associated with the asset when the format contemplates
```

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+            them (e.g. mp4)
+          items:
+            type: object
```
victorges, line 934: Need `additionalProperties: false` as well. Could have a `required: [kind]` as well, or including any other fields that are always present.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+                  audio tracks
+                example: 44100
+              bitDepth:
+                type: number
+                description: Bit depth of the track - only for audio tracks
+                example: 16
+              codec:
+                type: string
+                description: Codec of the track
+                example: aac
```
victorges: Is this only for audio as well?

Since the are so different in the specs, should we have separate `audioTracks` and `videoTracks` perhaps? Or would that be counter intuitive considering how other tools and platforms represent this?

victorges (commented):
Some preliminary comments just in case GitHub app loses the local state

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+        profiles:
```
victorges: Now this should be a field under an `hlsSpec` instead!

Possibly `hlsSpec` lives inside `videoSpec` as well, in case it still uses many of the video fields as well.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+          type: object
```
victorges, line 946: Gotta disable additional properties

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+              description: Duration of the profile in seconds
+              example: 23.8238
+            bitrate:
+              type: number
+              description: Bitrate of the profile in bits per second
+              example: 1000000
+        playbackUrl:
+          type: string
+          description: Playback url of the playlist manifest
+          example: https://cdn.livepeer.com/ABC123/playlist.m3u8
```
victorges: We should follow the pattern from streams and only store the playbackId as well. The url should be built and could change over time depending on the infra. From us getting a new domain to even changing the url structure somehow.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +813,354 @@ components:
...
+              example: ABC123
+            duration:
+              type: number
+              description: Duration of the profile in seconds
+              example: 23.8238
+            bitrate:
+              type: number
+              description: Bitrate of the profile in bits per second
+              example: 1000000
+        playbackUrl:
```
victorges: This should be under `hlsSpec` as well (or just `hls` perhaps? Maybe we can drop the `Spec` suffix on these sub-objects)

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,376 @@ components:
...
+              type: number
+              description: Frame rate of the profile
+              example: 30
+            id:
+              type: string
+              description: ID of the profile
+              example: ABC123
+            duration:
+              type: number
+              description: Duration of the profile in seconds
```
victorges: Also `readOnly`. Actually I think the whole asset spec is readonly right? Users won't be able to modify an asset via a PUT or something like that, only through specific tasks which then CRUD the asset 🤔

packages/api/src/schema/schema.yaml
gioelecerati: I saw some of the other schemas have not `readOnly: true` on the `userId` field, only the webhook and the multistream schema have it. Is it intended or it should be added also there?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,376 @@ components:
...
+            fps:
+              type: number
+              description: Frame rate of the profile
+              example: 30
+            id:
+              type: string
+              description: ID of the profile
+              example: ABC123
+            duration:
+              type: number
```
victorges, line 956: Instead of a single example, I think this should have/be an `enum` instead and we list all the possible values here

packages/api/src/schema/schema.yaml
victorges: I think we should have unidirectional links, of children pointing to the parent. So the tasks should have 1 indexed field for the source asset, and if we want to list all related (children) tasks we just query the task table.

Dealing with multiple sources of truth is always a headache, so let's try to keep all of them in a single place.

i.e. IMO we should remove this field

packages/api/src/schema/schema.yaml
gioelecerati: Something like this looks good?
```
"hash": {
       "hash": "9b560b28b85378a5004117539196ab24e21bbd75b0e9eb1a8bc7c5fd80dc5b57",
        "algorithm": "sha256sum"
}
```  
So we can support different types of hash algorithms wdyt?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+          description: List of task IDs associated with the asset
+          items:
+            $ref: "#/components/schemas/task"
+        sourceAssetId:
+          type: string
+          description:
+            ID of the source asset (root) - If missing, this is a root asset
+          example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+        nft:
+          type: array
```
victorges: I think an nft should be just another immutable asset pointing to the file on a decentralized storage, with whatever extra nft parameters saved on an equivalent "nft spec" field (which could be `nftSpec` or just `nft` as I'm realizing now).

If there are multiple NFTs minted from the same asset, there would be multiple separate assets created, each with their own NFT spec (but possibly the same storage?)

victorges (commented):
Finishing the asset schema now

packages/api/src/schema/schema.yaml
gioelecerati: I agree, I even think `aspectRatio` is redundant, as we have the `width` and `height` fields in both the track and profile objects?

packages/api/src/schema/schema.yaml
gioelecerati: I think the `format` should stay on the top level of the asset, as this is pretty specific to the container and other fields depend on the container of the asset.

About the enum sounds great, we actually could make separate types for the formats, as they come with different logic / parameters sometimes, wdyt?

packages/api/src/schema/schema.yaml
gioelecerati: Are you referring to `tracks`? This is related not only to the videoSpec, as tracks is an array that can contain also audio tracks (e.g. MP4 container with 1 video track and multiple audio tracks - audio languages)

packages/api/src/schema/schema.yaml
gioelecerati: Yeah as I mentioned in another comment, a track can be either a audio or a video track. For example, this command:
```
ffprobe -v quiet -print_format json -show_format -show_streams example.mp4
```
would output a similar json, with one or more separate streams (`tracks`), either audio or video. In this case all the params are even the same for both the tracks.  
In an example case, `codec` could be `aac` or `h264` depending on which track it is

packages/api/src/schema/schema.yaml
victorges: Yeah agreed! I think I'm good without that field as well.

packages/api/src/schema/schema.yaml
victorges: Hmm, it should be in all the other `userId` fields yeah! Feel free to add to the existing ones as well, but don't feel obliged. Let's just keep the new ones on the best shape.

packages/api/src/schema/schema.yaml
victorges: But aren't containers a video-specific thing? So they could be inside the params/spec about `video`-kind of assets, but not necessarily every kind of asset should have a `format`. Even if they do tho, like for example an `image` one might still have a format `jpeg`, the enum values will end up being different anyway, so it's not really the same field, it only has the same name. 

What I am thinking overall is to have the same pattern that we discussed for `Task`s input/output parameters, so that there is both a top-level "type" field which specifies what is the type of the asset, and then there is a specific field with the name of that `type`, with the specific parameters that are only relevant for that specific type. And that can be infinitely nested. 

So for example could have a `type: video` then `video.format: mp4` and then additional `mp4` fields under `video.mp4: {}`, like this:
```
id: 123
name: People Are Awesome
objectStoreId: 345
size: 3476384
type: video
video:
  format: mp4
  width: ...
  height: ...
  bitrate: ...
  mp4:
    anyCustomMp4Stuff: ...
```

That way, on the root of the `asset` object, we only hold directly the fields that would be valid for almost any kind of "file" in general. Then we specify what type of file it is, then have a dedicated sub-object for that type, etc.

On our MVP we will only have `video` files. But soon we might already have `hls` as a separate `type` altogether (or just as a separate `format`? Maybe not since it's not even a single file). And later we might want to have other things like `image` perhaps, for thumbnails or even GIFs or something. Maybe eventually `audio` as well for audio-only formats? Or would those be stored as "video"?

WDYT?

packages/api/src/schema/schema.yaml
victorges: I tried to explain better what I mean by that `videoSpec` here: https://github.com/livepeer/livepeer-com/pull/865#discussion_r790007068

I know that `tracks` can also be audio, but at least in this context I understood that they are tracks in a video-file / container. So they could be under the specific "video" segment perhaps (or something else that we decide to call it).

What would happen if/when we supported audio-files for example? Would they re-use all this same fields, with the only difference that it only has audio tracks? So what could be an asset type that could be either a "video"-file or an audio-file? We still have the case for images tho, which definitely don't have `tracks` (right? 😳), and so this field wouldn't make sense for them.

Idk, check my other comment first that I linked here, maybe the idea I'm suggestion does not make that much sense if the things are not that well separated in the end.

packages/api/src/schema/schema.yaml
victorges: Hmm I see. Yeah so let's follow the standard from `ffmpeg` then. I guess it could be useful for "newcomers" to have tracks explicitly separated about what are the audio and video tracks, but maybe it would feel alien for more experienced video-developers 😆

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+      properties:
+        id:
+          type: string
+          readOnly: true
+          description: Task ID
+          example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+        kind:
+          type: string
+          description: Type of the task
+          example: import
```
victorges: I think `kind` already has a legacy/tech debt meaning in some of the other resources, which like just stores what is the API resource of the object. I think it's a reminiscent from when we did not use PostgreSQL with different tables for each API resource, but maybe a single Firestore "table" if all the objects from the API. So for example all `stream` objects have a `kind: "stream"`, all `user` objects have a `kind: "user"` etc.

Soo, to avoid confusion, I think we might want to use `type` here instead of a`kind`.

Also, we should have an `enum` here for the exact types of task that we support. As we add more task types, we append to the `enum` here as well.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+        status:
+          type: string
+          description: Status of the task
+          example: ready
```
victorges: This should/could also be an `enum` instead of only an `example`

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+        createdAt:
+          type: string
+          readOnly: true
+          description: Timestamp (in milliseconds) at which task was created
+          example: 1587667174725
+        updatedAt:
+          type: string
+          description: Timestamp (in milliseconds) at which task was updated
+          example: 1587667174725
```
victorges: These should be `number`s as well

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+              "description": "This is a video of my awesome life",
+              "tags": ["awesome", "life", "video"],
+            }
+        updatedAt:
+          type: string
+          description:
+            Timestamp (in milliseconds) at which the asset was last updated
+          example: 1587667174725
+        createdAt:
+          type: string
```
victorges: Forgot to update the type to `number`.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+            import:
+              type: object
+              additionalProperties: false
+              description: Parameters for the import task
+              properties:
+                url:
+                  type: string
+                  description: URL of the asset to import
+                  example: https://cdn.livepeer.com/ABC123/filename.mp4
+            export:
```
victorges: Same as NFT, let's not create the schemas for these until we actually start implementing them. Otherwise they might end up with extraneous data or fields that we don't use or even notice.

I'd say we should start with only `import`, then right after we will implement `export` and maybe later `transcode`.

packages/api/src/schema/schema.yaml
victorges: Actually, for now let's just remove this field. We will not be handling NFTs directly on the first version (maybe soon, but not on the import/export MMMVP) and we shouldn't have fields in the schema that we don't have the code for. 

When we do implement it, everything will also be clearer and we should know better how to model it on top of the existing asset. (was useful having it here to think about/raise discussion tho)

packages/api/src/schema/schema.yaml
victorges: I think the params should be broken down between each `type`, no matter if we have the top-level `params` field or not.

Initially, I had thought about the `import`, `export` and transcode` fields directly on the root of the `task` object, but then they'd probably need a prefix or sufix like `importParams` or `paramsImport`, which looks a bit cumbersome. So I did like having the `params` on the root and then the field with the task type as a key! Looks neat. 

We could do with the params fields directly on the root though, but I do have a strong preference for the params for each separate task contained within their own sub-field (`import`, `export`, etc)

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+              type: object
+              additionalProperties: false
+              description: Parameters for the export task
+              properties:
+                kind:
```
victorges: WDYT of `protocol` or `service` instead of `kind`? I think "kind" would be too generic here

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+                  type: string
+                  description: Type of the export
+                  example: arweave|ipfs|s3
```
victorges: Consider an `enum` as well, I guess we'll have to implement specific support for any additional protocol anyway, so could have the supported ones enumerated here as well.

victorges: Also, let's keep the enum to only `s3` for now if that's the only one we are going to support.

packages/api/src/schema/schema.yaml
victorges: I think the properties here should be very similar, or maybe 100% equivalent to the ones from `cli-transcoder`. Let's just copy the CLI flags from it and increment this as it grows.

We shouldn't worry about this task type anyway for the MVP. Let's focus on creating the task infra only with the `import` task for now, and we will create the `export` and `transcode` task types right after.

packages/api/src/schema/schema.yaml
victorges: Let's mimic the `cli-transcoder` for the MVP! If it doesn't support muting audio we might want to add the feature later. But I think it (and Livepeer) does not support audio-transcoding right now, so it will be unlikely that we support it **soon** (might support it sometime tho)

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+                      example: 2
+                    sampleRate:
+                      type: number
+                      description: Sample rate of the audio
+                      example: 44100
+                    bitDepth:
+                      type: number
+                      description: Bit depth of the audio
+                      example: 16
+                video:
```
victorges: I **think** `cli-transcoder` only supports some preset profiles for now.  We don't need to check that now anyway, let's create this with only the import task.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+              properties:
+                url:
+                  type: string
+                  description: URL of the asset to import
+                  example: https://cdn.livepeer.com/ABC123/filename.mp4
+            export:
+              type: object
+              additionalProperties: false
+              description: Parameters for the export task
+              properties:
```
victorges: Might need an `assetId` / `sourceAssetId` on this one as well?

packages/api/src/schema/schema.yaml
```diff
@@ -1795,6 +2174,42 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/vod-asset"
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
```
victorges: I think we won't have a raw "create" like this. You can only create an asset from an import task or by uploading a new file. Uploading a file might require a little protocol like calling the API, receiving a pre-signed S3 URL to upload the file, then creating the `asset` with the link/pre-signed for that S3 file. So we might want to have separate/explicit API endpoints like `POST /assets/upload` perhaps which returns the S3 link, then another call which will basically import that file to the regular object store? The "import" would be only so that we run the `ffprobe` on the file and fill the asset spec, but could avoid the actual download&upload from S3 but only renaming the file which would already be in our own bucket. WDYT?

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,381 @@ components:
...
+              token:
+                type: number
+                description: Token id of the NFT
+                example: 123
+
+    task:
+      type: object
+      table: task
+      additionalProperties: false
+      properties:
```
victorges, line 1054: I think some fields are missing:
 - An indexed `sourceAssetId`, so that we can always search for tasks related to a given task. It would work differently from the `sourceAssetId` of the assets themselves tho, pointing to the direct parent and not to the tree root, so we might want to call it something else. I think `parentAssetId` could make sense then. It can also be `null` in case of "import" tasks for example. In the future it might even be a list of assets, in case of stitching 2 videos together for example, but for now I think we're good with a single parent ID, which will be simpler (it's kinda complex to get indexes on JSON arrays as well (tho definitely possible))
 - The output of the task. I think on our design we had thought of a `status` **object** instead of just a string, and inside the status object we would have both the current "phase" (k8s-inspired) which is that `Pending`/`Running` etc, a timestamp of the last change, and then the `output` of the task after it's down (which should follow the same pattern as `params` IMO, with the separate sub-fields for each task type). So I think the current `status` and `updatedAt` should be moved to an actual `status` object (`status`=>`phase`) and then we add the `output` there as well. WDYT?

victorges (commented):
now some ideas on the task schema!

packages/api/src/schema/schema.yaml
gioelecerati: I agree that NFTs should be another asset, let's remove it and implement it later

packages/api/src/schema/schema.yaml
gioelecerati: I agree to remove this field, unidirectional links are more than enough

packages/api/src/schema/schema.yaml
gioelecerati: Sounds good, we could put all these fields and the ones you cited in the other comments under a video field. It is for sure clearer, if we plan on using the same Asset for other type of assets (images, NFT, audios..)  
At this point I would rename from `vod-asset` to `asset`, as vod is video specific, wdyt?
I also wonder if in a second step we should separate all the metadata in a separate place in the schema

packages/api/src/schema/schema.yaml
gioelecerati: I put also `tracks` under the video field (as well as `profiles`). I guess we can later decide if all the specs should be separate and linked to asset, so the `asset` object becomes totally agnostic to what kind of asset it is

packages/api/src/schema/schema.yaml
hjpotter92: > 

I think you mean:

```jsonc
"hash": [{
    "hash": "9b560b28b85378a5004117539196ab24e21bbd75b0e9eb1a8bc7c5fd80dc5b57",
    "algorithm": "sha256sum"
},
.
..
]
```
i was going to suggest:

```json
"hash": {
  "sha256sum": "9b560b28b85378a5004117539196ab24e21bbd75b0e9eb1a8bc7c5fd80dc5b57",
  "md5sum": "md5 value here",
  "xxhash": "xxhash value"
}
```

victorges: Like all the ideas!

packages/api/src/schema/schema.yaml
victorges: Just asset instead of vod-asset makes sense to me as well! And we can separate the metadata, but I'm ok with keeping it on the root now as, after all, this whole object is metadata anyway. So we got the common file metadata on the root and the specialized metadata in the category-specific fields.

packages/api/src/schema/schema.yaml
gioelecerati: I removed the export task for now to define it later when we implement it!

packages/api/src/controllers/task.js
```diff
@@ -0,0 +1,180 @@
...
+
+  if (output.length > 0) {
+    res.links({ next: makeNextHREF(req, newCursor) });
+  }
+
+  return res.json(output);
+});
+
+app.get("/:id", authMiddleware({}), async (req, res) => {
+  const os = await req.store.get(`task/${req.params.id}`);
```
iameli: Should have `db.task` in all cases

packages/api/src/controllers/task.js
```diff
@@ -0,0 +1,180 @@
...
+
+app.post("/", authMiddleware({}), validatePost("task"), async (req, res) => {
+  const id = uuid();
+  const doc = validateTaskPayload(id, req.user.id, Date.now(), req.body);
+  try {
+    await db.task.create(doc);
+  } catch (e) {
+    console.error(e);
+    throw e;
+  }
```
iameli: Shouldn't need the try/catch, will just get picked up by our middleware

```suggestion
  await db.task.create(doc);
```

packages/api/src/controllers/task.js
```diff
@@ -0,0 +1,180 @@
...
+    return res.json({
+      errors: ["users may only delete their own tasks"],
+    });
+  }
+  await db.task.delete(id);
+  res.status(204);
+  res.end();
+});
+
+app.put("/:id", authMiddleware({}), validatePost("task"), async (req, res) => {
```
iameli: Maybe we just need an admin-only `PATCH` endpoint for the worker to update

victorges: Like this idea, at least for a start. Same for the assets, which at first can/may be editable only by the task runner itself (apart from very surface metadata like name/title).

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,289 @@ components:
...
+        userId:
+          type: string
+          readOnly: true
+          example: 66E2161C-7670-4D05-B71D-DA2D6979556F
+          index: true
+          description: owner of the asset
+        deleted:
+          type: boolean
+          description: Set to true when the asset is deleted
+        objectStoreId:
```
victorges, line 895: So I found that this field is considered "private" on the streams, which is likely because most of the time it will be referencing an object that cannot be retrieved by the user (an object store in Ivan's account). I believe we should follow the same pattern here, and make this field a `writeOnly` property, which will never be returned by the API **unless** an admin is making the request (so that `task-runner` can access it).

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,289 @@ components:
...
+          type: object
+          additionalProperties: false
+          description: Output of the task
+          properties:
+            import:
+              type: object
+              additionalProperties: false
+              description: Output of the import task
+              properties:
+                status:
```
victorges: The `status` field should be shared by all task types instead. I think it's more of a metadata than an output instead as well, so I think it can be moved to the root. 

Following the Kubernetes API, the status could also be an object that has a `phase` of `Pending/Running/Failed/Completed` instead, and we can move the `updatedAt` to inside the status as well WDYT?

```
status:
  phase: Pending
  updatedAt: 12437328123
  // later we can add more fields if appropriate
```

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,294 @@ components:
...
+              description:
+                List of tracks associated with the asset when the format
+                contemplates them (e.g. mp4)
+              items:
+                type: object
+                additionalProperties: false
+                required:
+                  - kind
+                properties:
+                  kind:
```
victorges: For consistency with ffmpeg and the rest of the schema we're avoiding `kind`, maybe this should be `type` instead

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,294 @@ components:
...
+                    example: 44100
+                  bitDepth:
+                    type: number
+                    description: Bit depth of the track - only for audio tracks
+                    example: 16
+                  codec:
+                    type: string
+                    description: Codec of the track
+                    example: aac
+            hlsSpec:
```
victorges: Let's keep this off/commented until we actually implement it as well

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,294 @@ components:
...
+              type: object
+              additionalProperties: false
+              description: HLS spec for the asset
+              properties:
+                profiles:
+                  type: object
```
victorges: I think you meant array of objects

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,294 @@ components:
...
+          description: owner of the asset
+        deleted:
+          type: boolean
+          description: Set to true when the asset is deleted
+        objectStoreId:
+          type: string
+          description: Object store ID where the asset is stored
+          writeOnly: true
+          example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+        name:
```
victorges, line 900: Hmm, I think we will want to get the filename somehow on the import asset flow in the end to populate this. Or should it be populated on the request to start the import instead? Maybe both, I'll do it on the import task as well since it's easy but the user can provide one on the request to start the import, which if provided should be used instead of the one that the import task gets.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,294 @@ components:
...
+          description: Timestamp (in milliseconds) at which asset was created
+          example: 1587667174725
+        hash:
+          type: array
+          description: Hash of the asset
+          items:
+            type: object
+            additionalProperties: false
+            properties:
+              hash:
```
victorges, line 937: Reminder for myself that I need a third processing of the file contents for calculating this hash. Or get it from the response from SE, but I'm not sure if it's available (currently not exposed on the OS drivers lib at least)

packages/api/src/schema/schema.yaml
gioelecerati: Yeah, I agree, we could move it to the root, I was thinking that different kinds of task in the future could have new types of status phases, but I guess we can add them in the phase enum and support all cases

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,310 @@ components:
             created
           example: 1587667174725
 
+    asset:
+      type: object
+      table: asset
+      additionalProperties: false
+      required:
+        - name
+      properties:
```
victorges, line 868: I think we are missing a `size` property on the root of the asset object as well.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,310 @@ components:
...
+                  height:
+                    type: number
+                    description: Height of the track - only for video tracks
+                    example: 1080
+                  fps:
+                    type: number
+                    description: Frame rate of the track - only for video tracks
+                    example: 30
+                  id:
+                    type: string
```
victorges: What is this track ID? ffprobe apparently doesn't return anything like that.

packages/api/src/schema/schema.yaml
```diff
@@ -811,6 +815,310 @@ components:
...
+                    example: video
+                  width:
+                    type: number
+                    description: Width of the track - only for video tracks
+                    example: 1920
+                  height:
+                    type: number
+                    description: Height of the track - only for video tracks
+                    example: 1080
+                  fps:
```
victorges, line 1011: I think we should also have the `pixelFormat` here since it's one of the things that can break the transcoding flow (`pixfmt` from ffmpeg)

packages/api/src/schema/schema.yaml
gioelecerati: Yeah we don't need it, there's not reference to it elsewhere, removed!

packages/api/src/schema/schema.yaml
gioelecerati: Good point!

packages/api/src/controllers/asset.ts
```diff
@@ -0,0 +1,351 @@
...
+    });
+    if (!qres.data.length && !otherKeys.includes(playbackId)) {
+      return playbackId;
+    }
+  }
+}
+
+function validateAssetPayload(id, userId, createdAt, payload) {
+  try {
+    if (payload.meta && JSON.stringify(payload.meta).length > META_MAX_SIZE) {
```
victorges, line 51: nice :)

packages/api/src/task/task.ts
```diff
@@ -0,0 +1,95 @@
...
+          let assetSpec;
+          try {
+            assetSpec = event.output.import.assetSpec;
+          } catch (e) {
+            console.log(
+              `task event process error: assetSpec not found in TaskResult for task ${event.task.id}`
+            );
+          }
+
+          db.asset.update(task.parentAssetId, {
```
victorges: The task status also needs to be updated with this output.

packages/api/src/controllers/asset.ts
```diff
@@ -0,0 +1,431 @@
...
+  const presignedUrl = await getS3PresignedUrl({
+    objectKey: `directUpload/${playbackId}/source`,
+    vodObjectStoreId,
+  });
+
+  const b64SignedUrl = Buffer.from(presignedUrl).toString("base64");
+  const lpSignedUrl = `https://${req.frontendDomain}/api/asset/upload/${b64SignedUrl}`;
+
+  await db.asset.create({
+    id,
```
victorges, line 286: Seems to be missing a couple of fields here. Either use the `validateAssetPayload` which I think should become a function to "create new asset payload with all the required fields". Or you can repeat them here, but I think it's worth keeping all those default stuff in a single place, since there will be a lot of new asset creation flows to come.

packages/api/src/task/task.ts
```diff
@@ -0,0 +1,102 @@
...
+            hash: assetSpec.hash,
```
victorges, line 80: need `await` here and above

packages/api/src/task/task.ts
```diff
@@ -0,0 +1,102 @@
...
+        if (event.output) {
+          let assetSpec;
+          try {
+            assetSpec = event.output.import.assetSpec;
+          } catch (e) {
+            console.log(
+              `task event process error: assetSpec not found in TaskResult for task ${event.task.id}`
+            );
+          }
+
```
victorges: could add a TODO here to do these in a transaction

victorges (approved):
Ship it!

PR #866, opened 2022-01-19
www: added status & status details on webhooks dashboard

### LINKED ISSUE #867: www: Add StatusBadge on webhooks dashboard

gioelecerati:

Add a StatusBadge on the dashboard in the webhooks table

- Add StatusBadge
- Last triggered time shown on hover in tooltip

yarn test & on local dashboard

Fixes #867 

**Screenshots (optional):**

status
status_detail

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/StatusBadge/index.tsx
+++ b/packages/www/components/Dashboard/StatusBadge/index.tsx
@@ -50,12 +50,17 @@ const StatusBadge = ({
       typeof timestamp === "string" ? Date.parse(timestamp) : timestamp;
     return !tsNum ? moment.invalid() : moment.unix(tsNum / 1000);
   }, [timestamp]);
-  if (!timeAgo.isValid() || style.noTooltip) {
+  if (
+    (!timeAgo.isValid() && !tooltipText) ||
+    (style.noTooltip && !tooltipText)
+  ) {
     return badge;
   }
-  let contentText = timeAgo.fromNow();
-  if (tooltipText) {
-    contentText = `${tooltipText} ${contentText}`;
+  let contentText: string;
+  if (tooltipText && timestamp) {
+    contentText = `${tooltipText} ${timeAgo.fromNow()}`;
+  } else {
+    contentText = tooltipText;
   }
   return <Tooltip content={contentText}>{badge}</Tooltip>;
 };
--- a/packages/www/components/Dashboard/StatusBadge/index.tsx
+++ b/packages/www/components/Dashboard/StatusBadge/index.tsx
@@ -50,14 +50,11 @@ const StatusBadge = ({
       typeof timestamp === "string" ? Date.parse(timestamp) : timestamp;
     return !tsNum ? moment.invalid() : moment.unix(tsNum / 1000);
   }, [timestamp]);
-  if (
-    (!timeAgo.isValid() && !tooltipText) ||
-    (style.noTooltip && !tooltipText)
-  ) {
+  if (!tooltipText && (!timeAgo.isValid() || style.noTooltip)) {
     return badge;
   }
   let contentText: string;
-  if (tooltipText && timestamp) {
+  if (tooltipText && timeAgo.isValid()) {
     contentText = `${tooltipText} ${timeAgo.fromNow()}`;
   } else {
     contentText = tooltipText;
--- a/packages/www/components/Dashboard/StatusBadge/index.tsx
+++ b/packages/www/components/Dashboard/StatusBadge/index.tsx
@@ -30,9 +30,11 @@ const styleByVariant: Record<Variant, Style> = {
 const StatusBadge = ({
   variant,
   timestamp,
+  tooltipText,
 }: {
   variant: Variant;
   timestamp?: string | number;
+  tooltipText?: string;
 }) => {
   const style = styleByVariant[variant];
   const badge = (
@@ -51,7 +53,11 @@ const StatusBadge = ({
   if (!timeAgo.isValid() || style.noTooltip) {
     return badge;
   }
-  return <Tooltip content={timeAgo.fromNow()}>{badge}</Tooltip>;
+  let contentText = timeAgo.fromNow();
+  if (tooltipText) {
+    contentText = `${tooltipText} ${contentText}`;
+  }
+  return <Tooltip content={contentText}>{badge}</Tooltip>;
 };
 
 export default StatusBadge;
--- a/packages/www/components/Dashboard/WebhooksTable/index.tsx
+++ b/packages/www/components/Dashboard/WebhooksTable/index.tsx
@@ -135,7 +135,10 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
                     <StatusBadge
                       variant={
                         webhook.status.lastFailure
-                          ? StatusVariant.Unhealthy
+                          ? webhook.status.lastFailure.timestamp >=
+                            webhook.status.lastTriggeredAt
+                            ? StatusVariant.Unhealthy
+                            : StatusVariant.Healthy
                           : StatusVariant.Healthy
                       }
                       timestamp={webhook.status.lastTriggeredAt}
--- a/packages/www/components/Dashboard/WebhooksTable/index.tsx
+++ b/packages/www/components/Dashboard/WebhooksTable/index.tsx
@@ -3,6 +3,7 @@ import { useApi } from "../../../hooks";
 import Table, { Fetcher, useTableState } from "components/Dashboard/Table";
 import TextCell, { TextCellProps } from "components/Dashboard/Table/cells/text";
 import DateCell, { DateCellProps } from "components/Dashboard/Table/cells/date";
+import StatusBadge, { Variant as StatusVariant } from "../StatusBadge";
 import { dateSort, stringSort } from "components/Dashboard/Table/sorts";
 import { SortTypeArgs } from "components/Dashboard/Table/types";
 import { Column } from "react-table";
@@ -32,6 +33,7 @@ type WebhooksTableData = {
   name: TextCellProps;
   url: TextCellProps;
   created: DateCellProps;
+  status: TextCellProps;
 };
 
 const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
@@ -68,6 +70,12 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
         Cell: DateCell,
         sortType: (...params: SortTypeArgs) =>
           dateSort("original.created.date", ...params),
+      },
+      {
+        Header: "Status",
+        accessor: "status",
+        Cell: TextCell,
+        disableSortBy: true,
       },
     ],
     []
@@ -117,6 +125,24 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
               css: {
                 cursor: "pointer",
               },
+            },
+            status: {
+              children: (
+                <Box>
+                  {!webhook.status ? (
+                    "-"
+                  ) : (
+                    <StatusBadge
+                      variant={
+                        webhook.status.lastFailure
+                          ? StatusVariant.Unhealthy
+                          : StatusVariant.Healthy
+                      }
+                      timestamp={webhook.status.lastTriggeredAt}
+                    />
+                  )}
+                </Box>
+              ),
             },
           };
         }),
--- a/packages/www/components/Dashboard/WebhooksTable/index.tsx
+++ b/packages/www/components/Dashboard/WebhooksTable/index.tsx
@@ -36,6 +36,9 @@ type WebhooksTableData = {
   status: TextCellProps;
 };
 
+// 1 hour
+const WARNING_TIMEFRAME = 1000 * 60 * 60;
+
 const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
   const router = useRouter();
   const { user, getWebhooks, deleteWebhook, deleteWebhooks, createWebhook } =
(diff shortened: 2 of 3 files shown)


packages/www/components/Dashboard/WebhooksTable/index.tsx
```diff
@@ -118,6 +126,28 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
                 cursor: "pointer",
               },
             },
+            status: {
+              children: (
+                <Box>
+                  {!webhook.status ? (
+                    "-"
```
gioelecerati: I think the `Idle` status badge would be not the correct one here (when the status is missing), in the stream `-` is displayed, but probably would be better an empty string?

packages/www/components/Dashboard/StatusBadge/index.tsx
```diff
@@ -51,7 +53,11 @@ const StatusBadge = ({
   if (!timeAgo.isValid() || style.noTooltip) {
     return badge;
   }
-  return <Tooltip content={timeAgo.fromNow()}>{badge}</Tooltip>;
+  let contentText = timeAgo.fromNow();
+  if (tooltipText) {
+    contentText = `${tooltipText} ${contentText}`;
+  }
+  return <Tooltip content={contentText}>{badge}</Tooltip>;
```
victorges: **nit:** Just naming, wdyt of `tootlipPrefix` for the arg received externally, and then we use that to create `tooltipText`?
```suggestion
  let tooltipText = timeAgo.fromNow();
  if (tooltipPrefix) {
    tooltipText = `${tooltipPrefix} ${tooltipTexst}`;
  }
  return <Tooltip content={tooltipText}>{badge}</Tooltip>;
```

packages/www/components/Dashboard/WebhooksTable/index.tsx
victorges: I think `Idle` makes sense tho, as the meaning of a webhook that hasn't received any events. Maybe the tooltip could even explain it like `"This webhook hasn't received any events yet"` (gotta update the `tooltipText` logic to allow one with no timestamps tho).

If you think there would be some other more accurate term, you can also create new `StatusVariant`s for that. I'm also ok with `-` anyway if that looks ok in the table.

packages/www/components/Dashboard/WebhooksTable/index.tsx
```diff
@@ -118,6 +126,28 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
...
+                  ) : (
+                    <StatusBadge
+                      variant={
+                        webhook.status.lastFailure
+                          ? webhook.status.lastFailure.timestamp >=
+                            webhook.status.lastTriggeredAt
```
victorges: I think we need a more "stable" logic here like for example showing the webhook as unhealthy, or at least a warning, if there has been any failure on the past X hours. Receiving a success does not mean that it is healthy, as we chatted already.

packages/www/components/Dashboard/StatusBadge/index.tsx
gioelecerati: It does make sense, but if I update the logic allowing `tooltipText` to be also without a timestamp for different kind of status badges, it's not necessarily a prefix, wdyt?

packages/www/components/Dashboard/WebhooksTable/index.tsx
gioelecerati: The thing is, when we have a webhook that failed at lease once in the past X hours, we can display it unhealthy with a warning of the failure, but should we display both the last failure timestamp and the last trigger? Because a webhook with a failure in the last X hour would stay unhealthy until enough time as passed, without any clear indication that some hooks are successful. I have pushed a [commit](https://github.com/livepeer/livepeer-com/pull/866/commits/08c6ae5b0081caba10f2a41c81a01c593f2de650#diff-3844b3b4667b4fbea621d6c24eac9c307c9616c60a3c1135562adca5ab83c1d3R142) with this for clarity on what I mean!

packages/www/components/Dashboard/StatusBadge/index.tsx
victorges: What if it's a function that receives the timeAgo (if a timestamp was provided) and returns the final tooltip text? 

Or maybe it should just be a string, not treated as a prefix but as the actual text, and we can repeat the timeAgo logic on the webhooks table when we wanna have it. Could create a helper if it feels cumbersome to repeat, but I think it should be ok either way (callback might be overengineering and suffer from React callback hell)

packages/www/components/Dashboard/WebhooksTable/index.tsx
victorges: We can display only the most relevant information for the webhooks on the table, and if you want more details you go into the detail page and can see each value individually.

In the case of the unhealthy state, I think it's more relevant when was the last failure than when was the last trigger, so that's what I would display here.

packages/www/components/Dashboard/StatusBadge/index.tsx
```diff
@@ -48,10 +50,19 @@ const StatusBadge = ({
       typeof timestamp === "string" ? Date.parse(timestamp) : timestamp;
     return !tsNum ? moment.invalid() : moment.unix(tsNum / 1000);
   }, [timestamp]);
-  if (!timeAgo.isValid() || style.noTooltip) {
+  if (
+    (!timeAgo.isValid() && !tooltipText) ||
+    (style.noTooltip && !tooltipText)
```
victorges: **nit:**

```suggestion
    !tooltipText && (!timeAgo.isValid() || style.noTooltip)
```

packages/www/components/Dashboard/StatusBadge/index.tsx
```diff
@@ -48,10 +50,19 @@ const StatusBadge = ({
...
+  ) {
     return badge;
   }
-  return <Tooltip content={timeAgo.fromNow()}>{badge}</Tooltip>;
+  let contentText: string;
+  if (tooltipText && timestamp) {
```
victorges: Gotta check if timeAgo is valid instead

```suggestion
  if (tooltipText && timeAgo.isValid()) {
```

packages/www/components/Dashboard/WebhooksTable/index.tsx
```diff
@@ -118,6 +129,32 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
...
+                      variant={StatusVariant.Idle}
+                      tooltipText="This webhook hasn't received any events yet"
+                    />
+                  ) : webhook.status.lastFailure &&
+                    +Date.now() - webhook.status.lastFailure.timestamp <
+                      WARNING_TIMEFRAME ? (
+                    <StatusBadge
+                      variant={StatusVariant.Unhealthy}
+                      timestamp={webhook.status.lastTriggeredAt}
+                      tooltipText="A trigger has failed in the last hour. Last triggered"
```
victorges: Followup on my other comment, I think this would be cleaner and give the most useful information:

```suggestion
                      timestamp={webhook.status.lastFailure.timestamp}
                      tooltipText="Last error"
```

I think tooltips should ideally be really short.

packages/www/components/Dashboard/WebhooksTable/index.tsx
```diff
@@ -118,6 +129,32 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
                 cursor: "pointer",
               },
             },
+            status: {
+              children: (
+                <Box>
+                  {!webhook.status ? (
+                    <StatusBadge
+                      variant={StatusVariant.Idle}
+                      tooltipText="This webhook hasn't received any events yet"
```
victorges: **nit:** For a short tooltip:

```suggestion
                      tooltipText="No triggers yet"
```
(Or events)

victorges (commented):
Nice!

I think you're only missing the webhook detail page to show all the information about last trigger, error, error message etc

packages/www/components/Dashboard/WebhooksTable/index.tsx
gioelecerati: I like it, I added the Idle badge and displayed the detailed infos in the detail page

packages/www/components/Dashboard/WebhooksTable/index.tsx
```diff
@@ -118,6 +129,36 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
...
+                      tooltipText="Last failure"
+                    />
+                  ) : (
+                    <StatusBadge
+                      variant={StatusVariant.Healthy}
+                      timestamp={webhook.status.lastTriggeredAt}
+                      tooltipText="Last triggered"
+                    />
+                  )}
+                </Box>
```
victorges: Neat!

packages/www/pages/dashboard/developers/webhooks/[id].tsx
```diff
@@ -217,6 +217,34 @@ const WebhookDetail = () => {
...
+                      )
+                    : "Never"}
+                </Cell>
+                <Cell variant="gray"></Cell>
+                <Cell>
+                  {data.status?.lastFailure
+                    ? data.status.lastFailure.response
+                      ? data.status.lastFailure.response
+                      : data.status.lastFailure.error
+                    : ""}
```
Comment (victorges) [inline packages/www/pages/dashboard/developers/webhooks/[id].tsx]: We probably want to include the status code here instead of only the response body?

victorges (approved):
LGTM!

Do you mind including in the PR some screenshots about how the status badges and info in the webhook details page looks? I was not able to see it in action on the preview.

packages/www/components/Dashboard/WebhooksTable/index.tsx
gioelecerati: Spotted a issue! If `lastTriggeredAt` is higher than `WARNING_TIMEFRAME` and is failed, it becomes `Healthy`! Fixed now

packages/www/pages/dashboard/developers/webhooks/[id].tsx
Comment (gioelecerati) [reply packages/www/pages/dashboard/developers/webhooks/[id].tsx]: Added it!

gioelecerati:
> LGTM!
> 
> Do you mind including in the PR some screenshots about how the status badges and info in the webhook details page looks? I was not able to see it in action on the preview.

Sure, added some screenshots right now!

packages/www/components/Dashboard/WebhooksTable/index.tsx
```diff
@@ -118,6 +129,38 @@ const WebhooksTable = ({ title = "Webhooks" }: { title?: string }) => {
...
+                <Box>
+                  {!webhook.status ? (
+                    <StatusBadge
+                      variant={StatusVariant.Idle}
+                      tooltipText="No triggers yet"
+                    />
+                  ) : (webhook.status.lastFailure &&
+                      +Date.now() - webhook.status.lastFailure.timestamp <
+                        WARNING_TIMEFRAME) ||
+                    webhook.status.lastFailure.timestamp >=
```
victorges, line 143: Good catch :)

packages/www/pages/dashboard/developers/webhooks/[id].tsx
```diff
@@ -217,6 +217,34 @@ const WebhookDetail = () => {
...
+                        data.status.lastFailure.timestamp,
+                        "MMMM dd, yyyy h:mm:ss a"
+                      )
+                    : "Never"}
+                </Cell>
+                <Cell variant="gray"></Cell>
+                <Cell>
+                  {data.status?.lastFailure
+                    ? data.status.lastFailure.response
+                      ? `Status code: ${data.status.lastFailure.statusCode} - ${data.status.lastFailure.response}`
```
Comment (victorges) [inline packages/www/pages/dashboard/developers/webhooks/[id].tsx]: Hmm, now I think this will look too crammed up on a single line. Do you think it's possible to add a line break between the status code and the response body? 

Also, is that response body displayed as code / with a monospaced font? I wonder if we are being explicit enough about what this is or if it might end up looking like a random string that doesn't make sense to the user.

Here is what I (later) managed to get on my webhook:
Screen Shot 2022-01-26 at 16 59 28
The response body is exactly "500 Internal Server Error", with the update here it becomes `Status Code 500 - 500 Internal Server Error`, none of which seems to be really clear to the user what they actually mean, and it could be much worse depending on the response body.

Soo... WDYT of adding a row "header" just like we have `Last failure`, `Last trigger` etc above so we are clear about what this means? Could be
 - Last failure: `<date>`
 - Error response: `<status code> <status code text> <new line> <response body>`

Or the error response fully separate like
 - Error Status Code: `<status code> <status code text>`
 - Error Response: `<response body>` (with proper line wrapping and ideally formatting as code)

WDYT?

Comment (victorges) [reply packages/www/pages/dashboard/developers/webhooks/[id].tsx]: I think on JavaScript you can get the status text like this:
```
import { STATUS_CODES } from "http";
```
then later
```
STATUS_CODES[status]
```
(not sure how perfectly that will work on the browser tho, gotta test)

Comment (gioelecerati) [reply packages/www/pages/dashboard/developers/webhooks/[id].tsx]: Good points, the thing I had to consider is that if the request fails, there are no status code and error responses, but just and error message. In any case, I added separate rows and a condition check where if no status code is present, I display `Error message` instead of `Error response`, without the status code row - both monospaced!

This is the result:
mono

I also added the status text as you suggested, thanks!

Comment (victorges) [reply packages/www/pages/dashboard/developers/webhooks/[id].tsx]: Looks great!

victorges (approved):
Great!

I think the only limitation to merge this is getting the backend deployed as well. Did a little check and I think we haven't released it on this Tuesday's load test right?

gioelecerati:
> Great!
> 
> I think the only limitation to merge this is getting the backend deployed as well. Did a little check and I think we haven't released it on this Tuesday's load test right?

Yes, the backend is merged but not released yet

PR #868, opened 2022-01-19
fixed player refreshing issue

### LINKED ISSUE #726: Switching between the "overview" and "health" tabs results in the test player refreshing
(Minor): Switching between the "overview" and "health" tabs results in the test player refreshing

https://user-images.githubusercontent.com/257909/141511713-b0cf16ed-687f-4984-8dae-e6a5ade149ad.mov

FullStackTiger:

## Switching between the "overview" and "health" tabs results in the test player refreshing

Fixes #726 

**Screenshots (optional):**

**Checklist:**


packages/www/pages/dashboard/streams/[id]/health.tsx
```diff
@@ -1,148 +0,0 @@
-import React, { useCallback } from "react";
```
Comment (victorges) [inline packages/www/pages/dashboard/streams/[id]/health.tsx:1]: Is there a way to still keep this as a separate file, even if not a full `page` anymore? Could be a component that is included by the stream details page.

Just feels like they currently have separate logics and were already kept separate, and this would be merging all of them into a single thing, with no clear separation on the purpose of the code.

packages/www/pages/dashboard/streams/[id]/index.tsx
```diff
@@ -38,39 +49,131 @@ const Overview = () => {
...
+    if (!id) {
+      return;
+    }
+    doGetInfo(typeof id === "string" ? id : null);
+  }, [doGetInfo]);
+
+  useEffect(() => {
+    const intervalId = window.setInterval(() => {
+      if (info) {
+        getIngestRate(typeof id === "string" ? id : null);
```
Comment (victorges) [inline packages/www/pages/dashboard/streams/[id]/index.tsx]: Edit: Just noticed that this was copied verbatim from the previous stream health implementation. I guess it's not necessary to change it but would still be a welcome addition/improvement 😄 

I don't think we should ever be calling neither `doGetInfo` nor `getIngestRate` with `null` as the input. That means they will do an useless roundtrip on the server just to get nothing back. If there's no use for that call, I think you should just check that `typeof` on the `if` that you have right before calling them. Like `if (typeof id !== "string") return`
```suggestion
    if (typeof id !== "string") {
      return;
    }
    doGetInfo(id);
  }, [doGetInfo]);

  useEffect(() => {
    const intervalId = window.setInterval(() => {
      if (typeof id === "string") {
        getIngestRate(id);
```
(I think the `if` before `getIngestRate` was wrong? Should it really be checking that `info` var instead which is not even used here?)

packages/www/pages/dashboard/streams/[id]/index.tsx
```diff
@@ -1,20 +1,31 @@
...
+const maxItems = 6;
 const refetchInterval = 5 * 1000;
 
 const Overview = () => {
   const router = useRouter();
   const queryClient = useQueryClient();
-  const { getStream } = useApi();
+  const { getStream, getStreamInfo } = useApi();
   const { getHealth } = useAnalyzer();
+  const [switchTab, setSwitchTab] = useState<"Overview" | "Health">("Overview");
```
Comment (victorges) [inline packages/www/pages/dashboard/streams/[id]/index.tsx]: **nit:**
```suggestion
  const [currentTab, setCurrentTab] = useState<"Overview" | "Health">("Overview");
```
(or maye `setCurrentTab` could be `switchTab`, but I think `setSwitchTab` doesn't make sense)

packages/www/pages/dashboard/streams/[id]/index.tsx
```diff
@@ -38,39 +49,131 @@ const Overview = () => {
...
+            }}>
+            <Heading size="1" css={{ fontWeight: 500, mb: "$1" }}>
+              Session ingest rate
+            </Heading>
+            <Text variant="gray" size="3">
+              After the stream loads, ingest rate updates every 10 seconds.
+            </Text>
+          </Box>
+          <Chart data={dataChart} />
+        </>
```
Comment (victorges) [inline packages/www/pages/dashboard/streams/[id]/index.tsx]: Here, instead of copying the implementation of the stream health tab, we could just insert a `StreamHealthTab` component or something like that! WDYT?

Ideally we would also have a separate component for the `"Overview"` tab, and this whole page here would only need to worry about the tab-switching logic. That would be awesome, but don't feel obliged to do that refactor. Just keeping the stream health tab separate, as it was before, already LGTM.

victorges (approved):
LGTM

Issue #851, opened 2022-01-12
Add stream health log for orchestrator where stream is being transcoded


victorges:
In the stream health dashboard, show a log like 
`Stream is being transcoded on orchestrator XYZ`

Then if the orchestrator changes, add a new log like
`Stream has been switched to be transcoded on orchestrator XYZ`

PR #869, opened 2022-01-20
www: Add some additional Stream Health logs

victorges:
This is to add some additional stream health logs in the dashboard from the data
that we already have available. 

Add logs for:
 - Which orchestrator a stream is currently being transcoded. Nice transparency outcome there plus might diagnose some issues like if the stream is hopping between different orchestrators too much (apparently happens a lot in staging, from my tests).
 - When a segment gets successfully transcoded, after we have logged errors about it. Just to not look like the transcoding never happened for some segments, when a later retry actually succeeded.

## -

Stream to staging and prod to check and iterate on logs. 

Implements #851 

**Screenshots (optional):**
Screen Shot 2022-01-20 at 14 14 07

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/Logger/index.tsx
+++ b/packages/www/components/Dashboard/Logger/index.tsx
@@ -64,7 +64,9 @@ function orchestratorName({
 
 function createEventHandler() {
   const [lastOrchestrator, setLastOrchestrator] = useState<string>();
-  const [failedSegments, setFailedSegments] = useState<number[]>([]);
+  const [failedSegments, setFailedSegments] = useState<Record<number, boolean>>(
+    {}
+  );
 
   return function handleEvent(
     evt: events.Any,
@@ -81,7 +83,7 @@ function createEventHandler() {
 
         // non-admin users should only see fatal errors.
         if (!evt.success || userIsAdmin) {
-          setFailedSegments([...failedSegments, seqNo]);
+          setFailedSegments({ ...failedSegments, [seqNo]: true });
           const errLogs = evt.attempts
             .filter((a) => a.error)
             .map((a, idx) => {
@@ -107,7 +109,7 @@ function createEventHandler() {
           );
         }
 
-        if (evt.success && failedSegments.includes(seqNo)) {
+        if (evt.success && failedSegments[seqNo]) {
           logs.push(
             infoLog(
               evt,
@@ -115,7 +117,7 @@ function createEventHandler() {
               "segment-success"
             )
           );
-          setFailedSegments(failedSegments.filter((s) => s !== seqNo));
+          setFailedSegments({ ...failedSegments, [seqNo]: false });
         }
 
         return logs;
--- a/packages/www/components/Dashboard/Logger/index.tsx
+++ b/packages/www/components/Dashboard/Logger/index.tsx
@@ -1,5 +1,5 @@
 import { useAnalyzer, useApi } from "hooks";
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 import { Stream } from "@livepeer.com/api";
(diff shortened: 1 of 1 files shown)


packages/www/components/Dashboard/Logger/index.tsx
```diff
@@ -44,52 +62,83 @@ function orchestratorName({
...
+            if (errLogs.length > 0) {
+              failedSegments.current.add(seqNo);
+              logs.push(...errLogs);
+            }
+          }
+
+          const lastAttempt = evt.attempts?.length
+            ? evt.attempts[evt.attempts.length - 1]
+            : null;
+          const orchestrator = orchestratorName(lastAttempt);
```
gioelecerati: This part is not clear to me, getting the orchestrator name passing the `lastAttempt` parameter to the `orchestratorName` function? What happens if `lastAttempt` is null? I am probably missing some context, so this is just a question

packages/www/components/Dashboard/Logger/index.tsx
```diff
@@ -44,52 +62,83 @@ function orchestratorName({
...
+          const seqNo = evt.segment.seqNo;
+
+          // non-admin users should only see fatal errors.
+          if (!evt.success || userIsAdmin) {
+            const errLogs = evt.attempts
+              .filter((a) => a.error)
+              .map((a, idx) => {
+                const orch = orchestratorName(a);
+                const msg = `Transcode error from ${orch} for segment ${seqNo}: ${a.error}`;
+                return errorLog(evt, msg, `error-${idx}`);
```
gioelecerati, line 89: On this error log, is the orchestrator error sent on the dashboard? Do we need this kind of extensive information about the orchestrator failure to transcode/deliver the segment to be provided to the user? Or is it in any case the most generic piece of information we have about the error?

gioelecerati (approved):
LGTM, just a couple of questions to get more context

packages/www/components/Dashboard/Logger/index.tsx
victorges: So the `orchestratorName` function takes an attempt object, for easy usage here and especially on the `errLogs` transformations above, which were why they were originally created. If we want to be extra precise on the interface here, we should change it to receive only the `orchestrator` field of the attempt instead of the whole attempt, which is all that it needs. I haven't done it like this before cause the `errLogs` were very minimal, created on a single-line lambda, but now they're not so I'll update the interface for clarity :)

Also good catch on the `null` case, it was defeinitely not being handled and would explode here. It is kind of a overcaution here since we'll likely never have a null or empty attempts array anyway, but I'll keep it anyway. So I'm fixing those 2 things!

packages/www/components/Dashboard/Logger/index.tsx
victorges: We only provide the error to user in case all attempts to transcode the segment have failed (`!evt.success`). That likely means that they are ingesting some bad media to us, and should definitely see the error that comes out of the transcoding process. This was one of the main goals we had with stream health, so that users can diagnose any issues they might have with their stream that they were previously unable to identify before. One example of those logs is if you send the video with an unsupported pixel format, and you'll get that error on the dashboard through this log.

On the worst case, it might be the case that we are having some issues in our infrastructure and the transcoding is failing because of that. That still means that the users' stream is going to fail, and I think it's better for us to show some error to them anyway, even if it is internal. If they want high-level information, they can also check only the healthchecks anyway and will see the transcoding/realtime/etc "boolean" health conditions about the stream, which should be easier for the less technical users. 

This is all pretty new and not iterated enough anyway, so we're likely changing strategies in the future on how to display some of this information for the user. For now, that is the mindset tho, high-level status on the healthchecks and low-level info in the logs.

PR #870, opened 2022-01-20
add pkg capabilities (act 2)

iameli:
Accidentally merged the previous one with an errant `lerna version` command. Everything I typed here still applies: https://github.com/livepeer/livepeer-com/pull/852

Ready for final review. Preview of the released binaries here: https://github.com/iameli/livepeer-com/releases/tag/v0.1.3

Couple hacky things in here. Open to suggestions.


packages/www/server.js
```diff
@@ -0,0 +1,47 @@
...
+    friendly: "Livepeer Dashboard",
+    desc: "Livepeer Next.js Dashboard Front-end Server",
+    optional: {
+      port: {
+        option: "--port",
+        default: 3012,
+        help: "Port to run on",
+      },
+    },
+    version: pkg.version,
```
hjpotter92, line 36: should we use `pkg.description` and `.name` fields as well?

hjpotter92 (approved):
LGTM!

PR #875, opened 2022-01-21
api: Set some default profiles to streams

victorges:
It's kind of a roadblock for initial developers trying to test
Livepeer to be obliged to provide transcoding profiles for the
stream, especially when they might not even know how to configure
those.

When I went through it, I actually went on the dashboard and open
the developer network tab to find out what the dashboard was setting
when creating a stream from the UI, since from there, there is no
requirement for configuring anything.

With this change, it becomes simpler for API users as well, while
still maintaining retro-compatibility and allowing to configure the
profiles manually if desired.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -1076,6 +1076,22 @@ describe("controllers/stream", () => {
       }
     });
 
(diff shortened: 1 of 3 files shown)


gioelecerati (approved):
LGTM

PR #893, opened 2022-02-09
api: Add export and transcode to Asset and Task 

### LINKED ISSUE #890: Stream recordings --> Assets
Create import task with the URL where we serve the MP4

### LINKED ISSUE #894: Add API export and transcode

gioelecerati:

Add export and transcode for the vod flow

- Add export and transcode endpoints
- Add TaskResult handler for these events

Fixes #894 
Fixes #890 

**Screenshots (optional):**

**Checklist:**


packages/api/src/controllers/asset.ts
```diff
@@ -169,49 +173,43 @@ app.get("/", authMiddleware({}), async (req, res) => {
...
+  const asset = await db.asset.get(assetId);
+  if (!asset) {
+    throw new NotFoundError(`Asset not found with id ${assetId}`);
+  }
+  if (req.user.id !== asset.userId) {
+    throw new ForbiddenError(`User can only export their own assets`);
   }
-);
+
+  let task = await req.taskScheduler.scheduleTask(asset, "export", {
```
victorges: Nice `scheduleTask` helper 😌 

I think the `update` to `phase: "waiting"` should be inside it as well though.

packages/api/src/schema/schema.yaml
```diff
@@ -1109,6 +1110,24 @@ components:
...
+              type: object
+              additionalProperties: false
+              description: Parameters for the export task
+              properties:
+                url:
+                  type: string
+                  description: s3 presigned url of the asset to export
+                  example: s3+https://s3.amazonaws.com/livepeer-test/ABC123/filename.mp4
+                ipfs:
+                  type: boolean
```
victorges: This **must** be an object, not a boolean, since it will later contain sub-fields. Like for example custom service, ERC1155 metadata, etc.

packages/api/src/schema/schema.yaml
```diff
@@ -1151,6 +1170,22 @@ components:
...
+                  type: string
+                  description: could be a url or a IPFS CID
+                metadataFilePath:
+                  type: string
+                  description: could be a url or a IPFS CID
+                assetSpec:
+                  type: object
+                  $ref: "#/components/schemas/asset"
+                ipfs:
+                  type: boolean
```
victorges: I'd go with
```suggestion
                ipfs:
                  type: object
                  properties:
                    videoFileCid:
                      type: string
                      description: IPFS CID of the exported video file
                    erc1155MetadataCid:
                      type: string
                      description: IPFS CID of the default metadata exported for the video
```

custom export has no output I guess. Can you think of any?

packages/api/src/task/task.ts
```diff
@@ -84,21 +88,65 @@ export default class TaskScheduler {
...
-              updatedAt: Date.now(),
-            },
-            output: event.output,
-          });
           return true;
         }
       }
-      console.log(`task type unknown: ${event.task.type}`);
+
+      if (allowedTasks.includes(event.task.type)) {
```
victorges: What is this check for? Shouldn't we update every task with its output even if we seemingly don't support it here? e.g. version migration

If we don't understand what is a given event, we should probably `nack` it instead of processing but ignoring all its information. But I think we should be fine just doing a best effort processing and at least always updating the task object.

packages/api/src/task/task.ts
```diff
@@ -84,21 +88,65 @@ export default class TaskScheduler {
...
           });
 
-          await db.task.update(task.id, {
-            status: {
-              phase: "completed",
-              updatedAt: Date.now(),
-            },
-            output: event.output,
-          });
           return true;
```
victorges: Transcode task also generates an output asset. We need a similar logic for it here. (could be related to the confusing `originTaskId` as well)

packages/api/src/task/task.ts
```diff
@@ -84,21 +88,65 @@ export default class TaskScheduler {
...
+      timestamp: Date.now(),
+      task: {
+        id: task.id,
+        type: task.type,
+        snapshot: task,
+      },
+    });
+    await db.task.update(task.id, {
+      status: { phase: "waiting", updatedAt: Date.now() },
+    });
```
victorges: Oh you do already have that here! Please remove the repeated update on every place you call `scheduleTask`.

packages/api/src/task/task.ts
```diff
@@ -84,21 +88,65 @@ export default class TaskScheduler {
...
   }
+
+  async scheduleTask(
+    asset: Asset,
+    type: "import" | "export" | "transcode",
+    params: object
+  ) {
+    let newTask: WithID<Task> = {
+      id: uuid(),
+      name: `asset-upload-${asset.name}-${asset.createdAt}`,
```
victorges: Name is inaccurate for the other task types. Either receive it as an argument to the function here or try to use the task type to build the name instead? 

From another perspective, feels like this name is unnecessary, perhaps something we'll just want to compose on the frontend but doesn't have any meaning/necessity in the backend. WDYT of just removing it?

packages/www/public/sitemap.xml
```diff
@@ -1,117 +1,117 @@
...
-<url><loc>https://livepeer.com/jobs/1466562</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-08T18:27:53.187Z</lastmod></url>
-<url><loc>https://livepeer.com/jobs/1454194</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-08T18:27:53.187Z</lastmod></url>
-<url><loc>https://livepeer.com/jobs/1414584</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-08T18:27:53.187Z</lastmod></url>
-<url><loc>https://livepeer.com/jobs/1412804</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-08T18:27:53.187Z</lastmod></url>
-<url><loc>https://livepeer.com/jobs/1412803</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-08T18:27:53.187Z</lastmod></url>
-<url><loc>https://livepeer.com/jobs/1412799</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-08T18:27:53.187Z</lastmod></url>
+<url><loc>https://livepeer.com</loc><changefreq>daily</changefreq><priority>0.7</priority><lastmod>2022-02-09T14:28:24.457Z</lastmod></url>
```
victorges, line 3: These changes desn't need to be committed. Just an annoying output from the www build.

packages/api/src/schema/schema.yaml
gioelecerati: Yeah I agree, this was just a placeholder until we know the exact data, but I guess it can be a object placeholder for now

packages/api/src/schema/schema.yaml
gioelecerati: Oh you already defined the IPFS  object for the output! Cool, looks great. Nothing comes to mind right not for the custome export output

packages/api/src/task/task.ts
gioelecerati: I wanted to add a "completed" status phase in a single place, making sure the task type is allowed, but now that you mention it the task is already created so it's already an allowed type of task. I remove it!

packages/api/src/task/task.ts
gioelecerati: I agree. Are you thinking about removing the name field or it could be just something like `task-$type-$createdAt`?

PR #895, opened 2022-02-11
VOD transcode task

darkdarkdragon:
**What does this pull request do? Explain your changes. (required)**

**Specific updates (required)**

## -

- **How did you test each of these updates (required)**

**Does this pull request close any open issues?**

**Screenshots (optional):**

**Checklist:**

- [ ] I have read the **CONTRIBUTING** document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

PR #900, opened 2022-02-11
api: Fix some schemas from asset APIs

victorges:
Some nitpicks on the assets and tasks schema.

 - Validate the url field from the schema, instead of just allowing any additional properties
 - Make videoFilePath and metadataFilePath from import task output writeOnly (private)

## -

`yarn test`

Only those discussed on Discord 😆 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -209,7 +209,7 @@ app.get("/", authMiddleware({}), async (req, res) => {
 
   res.status(200);
 
-  if (output.length > 0) {
+  if (output.length > 0 && newCursor) {
     res.links({ next: makeNextHREF(req, newCursor) });
   }
 
@@ -323,9 +323,18 @@ app.post(
       req.config.vodObjectStoreId,
       { name: `asset-upload-${id}`, ...req.body }
     );
+    const uploadedObjectKey = `directUpload/${playbackId}/source`;
     const presignedUrl = await getS3PresignedUrl(
       asset.objectStoreId,
-      `directUpload/${playbackId}/source`
+      uploadedObjectKey
+    );
+    const task = await req.taskScheduler.createTask(
+      "import",
+      {
+        import: { uploadedObjectKey },
+      },
+      null,
+      asset
     );
 
     const b64SignedUrl = encodeURIComponent(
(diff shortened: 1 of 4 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -383,7 +388,8 @@ app.put("/upload/:url", async (req, res) => {
           },
         },
         undefined,
-        asset
+        asset,
+        taskId
```
gioelecerati: We should take care of this as it conflicts with this #895 on [L141](https://github.com/livepeer/livepeer-com/blob/4eb6e2dfa6912fc15c351fb30a5706663f4e3e89/packages/api/src/task/scheduler.ts#L141)?

gioelecerati (commented):
All looking good, just a comment on `forcedTaskId`

packages/api/src/controllers/asset.ts
victorges: cool thanks for noticing that! might rebase on top of it

packages/api/src/schema/schema.yaml
```diff
@@ -1221,11 +1221,11 @@ components:
                       readOnly: true
                       description:
                         URL to access file through HTTP on an IPFS gateway
-                    erc1155MetadataCid:
```
gioelecerati, line 1215: nit: What about always use `CID` in uppercase, so erc1155MetadataCID?
EDIT: **nftMetadataCID**

gioelecerati: * nftMetadataCID

gioelecerati (approved):
LGTM

packages/api/src/schema/schema.yaml
victorges: Hmm are you reviewing an outdated version? Here it is `nftMetadataCid` already

Regarding casing, I'd prefer to keep it consistent with our current schema. In Go, the pattern is to keep acronyms all upper or all lowercase so that would make sense. I'm not sure if there is an officially recommended pattern for node/JSON, but we already have things like `playbackId` (not `ID`), `recording/playbackUrl` on our schemas, so I'd prefer to treat them as regular words and do upper only on the first letter as `Cid`

PR #911, opened 2022-02-14
return orchestrators pending activation for Confluence upgrade

### LINKED ISSUE #908: subgraph middleware return orchestrators pending activation for Confluence upgrade
**What's the problem? (required)** Describe the problem or need in exactly 1
sentence. Ex. "I'm unable to [...]", or "I am confused by [...]"

For the Confluence upgrade, we want to make sure that Os that migrate can immediately start being eligible for work [1].

At the moment, the subgraph middleware can be used by a B using a webhook to discover Os by setting `LP_SUBGRAPH_URL` to the subgraph URL that Os should be discovered from. The GraphQL query currently filters for active Os - see https://github.com/livepeer/livepeer-com/blob/75fca89f54fa36b0b76be344124326f31894eb3e/packages/api/src/middleware/subgraph.ts#L31. 

We could update the filter for the query to return Os that are pending activation as well. This will require fetching the current round as well which I believe we can do by querying the subgraph. Assuming that we have the current round, the updated filter for the query to fetch Os could look something like this:

```
transcoders(where: { activationRound_lte: currentRound + 1, deactivationRound_gt: currentRound })
```

[1] As mentioned in https://github.com/livepeer/go-livepeer/commit/9ab7a8842f2f4c12a0ca9b88cd8ef9b6a1bc1b73#commitcomment-66501975 we may want to revert this behavior after the round after the upgrade, but for now this makes the upgrade transition smoother.

**What's the current behavior? (required)** A clear and concise description of
the current behavior of the DApp or package (or lack thereof).

subgraph middleware currently returns active Os which means that a B using this middleware for discovery immediately after the Confluence upgrade will have an empty list of Os received until the start of the next round.

**Do you have any personal experiences dealing with this problem? (optional)**
Describe your personal experiences related this problem in 1-4 sentences.

**Describe a possible solution you've considered (optional)** A clear and
concise description of what you want to happen.

**Additional context (optional)** Add any other context or screenshots about the
enhancement or feature request here.

adamsoffer:
This PR ensures that Os who migrate can immediately start being eligible for work by updating the subgraph middleware query to return Os that are pending activation as well.

Manually tested the graphql requests.

Closes #908

### DIFF
--- a/packages/api/src/middleware/subgraph.ts
+++ b/packages/api/src/middleware/subgraph.ts
@@ -25,18 +25,37 @@ export default function subgraphMiddleware({
   let cachedResp: Array<OrchestratorNodeAddress> = [];
   let lastCachedRespUpdate = 0;
 
-  const getOrchestrators = async () => {
-    const query = `
-      {
-        transcoders(where: { active: true }) {
-          id
-          serviceURI
-        }
-      }
-    `;
+  const getCurrentRound = async () => {
+    const res = await fetchWithTimeout(subgraphUrl, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Accept: "application/json",
+      },
+      body: JSON.stringify({
+        query: `{ protocol(id: "0") { currentRound { id } } }`,
+      }),
+      timeout: SUBGRAPH_TIMEOUT,
+    });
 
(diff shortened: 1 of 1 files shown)


yondonfu (approved):
LGTM

PR #914, opened 2022-02-15
api/task: Add separate IPFS URL as well as gateway's

victorges:
Adding the raw IPFS URL as well for easy usage to the hackers.


gioelecerati (approved):
LGTM

PR #922, opened 2022-02-20
Disable cors on vod api

gioelecerati:

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -28,6 +28,9 @@ import cors from "cors";
 
 var corsOptions = {
   origin: "*",
+  methods: "GET,HEAD,PUT,PATCH,POST,DELETE",
+  preflightContinue: false,
+  credentials: true,
   optionsSuccessStatus: 200,
 };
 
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -130,6 +130,7 @@ const fieldsMap: FieldsMap = {
 };
 
 app.get("/", authMiddleware({}), async (req, res) => {
+  res.header("Access-Control-Allow-Origin", "*");
```
victorges: You should look into the cors plugin/library that we use somewhere in this repo. Should be in the places where we set up the whole express application like app-router or something like that

gioelecerati: Wouldn't adding the wildcard on the cors definition in `index.ts` add it to all the endpoints and not the VOD ones? I saw it was specifically added on some endpoints (e.g. the stream proxy)

gioelecerati: NVM, added only on vod via lib

iameli: You need access-control-allow-method too

gioelecerati: added as well

iameli:
LGTM! go ahead and merge.

PR #923, opened 2022-02-21
www: Page for minting an NFT from an exported asset

victorges:

Create a page from which to mint NFTs out of an asset that has been exported to IPFS.

New page with all the Metamask integration and all.

## -

`npx @livepeer/video-nft --mint-nft`

Part of #929 but doesn't really close it.

**Screenshots (optional):**

Screen Shot 2022-03-02 at 17 26 07

**Checklist:**


packages/www/pages/transact/eth/index.tsx
```diff
@@ -0,0 +1,454 @@
...
+                      <TextField
+                        size="2"
+                        type="text"
+                        id="contractAddress"
+                        value={
+                          state.contractAddress === defaultContractAddress
+                            ? ""
+                            : state.contractAddress
+                        }
+                        placeholder={`Livepeer Video NFT (${defaultContractAddress})`}
```
gioelecerati: If metamask is not installed in the browser (or connected to unsupported network), this returns undefined. Maybe this could be an empty string if metamask is not present?

packages/www/pages/transact/eth/index.tsx
```diff
@@ -0,0 +1,454 @@
...
+          <Flex
+            css={{
+              alignItems: "center",
+              justifyContent: "center",
+              flexGrow: 1,
+              flexDirection: "column",
+            }}>
+            <AlertDialog open={true}>
+              <AlertDialogContent
+                css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4" }}
```
gioelecerati: Instead of having the header and the footer with an empty page on the background, could this dialog cast a box shadow? Or is there something already in place for this kind of standalone dialogs? 
```
boxShadow: "0 0 100vw 100vw rgb(200 200 200)"
```

Although I guess it will be a separate full page and / or a dialog in the correct place (e.g. Asset Table dashboard) and never a standalone dialog?

packages/www/pages/transact/eth/index.tsx
```diff
@@ -0,0 +1,454 @@
...
+  }
+}
+
+const displayAddr = (str: string) => (
+  <code>
+    {str.slice(0, 5)}…{str.slice(-4)}
+  </code>
+);
+
+async function switchNetwork(
```
gioelecerati: Love it!

packages/www/pages/transact/eth/index.tsx
```diff
@@ -0,0 +1,454 @@
...
+  to: string,
+  tokenUri: string,
+  logger: (log: JSX.Element | string) => void
+) {
+  try {
+    logger("Started mint transaction...");
+    const videoNft = new web3.eth.Contract(videoNftAbi as any, contractAddress);
+    const transaction = {
+      to: contractAddress,
+      gas: 500000,
```
gioelecerati: Maybe it should calculate the estimated gas limit for this contract payable call (I think with web3.eth.estimateGas?)

gioelecerati (commented):
Looking good!

packages/www/pages/transact/eth/index.tsx
victorges: good catch

packages/www/pages/transact/eth/index.tsx
victorges: Yeah my idea is that soon we will extract a dialog component from this and inject it in some other pages like asset. Didn't want to spend tooo much time on the styling there, so the one thing I'm gonna try to do is at least to move the footer down in the page, so it doesn't look that weird

packages/www/pages/transact/eth/index.tsx
victorges: ideally it should, but the estimated gas is apparently really bad and the transactions take so long to execute on Polygon. Although maybe that was only a problem cause I was testing on mainnet and wanted to iterate quickly 🙈 

I'm gonna see what can be done here, thanks for pointing out the function for estimating it (I tried it by not suggesting anything and letting metamask do it)

gioelecerati: Mmm strange. I wonder if it's a behavior related to Polygon. In any case, I would think that letting the user input the gas limit would be better anyway than setting a very high default. Or maybe we could manually calculate it on our function call and hardcode the correct one.

Could possibly be that estimateGas is failing because we are also specifying the gasPrice, I would try to remove also that and retry with the estimateGas.

Another thing, maybe we could try to calculate the gas directly from the contract, like this?

```
const contract = web3.eth.contract(abi).at(address);
const gasEstimate = contract.$myFunctionMethod.estimateGas(arg1, arg2, {from: account});
```

gioelecerati: I think Yondon could help on this matter, or tell us what would be the best thing to do (hardcoded gasLimit or manual user input?)

victorges: > I would think that letting the user input the gas limit would be better anyway than setting a very high default. Or maybe we could manually calculate it on our function call and hardcode the correct one.

Oh, sorry. I thought you were talking about the gas price in the whole conversation. This gas limit has been chosen from the transaction that we run here. It's set to ~2x the gas that is usually spent just in case, since our contract implementation could also change in the future (especially if we do a proper openzeppeling proxy thing so we are able to update it).


The gas price thing that was confusing and the values suggested by MetaMask were bad.

But nevertheless, none of those values are final. The user will still choose on the MetaMask popup how much they want to pay for gas and what is the max gas allowed to be spent. I don't think having another form input for gas here is good since they can already tune it in MetaMask if they want to.

gioelecerati: Oh cool, so the gasLimit is already set based on the contract function! :) nice, then that's great. 

About the gasPrice, yeah, is not always reliable. There could be an "advanced tab" but I don't think it's something we should cover right now, and MetaMask let you customize the gas anyway, if needed, so yeah I agree

PR #925, opened 2022-02-22
api: vod asset Delete API & event

gioelecerati:

Spawn delete event for an Asset on api call

- Delete event
- All children of deleted asset become orphans

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -471,9 +471,36 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
   if (!req.user.admin && req.user.id !== asset.userId) {
     throw new ForbiddenError(`users may only delete their own assets`);
   }
-  await db.asset.delete(id);
-  res.status(204);
-  res.end();
+  await db.asset.markDeleted(id);
+  const childrenAssets = await db.asset.find(
+    { sourceAssetId: id },
+    { useReplica: false }
+  );
+
+  const task = await req.taskScheduler.scheduleTask(
+    "delete",
+    {
+      delete: {
+        assetId: id,
+      },
+    },
+    asset
+  );
+
+  if (!childrenAssets?.length || !childrenAssets[0]?.length) {
+    res.status(200);
+    return res.json({ task });
+  }
+
+  for (var i = 0; i < childrenAssets[0].length; i++) {
+    // remove the sourceAssetId from the children assets making them orphans
+    db.asset.update(childrenAssets[0][i].id, {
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1140,6 +1141,15 @@ components:
               properties:
                 profile:
                   $ref: "#/components/schemas/ffmpeg-profile"
+            delete:
```
gioelecerati, line 1144: Should the delete task also have an output object?

packages/api/src/controllers/asset.ts
```diff
@@ -471,9 +471,36 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
...
+  );
+
+  if (!childrenAssets?.length || !childrenAssets[0]?.length) {
+    res.status(200);
+    return res.json({ task });
+  }
+
+  for (var i = 0; i < childrenAssets[0].length; i++) {
+    // remove the sourceAssetId from the children assets making them orphans
+    db.asset.update(childrenAssets[0][i].id, {
```
gioelecerati, line 497: Should be made in a single query

packages/api/src/controllers/asset.ts
```diff
@@ -471,9 +471,36 @@ app.delete("/:id", authMiddleware({}), async (req, res) => {
...
+    },
+    asset
+  );
+
+  if (!childrenAssets?.length || !childrenAssets[0]?.length) {
+    res.status(200);
+    return res.json({ task });
+  }
+
+  for (var i = 0; i < childrenAssets[0].length; i++) {
```
victorges, line 495: **nit:**
```suggestion
  for (var i in childrenAssets[0]) {
```
to keep `i` as the index, or
```suggestion
  for (var child of childrenAssets[0]) {
```
to already get the asset in the `child` var

packages/api/src/controllers/asset.ts
victorges: Yeah that's one pretty doable actually. We could first get the id of the children, then update all the children + the source on a single query and avoid any dangling errors. Might not be super trivial to implement it right now tho, so maybe just a `// TODO` here should be ok

packages/api/src/schema/schema.yaml
victorges: Hmm I don't think so

victorges (approved):
lgtm

pglowacky:
@gioelecerati what's the status of this?

PR #928, opened 2022-02-22
Add docs link to dashboard header

### LINKED ISSUE #917: Cannot Navigate Between "https://livepeer.com/docs" and "https://livepeer.com/dashboard"
Currently, a developer cannot easily navigate between the dashboard and the docs.  The only way to go from dashboard to docs is through the "Get started tutorial" button in the welcome page.

We should have top-level navs that helps developers easily go between these two important sections of Livepeer Video Services.

adamsoffer:
Closes #917

Screen Shot 2022-02-22 at 9 21 47 PM

### DIFF
--- a/packages/www/components/Dashboard/Header/index.tsx
+++ b/packages/www/components/Dashboard/Header/index.tsx
@@ -117,10 +117,12 @@ const Header = ({ breadcrumbs = [] }) => {
               mr: "$5",
               background: "transparent",
               appearance: "none",
-              WebkitAppearance: "none",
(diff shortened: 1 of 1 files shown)


pglowacky:
Does it make more sense to have the documentation open in a new tab? I'm assuming that the user is trying to do something in the dashboard, can't figure it out or wants to see the API, I'd assume they'd still want to keep their place. 

Not married to that, up to you!

adamsoffer:
@pglowacky ah yes good call!

PR #942, opened 2022-02-28
vod: Add clip parameters to vod transcode task

gioelecerati:

Adds clip (trim) ffmpeg parameters to the transcode task.
To evaluate if ffmpeg-trim and ffmpeg-profile can belong to the same object, and if it's possible to create a clip in keyframes without re-encode the asset.

- Added parameters

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -80,6 +80,26 @@ components:
             - hevc
             - vp8
             - vp9
+    ffmpeg-trim:
+      type: object
+      description: ffmpeg cut/trim parameters
+      additionalProperties: false
+      required:
+        - start
+      properties:
+        start:
+          type: number
+          description: start time in seconds
+          minimum: 0
+        time:
+          type: number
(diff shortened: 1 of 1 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1070,6 +1090,8 @@ components:
           example: filename.mp4
         profile:
           $ref: "#/components/schemas/ffmpeg-profile"
+        clip:
```
gioelecerati, line 1093: To evaluate if ffmpeg-trim and ffmpeg-profile can belong to the same object, and if it's possible to create a clip on keyframes without re-encode the asset (currently ffmpeg-profile is required, could be optional)

victorges: I think either here or in `task-runner`, we could default the `profile` to something like "source" (`-c copy`) in case its not present 🤔

victorges (approved):
lgtm

PR #946, opened 2022-03-01
www: Docs - added video on demand api reference 

gioelecerati:

Add video on demand api reference under docs/api-reference/vod

- Overview
- List Assets
- Import
- Upload
- Export
- List Tasks
- Mint video NFT

**Screenshots (optional):**

**Checklist:**


Shih-Yu (commented):
Everything looks fine from my perspective

packages/www/docs/api-reference/vod/export.mdx
```diff
@@ -0,0 +1,83 @@
...
+    },
+    {
+      parameter: <code>custom.url</code>,
+      type: <code>string</code>,
+      description: "URL where to export the `Asset`",
+    },
+    {
+      parameter: <code>custom.method</code>,
+      type: <code>string</code>,
+      description: "Method to use on the export request",
```
victorges: Might want to specify that this is optional too, that by default we will use `PUT`: https://github.com/livepeer/task-runner/blob/64d0d1ae6a2ec4a2cd5b811ba7dd9fde2fb56537/task/export.go#L73

```suggestion
      description: "Alternative method to use on the export request. Will use PUT if not specified (Optional)",
```

packages/www/docs/api-reference/vod/export.mdx
```diff
@@ -0,0 +1,83 @@
...
+    },
+    {
+      parameter: <code>custom.headers</code>,
+      type: <code>object</code>,
+      description: "Headers to add to the export request",
```
victorges: This is also optional. Maybe:
```suggestion
      description: "Additional headers to add to the export request (Optional)",
```

packages/www/docs/api-reference/vod/import.mdx
```diff
@@ -0,0 +1,48 @@
...
+      description:
+        "The name of the `Asset` containing a custom human-readable description or title.",
+    },
+  ]}
+/>
+
+## How do I import a Live stream into my Assets?
+
+Livepeer live stream recordings are automatically imported as `Asset` objects.
+All you need to do is enable recording on your streams.
```
victorges, line 44: Btw, not something that should be fixed on this PR, but I'd say that today it is impossible to use the exported assets programmatically. I think we need to:
 - Create a relationship between the session and the exported asset, so one doesn't have to look for the asset by name. Probably a property in the session with the ID of the recording asset.
 - Ideally, also have a webhook when this export is done.
 - Also bonus would be that option to specify a custom Object Store for the exported asset. Some clients and/or hackers have kinda been wanting that already.

packages/www/docs/api-reference/vod/import.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+After the stream session ends and the recording is processed (`~6min`) you
```
victorges: I think we could be more explicit that it is a hardcoded delay for the recording, not just some processing time. WDYT?
```suggestion
When the stream session ends, the platform waits `6min` to give a chance 
for the user to reconnect on the same recording session. After that delay, the
recording is processed and you should have a new asset in your account
which you can retrieve from the list assets API.
```
(also removed the "above" from "assets AP", since each API is displayed on a separate page)

packages/www/docs/api-reference/vod/mint-nft.mdx
```diff
@@ -0,0 +1,162 @@
...
+description: Livepeer.com API Reference | Mint
+metaTitle: Mint - Livepeer.com
+metaDescription: Livepeer.com API Reference | Mint
+---
+
+# How do I use Livepeer to mint a video NFT?
+
+You can use
+[this script](https://gist.github.com/victorges/3369cc04c0002a63b2fab33079d8827e)
+(`bash`) to have a local file uploaded to Livepeer and then export it to IPFS.
```
victorges, line 12: We should update this to use the `@livepeer/video-nft` script instead. I think we can merge this is as is though, and I can make a follow-up PR with the updated script and all

victorges, line 12: We should update this to use the `@livepeer/video-nft` script instead. I think we can merge this is as is though, and I can make a follow-up PR with the updated script and all

victorges (approved):
LGTM!

packages/www/docs/api-reference/vod/mint-nft.mdx
victorges: Alright, I updated [the Mint NFT docs on Notion](https://livepeer.notion.site/Tutorial-How-to-Mint-a-Video-NFT-368a04d00f0f4ca9bb0e59b98ec04c40) with the latest info on how to use the `video-nft` script. It actually already referenced the npm script, this version here seems to have come from a really outdated version of it.

Do you think you could update the docs here with the content in there now? Or I can do on a follow-up PR as I proposed.

PR #947, opened 2022-03-02
api: allow suspending users from admin token

iameli:
This from an API key is fine. Makes automation easier.


gioelecerati (approved):
LGTM

PR #948, opened 2022-03-02
www: make stripe optional

iameli:
Lots of errors still but this prevents the biggest issues from blocking box development

### DIFF
--- a/packages/www/hooks/use-api.tsx
+++ b/packages/www/hooks/use-api.tsx
@@ -14,6 +14,7 @@ import {
 import qs from "qs";
 import { isStaging, isDevelopment, HttpError } from "../lib/utils";
 import Head from "next/head";
+import { products } from "@livepeer.com/api/src/config";
 
 /**
  * Primary React API client. Definitely a "first pass". Should be replaced with some
@@ -30,6 +31,7 @@ type ApiState = {
   user?: User;
   token?: string;
   userRefresh?: number;
+  noStripe?: boolean;
 };
 
 export interface UsageData {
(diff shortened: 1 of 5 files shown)


packages/www/hooks/use-api.tsx
```diff
@@ -313,6 +317,14 @@ const makeContext = (state: ApiState, setState) => {
       return [res, user as User | ApiError];
     },
 
+    // Get current Stripe product, allowing for development users that don't have any
+    getUserProduct(user: User) {
+      if (user.stripeProductId) {
+        return products[user.stripeProductId];
+      }
+      return Object.values(products)[0];
```
victorges: I think we should not have this logic in production/if stripe is configured. Also, maybe it's better to access some specific product not to rely on object keys ordering. WDYT of this?
```suggestion
      if (hasStripe) {
        return products[user.stripeProductId];
      }
      return products[user.stripeProductId || "prod_0"];
```

packages/www/lib/utils/index.tsx
```diff
@@ -145,6 +145,10 @@ export function blocksToText(blocks, opts = {}) {
  */
 let stripePromise: Promise<Stripe | null>;
 export const getStripe = () => {
+  const key = process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY;
+  if (!key) {
+    return null;
```
victorges: **nit:** I think this should be OK for JS craziness and wont cause a NRE on someone `await`ing on a null promise, but WDYT of also returning a promise here?
```suggestion
    return Promise.resolve(null);
```

victorges (approved):
lgtm

PR #953, opened 2022-03-07
Added multistream destination bitrate on stream health

### LINKED ISSUE #685: Include multistream destination bitrate on Stream Health dashboard
In our original spec of Stream Health, we discussed including multistream destination bitrate in the Session Ingest Rate graph. Something like this:

With the recent changes in the stream health API, this is now possible!

Completing this will require input from @victorges for technical implementation as well as @adamsoffer for design.

FullStackTiger:

## - Include multistream destination bitrate on Stream Health dashboard

- Fixes #685 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/Chart/index.tsx
+++ b/packages/www/components/Dashboard/Chart/index.tsx
@@ -100,7 +100,7 @@ const Chart = ({ data, multiData }) => {
           <YAxis domain={[0, 1600]} />
           <CartesianGrid vertical={false} />
           <Tooltip content={<CustomTooltip />} />
-          <Legend />
+          <Legend wrapperStyle={{ fontSize: "10px" }} />
           <Line
             data={data}
             cursor="pointer"
--- a/packages/www/components/Dashboard/Chart/index.tsx
+++ b/packages/www/components/Dashboard/Chart/index.tsx
@@ -1,7 +1,7 @@
 import { Box, Text } from "@livepeer.com/design-system";
 import {
-  AreaChart,
-  Area,
+  LineChart,
+  Line,
   CartesianGrid,
   XAxis,
   YAxis,
@@ -9,6 +9,21 @@ import {
   ResponsiveContainer,
   Legend,
 } from "recharts";
+
+const getMultistreamColor = (number) => {
+  switch (number) {
+    case 1:
+      return "#aa99ec";
+    case 2:
+      return "#8da4ef";
+    case 3:
+      return "#2da4ef";
+    case 4:
+      return "#7da41f";
+    default:
+      return "#7d241f";
+  }
+};
 
 const CustomTooltip = ({ active, payload }: any) => {
   if (active && payload) {
@@ -24,7 +39,11 @@ const CustomTooltip = ({ active, payload }: any) => {
             fontSize: "12px",
             color: "white",
           }}>
-          Rate: <b>{payload[0].value} kbps</b>
+          {payload.map((item) => (
+            <div>
+              {item.dataKey}: <b>{item.value} kbps</b>
+            </div>
+          ))}
         </Text>
       </Box>
     );
@@ -32,8 +51,8 @@ const CustomTooltip = ({ active, payload }: any) => {
 
   return null;
 };
-
-const Chart = ({ data }) => {
+const Chart = ({ data, multiData }) => {
+  const multistreamNames = multiData[0] && Object.keys(multiData[0]);
   return (
     <Box
       css={{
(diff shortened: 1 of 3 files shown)


packages/www/components/Dashboard/Chart/index.tsx
```diff
@@ -24,16 +39,29 @@ const CustomTooltip = ({ active, payload }: any) => {
...
 };
-
-const Chart = ({ data }) => {
+const Chart = ({
+  data,
+  multiData,
+}: {
+  data: Array<{ name: number; "Session bitrate": number }>;
+  multiData?: Array<{
+    [kbps: string]: number;
```
victorges: Isn't the key the multistream name here?
```suggestion
    [name: string]: number;
```

packages/www/components/Dashboard/Chart/index.tsx
```diff
@@ -24,16 +39,29 @@ const CustomTooltip = ({ active, payload }: any) => {
...
 
   return null;
 };
-
-const Chart = ({ data }) => {
+const Chart = ({
+  data,
+  multiData,
+}: {
+  data: Array<{ name: number; "Session bitrate": number }>;
```
victorges, line 58: It's weird that `name` is actually a `number`. Is it actually an `index` instead? And if so, should this just be an array of bitrates?

packages/www/components/Dashboard/Chart/index.tsx
```diff
@@ -65,26 +93,45 @@ const Chart = ({ data }) => {
...
-          <XAxis dataKey="name" />
+        <LineChart>
+          <XAxis
+            type="number"
+            dataKey="name"
+            domain={[
+              data[0]?.name,
+              data[data.length - 1].name === 0
+                ? 10
+                : data[data.length - 1].name,
```
victorges: Is this the same as checking if there is more than 1 element in the `data` array? If so, maybe this is clearer:
```suggestion
              data.length > 1
                ? 10
                : data[data.length - 1].name,
```
Additionally, what does this `10` mean?

packages/www/components/Dashboard/StreamDetails/StreamHealthTab.tsx
```diff
@@ -76,6 +95,24 @@ const StreamHealthTab = ({ stream, streamHealth, invalidateStream }) => {
...
+        const { dimensions, last } = item;
+        if (i === 0) {
+          data.name = Math.round((last[0] - startTime) / 1000);
+        }
+        data[`${dimensions.targetName} (${dimensions.targetProfile})`] =
+          Math.round(last[1] / 1000);
+      });
+
+      const newData = [...multiDataChart, data].slice(-(maxItems + 10));
+      setMultiDataChart(newData);
```
victorges: I believe it's more appropriate to send a function to the `setMultiDataChart` that receives the current value and returns the new value. This will avoid any closure hell that could happen here. Like this: https://github.com/livepeer/livepeer-com/blob/59d54adb8c5265b1236fb502429ae75ff4277a1f/packages/www/components/Dashboard/Logger/index.tsx#L152

packages/www/components/Dashboard/StreamDetails/StreamHealthTab.tsx
```diff
@@ -43,13 +59,16 @@ const StreamHealthTab = ({ stream, streamHealth, invalidateStream }) => {
         return;
       }
       const newInfo = rinfo as StreamInfo;
+
       setDataChart((prev) => {
         const lastItem = prev[prev.length - 1];
         return [
           ...prev,
           {
             name: lastItem ? lastItem.name + ingestInterval / 1000 : 0,
```
victorges, line 67: Oh I see, so you are actually not creating these specs on this PR right. Don't mind refactoring it since it's not a regression, but it'd be great to improve these interfaces a bit anyway if you're open to it.

victorges (approved):
LGTM!

Only caveat is that I tried to test this but no multistream metrics showed up. I checked on the stream health though and apparently the multistream bitrate metrics were not being included, so it's not a problem with the frontend. 

Were you able to test this with the current API? Or did you possibly get these errors as well?

packages/www/components/Dashboard/Chart/index.tsx
FullStackTiger: This is the end value of X-Axis when loads init chart data
Screenshot

FullStackTiger:
When I create this PR, I got multistream metrics from the current backend API and tested it, but it can't get the data from the backend API now.

PR #956, opened 2022-03-08
api: Allow user to retrieve its informations

gioelecerati:

Some api endpoints require the `userId`.
There is no way for the user to retrieve its information using the api token, so a user would not be able to get its `userId` and call the needed endpoint.  
  
This little pr modifies the `get user` endpoint returning the user info if the user is not admin, keeping the list response if admin.

- Added non-admin retrieve of user info
- Changed tests to allow non-admin user to retrieve personal information

Yarn test and dev frontend check

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/user.test.ts
+++ b/packages/api/src/controllers/user.test.ts
@@ -221,7 +221,9 @@ describe("controllers/user", () => {
       }
 
       let res = await client.get("/user");
-      expect(res.status).toBe(403);
+      let resJson = await res.json();
+      let condition = Array.isArray(resJson);
+      expect(condition).toBe(false);
 
       // should not be able to make users admin with non-admin user
       const resAdminChange = await client.post(`/user/make-admin/`, {
@@ -388,11 +390,14 @@ describe("controllers/user", () => {
       client.apiKey = nonAdminApiKey;
       let res = await client.get("/user");
       let resJson = await res.json();
+      let condition = Array.isArray(resJson);
       expect(res.status).toBe(403);
+      expect(condition).toBe(false);
       expect(resJson.errors[0]).toBe(
         `useremail ${nonAdminUser.email} has not been verified. Please check your inbox for verification email.`
       );
 
+      // should return nonverified error
       client.apiKey = adminApiKey;
       res = await client.get("/user");
       resJson = await res.json();
@@ -404,29 +409,26 @@ describe("controllers/user", () => {
       // adding emailValid true to user
       await db.user.update(nonAdminUser.id, { emailValid: true });
 
-      // should return admin priviledges error
       client.apiKey = nonAdminApiKey;
       res = await client.get("/user");
       resJson = await res.json();
-      expect(res.status).toBe(403);
-      expect(resJson.errors[0]).toBe("user does not have admin priviledges");
+      condition = Array.isArray(resJson);
+      expect(res.status).toBe(200);
+      expect(condition).toBe(false);
 
       // adding emailValid true to admin user
       await db.user.update(adminUser.id, { emailValid: true });
 
-      // only jwt auth should be allowed access to this list API
       client.apiKey = adminApiKey;
       res = await client.get("/user");
       resJson = await res.json();
-      expect(res.status).toBe(403);
-      expect(resJson.errors[0]).toBe("user does not have admin priviledges");
+      expect(res.status).toBe(200);
 
       client.apiKey = undefined;
       client.basicAuth = `${adminUser.id}:${adminApiKey}`;
       res = await client.get("/user");
       resJson = await res.json();
-      expect(res.status).toBe(403);
-      expect(resJson.errors[0]).toBe("user does not have admin priviledges");
+      expect(res.status).toBe(200);
     });
 
     it("should return verified user", async () => {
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/user.test.ts
```diff
@@ -221,7 +221,9 @@ describe("controllers/user", () => {
       }
 
       let res = await client.get("/user");
-      expect(res.status).toBe(403);
+      let resJson = await res.json();
+      let condition = Array.isArray(resJson);
+      expect(condition).toBe(false);
```
victorges: I think `condition` is not a helpful name for a helper variable, as it only creates an indirection that we then need to go back to the original place where the condition was measured to know what it is. I'd either rename this to something clear like `isArray` or `arrayRes` or anything like that, or just skip the helper variable altogether which seems ok imo
```suggestion
      expect(Array.isArray(resJson)).toBe(false);
```

packages/api/src/controllers/user.test.ts
```diff
@@ -404,29 +409,26 @@ describe("controllers/user", () => {
...
-      expect(resJson.errors[0]).toBe("user does not have admin priviledges");
+      expect(res.status).toBe(200);
 
       client.apiKey = undefined;
       client.basicAuth = `${adminUser.id}:${adminApiKey}`;
       res = await client.get("/user");
       resJson = await res.json();
-      expect(res.status).toBe(403);
-      expect(resJson.errors[0]).toBe("user does not have admin priviledges");
+      expect(res.status).toBe(200);
```
victorges: Should probably check if the response is the own user JSON. Maybe check only the ID?

packages/api/src/controllers/user.ts
```diff
@@ -131,7 +131,12 @@ const fieldsMap: FieldsMap = {
   stripeProductId: `data->>'stripeProductId'`,
 };
 
-app.get("/", authMiddleware({ admin: true }), async (req, res) => {
+app.get("/", authMiddleware({}), async (req, res) => {
+  if (req.user.admin !== true) {
+    const user = await db.user.get(req.user.id);
+    res.status(200);
+    return res.json(cleanUserFields(user, req.user.admin));
+  }
```
victorges: Hm, I'm not so sure about this since it will create a completely different schema for the response depending on a non-functional parameter sent to the API which is the authorization token. Think about how clients would handle this for example, where a `getUser()` function or something like that might return completely different results depending on the type of token you've created the client with.

One thing that we could do is move this "list" API to something like `/api/user/list` for example, and then have the "get myself" API as the only thing on `/api/user`. The only component that uses this API is our own (admin) dashboard anyway, so it should be pretty chill to update it to call the new API (on this same repository, on `www` package).

Alternatively, we can also create this new API under a `/api/user/me` or `/api/user/self` or something like that. WDYT?

packages/api/src/controllers/user.ts
```diff
@@ -131,7 +131,12 @@ const fieldsMap: FieldsMap = {
   stripeProductId: `data->>'stripeProductId'`,
 };
 
-app.get("/", authMiddleware({ admin: true }), async (req, res) => {
+app.get("/", authMiddleware({}), async (req, res) => {
```
victorges: It would be awesome to not only allow users to find out their own `userId`s, but also to stop requiring them to know it in the first place. WDYT of going through the APIs that require an `userId` querystring parameter and making it an admin-only feature for querying other users instead? And then for an admin to list all streams (for example) they would actually need to "opt-in" for "all users" instead, like with an `allUsers=true` querystring instead. WDYT?

victorges (commented):
some discussions

packages/api/src/controllers/user.ts
figintern: Victor put it really well, my main concern here was the early return, outside of user experience I think its better to have a single return in the API handler with a consistent response for readability reasons

Moving the API here also makes sense

packages/api/src/controllers/user.ts
figintern: Hopefully I'm fully seeing eye-to-eye with the proposal here and don't have a misconception of our existing API - I am in favour of separating out the admin and user API's into separate paths, with Admin endpoints requiring userId or userId[], and end-user endpoints acting on the userId derived from auth header, this doesn't really follow REST convention for retrieving of resources but makes for a friendly API

packages/api/src/controllers/user.ts
gioelecerati: I moved the api to a separate "/me" endpoint and rollback the user list api to its status. I agree it's much better

victorges (approved):
LGTM!

figintern (approved):
🚢

packages/api/src/controllers/user.test.ts
```diff
@@ -383,6 +383,12 @@ describe("controllers/user", () => {
       client.apiKey = uuid();
     });
 
+    it("should return personal user info", async () => {
+      client.apiKey = nonAdminApiKey;
+      let res = await client.get("/user/me");
+      expect(res.status).toBe(200);
```
iameli, line 391: Probably should check that a field of the user itself matches

PR #960, opened 2022-03-11
www: Disable contract field when not on right network

victorges:
It can be confusing since we were showing an empty placeholder
when the network was wrong (no default contract address).

With this, we disable the field and also show an unsupported network
placeholder.


gioelecerati (approved):
LGTM

PR #962, opened 2022-03-13
www: Add support for polygon testnet on mint-nft

victorges:


packages/www/pages/mint-nft/index.tsx
```diff
@@ -41,23 +41,26 @@ const networks = {
         "https://cloudflare-ipfs.com/ipfs/bafkreiduv5pzw233clfjuahv5lkq2xvjomapou7yarik2lynu3bjm2xki4",
       ],
     },
-    defaultContract: "0x69C53E7b8c41bF436EF5a2D81DB759Dc8bD83b5F", // TODO: Final address here
+    defasultContract: "0x69C53E7b8c41bF436EF5a2D81DB759Dc8bD83b5F",
```
gioelecerati: `defasultContract` instead of `defaultContract` here :D

victorges: Oh that's weird I think I added it on the final attempt to save lol

PR #972, opened 2022-03-22
api/queue: Fix delayed messages logic

victorges:

In #800 I actually broke the logic of the API for delaying messages.

This was caused by this behavior of RabbitMQ: 
> Only when expired messages reach the head of a queue will they actually be discarded (or dead-lettered)

> When setting per-message TTL expired messages can queue up behind non-expired ones until the latter are consumed or expired. Hence resources used by such expired messages will not be freed, and they will be counted in queue statistics (e.g. the number of messages in the queue).

That means that the message with the highest TTL in the queue would block any other messages from being
processed (expired and then deadlettered/rescheduled) until that highest TTL message in front of the queue
expired. Since we have a maximum webhook delay of 1 hour, this means that messages could be delayed for
up to 1 hour even if they had been scheduled for a few seconds after the publishing.

This is especially relevant for recordings, since those are processed through a RabbitMQ message delayed for
5 minutes. If at the same time/region there was a webhook delayed for more than that, we could easily observe
delays in the stream recording processing. We actually had gotten reports from users about that but had no idea
how it could occur. From investigating the event surges in MDW I realized what was going on.

This reverts #800 and also tries to make some improvements on the previous logic, by not creating a separate
queue for each routing key but only for each expiration time. Details better explained in the commit order and
messages.

## -

For now `yarn test` but I intend on testing it running locally and/or in staging.

Last step to close https://github.com/livepeer/livepeer-infra/issues/808

**Checklist:**

### DIFF
--- a/packages/api/src/store/queue.ts
+++ b/packages/api/src/store/queue.ts
@@ -17,6 +17,8 @@ const QUEUES = {
   delayed: "webhook_delayed_queue",
   task: "task_results_queue",
 } as const;
+const delayedWebhookQueue = (routingKey: string, delaySec: number) =>
+  `delayed_webhook_${routingKey}_${delaySec}s`;
 
 type QueueName = keyof typeof QUEUES;
 type ExchangeName = keyof typeof EXCHANGES;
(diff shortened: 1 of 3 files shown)


packages/api/src/store/queue.ts
```diff
@@ -6,17 +6,17 @@ import { EventKey } from "./webhook-table";
...
   events: "webhook_events_queue_v1",
   webhooks: "webhook_cannon_single_url_v1",
-  events_old: "webhook_events_queue",
-  webhooks_old: "webhook_cannon_single_url",
-  delayed: "webhook_delayed_queue",
   task: "task_results_queue",
+  delayed_old: "webhook_delayed_queue",
 } as const;
+const delayedWebhookQueue = (delaySec: number) =>
+  `delayed_webhook_${delaySec}s`;
```
iameli:
```suggestion
const delayedWebhookQueue = (delayMs: number) =>
  `delayed_webhook_${delayMs}ms`;
```

packages/api/src/store/queue.ts
```diff
@@ -240,15 +217,52 @@ export class RabbitQueue implements Queue {
...
     msg: messages.Any,
     delay: number
   ): Promise<void> {
-    console.log(`emitting delayed message: delay=${delay / 1000}s msg=`, msg);
-    await this.channel.publish(EXCHANGES.delayed, routingKey, msg, {
-      persistent: true,
-      expiration: delay,
-    });
+    const delaySec = delay / 1000;
+    const delayedQueueName = delayedWebhookQueue(delaySec);
```
iameli:
```suggestion
    const delayedQueueName = delayedWebhookQueue(delay);
```

iameli (approved):

PR #976, opened 2022-03-23
api: Replace web3 with ethers in /mint-nft

### LINKED ISSUE #961: Replace web3 with ethers
https://github.com/livepeer/livepeer-com/blob/ee25c5c00cb628f2cf337ddac89d1d12bc1a022e/packages/www/pages/mint-nft/index.tsx#L8

victorges:

This replaces the `web3` library with the superior `ethers`, as suggested by @adamsoffer in
the linked issue below (thanks!). Just from the code I can already agree how it provides a better
interface for interacting with the blockchain. The event subscription API also works, which just
didn't on web3 and I had to resort to that historic events polling.

 - Installed `ethers`
 - Removed imports of web3
 - Started translating all the code

## -

Access `/mint-nft` and try to mint an NFT, ensure logs are good and provides the OpenSea link. 
Also ensure that network changes are handled properly.

Closes #961 

**Checklist:**

### DIFF
--- a/packages/www/pages/mint-nft/index.tsx
+++ b/packages/www/pages/mint-nft/index.tsx
@@ -3,9 +3,7 @@ import { useMetaMask } from "metamask-react";
 import { Container } from "next/app";
 import Link from "next/link";
 import { useCallback, useMemo, useState } from "react";
-import { AbstractProvider as MetaMask, TransactionReceipt } from "web3-core";
-import { Contract } from "web3-eth-contract";
-import Web3 from "web3";
+import { ethers } from "ethers";
 
 import Guides from "components/Marketing/Guides";
 import Spinner from "components/Dashboard/Spinner";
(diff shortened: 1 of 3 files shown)


packages/www/pages/mint-nft/index.tsx
```diff
@@ -62,89 +60,55 @@ const networks = {
...
+        videoNft.off(filter, handler);
+        resolve(tokenId.toNumber());
+      };
+      cancelTokenId = () => {
+        videoNft.off(filter, handler);
+        resolve(null);
+      };
+      const filter = videoNft.filters.Mint(selfAddr, selfAddr, null, null);
+      videoNft.once(filter, handler);
+    });
```
victorges: @adamsoffer so even though this is so much better than the polling of historic events, I still don't like this code. Seems very much like a hack on top of an API meant for streaming but not for getting a single event. 

I could not add a filter by `tokenUri` and just use `once` instead of `on`/`off` since the `tokenUri` field is un-indexable. I theoretically could make the field indexable in the smart contract but that would seem like a overfit into a bad interface or bad code here.

The cancel/timeout logic also seems pretty hacky, with the leak of the `resolve` function of the promise which we set below. 

All ideas welcome!

adamsoffer: Gotcha - yeah that field would have to be indexable. I'm cool with this hack for now if you are.

adamsoffer (approved):
Much cleaner! ethers FTW 🚢

packages/www/pages/mint-nft/index.tsx
victorges: Just found a much cleaner way to get the `Mint` event from the transaction receipt! ethers ❤️ 

Cominted here: https://github.com/livepeer/livepeer-com/pull/976/commits/540ef4279e6030fb36c20ebad28966071713129d

PR #977, opened 2022-03-23
api: Fix RabbitMQ tests flakiness

victorges:
We have a lot of tests that fail because one event
leaks from one test to another. With this change, we
create a new vhost in RabbitMQ for every test suite
run, and additionally also create one vhost for every
test run in the specific queue test, which tests the
delayed events and thus could have more leaks.

This will hopefully make them more independent and also
not affect the local development environments (e.g. the box)
whenever we run tests on localhost.


iameli (approved):
LGTM

PR #979, opened 2022-03-25
api: Change user session timeout to 1 min

### LINKED ISSUE #949: Reduce recording wait time to 10 seconds
Currently, there's a 5 minute delay from the completion of a stream to when Playback URLs / MP4 download links are ready. 

We've heard from users, especially those who are building FE applications with Livepeer, that the wait time is too long and breaks the user experience. 

If this is an easy change, let's change it from 5 minutes --> 10 seconds.

gioelecerati:

Changed user session timeout to from 5 to 1 minute

Fixes #949 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -39,7 +39,7 @@ type Profile = DBStream["profiles"][number];
 type MultistreamOptions = DBStream["multistream"];
 type MultistreamTargetRef = MultistreamOptions["targets"][number];
 
-export const USER_SESSION_TIMEOUT = 5 * 60 * 1000; // 5 min
+export const USER_SESSION_TIMEOUT = 60 * 1000; // 1 min
 const HTTP_PUSH_TIMEOUT = 60 * 1000; // value in the go-livepeer codebase
 const ACTIVE_TIMEOUT = 90 * 1000;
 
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -323,7 +323,7 @@ function setRecordingStatus(
   if (session.record && session.recordObjectStoreId && session.lastSeen > 0) {
     const isReady = session.lastSeen > 0 && session.lastSeen < olderThen;
     session.recordingStatus = isReady ? "ready" : "waiting";
-    if (isReady || (req.user.admin && forceUrl)) {
```
figintern, line 326: How come this `req.user.admin` check is removed?

victorges (approved):
:shipit:

packages/api/src/controllers/stream.ts
victorges: So I think the context is this: https://github.com/livepeer/livepeer-com/issues/949#issuecomment-1064580349

Removing that filter, we keep it a safer threshold for users to recover their sessions, while still allow other users to force getting the recording URL through that query-string. I've linked the code way too many times to users so they can copy the way we create the URLs, might as well expose the `forceUrl` option to them in case they know what they are doing.

PR #985, opened 2022-04-02
api: Allow hackers to create dev API keys with CORS

### LINKED ISSUE #996: Allow direct cross-origin uploads
### Problem:
Livepeer.com API currently applies restricted CORS policy in its **/api/asset/upload**, Therefore, it is not possible to upload from websites platforms other than livepeer or localhost:3000.
Thus forcing devs to create and maintain their own backend structure to get the video from the client and send it to the livepeer api, acting as a bridge, having unnecessary costs and extra time of  execution.

### Current Behavior (bad):

1. User upload all the content to my Backend 
2. My Backend receive the content 
3. My Backend request upload token to livepeer
4. My Backend uploads to livepeer using https://origin.livepeer.com/api/asset/upload/$TOKEN

### Solution / Proposal.
Remove CORS restrictions from `/api/asset/upload`, allowing content to be uploaded from any domain.

So the process can be: 
1. Users request upload url from my Backend 
5. My Backend request upload token from livepeer and returns to user
6. User uploads directly to livepeer url: https://origin.livepeer.com/api/asset/upload/$TOKEN

### Advantages

- By enabling users to upload files to LivePeer, this **serverless** pattern moves the network load away from integrations. This can make our application much more scalable, and capable of handling spiky traffic.

- Such a change does not create any conflict with current integrations. Anyone who wants to continue using the backend throughout the entire process can.

- Since **/api/asset/request-upload** already creates the unique token, the process remains fully secure, there is no need to expose the api-id and there is no way to abuse it.

- The Authorization header is not currently needed in the /api/asset/upload currently, so just edit the CORS solves the problem.

### Example

This is how direct-uploads to **AWS S3** works, for example: You can read more here:

https://aws.amazon.com/pt/blogs/compute/uploading-to-amazon-s3-directly-from-a-web-or-mobile-application/

https://devcenter.heroku.com/articles/s3#uploading-files-from-a-heroku-app

victorges:

This creates a hopefully complete solution for all the CORS problems we have been having with hackers.
During hackathons, the hackers very frequently don't have the time nor the expertise to create backends
for their applications. In those cases, they frequently encounter issues when deploying their applications
for the first time to a hosted platform, which is something they normally do in the last minute of hackathons
and then render some broken demos and frustration on hackers and our side with the CORS FAQs.

This will also reduce the complexity for getting started with Livepeer, by not requiring hackers to run some
[dummy proxies like this](https://github.com/victorges/livepeer-web-api-proxy) with in the end just expose
their API/accounts anyway...

I also aimed for an experience in the dashboard that will not only allow them to create API keys that are 
enabled for CORS usage, but also provide some helpful/educational text so they can understand what they
are doing when they enable an API key for CORS access.

Finally, to reduce the impact of hackers enabling CORS on their keys and just forgetting about it until their
app goes live, I've also limited the power that an API key has by default if it has CORS enabled (but if they
(hopefully) know what they''re doing they can also create it with full access). This will make sure that the
API key won't have powers like listing all streams/assets that exist in the account nor deleting resources
they don't already know the ID for. I tried only allowing access to creation APIs and APIs that require the
caller to know at least one resource ID, like an asset or a stream (meaning they probably got that ID from
somewhere else like a stream that they just created or saved in cookies).

Made sure that all APIs called by the Video NFT SDK are also allowed for those restricted keys.

Implementing this was a little complex since it require some structural changes in our authorization logic. 
To allow changing the CORS allow list depending on the API key, we needed to parse and fetch the API key
before the CORS middleware. Also didn't want to move CORS all the way to our API handlers, so 
the first step was to:
 - Split authorization middleware in 2: an `authenticator` (global), which parses/validates the credentials, and an
 `authorizer`, which checks if those credentials give access to the currently called API.

Apart from that refactor:
 - Created param in the API to give them CORS access to specified allowed origins;
 - Restricted access of those API keys by default, only allowing access to APIs that have `allowCorsApiKey: true` in the authorization parameters
 - Created another param in the API keys to give them full access even with CORS configured
 - Updated API key creation dialog in the dashboard to set the above params in API keys

Also fixed some other CORS issues in the API:
 - Remove the static list of allowed origins, which contained all of prod, staging and development values, and receive them as a CLI arg/env instead
 - Remove CORS middlewares from VOD. They were never used anyway since afaik the outer/global middleware would already return the CORS stuff in the response (at least in the complex OPTIONS requests). Even if that's not the case, it is not necessary anymore anyway.
 - Allow any origin to access specifically the `/api/asset/upload/:url` API, since that's one that we expect frontends to call with no need for any kind of authorization.

## -

 - Created API keys with some combination of allowed origins like localhost, localhost:3000, example.com etc
 - Called different APIs with those keys, checking that the restricted/full access works and also that the CORS headers in the response come OK

Kinda still need testing if OPTIONS requests will work as I hope. My concern here is that the browser won't include
the `Authorization` header in the request and then the server doesn't know what is the API key to know what origins
it can come from.

Implements #986 

**Screenshots (optional):**
Screen Shot 2022-04-04 at 18 39 28
Screen Shot 2022-04-04 at 18 40 12
Screen Shot 2022-04-04 at 18 40 43

**Checklist:**


packages/api/src/app-router.ts
```diff
@@ -35,6 +41,26 @@ const GEOLOCATION_ENDPOINTS = [
   "geolocate",
 ];
 
+const CORS_WHITELIST = [
+  "https://livepeer.com",
+  "https://livepeer.monster",
+  "https://explorer.livepeer.org",
+  "http://localhost:3000",
```
hjpotter92: local port 3004 as well (default for livepeer box)

victorges: Actually the localhost here is unnecessary afaik since browsers already allow cors from localhost by default. I'll remove the 3000 as well.

victorges: Actually, maybe it only worked on localhost becaus we had this here? But probably browsers/the cors library just ignored the port cause I doubt all hackers that ever depended on it ran their applications on port 3000.

Nevertheless, I'll try to improve this list as well. I don't think we should be running production with CORS enabled for localhost. Probably enable it only development and somehow for the box?

victorges: Created the `LP_API_CORS_JWT_ALLOWLIST` which should be set in custom environments where the dashboard is going to be somewhere different from `locahost:3000`. Already set it in the staging environment like this: https://github.com/livepeer/livepeer-infra/blob/f43702e3fdf34adff59090629a76a3411b478e62/values/staging/livepeer.values.yaml#L12

WDYT? Think we can set an env like that in the box with the custom `3004` port?

hjpotter92: > Created the `LP_API_CORS_JWT_ALLOWLIST` which should be set in custom environments where the dashboard is going to be somewhere different from `locahost:3000`. Already set it in the staging environment like this: [livepeer/livepeer-infra@`f43702e`/values/staging/livepeer.values.yaml#L12](https://github.com/livepeer/livepeer-infra/blob/f43702e3fdf34adff59090629a76a3411b478e62/values/staging/livepeer.values.yaml#L12)

I was concocting the same theory while flying back today ^_^ :+1: a set of csv domains to be populated through environment variables. It accepting regex patterns is super cool :star2: :star_struck: 

> 
> WDYT? Think we can set an env like that in the box with the custom `3004` port?

it should theoretically be possible, but since MistController is the one overtaking spawns, might need to confirm! Maybe @iameli might know right off about this behaviour

packages/api/.env
```diff
@@ -2,7 +2,7 @@ LP_API_BROADCASTERS=[{"address":"http://127.0.0.1:3085","cliAddress":"http://127
 LP_API_FALLBACK_PROXY=http://localhost:3008
 LP_API_JWT_SECRET=secretsecret
 LP_API_JWT_AUDIENCE=livepeer
-LP_API_CORS_JWT_ALLOWLIST=http://localhost:3000
+LP_API_CORS_JWT_ALLOWLIST=["http://localhost:3000"]
```
hjpotter92, line 5: nit: I don't know if the name `CORS_JWT_ALLOWLIST` is technically correct. CORS is a browser specific limitation/feature, having nothing to do with jwt

victorges: So the JWT part is not from the CORS definition, but from our own implementation of our CORS rules. This allow-list only applies for requests authenticated with a JWT.

If you prefer, I can rename this to `LP_API_JWT_CORS_WHITELIST` instead. Do you think it makes more sense?

hjpotter92: oh nice! that clears up my doubts. I was just confused with the jwt being present in there.

packages/api/src/controllers/task.ts
```diff
@@ -171,7 +171,7 @@ app.get("/", authMiddleware({}), async (req, res) => {
   return res.json(output);
 });
 
-app.get("/:id", authMiddleware({}), async (req, res) => {
+app.get("/:id", authorizer({}), async (req, res) => {
```
victorges: Gotta allow CORS in this API as well

packages/api/src/controllers/task.ts
```diff
@@ -205,39 +205,35 @@ app.post(
...
+    updatedAt: Date.now(),
+  };
+  await db.task.update(id, { status });
+
+  res.status(200);
+  res.json({ id, status });
+});
 
-app.delete("/:id", authMiddleware({}), async (req, res) => {
+app.delete("/:id", authorizer({}), async (req, res) => {
```
victorges: This should probably be admin as well

victorges (commented):
some reminders

packages/api/src/app-router.ts
```diff
@@ -143,6 +157,25 @@ export default async function makeApp(params: CliArgs) {
...
+      corsOptsProvider({
+        anyOriginPathPrefixes: [pathJoin("/", httpPrefix, "/asset/upload/")],
+        jwtOrigin: corsJwtAllowlist,
+        baseOpts: {
+          credentials: true,
+          exposedHeaders: ["*"],
+        },
+      })
+    )
+  );
```
victorges: There is currently 1 problem with this middleware order:

If an invalid token or JWT is used for the request, the `authenticator` middleware will explode and then the `cors` middleware will never run. This means that if you use an invalid token from the browser you might never even see the response, because the response won't have CORS headers and then the browser will explode the request in a different way for you. 

To fix this, we have to make the `authenticator` middleware never throw an exception. It can be made either by actually never exploding and just letting the `authorizer` middleware explode later with "request not authenticated", or we can do some kind of deferred explode, saving the error in the request to throw after the `cors` middleware. Orr we can also make the CORS an error handler somehow?

I'll explore these possibilites tomorrow.

victorges (commented):
Just one other little thing for this to be ready to merge

packages/api/src/app-router.ts
victorges: Fixed it here! 
https://github.com/livepeer/livepeer-com/pull/985/commits/89d1fac7299d04b9d9e79600976ed7ba798bce6e

anarkrypto:
CORS restrictions create many integration problems. But adapting the whole process certainly seems like a lot of work.

I suggested something simpler, but that will certainly bring great advantages. Remove only CORS from /api/asset/upload, thus allowing direct-uploads: #996

The backend can continue to manage the api-id, but after performing the request-upload, it can send the upload token to the user who can upload it directly to the livepeer without CORS restrictions

packages/api/src/app-router.ts
```diff
@@ -143,6 +155,24 @@ export default async function makeApp(params: CliArgs) {
     req.stripe = stripe;
     next();
   });
+  app.use(
+    authenticateWithCors({
+      cors: {
+        anyOriginPathPrefixes: [pathJoin("/", httpPrefix, "/asset/upload/")],
```
victorges, line 161: Hey @anarkrypto! This line here is the one that will make these code changes have exactly the effect that you suggested!

The whole point of the separate `/request-upload` from the `/upload/$token` APIs was that users of the API would not need to proxy the file contents themselves. The `/upload` API doesn't even require authentication, so this behavior of restricting CORS was certainly a bug. 

I've also been testing this PR for the past week and I'm pretty confident that it is good to go. So I'll be merging it tomorrow (just after adding some docs) and will be hopefully deploying it on tuesday! Rest assured that you won't have to deal with file upload traffic from your backend integrating with our VOD API :)

And off course, thanks a lot for your input and detailed issue! If I didn't already have this PR close to merging I'd certainly work on a hotfix just for your use case, knowing you were blocked/affected by this behavior. Cheers!

hjpotter92 (approved):
lgtm!

iameli (approved):
Went through all the `{allowCorsApiKey: true}` instances to me and they seem reasonable. And I love the tooltip text. LGTM!

victorges:
Hey @anarkrypto ! Just letting you know that we've just deployed these changes today! You should be able to use the `/upload` API now directly from your frontend :)

You can also test the new feature of creating CORS API keys, but it seems like you've got all that figured out with a proper backend already!

Let me know if you have any issues using any of those! Cheers

anarkrypto:
Excellent news @victorges !

It is an honor for sea.tube to be one of the first to test this new feature.

I integrated in mvp.sea.tube and the direct-upload to the livepeer is working perfectly

However I was having some issues when transcoding or exporting to IPFS: The **Task** phase simply continues in "**waiting**" forever, 

The issue does not appear to be CORS related, as it also occurs locally.

I'm using the api code in js from https://github.com/livepeer/video-nft

Here is our livepeer integration code, I don't see any issue with the code.
https://github.com/sea-tube/seatube/tree/main/src/pages/api
https://github.com/sea-tube/seatube/tree/main/src/services/livePeer

You can test locally or on https://mvp.sea.tube/home by clicking on the "add video" icon in the navbar.
The project is really a pre-alpha, therefore, for now, I don't have an interface just to test this functionality, but you can check the network logs in the console.

Can you verify this behavior ? If necessary I can open a new issue.

Thank you!

victorges:
@anarkrypto Thanks for trying it out. Also thanks for reporting the issue with the tasks, I've just fixed it! :)

Long story:
We indeed had problems in the NYC region which your backend seems to be using (probably just the closest one to where it is deployed). We have been handling a couple weird reliability issues this week, and in that process we had to basically reprovision all of our clusters. NYC was one of those, and after the reprovision it was left with a problem that affected only the VOD API. We don't have E2E tests for that part of our API yet so they weren't detected automatically either :/

I've at least create a specific alert for that error now, so if it happens again along these reprovisionings we will be aware of it ASAP. In the short to medium term we will also deploy some kind of E2E testing so we are also aware of any other issues that may come up.

Thanks again for reporting it! For any other issues that you find feel free to contact me directly not to use this PR. I think we're connected on email and I've got some other contact info [here](https://hihello.me/p/0ed452b3-3584-4d64-9145-b36e87f8bcc1).

anarkrypto:
Everything is working perfectly now, @victorges 

Thank you very much, stay in touch.

anarkrypto:
Great news, the new solution is working perfectly.

Thank you very much!

Stay in touch.

___

Kaique Anarkrypto

Em qua., 13 de abr. de 2022 às 15:31, Victor Elias ***@***.***>
escreveu:

PR #989, opened 2022-04-05
api: Send user suspension email automatically + suspend all streams

victorges:
This is to send the suspension email automatically when we suspend the user.
Callers only need to update their calls to include the `emailTemplate: copyright` field.

In the end also realized it was easy to add a suspension to all the user streams as well.
It won't work 100%, but it will make bans from the admin dashboard already a tad more
reliable.

 - Send suspension email when user is suspended
 - Suspend and terminate all streams of the user when the user is suspended

## -

Trust the code.

Implements #987 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -32,6 +32,7 @@ import {
   toStringValues,
   FieldsMap,
 } from "./helpers";
+import { terminateStreamReq } from "./stream";
 
 const adminOnlyFields = ["verifiedAt", "planChangedAt"];
 
@@ -70,6 +71,29 @@ const frontendUrl = (
 ) => `${proto || "http"}://${frontendDomain}${path}`;
 
 const unsubscribeUrl = (req: Request) => frontendUrl(req, "/contact");
+
+export async function suspendUserStreams(
+  req: Request,
+  userId: string,
+  suspended: boolean
+): Promise<void> {
+  const [streams] = await db.stream.find({ userId });
+  for (const stream of streams) {
+    try {
+      const promises: Promise<any>[] = [
+        db.stream.update(stream.id, { suspended }),
+      ];
+      if (suspended) {
+        promises.push(terminateStreamReq(req, stream));
+      }
+      await Promise.all(promises);
+    } catch (err) {
+      console.error(
+        `error suspending stream id=${stream.id} userId=${userId} err=${err}`
+      );
+    }
+  }
+}
 
 type StripeProductIDs = CreateSubscription["stripeProductId"];
 
@@ -338,12 +362,19 @@ app.patch(
     const { id } = req.params;
     const user = await db.user.get(id);
     if (!user) {
-      res.status(404);
-      return res.json({ errors: ["not found"] });
+      return res.status(404).json({ errors: ["not found"] });
     }
     const { email } = user;
+
     logger.info(`set user ${id} (${email}) suspended ${suspended}`);
     await db.user.update(id, { suspended });
+
+    suspendUserStreams(req, id, suspended).catch((err) => {
+      logger.error(
+        `error suspending user streams id=${id} email=${email} err=${err}`
+      );
+    });
+
     if (suspended && emailTemplate === "copyright") {
       const {
         frontendDomain,
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -36,6 +36,9 @@ import { terminateStreamReq } from "./stream";
 
 const adminOnlyFields = ["verifiedAt", "planChangedAt"];
 
+const salesEmail = "sales@livepeer.org";
+const infraEmail = "infraservice@livepeer.org";
+
 function cleanAdminOnlyFields(fields: string[], obj: Record<string, any>) {
   for (const f of fields) {
     delete obj[f];
@@ -320,6 +323,7 @@ app.post("/", validatePost("user"), async (req, res) => {
       await sendgridEmail({
         email,
         supportAddr,
+        bcc: infraEmail,
         sendgridTemplateId,
         sendgridApiKey,
         subject: "Verify your Livepeer.com Email",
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -60,6 +62,39 @@ async function findUserByEmail(email: string, useReplica = true) {
...
 
+const frontendUrl = (
+  {
+    headers: { "x-forwarded-proto": proto },
+    config: { frontendDomain },
+  }: Request,
+  path: string
+) => `${proto || "http"}://${frontendDomain}${path}`;
+
+const unsubscribeUrl = (req: Request) => frontendUrl(req, "/contact");
```
gioelecerati, line 76: This is also in the other sendGrid notification pr we merged, we should at some point put a correct unsubscribe url for `{{{unsubscribe}}}` in the sendGrid template? Or at least not use an "unsubscribe" link into the email template but a "contact us" one, as it's a notification for the service and not a newsletter

packages/api/src/controllers/user.ts
```diff
@@ -60,6 +62,39 @@ async function findUserByEmail(email: string, useReplica = true) {
...
+
+export async function suspendUserStreams(
+  req: Request,
+  userId: string,
+  suspended: boolean
+): Promise<void> {
+  const [streams] = await db.stream.find({ userId });
+  for (const stream of streams) {
```
gioelecerati, line 84: Nice! Once for all of them

packages/api/src/controllers/user.ts
```diff
@@ -326,23 +355,55 @@ app.post("/", validatePost("user"), async (req, res) => {
 
 app.patch(
   "/:id/suspended",
+  validatePost("suspend-user-payload"),
   authMiddleware({ anyAdmin: true }),
   async (req, res) => {
+    const { suspended, emailTemplate } = req.body as SuspendUserPayload;
```
gioelecerati, line 379: Just a question, why is the emailTemplate sent as a body parameter instead of being defined somewhere in livepeer-com? I think the only one is `copyright` anyway, right?

packages/api/src/controllers/user.ts
```diff
@@ -326,23 +355,55 @@ app.post("/", validatePost("user"), async (req, res) => {
...
+    if (suspended && emailTemplate === "copyright") {
+      const {
+        frontendDomain,
+        supportAddr,
+        sendgridTemplateId,
+        sendgridApiKey,
+      } = req.config;
+      try {
+        await sendgridEmail({
+          email,
```
gioelecerati, line 405: Should we CC also infraservice as we are manually doing right now?

packages/api/src/controllers/user.ts
```diff
@@ -60,6 +62,39 @@ async function findUserByEmail(email: string, useReplica = true) {
   return users[0];
 }
 
+const frontendUrl = (
```
gioelecerati, line 68: Nice! Maybe `buttonUrl = (` as it's always used for the `buttonUrl` field?

gioelecerati (commented):
Adding the pending comments before losing them

packages/api/src/controllers/user.ts
```diff
@@ -326,23 +355,55 @@ app.post("/", validatePost("user"), async (req, res) => {
...
+      logger.error(
+        `error suspending user streams id=${id} email=${email} err=${err}`
+      );
+    });
+
+    if (suspended && emailTemplate === "copyright") {
+      const {
+        frontendDomain,
+        supportAddr,
+        sendgridTemplateId,
```
gioelecerati, line 400: Same as above, we have the `sendgridTemplateId` defined anyway, so only copyright with this Id is possible

gioelecerati (approved):
LGTM

packages/api/src/controllers/user.ts
victorges: Yeah... we need a real unsubscribe URL at some point I guess (tho still weird cause it's not a newsletter. we need this communication access and if you don't like the webhooks just disable them)

packages/api/src/controllers/user.ts
victorges: Hopefully. The only problem I guess is that it's not really reliable, which is why we need that MistUtilNukeProc or something like that, running directly on the container :/

But maybe the problem is just something on our side, idk.

packages/api/src/controllers/user.ts
victorges: Yeah but maybe we don't want to send the email when we suspend, so the template is optional. And later/soon we might have other reasons to suspend the user so I thought it made sense to be more than just an opt-in boolean here.

packages/api/src/controllers/user.ts
victorges: Ah, right, forgot about that

packages/api/src/controllers/user.ts
victorges: I actually had `emailUrl` since they're always included in emails. I don't think that makes a lot of sense tho since you would then have to look at where this is used to make sense out of it, instead of it making sense on its own. Then I thought that the most appropriate name would be this, since what it really does is create a URL pointing to a frontend page.

packages/api/src/controllers/user.ts
victorges: Hmm, we have a template for the outer portion of the email, but we still need to provide the content for it including the subject, text, etc, right? So that's what I meant with the copyright "email template". We can rethink any of this later when we either add more sendgrid templates or more "text templates", I guess.

I don't think automatically sending an email about copyright infringiment makes sense tho, so that's why I wanted to specify what email is supposed to be sent. There are various reasons why we might want to suspend an user, and forgetting that it automatically sends a copyright email and sending it when it was not the case seems worse to me than forgetting about including the param here and not sending the email sometimes (which won't happen with the automatic tooling anyway).

victorges: OK I changed the code a bit to make my intent clearer, and also made sure that we always email users, using just a simple message when the copyright is not specified.

LMK what you think!
https://github.com/livepeer/livepeer-com/pull/989/commits/1d1d529ea4c38e2cd4a8d11b52138eecc9012256

gioelecerati: Oh ok I see! Much clearer and makes sense!

victorges: Default email template powered by GitHub Autopilot®! 😂

PR #992, opened 2022-04-06
api/stream: Improve reliability of `/setactive` API

victorges:
This is to do a couple improvements in the `/setactive` API, specifically made for improving
reliability of it. This is especially important for our stream nuking APIs, since for those we need
a reliable source of "where is the stream?". In the worst case we can start fetching this info
from the Stream Health API, but I believe a very simple change to the `/setactive` API should 
already make this a lot more reliable, especially against malicious actors like our soccer
streamers who have learned how to explore the bugs in that API.

Apart from that very simple change, inplemented here https://github.com/livepeer/livepeer-com/commit/8045d8d9514004df8f203e64360fe8bb683a90b0, also made a bunch of
improvements to the API performance, since I noticed from our metrics that it's already 
running dangerously close to the `mist-api-connector` timeout a lot of the times: [[grafana](http://eu-metrics-monitoring.livepeer.live/grafana/explore?orgId=1&left=%5B%22now-6h%22,%22now%22,%22Prometheus%22,%7B%22refId%22:%22A%22,%22instant%22:true,%22range%22:true,%22exemplar%22:true,%22expr%22:%22histogram_quantile(0.9,%20sum%20by%20(region,%20le)%20(rate(livepeer_api_http_request_duration_seconds_bucket%7Bpath%3D%5C%22%2Fapi%2Fstream%2F%23val%2Fsetactive%5C%22%7D%5B10m%5D)))%22%7D%5D)]

Also did a small fix in the `/user/suspended` API here, just to avoid re-emailing the same user
that has already been suspended.

 - Avoid responding 403 to a `/setactive: false` request for a suspended stream (only when active=true)
 - Ignore `/setactive` calls from sessions that are not the last one
 - Optimize API handler execution by moving Webhooks and aux DB updates to background
 - Only send account suspended emails to not already suspended users

## -

`yarn test`
also deployed to staging and checked things still work

It is related to but does not close https://github.com/livepeer/livepeer-infra/issues/830

I'm hoping that this change will make a full backend calling the `kube-api` unnecessary.
Our stream termination logic from the API should work perfectly.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -952,6 +952,16 @@ app.put(
       }
     }
 
+    const sameFromStream =
+      stream.region === req.config.ownRegion && stream.mistHost === hostName;
+    if (stream.isActive && !active && !sameFromStream) {
+      // Don't clear the isActive field if this is being called for a different
+      // session than the last one to update the stream object. This probably
+      // means the user is doing more than 1 session with the same stream key.
+      // We only support 1 conc. session tho so ignore all but the last one.
+      return res.status(204).end();
+    }
+
     const patch = {
       isActive: active,
       lastSeen: Date.now(),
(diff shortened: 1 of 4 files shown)


hjpotter92 (approved):
lgtm!

PR #995, opened 2022-04-07
schema: add user.new event

gioelecerati:

Added mist user.new in the event schema and the frontend subscription dropdown.

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -209,7 +209,7 @@ components:
               - multistream.connected
               - multistream.error
               - multistream.disconnected
-              - user.new
+              - playback.user.new
         url:
           type: string
           format: uri
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -209,6 +209,7 @@ components:
               - multistream.connected
               - multistream.error
               - multistream.disconnected
+              - user.new
         url:
           type: string
           format: uri
--- a/packages/www/components/Dashboard/WebhookDialog/index.tsx
+++ b/packages/www/components/Dashboard/WebhookDialog/index.tsx
@@ -55,7 +55,7 @@ const eventOptions = [
   "multistream.connected",
   "multistream.error",
   "multistream.disconnected",
-  "user.new",
+  "playback.user.new",
 ];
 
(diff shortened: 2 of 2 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -209,6 +209,7 @@ components:
               - multistream.connected
               - multistream.error
               - multistream.disconnected
+              - user.new
```
victorges: @iameli I had talked to Gioele about this in a call and he mentioned that you were against naming this event differently in our API. I think I understand where you are coming from, not to have too much divergence between our API and Mist's, but I think on this case it is not a good idea to use the same terminology here. Simply because the "user" concept already exists and has a totally different meaning in the API, even with it's own `/api/user` root and all...

So when we do work on the Livepeer+Mist common API we can think of a better name for this, which makes sense in both contexts, but IMO right now using the name from one in another will only make things more confusing. Especially because currently the LP API and the Mist API have their own universe/namespaces so they will both need to be rethinked.

So @iameli WDYT? Could we rename this to something that makes more sense in the livepeer.com/api? If we were going totally free I'd go maybe with:
 - `stream.playback.new`, but if we want to keep some level of consistency with Mist, WDYT of;
 - `stream.user.new` or maybe even 
 - `stream.playback.user.new`? (I actually like this last one)

iameli: Yeah, that makes sense. Do you think there's anything in the fact that this same hook would likely be (eventually) used to authorize VoD playbacks as well, perhaps making the `stream` prefix less meaningful? What about just `playback.new`?

victorges: Makes sense! Could also be `playback.user.new`

gioelecerati: Changed it to `playback.user.new` both here and [here](https://github.com/livepeer/stream-tester/pull/140/files)!

victorges (approved):
lgtm!

hjpotter92 (approved):

iameli:
@gioelecerati Looks great. Let's sit on this for a bit until we're ready with the backend for it :+1:

iameli:
Mergin'.

PR #997, opened 2022-04-11
workflows: Enable codesign and notarize for macOS binaries

hjpotter92:


.github/workflows/api.yaml
```diff
@@ -36,23 +41,34 @@ jobs:
...
-      - name: yarn test
-        run: |
-          yarn test
+      - uses: actions/cache@v3
+        id: yarn-cache
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
```
victorges: Doesn't `actions/setup-node` already handle deps caching with the `cache: yarn` arg?

Yeah I checked on the builds in this branch, and now there are 2 steps doing possibly redundant caching:
Screen Shot 2022-04-11 at 19 03 58

.github/workflows/build.yaml
```diff
@@ -0,0 +1,167 @@
...
+      - name: restore lerna
+        id: cache-lerna
+        uses: actions/cache@v3
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: restore pkg cache
```
victorges, line 40: Same regarding cache 🤔

.github/workflows/test.yaml
```diff
@@ -1,36 +1,46 @@
-name: main
-on: push
+name: Test project
```
victorges, line 1: Nice :)

.github/workflows/www.yaml
```diff
@@ -1,20 +1,33 @@
 name: Build web service docker image
```
victorges: WDYT of 
```suggestion
name: Build www docker image
```
?

.github/workflows/www.yaml
```diff
@@ -1,20 +1,33 @@
...
 
-on: push
+on:
+  pull_request:
+  push:
+    paths:
+      - "packages/www/*"
```
victorges, line 7: Might want to add `packages/design-system` and maybe `packages/admin` here

victorges (approved):
I may not have all the required context here, but LGTM!

.github/workflows/build.yaml
hjpotter92: In case of sub-packages (like what's happening in `api.yaml`), it does work as expected. With lerna; we have separate `node_modules` populated for each package (api/www/admin/design-system) and the restore doesn't work as expected. The cache from `setup-node` only restores the root `node_modules` dir (check cache sizes). Although, the yarn step still takes 10m despite having the packages available so I blame nodejs :smile: 

The caching of `pkg` is entirely for the built binaries. It has its own caching independent from yarn/npm. Check `PKG_CACHE_PATH` at <https://github.com/vercel/pkg#environment>.

.github/workflows/www.yaml
hjpotter92: done

PR #1000, opened 2022-04-11
www: Switch cdn.livepeer.com to livepeercdn.com

gioelecerati:
Change references of cdn.livepeer.com to livepeercdn.com in the dashboard and the docs

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/AssetsTable/index.tsx
+++ b/packages/www/components/Dashboard/AssetsTable/index.tsx
@@ -146,7 +146,9 @@ const AssetsTable = ({
           source: {
             children: (
               <Box>
-                {sourceUrl && sourceUrl.indexOf("https://livepeercdn.com") === 0
+                {sourceUrl &&
+                (sourceUrl.indexOf("https://livepeercdn.com") === 0 ||
+                  sourceUrl.indexOf("https://cdn.livepeer.com") === 0)
                   ? "Live Stream"
                   : "Import"}
               </Box>
--- a/packages/www/components/Dashboard/AssetsTable/index.tsx
+++ b/packages/www/components/Dashboard/AssetsTable/index.tsx
@@ -146,8 +146,7 @@ const AssetsTable = ({
           source: {
             children: (
               <Box>
-                {sourceUrl &&
-                sourceUrl.indexOf("https://cdn.livepeer.com") === 0
+                {sourceUrl && sourceUrl.indexOf("https://livepeercdn.com") === 0
                   ? "Live Stream"
                   : "Import"}
               </Box>
--- a/packages/www/docs/api-reference/ingest.mdx
+++ b/packages/www/docs/api-reference/ingest.mdx
@@ -31,8 +31,8 @@ playback URL pairs.
 [
   {
     ingest: "rtmp://<reg>-rtmp.livepeer.com/live",
-    playback: "https://<reg>-livepeercdn.com/hls",
-    base:"https://<reg>-livepeercdn.com/"
+    playback: "https://<reg>-cdn.livepeer.com/hls",
+    base:"https://<reg>-cdn.livepeer.com/"
   }
 ]
 ```
--- a/packages/www/docs/api-reference/ingest.mdx
+++ b/packages/www/docs/api-reference/ingest.mdx
@@ -31,8 +31,8 @@ playback URL pairs.
 [
   {
     ingest: "rtmp://<reg>-rtmp.livepeer.com/live",
-    playback: "https://<reg>-cdn.livepeer.com/hls",
-    base:"https://<reg>-cdn.livepeer.com/"
+    playback: "https://<reg>-livepeercdn.com/hls",
+    base:"https://<reg>-livepeercdn.com/"
   }
 ]
 ```
--- a/packages/www/docs/api-reference/session/list-recorded-sessions.mdx
+++ b/packages/www/docs/api-reference/session/list-recorded-sessions.mdx
@@ -26,7 +26,7 @@ curl -H 'authorization: Bearer {api-key}' \
     "parentId":"nnnnnnnn-nnnn-nnnn-nnnn-nnnn",
     "recordingStatus":"ready",
     "recordingUrl":"https://livepeercdn.com/recordings/aaaaaaaa-aaaa-aaaa-aaaa-aaaa/index.m3u8"},
-    "mp4URL":"https://mdw-livepeercdn.com/recordings/ccccccc-cccc-cccc-cccc-cccccccccccccc/2021_01_01_01_00_01-source.mp4",
+    "mp4URL":"https://mdw-cdn.livepeer.com.com/recordings/ccccccc-cccc-cccc-cccc-cccccccccccccc/2021_01_01_01_00_01-source.mp4",
     {other asset object keys, like createdAt and profiles},
   {
     "sourceSegmentsDuration":30.266,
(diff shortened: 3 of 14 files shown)


packages/www/components/Dashboard/AssetsTable/index.tsx
```diff
@@ -146,8 +146,7 @@ const AssetsTable = ({
           source: {
             children: (
               <Box>
-                {sourceUrl &&
-                sourceUrl.indexOf("https://cdn.livepeer.com") === 0
+                {sourceUrl && sourceUrl.indexOf("https://livepeercdn.com") === 0
```
gioelecerati: This is going to break the asset table source of import detection, should be handled differently

hjpotter92: perhaps we can have check for both domains here with a `NOTE` mentioning its deprecation?

packages/www/docs/api-reference/ingest.mdx
```diff
@@ -31,8 +31,8 @@ playback URL pairs.
 [
   {
     ingest: "rtmp://<reg>-rtmp.livepeer.com/live",
-    playback: "https://<reg>-cdn.livepeer.com/hls",
-    base:"https://<reg>-cdn.livepeer.com/"
+    playback: "https://<reg>-livepeercdn.com/hls",
+    base:"https://<reg>-livepeercdn.com/"
```
victorges: I don't think we support these domains since we only have `livepeercdn.com`

Maybe keep these regional ones as `*.livepeer.com`, or maybe even just remove this part of the doc as Eli suggested.

packages/www/components/Dashboard/AssetsTable/index.tsx
victorges: Yeah should probably check for both here.

packages/www/docs/api-reference/session/list-recorded-sessions.mdx
```diff
@@ -25,17 +25,17 @@ curl -H 'authorization: Bearer {api-key}' \
     "record":true,
     "parentId":"nnnnnnnn-nnnn-nnnn-nnnn-nnnn",
     "recordingStatus":"ready",
-    "recordingUrl":"https://cdn.livepeer.com/recordings/aaaaaaaa-aaaa-aaaa-aaaa-aaaa/index.m3u8"},
-    "mp4URL":"https://mdw-cdn.livepeer.com/recordings/ccccccc-cccc-cccc-cccc-cccccccccccccc/2021_01_01_01_00_01-source.mp4",
+    "recordingUrl":"https://livepeercdn.com/recordings/aaaaaaaa-aaaa-aaaa-aaaa-aaaa/index.m3u8"},
+    "mp4URL":"https://mdw-livepeercdn.com/recordings/ccccccc-cccc-cccc-cccc-cccccccccccccc/2021_01_01_01_00_01-source.mp4",
```
victorges: These regional URLs won't work like this. Maybe look for all resulting `-livepeercdn.com` to fix them back

packages/www/docs/guides/start-live-streaming/back-up-transcoding.mdx
```diff
@@ -95,18 +95,18 @@ And you should get a result similar to this:
 [
    {
       "ingest":"rtmp://{region-1}-rtmp.livepeer.com/live",
-      "playback":"https://{region-1}-cdn.livepeer.com/hls",
-      "base":"https://{region-1}-cdn.livepeer.com"
+      "playback":"https://{region-1}-livepeercdn.com/hls",
```
victorges: Same regarding regioinal

packages/www/pages/app/stream/[id].tsx
```diff
@@ -305,7 +305,7 @@ const ID = () => {
   const playbackId = (stream || {}).playbackId || "";
   const domain = isStaging() ? "monster" : "com";
   const globalIngestUrl = `rtmp://rtmp.livepeer.${domain}/live`;
-  const globalPlaybackUrl = `https://cdn.livepeer.${domain}/hls/${playbackId}/index.m3u8`;
+  const globalPlaybackUrl = `https://livepeercdn.${domain}/hls/${playbackId}/index.m3u8`;
```
Comment (victorges) [inline packages/www/pages/app/stream/[id].tsx:308]: Do we have `livepeercdn.monster` as well? We probably should cc @iameli

Comment (victorges) [inline packages/www/pages/app/stream/[id].tsx:308]: Hmm funny, I swear I saw this page using the `getIngest` API somewhere...


Hmm It actually does... don't know what's the difference between the 2
https://github.com/livepeer/livepeer-com/blob/1c9390d3490f6f7473d770d967c672599b3e7d51/packages/www/pages/app/stream/%5Bid%5D.tsx#L281

Comment (victorges) [inline packages/www/pages/app/stream/[id].tsx:308]: Either way, WDYT of adding a tooltip in these URLs explaining that we changed our playback domains, but old ones will continue working? If we aren't gonna have any docs or posts about it, I think it's kind of a must.

It's really easy to do so with the built-in `Tooltip` from our design system. e.g. https://github.com/livepeer/livepeer-com/blob/0442d063e7382f8e7af6c5497ac1bd9f5a638296/packages/www/components/Dashboard/MultistreamTargetsTable/SaveTargetDialog.tsx#L325

packages/www/components/Dashboard/AssetsTable/index.tsx
gioelecerati: I was almost pushing a commit with a check to both domains in this pr, but then I thought that even that's not ok, it would be delaying it again if we hardcode another domain. If for example at some point we decide to change again or we decide to use a different delivery method, we're going to need to come back here to fix this check that not necessarily depends on the domain.  
Couldn't we detect that it's a `Live Stream` or a `Import` relying on the domain? That's even not necessarily true, as I could import a .mp4 from another asset link and that's not a `Live Stream`

packages/www/docs/api-reference/session/list-recorded-sessions.mdx
gioelecerati: Right, so we keep the livepeer.com domain for the regional endpoints?

packages/www/components/Dashboard/AssetsTable/index.tsx
victorges: One thing I think we are missing in the API is some simple cross-linking between the asset and the stream/session when the asset comes from a livestreaming. The second one is absolutely necessary, since today there's no clear way or API of grabbing the recording `asset` after you finish a stream/your recording is ready.

So one simple thing we can do is: When we create the import task to create an asset from the recording, immediately save the ID of that asset in the stream session as something like `recordingAssetId`. We can save the back-reference in the asset, with `sessionId`, `recordedSessionId` or something like that. WDYT?

If it seems too much to finish, merge, and deploy tomorrow though, I think we should go with the simpler solution for tomorrow (adding the domain list here) and can do the actual referencing sometime later. Ideally not too long in the future tho, since we've held on that technical debt for too long now, about time someone will bump into it and get blocked.

gioelecerati: I think it's going to be handled not too much in the future as we are right now implementing the asset dashboard, I am going to just cc @FullStackTiger. For now I put the two domains and then we can do how you suggested in the very next future!

packages/www/pages/app/stream/[id].tsx
Comment (gioelecerati) [reply packages/www/pages/app/stream/[id].tsx:308]: Good idea!

Comment (iameli) [reply packages/www/pages/app/stream/[id].tsx:308]: no, we're still using `cdn.livepeer.monster`

iameli (approved):

iameli (approved):

PR #1010, opened 2022-04-16
api: no index plz

iameli:

## -

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/store/table.ts
+++ b/packages/api/src/store/table.ts
@@ -332,24 +332,24 @@ export default class Table<T extends DBObject> {
 
   // on startup: auto-create indices if they don't exist
   async ensureIndex(propName, prop) {
-    if (!prop.index && !prop.unique) {
-      return;
-    }
-    let unique = "";
-    if (prop.unique) {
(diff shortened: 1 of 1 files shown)


packages/api/src/store/table.ts
```diff
@@ -332,24 +332,24 @@ export default class Table<T extends DBObject> {
...
-          CREATE ${unique} INDEX "${indexName}" ON "${this.name}" USING BTREE ((data->>'${propName}'));
-        `);
-    } catch (e) {
-      if (!e.message.includes("already exists")) {
-        throw e;
-      }
-      return;
-    }
-    logger.info(`Created ${unique} index ${indexName} on ${this.name}`);
+    // if (!prop.index && !prop.unique) {
```
victorges, line 335: So this only means that we will not belê creating new indexes right? Or did this ensure here cause performance issues even when the indexes were already there? Or did you have to delete some existing indices by any chance and can't have them recreated?

victorges (approved):
LGTM

PR #1014, opened 2022-04-20
api/vod: Inline storage info in the asset resource

victorges:

This is an improvement proposal to the VOD API, from some feedback that we've already got from users
and evolutions we would like to make to the API for easier use, like. for example for abstraction-matching
the new VOD dashboard pages. The changes are specifically regarding the relation between assets and
tasks and more extensively with the export to IPFS task.

The idea is that instead of having an asset loosely tied to a bunch of export tasks, it will now have explicit
fields for the current desired storage for the asset and status for those storages (as it can take a while until
the desired state becomes reality). This storage is currently only IPFS, but we will add other d3 storages in
the future. The object store ID is not changeable for now.

There's a whole migration strategy for the `status` field of the assets that have changed:
 - for now, would deploy this change with backward compatibility code implemented directly in the DB layer
 - created a migration API which, once we're confident with the change, should be called until all assets are migrated
 - after that (migration API starts returning `0`), we can remove the compatibility code

We are actually breaking the public schema of assets, but I believe this should be OK since it's not being
used in many places. This new design is more extensible and will allow us to make other changes in the future
without breaking compatibility again.

 - Make the `status` field in the asset be an object just like from tasks. The previous `status` becomes `status.phase` and the previous `updatedAt` becomes `status.updatedAt`.
 - Create a `storage` field in the asset and allow patching it
 - Update `storage` and `status` field on the right places (when exporting an asset on IPFS, when tasks finish, etc)
 - Update `www` code to use the new `status` field (all responses will come with that from day1, old schema will be only in the DB)
 - Update VOD documentation with new asset schema

Pending:
 - TODO after deploying this: Update `video-nft` SDK with new asset schema

## -

`yarn test`, running all the new VOD tests!

I'll try to add some unit tests here while this gets reviewed, otherwise will test locally and/or on staging.

Implements #1020 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -95,11 +95,6 @@ const assetStatus = (asset: Asset) =>
         updatedAt: asset.updatedAt,
       };
 
-const assetStorage = (asset: Asset) =>
-  asset.storage ?? {
-    objectStoreId: asset.objectStoreId,
-  };
-
 function withPlaybackUrls(asset: WithID<Asset>, ingest: string): WithID<Asset> {
   if (asset.status !== "ready") {
     return asset;
@@ -332,7 +327,7 @@ app.post(
     if ("ipfs" in params && !params.ipfs.pinata) {
       await db.asset.update(assetId, {
         storage: {
-          ...assetStorage(asset),
+          ...asset.storage,
           ipfs: params.ipfs,
         },
         status: {
@@ -341,7 +336,7 @@ app.post(
             ...status.storage,
             ipfs: {
               ...status.storage?.ipfs,
-              taskId: task.id,
+              pendingTaskId: task.id,
             },
           },
         },
@@ -579,17 +574,10 @@ app.patch("/:id", authorizer({}), validatePost("asset"), async (req, res) => {
   if (storage?.ipfs?.pinata) {
     throw new BadRequestError("Custom pinata not allowed in asset storage");
   }
-  const currStorage = assetStorage(asset);
-  if (
-    storage.objectStoreId !== undefined &&
-    storage.objectStoreId !== currStorage.objectStoreId
-  ) {
-    throw new BadRequestError("Changing object store is not allowed");
-  }
 
   let status = assetStatus(asset);
   const ipfsParamsEq =
-    JSON.stringify(storage.ipfs) === JSON.stringify(currStorage.ipfs);
+    JSON.stringify(storage.ipfs) === JSON.stringify(asset.storage?.ipfs);
   if (storage.ipfs && !ipfsParamsEq) {
     const { id: taskId } = await req.taskScheduler.scheduleTask(
       "export",
@@ -602,14 +590,14 @@ app.patch("/:id", authorizer({}), validatePost("asset"), async (req, res) => {
         ...status.storage,
         ipfs: {
           ...status.storage?.ipfs,
-          taskId,
+          pendingTaskId: taskId,
         },
       },
     };
   }
   await db.asset.update(req.body.id, {
     name,
-    storage: { ...currStorage, ...storage },
+    storage: { ...asset.storage, ...storage },
     status,
   });
 
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -336,7 +336,10 @@ app.post(
             ...status.storage,
             ipfs: {
               ...status.storage?.ipfs,
-              pendingTaskId: task.id,
+              taskIds: {
+                ...status.storage?.ipfs?.taskIds,
+                pending: task.id,
+              },
             },
           },
         },
@@ -563,7 +566,8 @@ app.delete("/:id", authorizer({}), async (req, res) => {
 
 app.patch("/:id", authorizer({}), validatePost("asset"), async (req, res) => {
   // update a specific asset
-  const asset = await db.asset.get(req.params.id);
+  const { id } = req.params;
+  const asset = await db.asset.get(id);
   if (!asset) {
     throw new NotFoundError(`asset not found`);
   }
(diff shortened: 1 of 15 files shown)


packages/api/src/store/asset-table.ts
```diff
@@ -0,0 +1,93 @@
...
+// TODO: Update existing objects in DB with the new schema to remove this
+// compatibility code.
+const assetStatusCompat = (asset: DBAsset): WithID<Asset> =>
+  !asset || typeof asset.status === "object"
+    ? (asset as WithID<Asset>)
+    : {
+        ...{ ...asset, updatedAt: undefined },
+        status: {
+          phase: asset.status,
+          updatedAt: asset.updatedAt,
```
gioelecerati, line 36: Am I missing something or `asset.updatedAt` does not exist anymore as a field?
Or is this function updating non-object status to object status, so this field always exists?

packages/api/src/controllers/asset.ts
```diff
@@ -105,6 +115,40 @@ function withPlaybackUrls(asset: WithID<Asset>, ingest: string): WithID<Asset> {
...
+): Promise<{ storage: Asset["storage"]; status: Asset["status"] }> {
+  let { storage, status } = asset;
+  const ipfsParamsEq =
+    JSON.stringify(newStorage.ipfs) === JSON.stringify(storage?.ipfs);
+  if (!ipfsParamsEq) {
+    if (!newStorage.ipfs) {
+      throw new BadRequestError("Cannot remove asset from IPFS");
+    }
+    if (!task) {
+      task = await taskScheduler.scheduleTask(
```
gioelecerati, line 132: So, here it spawns a new export task with the new IPFS metadata, if I understood correctly. I suppose the only thing you could change is the metadata CID as the video file CID would be always the same for the same asset, does this mean we are redoing the full export on the same CID reuploading the file?

gioelecerati, line 132: So, here it spawns a new export task with the new IPFS metadata. I suppose the only thing you could change is the metadata CID as the video file CID would be always the same for the same asset, does this mean we are redoing the full export on the same CID reuploading the file?

packages/api/src/controllers/asset.ts
```diff
@@ -149,21 +193,31 @@ function parseUploadUrl(
...
+    if (!ingests.length) {
+      throw new InternalServerError("Ingest not configured");
     }
+    const ingest = ingests[0].base;
+    const toExternalAsset = (a: WithID<Asset>) =>
+      req.user.admin
+        ? withPlaybackUrls(a, ingest)
+        : db.asset.cleanWriteOnlyResponse(withPlaybackUrls(a, ingest));
+
     if (Array.isArray(data)) {
```
gioelecerati, line 210: This section of the code is not clear to me, what are we doing here? In what case `data` has no `id`?

gioelecerati, line 210: This section of the code is not clear to me, what are we doing here? In what case `data` has no `id`?

packages/api/src/schema/schema.yaml
```diff
@@ -906,6 +906,60 @@ components:
...
+                  additionalProperties: false
+                  required: [taskIds]
+                  properties:
+                    taskIds:
+                      type: object
+                      additionalProperties: false
+                      properties:
+                        pending:
+                          type: string
+                          description:
```
gioelecerati, line 959: Does this get cleared by `task-runner` on export completion? I cannot find any reference to it

gioelecerati, line 950: Does this get cleared by `task-runner` on export completion? I cannot find any reference to it

gioelecerati (commented):
Saving my comments mid review before I lose them

packages/api/src/controllers/asset.ts
victorges: Yeah, all correct! 

Right now it will work the same way as before since we are doing no modifications to the `task-runner`. It will re-do the export of the same CID to Piñata on every execution. One of the optimizations that will be possible with this change is that the `task-runner` will be able to skip re-exporting the asset in case it already has a CID in the spec, i.e. it has already been exported before. I believe that check should be made in the `task-runner` tho, so for this PR the re-export will happen indeed (which is also not a regression, it's just the way it already is).

packages/api/src/controllers/asset.ts
victorges: This modifies every response from the controller to make sure we transform the output asset fields. This means basically clearing the fields that are private/admin-only when the calling user is not an admin as well as adding the playback URLs.

The response from the controller is not always a simple JSON object representing the asset. It can be an array of assets. It can also be an object with an asset object under the `asset` field. These are all the branches you see here. `id` field means the asset is the whole response object.

packages/api/src/schema/schema.yaml
victorges: It's not cleared by `task-runner` since it is an object in the API DB. There are a couple references in this pull request:
 - created: https://github.com/livepeer/livepeer-com/blob/479aa3339402a4360760fa4f68cf65aed2a1af00/packages/api/src/controllers/asset.ts#L143
 - cleared on task success: https://github.com/livepeer/livepeer-com/blob/0a4d8711d9bad1ceb5edfb0d931304dd6699b756/packages/api/src/task/scheduler.ts#L123
 - cleared on task failure: https://github.com/livepeer/livepeer-com/blob/0a4d8711d9bad1ceb5edfb0d931304dd6699b756/packages/api/src/task/scheduler.ts#L166

packages/api/src/store/asset-table.ts
victorges: It does not exist on the "public" schema of the object, but it exists in several (700k+ 😱) objects that already exist in our database. So we need to have some backward compatibility for those already persisted objects until we migrate all of them to the new format. This is the main function that does it, converting the asset objects in the old format (`DBAsset`) to the new one (just `WithID<Asset>`, used everywhere else).

packages/api/src/controllers/asset.ts
gioelecerati: Yeah I agree, we should skip the re-importing of a same asset later. I wonder if two different customers upload the same video file what happens. Of course we skip the re-import, but do we let any of them know the export didn't actually happen? I would say no, as on Livepeer they would have two separate assets in any case.

packages/api/src/schema/schema.yaml
gioelecerati: Thanks, missed this somehow

packages/api/src/controllers/asset.ts
victorges: I don't think we should expose cross-user information like that. It might execute a little faster but we won't expose why to the user. Also, that's another level of optimization that I hadn't thought of doing at first.

One thing is that we can avoid re-exporting the same asset multiple times. Another is to avoid exporting different assets with the same content. The latter would require us to actually compute the CID of an asset before it is exported to IPFS, then to check if we have already exported it from another asset by querying the DB for that CID. I think it's not a super important optimization to do either. It's a little more complex to implement and the user would not even be expecting anything like that anyway. It's also a scenario that will mostly happen on development and tests only, not on the real world with users ideally creating and minting NFTs with original content, so I'd say we shouldn't bother about that.

gioelecerati (approved):
LGTM

PR #1022, opened 2022-04-27
Tq/task runner prepare

figintern:

Adds `playbackUrl` to assets for HLS playback (as part of `transcode` and `import` tasks), formed using `playbackRecordingId` which is a hidden `writeOnly` field.

https://github.com/livepeer/task-runner/issues/12

Update `assets` endpoint

## -

Manual local testing

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -98,7 +98,7 @@ function withDownloadUrl(asset: WithID<Asset>, ingest: string): WithID<Asset> {
 }
 
 function withRecordingUrl(asset: WithID<Asset>, ingest: string): WithID<Asset> {
-  if (asset.status !== "ready" || !asset.playbackUrl) {
+  if (asset.status !== "ready" || !asset.playbackRecordingId) {
     return asset;
   }
   return {
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -98,7 +98,7 @@ function withDownloadUrl(asset: WithID<Asset>, ingest: string): WithID<Asset> {
 }
 
 function withRecordingUrl(asset: WithID<Asset>, ingest: string): WithID<Asset> {
-  if (asset.status !== "ready") {
+  if (asset.status !== "ready" || !asset.playbackUrl) {
     return asset;
   }
   return {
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -875,6 +875,17 @@ components:
...
+        playbackRecordingId:
+          type: string
+          writeOnly: true
+          example: ea03f37e-f861-4cdd-b495-0e60b6d753ad
+          index: true
+          description: Used to form recording URL for HLS playback
+        playbackUrl:
+          type: string
+          readOnly: true
+          example: https://cdn.livepeer.com/asset/ea03f37e-f861-4cdd-b495-0e60b6d753ad/index.m3u8
```
hjpotter92: should be `livepeercdn.com` now :)

packages/api/src/controllers/asset.ts
```diff
@@ -213,6 +228,7 @@ app.get("/", authorizer({}), async (req, res) => {
         }
         return {
           ...withDownloadUrl(data, ingest),
+          ...withRecordingUrl(data, ingest),
```
victorges: This will re-set the same fields over and over, since both functions will return the whole asset object additionally from the individual new fields that we want to add. Although this theoretically generates the same output, I think it's a sub-ideal logic here since it can be confusing and create unexpected bugs in the future. For example, if the `data` already contains a `downloadUrl` for any reason, the `...withRecordingUrl` would overwrite whatever `withDownloadUrl` set. 

I see a couple of options here, but there may be others:
 - We don't need the 2 functions to do kind-of similar jobs: Can just have a single `withPlaybackUrls` or something like that which sets both the download and playback URLs. This is my preferred approach especially to avoid the most repetition 
 - Can have the `withDownload/RecordingUrl` return an object only with the fields that they create, and here we do `...data, ...downloadUrl(), ...recordingUrl()` (no `with` anymore)
 - Can have the functions return only the actual values, and here we actually set the fields as `...data, downloadUrl: downloadUrl(), playbackUrl: playbackUrl()` (this will have the most repetition I guess)
 
 Btw one other thing we could consider would be to make any transformation like this in the `cleanWriteOnlyResponses` middleware that we set up for all these endpoints. Then we only have to implement this logic in 1 place and it's ok to be a little more verbose like the last idea I guess. This would be a bonus tho since it would change this controller a little more, so feel free to skip this idea if you're not comfortable with that.

victorges (approved):
LGTM!

packages/api/src/controllers/asset.ts
gioelecerati: I agree with having a single `withPlaybackUrls` that does both things.

figintern: Updated

PR #1038, opened 2022-05-02
Update doc to include getting an API Key from the dashboard

codentell:
Update “Get an API Key” /doc to include getting an API Key from the dashboard with 
screenshots and step by step guide 

## -

**Screenshots (optional):**

**Checklist:**


packages/www/docs/guides/start-live-streaming/api-key.mdx
```diff
@@ -11,18 +11,42 @@ You need an API key to make requests of the Livepeer.com API. You do not need to
...
+
+- Click `Create API Key`
+
+![Api Guide 1](/img/docs/api-guide2.png)
+
+- A popup should show up. Enter the name for your API key and then Click `Create`.
+
+![Api Guide 3](/img/docs/api-guide3.png)
+
+- Click on `Create Stream` and your dashboard should look like:
```
victorges: This step seems copied from somewhere else. Also, if you change this to just "create", might as well remove the instruction on the step above?

packages/www/docs/guides/start-live-streaming/api-key.mdx
```diff
@@ -11,18 +11,42 @@ You need an API key to make requests of the Livepeer.com API. You do not need to
...
+
+![Api Guide 4](/img/docs/api-guide4.png)
```
victorges: This files should probably have a more specific name. We have plenty of API guides so it can get confusing if we have a bunch of these with no reference to which guide they are used from.

I would go with sth like `/img/docs/api-key-1/2/3/4.png`, with the same name as the `.mdx`

packages/www/docs/guides/start-live-streaming/api-key.mdx
```diff
@@ -11,18 +11,42 @@ You need an API key to make requests of the Livepeer.com API. You do not need to
...
 the Livepeer.com Dashboard.
 
+When you create a new API key, copy and secure the value displayed to you. This
+is the sole secret you will have to use when making requests to the Livepeer.com
+API, so share and store it with caution.
+
 Login to the Livepeer.com Dashboard, and navigate to
 [the API keys page within the Developers tab](https://livepeer.com/dashboard/developers/api-keys).
 On this page, you can manage your API keys including creating or deleting them
 and monitoring their usage.
```
victorges, line 21: Shouldn't this be inside the "how to get an API key..." section?

packages/www/docs/guides/start-live-streaming/api-key.mdx
```diff
@@ -11,18 +11,42 @@ You need an API key to make requests of the Livepeer.com API. You do not need to
...
+
+![Api Guide 1](/img/docs/api-guide1.png)
+
+- Click `Create API Key`
+
+![Api Guide 1](/img/docs/api-guide2.png)
+
+- A popup should show up. Enter the name for your API key and then Click `Create`.
+
+![Api Guide 3](/img/docs/api-guide3.png)
```
victorges: This screenshot is outdated since it doesn't have the CORS checkbox. This how-to is also not mentioning anything about CORS, which I think it must. Would you mind adding some information about it here? Maybe a screenshot about it as well, can copy the ones from the CORS section below (and remove from it, if it makes sense).

packages/www/docs/guides/start-live-streaming/api-key.mdx
```diff
@@ -11,18 +11,42 @@ You need an API key to make requests of the Livepeer.com API. You do not need to
...
+![Api Guide 4](/img/docs/api-guide4.png)
+
+
+We currently allow you to reveal the secret of any existing API key but this behavior may change in the future, so
 make sure to persist the API key safely. If you ever forget an API key you'll
 always be able to delete it and create a new one anyway.
 
+#### 🎉 Congrats! You have created an API key with Livepeer using the dashboard 🎉
+
 ## Usage
```
victorges, line 48: Hmmm, mixed feelings about mixing different kinds of documentations on the same file here. Previously what we had was something in between a "tutorial" and an "explanation" ([doc system](https://documentation.divio.com)), which are focused more for users learning about API keys. We did have short instructions on how to create an API key but just linked to the dashboard and assumed it was intuitive (did you get reports of users having problems with that?).

This new section is very much a "how-to" guide tho (literally even the title). I think it mixes the crowds in this document and can make it an unsatisfying experience both for people looking for learning, getting lost in the middle of a step by step how-to, as well as for those just looking to get started with a step by step guide since there will be a lot of information they don't really want here.

So I think that we should break those up. The new section you are creating is actually the one that looks more like a "how to" guide, while all the explanations could go into the API reference doc instead (which currently is really small, needs improvement anyway so this makes sense). Then here we can focus more on a simple how-to guide, and link to the API reference doc in any case we mention something that is a little more complex to explain inline (i.e. CORS 😆). WDYT?

pglowacky (approved):

packages/www/docs/guides/start-live-streaming/api-key.mdx
pglowacky: @victorges Totally agree with this -- we need to do a better job across all of our documentation of separating out references vs. tutorials. We're going to work on overhauling all the docs this quarter. For now, I think including the tutorial here is okay -- we currently don't have a place to put tutorials and how-tos anywhere. 

@codentell Can you remove the "Congrats" line and resubmit the PR, forgot to remove before approving 🙃

packages/www/docs/guides/start-live-streaming/api-key.mdx
pglowacky: @codentell can you update the screenshot and resubmit the PR?

packages/www/docs/guides/start-live-streaming/api-key.mdx
pglowacky: +1

packages/www/docs/guides/start-live-streaming/api-key.mdx
victorges: We already have a separation of guides vs reference, which I thought could be the distinction for higher-level/how-to things vs more technical details. If we wanna get this through ASAP for some reason I guess it's ok as is then. Would just try to keep them as separate as possible in the document. Right now we are doing explanation-howto-explanation which seems like a bad reading experience. Maybe move the howto to the very end or the very beginning of the doc instead (maybe beginning, since the name of this doc is "get an api key")

packages/www/docs/guides/start-live-streaming/api-key.mdx
codentell: @pglowacky i can updated the screenshot

packages/www/docs/guides/start-live-streaming/api-key.mdx
codentell: fixed the folder structure for screen shot

packages/www/docs/guides/start-live-streaming/api-key.mdx
codentell: fixed changes and rewording

Shih-Yu (approved):
Updated images look good

PR #1051, opened 2022-05-11
api: Make sure our playback IDs are globally unique

victorges:
This fixes a tech debt in our current code since deploying VOD, which is that
the playback ID generated for assets is completely independent of the playback ID
(and stream key) generated for streams.

This fixes it by creating common helpers to check that a playback ID is unique among
all of those (stream keys, playback IDs, and asset playback IDs).

This also fixes the logic in both the assets and streams checks, which had a few flaws like
checking for uniqueness **before** applying the shard prefixes. 

 - Create common helpers for generating these values
 - Check both streams and assets tables when ensuring a unique key
 - Create sharded keys consistently and before the uniqueness is checked
 - Add a log when we find a conflict. We should ideally never see this happen, but we wanna know if it does (especially if it starts becoming frequent)

## -

`yarn test`

No. Just prepares us for the future where we might want playback IDs to be a first class citizen.

**Checklist:**


packages/api/src/controllers/generate-keys.ts
```diff
@@ -0,0 +1,87 @@
...
+ * other keys sent.
+ */
+async function keyExists(key: string, otherKeys: string[]) {
+  const playbackId = formatPlaybackId(key);
+  if (otherKeys.includes(key) || otherKeys.includes(playbackId)) {
+    return true;
+  }
+
+  const streamByKey = db.stream.find({ streamKey: key });
+  const streamByPid = db.stream.find({ playbackId });
```
gioelecerati, line 59: Can these two queries get bundled in a single one?

gioelecerati (approved):
LGTM

packages/api/src/controllers/generate-keys.ts
victorges: They could but I didn't want to handle the complexity of a custom query here since we can already parallelize them anyway and only run this on already-complex transactions which is creating a new stream or asset. So I wouldn't say it is a requirement myself, but if you think it is I can try to include it in this PR.

PR #1052, opened 2022-05-11
api: Do not skip index creation on tests

victorges:
We cannot create indexes in production because that is affecting our DB reliability.

We need to create them on tests though because many of our tests rely on that, like for example the ones that check for user email uniqueness.

This fixes that by creating indexes at leas on the tests runs.


iameli (approved):
yeah, cool, LGTM

PR #1064, opened 2022-05-17
Mark half of the Orchestrators on Staging as 'untrusted'

thomshutt:

When a new flag is passed, mark 50% of the Orchestrators as "untrusted" to allow us to test Fast Verification on Staging, where there are no community Orchestrators.

* Add boolean "half-region-orchestrators-untrusted" parameter.
* Mark every other O as untrusted when this is passed.

* Deploying to Staging to test

No

**Checklist:**

### DIFF
--- a/packages/api/src/app-router.ts
+++ b/packages/api/src/app-router.ts
@@ -77,6 +77,7 @@ export default async function makeApp(params: CliArgs) {
     stripeSecretKey,
     amqpUrl,
     returnRegionInOrchestrator,
+    halfRegionOrchestratorsUntrusted,
   } = params;
 
   if (supportAddr || sendgridTemplateId || sendgridApiKey) {
@@ -197,7 +198,9 @@ export default async function makeApp(params: CliArgs) {
 
   if (returnRegionInOrchestrator) {
     app.use((req, res, next) => {
-      req.orchestratorsGetters.push(regionsGetter);
+      req.orchestratorsGetters.push(
+        regionsGetter(halfRegionOrchestratorsUntrusted)
+      );
       return next();
     });
   }
--- a/packages/api/src/app-router.ts
+++ b/packages/api/src/app-router.ts
@@ -198,7 +198,7 @@ export default async function makeApp(params: CliArgs) {
 
   if (returnRegionInOrchestrator) {
     app.use((req, res, next) => {
-      req.orchestratorsGetters.push(
+      req.orchestratorsGetters.push(() =>
         regionsGetter(halfRegionOrchestratorsUntrusted)
       );
       return next();
--- a/packages/api/src/controllers/region.js
+++ b/packages/api/src/controllers/region.js
@@ -6,20 +6,22 @@ const defaultScore = 1;
 
 const app = Router();
 
-function flatRegions(regions = []) {
+function flatRegions(regions = [], halfRegionOrchestratorsUntrusted = false) {
   return regions.flatMap((reg) =>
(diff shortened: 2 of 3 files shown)


packages/api/src/controllers/region.js
```diff
@@ -6,20 +6,22 @@ const defaultScore = 1;
...
       region: reg.region,
       ...orch,
     }))
   );
 }
 
-export async function regionsGetter() {
-  const [regions, cursor] = await db.region.find({}, { limit: 100 });
+export function regionsGetter(halfRegionOrchestratorsUntrusted = false) {
+  return async function regionsGetter2() {
```
thomshutt: @livepeer/video-user I feel like this is probably not the optimal way, so would appreciate any pointers on how I should be doing it!

packages/api/src/controllers/region.js
```diff
@@ -28,7 +30,9 @@ app.get("/", async (req, res, next) => {
     return res.json(regions);
   }
 
-  return res.json(flatRegions(regions));
+  return res.json(
+    flatRegions(regions, req.config.halfRegionOrchestratorsUntrusted)
```
victorges, line 32: Since flatRegions returns a function now, I think you need to call the returned function here.

packages/api/src/controllers/region.js
victorges: I think this works fine! One syntax sugar you could use to make this look a little better would be a lambda function, replacing `async function regionsGetter2() {` with `async () => {`. Mostly a stylistic choice anyway.

Another alternative, to avoid this helper function that returns a function (which I find a little confusing sometimes) is to curry the new boolean variable on the call site where you need this function with zero arguments instead. That is, instead of returning a function here you could just make this function receive the boolean and return the regions, and in `app-router` you do:
`orchestratorGetters.push(() => regionsGetter(halfUntrusted...))`.

This also fixes the issue of using the return value of this function (another function) when we actually want the final list of regions.

This is a totally optional/stylistic change anyway, only raising the possibility since you were a little concerned about the current version here.

packages/api/src/app-router.ts
```diff
@@ -197,7 +198,9 @@ export default async function makeApp(params: CliArgs) {
 
   if (returnRegionInOrchestrator) {
     app.use((req, res, next) => {
-      req.orchestratorsGetters.push(regionsGetter);
+      req.orchestratorsGetters.push(
+        regionsGetter(halfRegionOrchestratorsUntrusted)
```
victorges, line 202: Contextualizing the other comment, if regionsGetter is made to return the orchestrators directly instead of a function, here you could have:
`() => regionsGetter(halfRegionO...)`

Then this function indirection stays localized here, which is where we actually need it. Other places can just use regionGetter (like the API controllers are using now)

victorges (approved):
LGTM! ~~Only one required change to the JSON response, to return the Os instead of the function.~~ Nvm I had misunderstood that one. This is good to go!

packages/api/src/controllers/region.js
victorges: Oh! Actually nevermind. Just realized that this flatRegions function is different than the regionsGetter.

Feel free to ignore this comment and merge the PR as is! Only the stylistic/optional comment now.

packages/api/src/app-router.ts
thomshutt: Awesome, that's what I was looking for thanks! Much nicer to curry it at the calling site and keep the actual function clean

PR #1032, opened 2022-04-30
Use latest design-system library

adamsoffer:
**What does this pull request do? Explain your changes. (required)**
This PR removes the design system package in the mono repo and replaces it with the published @livepeer/design-system npm package.

PR #1068, opened 2022-05-18
www: Fix component imports in admin users table

victorges:
This fixes a regression caused by: https://github.com/livepeer/livepeer-com/pull/1032

Turns out we are mixing @theme-ui/components with the
@livepeer/design-system in that page.

We should probably migrate everything to design system
altogether, but while we don't do that let's keep it
working as before.

image


adamsoffer (approved):
LGTM :)

Issue #1058, opened 2022-05-16
New /api/playback/{playbackId} API


victorges:
To return playback URLs for a given stream, asset or recording.

https://www.notion.so/livepeer/Lens-Integration-e756c036e6c741de97d8e9ef1c351e84#28ab85e81fbd4512b660fb56bd1eb67f

Definition of done:
 - New API `/api/playback/{playbackId}`
    1. Can send any playback ID from the platform and get the corresponding URL for HLS playback. 
    2. Should work for: streams playback IDs, asset playback IDs, stream recordings IDs
    3. Unit tests for this new API testing resolution for any of the resources above
    4. Should be compatible with Mist's output format: https://gist.github.com/iameli/3e9d20c2b7f11365ea8785c5a8aa6aa6

PR #1075, opened 2022-05-19
api: New API to get playback URL from an identifier

victorges:

This creates an API to facilitate and create a consistent interface for creating a playback URL
given an identifier. It could be either a stream or asset playback ID, or also a recorded session
ID and it will return a response in the same format containing the corresponding HLS playback URL.

The context for this is the creation of the Lens publication video NFTs with an embeddable HLS player.
Ideally the NFT would have no internal identifiers from livepeer.com, maybe referencing assets by their
CIDs instead. While we don't do that, we can at least simplify that internal identifier and only provide a
playback ID (additionally from the CID of the original file). The playback ID will be used to playback 
a pre-processed version of the video in an optimized format. 

Having this API also makes these NFTs future-proof, since we will be able to change their format without
it breaking every previously minted NFT that had the broken references. The reference will be only a simple
playback ID and we can change the response of that API freely.

This also has a soft dependency on #1051 but they can be deployed independently. It's also based on top of
#1014 but I believe it can also be rebased, no strict dependencies in the code.

 - Create new API for resolving the playback URL
 - Add tests 🙌 

## -

`yarn test` ✨ 

Implements #1058

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -97,20 +97,25 @@ async function validateAssetPayload(
   };
 }
 
+export function getPlaybackUrl(ingest: string, asset: WithID<Asset>): string {
+  if (!asset.playbackRecordingId) {
+    return undefined;
+  }
+  return pathJoin(
+    ingest,
+    "recordings",
+    asset.playbackRecordingId,
+    "index.m3u8"
+  );
+}
+
 function withPlaybackUrls(asset: WithID<Asset>, ingest: string): WithID<Asset> {
   if (asset.status.phase !== "ready") {
     return asset;
   }
-  if (asset.playbackRecordingId) {
-    asset.playbackUrl = pathJoin(
-      ingest,
-      "recordings",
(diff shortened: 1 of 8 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -97,20 +97,25 @@ async function validateAssetPayload(
...
+  }
+  return pathJoin(
+    ingest,
+    "recordings",
+    asset.playbackRecordingId,
+    "index.m3u8"
+  );
+}
+
 function withPlaybackUrls(asset: WithID<Asset>, ingest: string): WithID<Asset> {
```
hjpotter92: we can have the same order of params for `getPlaybackUrl` and `withPlaybackUrls` methods? `(asset, ingest)` or `(ingest, asset)`

victorges: Makes sense! Done

hjpotter92 (approved):
lgtm

PR #1076, opened 2022-05-20
api: Add streamId to webhook fields

### LINKED ISSUE #150: Add stream ids to USER_NEW trigger

### LINKED ISSUE #1057: Add optional streamId field to webhooks
Currently, the behavior for the token gating MVP forces the user to apply token gating to all streams instead of individual streams. 

Token gating should be optionally applied to only individual streams, not to all streams. 

Setting this should be included in documentation as well

gioelecerati:

For token gate single streams onto separate webhooks, the `webhook` schema needs a `streamId` field to be read by mapic.  
This way mapic can retrieve all the webhooks on a specific stream without gating all of them if a `playback.user.new` webhook is configured. 

- Added `streamId` to webhook schema
- Added a `webhook-patch-payload` type
- Added a `patch` api on the webhook to let you edit separate fields. The current existing `put` endpoint requires you to send the full webhook object, which is not user friendly. Users can now specify a `streamId` on a webhook calling `PATCH /api/webhook/:id`
- Modified the `listSubscribed` endpoint to include `streamId`. The endpoints now accepts a `streamId` parameters. It returns all webhooks configured on a specific streamId and all the webhooks with no streamId configured.
- Moved webhook.js to webhook.ts

yarn test

Fixes [stream-tester 150](https://github.com/livepeer/stream-tester/issues/150)
Fixes #1057 
**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -4,17 +4,22 @@ import { validatePost } from "../middleware";
 import Router from "express/lib/router";
 import logger from "../logger";
 import uuid from "uuid/v4";
-import { makeNextHREF, parseFilters, parseOrder } from "./helpers";
+import { makeNextHREF, parseFilters, parseOrder, FieldsMap } from "./helpers";
 import { db } from "../store";
 import sql from "sql-template-strings";
-import { UnprocessableEntityError } from "../store/errors";
+import {
+  UnprocessableEntityError,
+  NotFoundError,
+  BadRequestError,
+} from "../store/errors";
+import { Webhook, WebhookPatchPayload } from "../schema/types";
 
 function validateWebhookPayload(id, userId, createdAt, payload) {
   try {
     new URL(payload.url);
   } catch (e) {
     console.error(`couldn't parse the provided url: ${payload.url}`);
-    throw new UnprocessableEntityError(`bad url: ${url}`);
+    throw new UnprocessableEntityError(`bad url: ${payload.url}`);
   }
 
   if (!payload.events && !payload.event) {
@@ -32,12 +37,13 @@ function validateWebhookPayload(id, userId, createdAt, payload) {
     events: payload.events ?? [payload.event],
     url: payload.url,
     sharedSecret: payload.sharedSecret,
+    streamId: payload.streamId,
   };
 }
 
 const app = Router();
 
-const fieldsMap = {
+const fieldsMap: FieldsMap = {
   id: `webhook.ID`,
   name: { val: `webhook.data->>'name'`, type: "full-text" },
   url: `webhook.data->>'url'`,
@@ -131,7 +137,9 @@ app.get("/subscribed/:event", authorizer({}), async (req, res) => {
   let userId =
     req.user.admin && req.query.userId ? req.query.userId : req.user.id;
 
-  const { data } = await db.webhook.listSubscribed(userId, event);
+  let streamId = req.query.streamId;
+
+  const { data } = await db.webhook.listSubscribed(userId, event, streamId);
 
   res.status(200);
   return res.json(data);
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/webhook.ts
```diff
@@ -4,17 +4,22 @@ import { validatePost } from "../middleware";
 import Router from "express/lib/router";
```
victorges, line 4: Thanks for making this typescript ❤️

packages/api/src/controllers/webhook.ts
```diff
@@ -4,17 +4,22 @@ import { validatePost } from "../middleware";
...
+} from "../store/errors";
+import { Webhook, WebhookPatchPayload } from "../schema/types";
 
 function validateWebhookPayload(id, userId, createdAt, payload) {
   try {
     new URL(payload.url);
   } catch (e) {
     console.error(`couldn't parse the provided url: ${payload.url}`);
-    throw new UnprocessableEntityError(`bad url: ${url}`);
+    throw new UnprocessableEntityError(`bad url: ${payload.url}`);
```
victorges, line 22: The bugs just start showing themselves with ts 😌

packages/api/src/controllers/webhook.ts
```diff
@@ -188,6 +196,48 @@ app.put("/:id", authorizer({}), validatePost("webhook"), async (req, res) => {
...
+    ) {
+      // do not reveal that webhooks exists
+      throw new NotFoundError(`webhook not found`);
+    }
+
+    const { id, userId, createdAt, kind } = webhook;
+
+    if (req.body.streamId) {
+      const stream = await db.stream.get(req.body.streamId);
+      if (!stream) {
```
victorges: Gotta check that the stream is not deleted and from the same user as well
```suggestion
      if (!stream || stream.deleted || stream.userId !== webhook.userId)  {
```

packages/api/src/schema/schema.yaml
```diff
@@ -658,6 +661,11 @@ components:
       required: []
       table: null
 
+    webhook-patch-payload:
+      $ref: "#/components/schemas/webhook"
+      required: []
+      table: null
```
victorges: This will allow requests with any webhook fields to go through, they will just silently fail because we ignore them. So I think it's a better experience for the API user if you explicitly list the editable fields here, like on the `stream-patch-payload` (and make sure to have a `additionalProperties: false`).

packages/api/src/controllers/webhook.ts
```diff
@@ -188,6 +196,48 @@ app.put("/:id", authorizer({}), validatePost("webhook"), async (req, res) => {
...
+    }
+
+    await db.webhook.update(req.params.id, {
+      ...req.body,
+      // non editable fields
+      id,
+      userId,
+      createdAt,
+      kind,
+    });
```
victorges: The `update` function works as a patch API itself, so you don't need to specify every field in the desired resulting object. So I believe a safer approach here, instead of "blocking" the uneditable fields, would be to explicitly say which fields are editable and setting only those instead. This will safeguard against potential bugs in the request payload schema validation (like for example forgetting an `additionalProperties: false`) and also be easier to evolve in the future as we add more fields. For example one field you forgot to list as uneditable here was the `deleted`, which is an internal field that should never be exposed to users.

So I think this would be safer instead:
```suggestion
    const { name, events, url, sharedSecret, streamId } = req.body;
    await db.webhook.update(req.params.id, {
      name, 
      events, 
      url, 
      sharedSecret, 
      streamId
    });
```
(undefined fields will also not be included in the request to the DB, so it won't clear their value if the user doesn't send them in the request!)

WDYT?

packages/api/src/controllers/webhook.ts
```diff
@@ -188,6 +196,48 @@ app.put("/:id", authorizer({}), validatePost("webhook"), async (req, res) => {
...
+
+    res.status(200);
+    res.json({ id: req.params.id });
```
victorges: If we only return input values, might as well return a `204 No Content` instead. Could fetch the updated webhook to return it here but I don't think it's necessary (if you do do that tho, remember to clean write only fields).
```suggestion
    res.status(204).end();
```

packages/api/src/schema/schema.yaml
```diff
@@ -221,6 +221,9 @@ components:
           type: string
           writeOnly: true
           description: shared secret used to sign the webhook payload
+        streamId:
+          type: string
+          description: streamId of the stream on which the webhook is applied
```
victorges, line 226: Hopefully this field won't need to be indexed right? Raising the discussion here just in case.

victorges (approved):
LGTM with some suggesstions!

packages/api/src/controllers/webhook.ts
gioelecerati: Good catch!

packages/api/src/schema/schema.yaml
gioelecerati: Yeah makes much more sense. Thanks for this suggestion and the different way to approach the update of the webhook, I am updating both

PR #1091, opened 2022-05-26
update contact form

Shih-Yu:

Updating the contact form by removing Product/User's Plan/Ticket category fields and changing the place holder for Ticket subject and the text area

 -Removed section for Product, User's Plan, and Ticket category on the form

 -Update placeholder text for ticket subject and textarea

Created a branch and ran it on the local environment

**Screenshots (optional):**

**Original**
Screen Shot 2022-05-26 at 2 59 47 PM

**Updated**
Screen Shot 2022-05-26 at 3 00 16 PM

**Checklist:**

### DIFF
--- a/packages/www/pages/contact.tsx
+++ b/packages/www/pages/contact.tsx
@@ -142,7 +142,8 @@ const ContactPage = () => {
               id="message"
               css={{ width: "100%", boxSizing: "border-box", mb: "$4" }}
               name="TICKET.content"
-              placeholder="Please include contexts like links, error messages, or logs based on the type of question you have. The more information, the better.*"
+              placeholder="How can we help you? To guide the discussion, please give us specific details on
+your interest in speaking with our team. The more information, the better.*"
               required
             />
             <Box css={{ textAlign: "center" }}>
--- a/packages/www/pages/contact.tsx
+++ b/packages/www/pages/contact.tsx
@@ -127,15 +127,6 @@ const ContactPage = () => {
               type="hidden"
               value={query?.utm_campaign}
             />
-            <TextField
-              css={{ width: "100%", boxSizing: "border-box", mb: "$4" }}
-              size="3"
-              id="subject"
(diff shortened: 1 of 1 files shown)


packages/www/pages/contact.tsx
```diff
@@ -150,187 +133,16 @@ const ContactPage = () => {
               id="subject"
               name="TICKET.subject"
               type="text"
-              placeholder="Ticket subject*"
+              placeholder="What's the topic?*"
```
adamsoffer: Thoughts on removing this field? Is it necessary if we have the message field below it?

packages/www/pages/contact.tsx
```diff
@@ -150,187 +133,16 @@ const ContactPage = () => {
...
-              </Box>
-            </Grid>
 
             <TextArea
               size="3"
               id="message"
               css={{ width: "100%", boxSizing: "border-box", mb: "$4" }}
               name="TICKET.content"
-              placeholder="Tell us what's happening*"
+              placeholder="Please include contexts like links, error messages, or logs based on the type of question you have. The more information, the better.*"
```
adamsoffer: What do you think about making this field more general for folks that might not have a question related to support?
```
How can we help you? To guide the discussion, please give us specific details on
your interest in speaking with our team. The more information, the better.
```

Shih-Yu: That makes sense

packages/www/pages/contact.tsx
Shih-Yu: The original thought of it was to help filter out the inquiry in relation to sales support, product feature, or tech support.

adamsoffer: Gotcha. Since it's an open ended text field and we don't explicitly allow users to choose between those topics should we remove now?

Shih-Yu: 👍🏼  I removed it

Issue #1005, opened 2022-04-13
Webhooks for VOD 


pglowacky:
# Product Spec: Webhooks for VOD

## Why this is important

Webhooks are automated messages sent from apps when something happens. This is useful for developers who want to build in logic into their applications dependent on events occurring. 

VOD currently has an `upload` process that creates assets in a `waiting` and then they are moved to a `ready` state after processing is done. This is the same for any other task that is performed asynchronously, like `transcode`. For `export`, we don’t generate a new asset as the output but the task itself can have its events for passive progress tracking.

As users of the [Livepeer.com](http://Livepeer.com) platform, the developers should:

- On the upload flow, be able to tell when their users have finished uploading a file **without** having to:
    - create glue code on the frontend to let their backend know that the upload is finished;
    - proxying the contents of the files just to know when the upload finished. (pls dont)
- On every other flow, they should also be able to know when the file/video/asset info is ready for display in their app **without** a long-running polling logic on their backend nor their frontend.
- On the export flow, they should be able to know when the IPFS info is ready without the need for polling as well.

## MVP

Webhooks should be created whenever there are changes to the asset: 

`asset.waiting`

`asset.ready`

`asset.created`

`asset.status.changed`

`asset.deleted`

`task.created`

`task.status.changed` / `task.progress` (both? only the first? the latter?)

`task.completed` (covered by status.changed above, but maybe we could have this as a separate event for cases when that’s all the user cares about? (most?))

pglowacky:
https://www.notion.so/livepeer/Product-Spec-Webhooks-for-VOD-49ddff8a9314442986755a427eeb0592

PR #1099, opened 2022-06-03
api: Webhooks for VOD

victorges:

This implements a couple of webhooks for the VOD feature in general, as
per [this short discussion/spec](https://www.notion.so/livepeer/Product-Spec-Webhooks-for-VOD-49ddff8a9314442986755a427eeb0592#eb03063670794b0d8f40fcc6999da798).

 - New event types for webhooks
 - Webhooks for tasks
 - Webhooks for assets

## -

`yarn test` (might write some explicit tests for these, idk)

Implements #1005 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -33,6 +33,7 @@ import {
 } from "../schema/types";
 import { WithID } from "../store/types";
 import { mergeAssetStatus } from "../store/asset-table";
+import Queue from "../store/queue";
 
 const app = Router();
 
@@ -108,6 +109,24 @@ function withPlaybackUrls(ingest: string, asset: WithID<Asset>): WithID<Asset> {
     playbackUrl: getPlaybackUrl(ingest, asset),
     downloadUrl: pathJoin(ingest, "asset", asset.playbackId, "video"),
   };
+}
+
+export async function createAsset(asset: WithID<Asset>, queue: Queue) {
+  asset = await db.asset.create(asset);
+  await queue.publishWebhook("events.asset.created", {
+    type: "webhook_event",
+    id: uuid(),
+    timestamp: asset.createdAt,
+    event: "asset.created",
+    userId: asset.userId,
+    payload: {
+      asset: {
+        id: asset.id,
+        snapshot: asset,
+      },
+    },
+  });
+  return asset;
 }
 
 async function reconcileAssetStorage(
@@ -381,7 +400,7 @@ app.post(
       });
     }
 
-    asset = (await db.asset.create(asset)) as WithID<Asset>;
+    asset = await createAsset(asset, req.queue);
 
     const task = await req.taskScheduler.scheduleTask(
       "import",
@@ -437,7 +456,7 @@ const transcodeAssetHandler: RequestHandler = async (req, res) => {
     }
   );
   outputAsset.sourceAssetId = inputAsset.id;
-  outputAsset = (await db.asset.create(outputAsset)) as WithID<Asset>;
+  outputAsset = await createAsset(outputAsset, req.queue);
 
   const task = await req.taskScheduler.scheduleTask(
     "transcode",
@@ -498,7 +517,7 @@ app.post(
     const baseUrl = ingests[0].origin;
     const url = `${baseUrl}/api/asset/upload/${signedUploadUrl}`;
 
-    asset = await db.asset.create(asset);
+    asset = await createAsset(asset, req.queue);
     const task = await req.taskScheduler.createTask(
       "import",
       {
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -352,7 +352,7 @@ app.post(
     );
     if ("ipfs" in params && !params.ipfs?.pinata) {
       const updates = await reconcileAssetStorage(req, asset, params, task);
-      await db.asset.update(assetId, updates);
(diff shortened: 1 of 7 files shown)


victorges:
PR description that Github is not showing:

<!-------------------------------------------------------------------------
 | Thanks for send a pull request! 🎉

This implements a couple of webhooks for the VOD feature in general, as
per [this short discussion/spec](https://www.notion.so/livepeer/Product-Spec-Webhooks-for-VOD-49ddff8a9314442986755a427eeb0592#eb03063670794b0d8f40fcc6999da798).

 - New event types for webhooks
 - Webhooks for tasks
 - Webhooks for assets

## -

`yarn test` (might write some explicit tests for these, idk)

Implements #1005 

**Checklist:**

victorges:
Aha! The PR template apparently came with a big comment that wasn't closed, then everything was hidden lol

Fixed now

packages/api/src/controllers/asset.ts
```diff
@@ -110,6 +111,24 @@ function withPlaybackUrls(ingest: string, asset: WithID<Asset>): WithID<Asset> {
   };
 }
 
+export async function createAsset(asset: WithID<Asset>, queue: Queue) {
+  asset = await db.asset.create(asset);
+  await queue.publishWebhook("events.asset.created", {
+    type: "webhook_event",
+    id: uuid(),
+    timestamp: asset.createdAt,
+    event: "asset.created",
```
figintern, line 120: Nit: Can these event names be made a enums?

packages/www/components/Dashboard/WebhookDialog/index.tsx
```diff
@@ -54,6 +54,11 @@ const eventOptions = [
   "multistream.connected",
   "multistream.error",
   "multistream.disconnected",
+  "asset.created",
```
figintern, line 57: how come these are not enums and used in the queue publish requests?

figintern (approved):
Webhook logic LGTM, just some nits

packages/api/src/controllers/asset.ts
victorges: They already are! The allowed values are enumerated [here](https://github.com/livepeer/livepeer-com/blob/1c212c6a02e707e66da138e373672ff6c7e77bf8/packages/api/src/schema/schema.yaml#L202).

Also everywhere we create any field that should represent these values we make a reference to this field, which is aliased [here](https://github.com/livepeer/livepeer-com/blob/ee0c2199d465c0f4d31554fe12a6096deffa9331/packages/api/src/store/webhook-table.ts#L7) as `EventKey`.

Finally, if we try to set anything here that is not one of the enum values, it results in a Typescript error 😄 
Screen Shot 2022-06-07 at 15 20 12

So I don't think it's worth repeating those enum values as "typescript enums" instead of a string literal union, as that would just obly us to keep those 2 separate declarations in sync. The string literal union also allows for cool things like [this](https://github.com/livepeer/livepeer-com/blob/5b4b5440c9d1a6bf8dc8542b5f2c45639d55771f/packages/api/src/store/queue.ts#L23), so I think they are actually better than the enums in some sense!

packages/www/components/Dashboard/WebhookDialog/index.tsx
victorges: They are enums but as a string literal union instead, as I explained in the other comment.

Unfortunately, the schema definition only becomes a Typescript type, which has no runtime value in itself that we could use here instead of redeclaring the possible events. This is also useful because we don't want every event to be available in the frontend, just like we comment out some of the values here like `stream.detection` and `playback.user.new` which are kept as a little more internal features for now.

Also, enums wouldn't solve this completely, since we'd still need some hacks to actually get a list of values as we need here: https://stackoverflow.com/a/48768775

If we were actually writing all the code from our schema objets, we could declare first the array and then derive the string literals from it ([like this](https://dev.to/remshams/derive-union-of-string-literal-types-with-lookup-types-in-typescript-1kkf)), but we don't. Our typescript schema object definitions are generated dynamically [using ajv](https://github.com/livepeer/livepeer-com/blob/7eeb04d5bdd819a5d7022dd6455a2b51d15fbbfb/packages/api/src/compile-schemas.js#L53), so I don't think it's worth moving away from that.

victorges: Oh, regarding this being used on the queue publish requests, this code is in the frontend (`www`) package, while the queue publish happens in the `api`. We normally try to keep the cross-package dependencies to a minimum, so currently there are only "type" dependencies and then [1 specific file](https://github.com/livepeer/livepeer-com/blob/a82d714ca7b801d9b6c8dc83fbb52996fd2b0b36/packages/api/src/hash.js#L1) for hashing passwords, which is useful to have a consistent code between front and backend. 

It could be done by declaring these on the `api` and then depending on it here (just so we don't have a cyclic dependency). But then that would also mean be a new code dependency between packages just to avoid this repetition (which would still need a filter anyway for the "private" webhooks), additionally from having to move away from schema types (and validators) code generation, which saves us a lot of manual fiddling of types as well.

So overall/tldr, I don't think it's worth moving away from string literals to some sort of manual enum here. But I've improved this list by at least adding a type to it, to make sure that every value on the array will actually match an existing webhook event. It won't check if every value is here nor if we don't have duplicates, but I think that wouldn't be guaranteed on any "enum" implementation anyway.

PR #1109, opened 2022-06-14
api: Optimize recording assets generation

victorges:

Implements the API side of: https://github.com/livepeer/task-runner/pull/45

We should avoid re-transcoding recordings for generating the assets playback URLs.

Should just re-use the same HLS playlists.

For that, pass the original session ID in the import task params so it can be reused.

 - Add recorded session ID field to import task params

## -

`yarn test` doesn't break

**Checklist:**


gioelecerati (approved):
LGTM

PR #1112, opened 2022-06-17
api: Implement resumable uploads for VOD

victorges:
## [DEV] How to test this
 - Go to `packages/api` folder and run `yarn dev`
 - Go to `packages/tus-test` folder, set `LP_API_KEY` env and run `yarn start`

## Useful links
 - [Reference tus implementations](https://tus.io/protocols/resumable-upload.html#creation)
 - [Tus protocol definition](https://tus.io/protocols/resumable-upload.html)
 - [tus-js-client usage examples](https://github.com/tus/tus-js-client/blob/master/docs/usage.md)
 - [tus-node-server repo/docs](https://github.com/tus/tus-node-server)
 - [GCS authentication docs](https://cloud.google.com/storage/docs/authentication)
 
## What's missing

The "only" thing we're really missing now is changing the local filesystem storage to a Google Cloud Storage storage,
which is what we use for VOD today. For that it might be necessary to change directUpload file names again, maybe 
losing all the folders and just saving something on the root of the bucket like `directUpload-<playbackId>` or the other
way around `<playbackId>-directUpload` just so it has the same prefix as the final asset folder.

It may seem like a simple change, but other problems might happen on the way to switching the storage. Let's just hope
that not too many.


packages/api/src/controllers/asset.ts
```diff
@@ -573,6 +573,18 @@ async function createTusServer(objectStoreId: string) {
...
 }
 
+export const setupTestTus = async (): Promise<void> => {
+  tusServer = await createTestTusServer();
+};
+
+async function createTestTusServer() {
+  const tusTestServer = new tus.Server();
+  tusTestServer.datastore = new tus.FileStore({
+    path: "/tmp",
```
victorges: Might wanna use `os.tmpdir()` here instead of the hardcoded `/tmp` just for compatibility with non-Unix platforms.

Can also consider using `fs.mkdtempSync` to create a folder inside `/tmp` just not to pollute devs machines too much. Could be just `/tmp/studio-tus` for example to keep everything in there. WDYT?

Found all those functions here: https://blog.mastykarz.nl/create-temp-directory-app-node-js/
But I don't think we need to bother deleting the directory after the tests are done. Seems OK to just write to the same folder every time and the dev can clean it up eventually if desired.

packages/api/src/controllers/asset.ts
```diff
@@ -527,23 +536,89 @@ app.post(
...
+}
+
+export const setupTestTus = async (): Promise<void> => {
+  tusServer = await createTestTusServer();
+};
+
+async function createTestTusServer() {
+  const tusTestServer = new tus.Server();
+  tusTestServer.datastore = new tus.FileStore({
+    path: "/tmp",
```
victorges: You're probably missing a `namingFunction` here as well. Keep in mind that you can't use the same one as below, since that one creates a file name with a directory, and the `FileStore` seemed to not like that when I tested it. You can just inline some function here like
```suggestion
    path: "/tmp",
    namingFunction: (req) => req.res.getHeader("livepeer-playback-id").toString(),
```

To keep the file name just equal to the playback ID. Then it's also easy on your tests to check if the file was created and is equal to what you sent. WDYT?

packages/api/src/controllers/asset.ts
```diff
@@ -527,23 +536,90 @@ app.post(
...
+
+  const opts: tus.S3StoreOptions | S3ClientConfig = {
+    path: "/upload/tus",
+    bucket: vodBucket,
+    accessKeyId: url.username,
+    secretAccessKey: url.password,
+    region: vodRegion,
+    partSize: 8 * 1024 * 1024,
+    tmpDirPrefix: "tus-tmp-files",
+    endpoint: "https://storage.googleapis.com",
```
victorges: This should actually come from the parsed `url` above, it doesn't have to/shouldn't be hardcoded. Can just follow the logic from [`getS3PresignedUrl`](https://github.com/livepeer/livepeer-com/blob/6b86fb011c94e9c6dee5c9837110e0bdf7f08663/packages/api/src/controllers/helpers.ts#L141), tho you can probably just omit the user/password from the endpoint since we are providing them raw above.

probably sth like this works
```suggestion
    endpoint: `${protocol}//${url.host}`,
```

packages/api/src/controllers/asset.ts
```diff
@@ -527,23 +536,90 @@ app.post(
...
+
+export const setupTestTus = async (): Promise<void> => {
+  tusServer = await createTestTusServer();
+};
+
+async function createTestTusServer() {
+  const tusTestServer = new tus.Server();
+  tusTestServer.datastore = new tus.FileStore({
+    path: "/upload/tus",
+    directory: "/tmp",
```
victorges: Still gotta make this use os tmp folder. I made an isolated comment here another day after our session to make the tests work.

victorges: Also missing naming function (same comment).

packages/api/src/test-helpers.ts
```diff
@@ -238,3 +240,21 @@ export async function setupUsers(
...
   };
 }
+
+export async function resumeTestTusUpload(upload: tus.Upload) {
+  upload.findPreviousUploads().then(function (previousUploads) {
+    if (previousUploads.length) {
+      upload.resumeFromPreviousUpload(previousUploads[0]);
+    }
+    upload.start();
+  });
```
victorges: Since the function is async you can write this more readable with async/await.
```suggestion
  const previousUploads = await upload.findPreviousUploads()
  if (previousUploads.length) {
    upload.resumeFromPreviousUpload(previousUploads[0]);
  }
  return upload.start(); // (just in case it returns something? if it doesn't, can avoid the `return`)
```

packages/tus-test/index.html
```diff
@@ -0,0 +1,119 @@
+<html>
```
victorges: This entire `tus-test` folder should be removed. It was only a local test during development and should not be committed/merged.

packages/api/src/webhooks/cannon.ts
```diff
@@ -31,7 +31,7 @@ export default class WebhookCannon {
   sendgridTemplateId: string;
   sendgridApiKey: string;
   supportAddr: [string, string];
-  taskScheduler: TaskScheduler;
+  taskScheduler: typeof TaskScheduler;
```
victorges: This is a little weird/unnecessary. The `UpperCase` token refers to the type while `loweCase` is a variable/value. To fix this the right way, you should import the type instead of the singleton instance.

So the `import` line in the beginning should be
`import { TaskScheduler } from "../task/scheduler";`

and this line should stay as is, `TaskScheduler` being the type.

Edit: Actually, we should just remove the field from this `WebhookCannon` class and use the singleton directly. It was only a field before cause we needed to receive it in the constructor ✨

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
         },
       });
     });
+    describe("chunked upload", () => {
+      jest.setTimeout(10000);
```
victorges: Comment this out before merging. Or if the tests are really taking that long, should tune them so they're faster.

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
...
+        const res = await client.post("/asset/request-upload", {
+          name: "tus-test",
+        });
+        await createMockFile(filePath, 1024 * 1024 * 10);
+        const file = fs.createReadStream(filePath);
+        const { size } = fs.statSync(filePath);
+        expect(res.status).toBe(200);
+        let { tusEndpoint } = await res.json();
+        tusEndpoint = tusEndpoint.split("/api/")[1];
+        tusEndpoint = `${client.server.host}/api/${tusEndpoint}`;
```
victorges: The test server should be configured with an `ingest` that matches the server that it listens on. Is this not happening? Hows the `tusEndpoint` coming from the API vs what it becomes here?

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
...
+            endpoint: tusEndpoint,
+            urlStorage: new (tus as any).FileUrlStorage("/tmp/metadata"),
+            metadata: {
+              filename,
+              filetype: "video/mp4",
+            },
+            uploadSize: size,
+            onError(error) {
+              reject(error);
+              throw error;
```
victorges: Shouldn't throw. Already `reject`-ing above.

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
...
+            },
+            onProgress(bytesUploaded, bytesTotal) {
+              const percentage = parseFloat(
+                ((bytesUploaded / bytesTotal) * 100).toFixed(2)
+              );
+              if (!aborted && percentage > 0) {
+                async () => {
+                  await upload.abort(true);
+                  aborted = true;
+                };
```
victorges: You're not actually calling this function, only declaring it.

Also, might want to `aborted = true` before calling abort, otherwise you may get more segments uploaded before `abort` actually gets to execute. Simplifying:
```suggestion
                aborted = true;
                await upload.abort(true);
```

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
...
+              }
+              async () => {
+                await resumeTestTusUpload(upload);
+              };
```
victorges: This function is also not being called, only declared. Even if it were, I believe it would have been called in the wrong place as well. 

AFAIK this must be called before calling `upload.start()`, after creating the `upload`, not during progress.

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
...
+            },
+            onSuccess() {
+              resolve();
+            },
+          });
+          upload.start();
+        });
+        fs.unlinkSync(filePath);
```
victorges: You should try to use the `async` versions of these filesystem calls as much as possible. Here you're in an `async` function already, could just use `await fs.unlink(filePath)` instead.

Sync I/O is an anti-pattern in JavaScript since it only has 1 thread. If you block doing sync I/O everything else in the process will wait until that's done, and even in tests this could cause issues.

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
...
+          name: "tus-test",
+        });
+        await createMockFile(filePath, 1024 * 1024 * 10);
+        const file = fs.createReadStream(filePath);
+        const { size } = fs.statSync(filePath);
+        expect(res.status).toBe(200);
+        let { tusEndpoint } = await res.json();
+        tusEndpoint = tusEndpoint.split("/api/")[1];
+        tusEndpoint = `${client.server.host}/api/${tusEndpoint}`;
+        await new Promise<void>((resolve, reject) => {
```
victorges: You are not really performing any tests from the process performed "inside" this promise apart from ensuring it doesn't throw any exceptions. Some things missing: 
 - Actually interrupting the upload and then starting a new one to resume it
 - Finishing the upload and checking if the file is now available in the `/tmp` directory (or wherever they are uploaded)

packages/api/src/controllers/asset.test.ts
```diff
@@ -316,5 +324,58 @@ describe("controllers/asset", () => {
         },
       });
     });
+    describe("chunked upload", () => {
+      jest.setTimeout(10000);
+      it("should start upload, stop it, resume it on tus test server", async () => {
```
victorges, line 399: This test is not really doing what it describes. It's missing some logic to interrupt the upload and then resuming it from where it stopped (the interrupt/resume logic are there but are never executed, just declared as a lambda that is never called).

It's also not really testing if the uploads work, it's only checking that the `tus` client finishes cleanly. You should also check if the files are actually uploaded to where they should, ideally ensuring that it's named after the asset playback ID as well and that the `task` gets scheduled after the uploads is complete (you get the `task` ID on the `/request-upload` response, just need to check if it's `state.phase` changes to `waiting` (from `pending`))

victorges (commented):
LGTM but I opened the PR so I can't approve 😛 

The tests need some fixes though, as it's missing a couple of important checks to ensure proper functioning. Both in the "setup test" logic as well as in the test code itself.

packages/api/src/controllers/asset.test.ts
victorges: Ok I got curious about this one and tested/fixed it myself 🙈  
WDYT?
https://github.com/livepeer/studio/pull/1112/commits/bc934cd17dd4da7d78684fe85e702a22cd04fa9e

PR #1144, opened 2022-07-06
Rename livepeer/livepeer-studio to livepeer/studio

victorges:
Just a find all & replace. Avoids the repetition cacophony.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ Thanks for your interest in this project. Every contribution is welcome and
 appreciated. We're super excited to help you to get started 😎
 
 > **Note:** If you still have questions after reading through this guide,
-> [open an issue](https://github.com/livepeer/livepeer-studio/issues) or
+> [open an issue](https://github.com/livepeer/studio/issues) or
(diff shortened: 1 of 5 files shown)


hjpotter92:
needs updates in `package.json`

victorges:
@hjpotter92 Done! Weird that the package.json was pointing to a `livepeer/livepeer.studio` repository which I think never existed (or at least github is not redirecting anymore).

The package is still named `livepeer.studio` and I think we can keep it like this tho, or we can change to `@livepeer/studio` in the future. I'm not so sure about the implications there tho, so rather leave it as is.

hjpotter92 (approved):
> @hjpotter92 Done! Weird that the package.json was pointing to a `livepeer/livepeer.studio` repository which I think never existed (or at least github is not redirecting anymore).

That might be because the last find-and-replace might've taken place using regex (instead of exact match), so `livepeer.com` matched `livepeer-com`. :thinking: 

> The package is still named `livepeer.studio` and I think we can keep it like this tho, or we can change to `@livepeer/studio` in the future. I'm not so sure about the implications there tho, so rather leave it as is.

lgtm!

PR #1153, opened 2022-07-12
api: Allow finding an asset by CID (of video or metadata)

victorges:

This is to allow finding an asset from its CID (of video or metadata).

It will allow clean-up of pins in any IPFS provider, so that we only actually delete
the Pin when there are no more assets pointing to that same file.

Will also allow unpinning of previous metadata when the NFT metadata changes,
following that same principle/api.

 - Allow indexes of nested properties
 - Add index to CID fields
 - Allow filtering by them on the `/asset` API

## -

 - `yarn test`
 - to-do: Query for an asset using its CIDs and check it's findable

Not really, but it's related to the current Piñata incident that we're having. Gotta create an issue for that.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -261,9 +261,8 @@ const fieldsMap: FieldsMap = {
 };
 
 app.get("/", authorizer({}), async (req, res) => {
-  let { limit, cursor, all, allUsers, order, filters, count } = toStringValues(
-    req.query
-  );
+  let { limit, cursor, all, allUsers, order, filters, playbackId, count } =
+    toStringValues(req.query);
   if (isNaN(parseInt(limit))) {
     limit = undefined;
   }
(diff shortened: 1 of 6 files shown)


packages/api/src/store/table.ts
```diff
@@ -331,23 +331,32 @@ export default class Table<T extends DBObject> {
...
       unique = "unique";
     }
-    const indexName = `${this.name}_${propName}`;
+    const indexName = `${this.name}_${[...parents, propName].join("_")}`;
+    const parentsAcc = parents.map((p) => `->'${p}'`).join("");
+    const propAccessor = `data${parentsAcc}->>'${propName}'`;
     try {
       await this.db.query(`
-          CREATE ${unique} INDEX "${indexName}" ON "${this.name}" USING BTREE ((data->>'${propName}'));
+          CREATE ${unique} INDEX "${indexName}" ON "${this.name}" USING BTREE ((${propAccessor}));
```
hjpotter92, line 365: should this be defined as:
```js
sql`CREATE ${unique} INDEX "${indexName}" ON "${this.name}" USING BTREE ((${propAccessor}));`
```

the `sql` prefix before string literal

packages/api/src/controllers/playback.test.ts
```diff
@@ -144,6 +144,37 @@ describe("controllers/playback", () => {
...
+          status: {
+            phase: "ready",
+            storage: {
+              ipfs: {
+                data: {
+                  videoFileCid: cid,
+                },
+              },
+            },
+          },
```
gioelecerati, line 161: In this tests it's complaining that a couple of fields are missing for this update
```suggestion
          status: {
            phase: "ready",
            updatedAt: Date.now()
            storage: {
              ipfs: {
                data: {
                  videoFileCid: cid,
                },
                taskIds:{}
              },
            },
          },
```

packages/api/src/controllers/asset.ts
```diff
@@ -247,32 +247,39 @@ const fieldsMap: FieldsMap = {
...
-    const query = parseFilters(fieldsMap, filters);
-    if (!all || all === "false") {
-      query.push(sql`asset.data->>'deleted' IS NULL`);
-    }
+  const query = parseFilters(fieldsMap, filters);
+  if (playbackId) {
+    query.push(sql`asset.data->>'playbackId' = ${playbackId}`);
+  }
+  if (!all || all === "false") {
+    // TODO: Consider keeping this flag as admin-only
```
gioelecerati: I agree this would be better as admin only, at least for consistency with the other apis that never allows the user to retrieve the `deleted` objects

packages/api/src/store/asset-table.ts
```diff
@@ -110,4 +110,19 @@ export default class AssetTable extends Table<DBAsset> {
     }
     return assetStatusCompat(res.rows[0].data as WithID<Asset>);
   }
+
+  async getByIpfsCid(cid: string): Promise<WithID<Asset>> {
```
gioelecerati, line 114: ❤️

packages/api/src/controllers/playback.test.ts
victorges: I'll ignore this for now since I believe I've fixed it in the follow-up PR.

PR #1178, opened 2022-07-23
api/asset: Keep storage spec and status in the same place

victorges:

This is to make the `storage` API from assets a little easier to understand, reason and use
by moving the `status` of the storage exports to inside the `storage` fields itself. This avoids
repeating the same storage types in different places and having to parse through the whole
object to find the result of something you patched somewhere else.

With this change, there are now 3 ways of adding an IPFS storage to an app:
 - `PATCH /asset/:id {"storage": { "ipfs": true } }`
 - `PATCH /asset/:id {"storage": { "ipfs": {} } }` (backward compatibility)
 - `PATCH /asset/:id {"storage": { "ipfs": { spec: {} } } }`

After doing so, the `status` of the `ipfs` storage will be shown in the same place, sth like:
```yaml
{ 
  "id": "...", 
  "storage": { 
    "ipfs": { 
      "spec": {}, 
      "status": { 
        phase: "waiting", 
        tasks: { last: "abc" }, 
        addresses: { videoFileCid: ... ... } 
    } 
  } 
}
```

To save to IPFS with custom options, like a custom `nftMetadata`, the API is:
 - `PATCH /asset/:id {"storage": { "ipfs": { spec: { nftMetadata: { "a": "example" } } } } }`

And the `storage` object in the `asset` is updated in a similar manner, keeping the provided `spec`.
If the task fails, it reverts to the previous task spec + addresses.

 - Change schame of asset with the new setup
 - Implement a migration logic from old format to the new. This includes:
   - A logic to transform the objects on-demand when reading them from DB (if old format)
   - And an API to actually change the objects in the DB, to be called manually until all are migrated

## -

Wrote a bunch of tests and `yarn test`ed'em out.

Implements #1167

**Checklist:**

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -84,6 +84,7 @@
     "isomorphic-webcrypto": "^2.3.8",
     "js-yaml": "^3.13.1",
     "jsonwebtoken": "^8.5.1",
+    "lodash": "^4.17.21",
     "m3u8-parser": "^4.4.0",
     "minio": "^7.0.12",
     "morgan": "^1.9.1",
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -32,7 +32,7 @@ import {
   Task,
 } from "../schema/types";
 import { WithID } from "../store/types";
-import { mergeAssetStatus } from "../store/asset-table";
+import { mergeStorageStatus } from "../store/asset-table";
 import Queue from "../store/queue";
 
 const app = Router();
@@ -155,16 +155,11 @@ async function reconcileAssetStorage(
     storage = {
       ...storage,
       ipfs: {
-        ...storage?.ipfs,
-        ...newIpfs,
-        status: {
-          ...storage?.ipfs?.status,
+        spec: newIpfs.spec,
+        status: mergeStorageStatus(storage?.ipfs?.status, {
           phase: "waiting",
-          tasks: {
-            ...storage?.ipfs?.status?.tasks,
-            pending: task.id,
-          },
-        },
+          tasks: { pending: task.id },
+        }),
       },
     };
   }
@@ -380,10 +375,7 @@ app.post(
         { ipfs: { spec: params.ipfs } },
         task
       );
-      await req.taskScheduler.updateAsset(asset, {
-        storage,
-        status: { ...asset.status, updatedAt: Date.now() },
-      });
+      await req.taskScheduler.updateAsset(asset, { storage });
     }
 
     res.status(201);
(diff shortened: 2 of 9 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -111,6 +112,48 @@ function withPlaybackUrls(ingest: string, asset: WithID<Asset>): WithID<Asset> {
...
+  }
+  return {
+    ...asset,
+    storage: {
+      ...asset.storage,
+      ipfs: {
+        ...asset.storage.ipfs,
+        status: {
+          ...asset.storage.ipfs.status,
+          addresses: withIpfsUrls(asset.storage.ipfs.status.addresses),
```
hjpotter92: there should be a better way to do this 😖

packages/api/package.json
```diff
@@ -84,6 +84,7 @@
     "isomorphic-webcrypto": "^2.3.8",
     "js-yaml": "^3.13.1",
     "jsonwebtoken": "^8.5.1",
+    "lodash": "^4.17.21",
```
hjpotter92, line 87: did the lock file not update because of this?

hjpotter92 (approved):

packages/api/package.json
victorges: It did not because `www` already uses `lodash`! So I believe yarn just added a dependency on the same version

packages/api/src/controllers/asset.ts
victorges: You mean the infinite recursive spread ...?

Just realized that since I added `lodash` I can just use `_.merge` now 😌 

(or did you mean something else?)

victorges: Just added some `lodash` sourcery here, lmk what you think!
https://github.com/livepeer/studio/pull/1178/commits/631ce6bd4bfb8f782c684d35b427dbd99889b7f9

hjpotter92: yeah, i meant the recursive spreading. looks better after `_.merge` :D

Shih-Yu (approved):
Looks good

packages/api/src/store/asset-table.ts
```diff
@@ -1,6 +1,7 @@
+import _ from "lodash";
```
gioelecerati, line 1: I am not getting in what context `lodash` is used

gioelecerati (approved):
LGTM

victorges:
Gotta merge this as well

packages/api/src/store/asset-table.ts
victorges: Only as a little helper when manipulating objects https://github.com/livepeer/studio/pull/1178/files/171bf9964a3088f5c0ffe8fc7a70dfa2763481ba#diff-d7e49e7719145cd61bdb5994019c95d2abb9c9fe8de1367dd4fb1ef97eb0e33dR91

victorges:
Just finished calling `/migrate-status` on staging and prod! All assets are now migrated and searchable from their CID!

Can play any asset on Livepeer that has been exported to IPFS with
https://lvpr.tv/?v=$CID ✨

PR #1179, opened 2022-07-30
api: Reduce amount of updates done by DB metrics logic

victorges:

This to reduce the amount of `UPDATE` queries that we do in our database. These queries
are performed mostly for some "metric" functionalities that we created on top of it, mainly: 
 - `100k/hour`: The logic to track the `lastSeen` timestamp of both users and API keys (but especially API keys)
 - `600k/hour`: The logic to track the amount of transcoded segments that have been streamed a given Stream object.

The goal of this pull request is to fix both of those update logics by creating some "buffer" in memory
and then combining updates done on the database. 

The expectation is that this will also fix a couple of issues we've been having with our databases regarding
replication lags everytime a VACCUUM  operation is run on the `streams` table. By doing 30x less updates
on it we'll hopefully be able to tune it better and avoid disruptions in replication.

 - Make the `stream-info-service` not send updates for every transcoded segment. Keep them in memory instead and only flush each record every 60 seconds instead.
 - Make `tracking` not send a transaction on the database on every observation of the API key. Instead, keep the last seen value in memory and flush every 60s as well.
 - Make sure to flush the metrics on process termination (so added some cleanup calls on `SIGTERM` handler)
 - Make `add` and `set` into a single query on `stream-info-service`

## -

Check metrics are still updated on the database with these versions running.

Hopefully fixes https://github.com/livepeer/livepeer-infra/issues/851

**Checklist:**


packages/api/src/app/stream-info/stream-info-app.ts
```diff
@@ -181,6 +181,63 @@ class statusPoller {
       }
       return storedInfo;
     };
+
+    const flushStreamMetrics = async (mid: string, si: streamInfo) => {
+      const storedInfo = await getObjectByMid(mid);
```
victorges: Gotta check if `storedInfo` is not null

packages/api/src/middleware/tracking.ts
```diff
@@ -0,0 +1,50 @@
...
+      try {
+        const lastSeen = this.updateLastSeenTasks.get(id);
+        this.updateLastSeenTasks.delete(id);
+
+        await table.update(
+          [
+            sql`id = ${id}`,
+            sql`coalesce((data->'lastSeen'})::bigint, 0) < ${lastSeen}`,
+          ],
+          { lastSeen } as any
```
victorges: Likely don't need as any

packages/api/src/middleware/tracking.ts
```diff
@@ -0,0 +1,50 @@
+import sql from "sql-template-strings";
+import { ApiToken } from "../schema/types";
+import { db } from "../store";
+import Table from "../store/table";
+
+const flushDelay = 60 * 1000; // 60s
+
+class Tracker {
+  updateLastSeenTasks: Map<string, number> = new Map();
```
victorges: Might wanna use something like `death` to flush these when the process is interrupted (same for stream info)
https://stackoverflow.com/a/31526978

packages/api/src/store/table.ts
```diff
@@ -218,18 +218,24 @@ export default class Table<T extends DBObject> {
...
         q.append(`, `);
       }
       q.append(`'${k}', COALESCE((data->>'${k}')::numeric, 0) + `);
-      q.append(sql` ${doc[k]}`);
+      q.append(sql` ${add[k]}`);
     });
-    q.append(sql`) WHERE id = ${id}`);
+    q.append(`)`);
+    if (set) {
+      q.append(sql` || ${JSON.stringify(set)}`);
```
iameli, line 236: The thinking here is just to do as much as possible in one query?

victorges: Yeah exactly! This small change reduces the number of transactions/updates in half as well! This was slightly necessary since we didn't get as much of a reduction only by buffering the updates. I expected 30x on the best case, but it was something like ~8x instead. With this we get ~16x less transactions which I think will be enough for now.

packages/api/src/store/table.ts
```diff
@@ -15,7 +15,7 @@ import {
   FieldSpec,
 } from "./types";
 
-const DEFAULT_SORT = "id ASC";
```
victorges, line 18: FTR: This was not really worth it. It (luckily) broke tests and when I went to do some further testing I noticed it's not really making queries faster (sometimes they're worse, cause not all objects have an indexed `data->>'id'`). Let's stick to just the `id` column.

gioelecerati (approved):
LGTM

PR #1182, opened 2022-08-01
api: Rename /import to /upload/url

victorges:
We renamed the concept and should rename the APIs accordingly.


Shih-Yu (approved):

Shih-Yu (commented):
Looks good to me!

PR #1184, opened 2022-08-03
api: bundle stream-info-service

### LINKED ISSUE #1196: catalyst: get stream-info-service
@iameli commented on [Tue Aug 09 2022](https://github.com/livepeer/livepeer-infra/issues/890)

This was supposed to do it, but this will break `-j` if we actually populate that environment variable during service discovery, so needs to be rethought a bit: https://github.com/livepeer/studio/pull/1184

---

@hjpotter92 commented on [Wed Aug 10 2022](https://github.com/livepeer/livepeer-infra/issues/890#issuecomment-1210272334)

I suppose the ticket is to add studio binaries under mist controller?

---

@leszko commented on [Wed Aug 10 2022](https://github.com/livepeer/livepeer-infra/issues/890#issuecomment-1210807971)

Moving this ticket to `livepeer/studio`.

iameli:
Simple environment variable that can be used to trigger the stream-info-service from inside the Catalyst bundle.

Actually this won't work because `-j` needs to work first. Dang. Small tweak incoming...

fix https://github.com/livepeer/studio/issues/1196


victorges (approved):
LGTM

packages/api/src/cli.ts
```diff
@@ -0,0 +1,15 @@
...
+import makeStreamInfoSvc from "./app/stream-info/stream-info-app";
+import parseCli from "./parse-cli";
+
+if (require.main === module) {
+  const args = parseCli();
+  if (args.streamInfoService) {
+    makeStreamInfoSvc(args);
+  } else {
+    makeApp(args);
+  }
```
victorges, line 14: @iameli I've made some changes to this in order to get `-j` hopefully working better. Do you think this fixes the issues you described on https://github.com/livepeer/livepeer-infra/issues/890 ?

iameli: Yep, probably!

iameli: I'll try it out - concerned that we might get errors from strict mode `yargs` about some of the environment variables but we'll see

PR #1185, opened 2022-08-05
webhooks: allow to update webhook status & response

### LINKED ISSUE #1171: Allow users to understand the health of `playback.user.new` webhook
As a Web3 developer, building a creator application that utilizes token gating
I need to know whether the `playback.user.new` webhook is healthy and if it's not, what the error message is
So that I understand whether or not my stream is working and if its not, why not. 

This issue currently blocks more users from using and testing the token gating functionality

From Victor:

> Right now users have to blindly try and either get playback "just working" or "not working at all" with no error message to know what's wrong
maybe the account confusion started from that, but not sure
hans is having similar issues
from that other issue that shih-yu opened
he has a webhook calling https://streambot.vercel.app/api/webhooks/livepeer which just returns a 500
meaning playback is broken. but mist only returns "stream open failed". it doesn't say what's wrong (and that's better, we shouldn't be leaking the internal response to any user trying to playback)
if we showed that the webhook was failing tho (and the response we got), users could iterate on it themselves

gioelecerati:

Not in the weekly planning, but needed while debugging token gating:
All the webhooks success & failures are currently updated by Studio, apart of the webhook `playback.user.new` (token gating) which is handled on mapic. This new api allows admins (& mapic) to update the status of a webhook and store the responses, allowing us to:
- Better debug problems with token gating
- Update the dashboard correctly
- Inform the user on the status of the webhook

This needs some changes on mapic to store this information.

- Added /webhook/:id/status POST endpoint (similar to the task status update endpoint used by task runner)
- Created new payload check in the schema for the response & status of a webhook

yarn test

FIxes #1171 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/webhook.test.js
+++ b/packages/api/src/controllers/webhook.test.js
@@ -193,10 +193,15 @@ describe("controllers/webhook", () => {
     it("update the status of a webhook", async () => {
       const { id } = generatedWebhook;
       const payload = {
-        status: {
-          lastTriggeredAt: Date.now(),
-          lastFailure: { timestamp: Date.now() },
+        response: {
+          webhookId: id,
+          statusCode: 400,
+          response: {
+            body: "",
+            status: 400,
+          },
         },
+        errorMessage: "Error test",
       };
       const res = await client.post(`/webhook/${id}/status`, payload);
       expect(res.status).toBe(204);
--- a/packages/api/src/controllers/webhook.test.js
+++ b/packages/api/src/controllers/webhook.test.js
@@ -190,6 +190,18 @@ describe("controllers/webhook", () => {
       expect(updated.name).toEqual("modified_name");
     });
 
+    it("update the status of a webhook", async () => {
+      const { id } = generatedWebhook;
+      const payload = {
+        status: {
+          lastTriggeredAt: Date.now(),
+          lastFailure: { timestamp: Date.now() },
+        },
+      };
+      const res = await client.post(`/webhook/${id}/status`, payload);
+      expect(res.status).toBe(204);
+    });
+
     it("disallows setting webhook for another user", async () => {
       const { id } = generatedWebhook;
       const modifiedHook = { ...generatedWebhook, userId: nonAdminUser.id };
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -8,8 +8,8 @@ import { makeNextHREF, parseFilters, parseOrder, FieldsMap } from "./helpers";
 import { db } from "../store";
 import sql from "sql-template-strings";
 import { UnprocessableEntityError, NotFoundError } from "../store/errors";
-import { WebhookResponse, WebhookStatusPayload } from "../schema/types";
-import { WithID } from "../store/types";
+import { WebhookStatusPayload } from "../schema/types";
+import { DBWebhook } from "../store/webhook-table";
 
 function validateWebhookPayload(id, userId, createdAt, payload) {
   try {
@@ -242,26 +242,55 @@ app.post(
     const { id } = req.params;
     const webhook = await db.webhook.get(id);
     if (!webhook || webhook.deleted) {
-      return res.status(404).json({ errors: ["not found"] });
+      return res.status(404).json({ errors: ["webhook not found or deleted"] });
     }
 
-    const { status, response } = req.body as WebhookStatusPayload;
+    const { response, errorMessage } = req.body as WebhookStatusPayload;
 
-    if (!status) {
-      return res.status(422).json({ errors: ["missing status in payload"] });
+    if (!response || !response.response) {
+      return res.status(422).json({ errors: ["missing response in payload"] });
+    }
+    if (!response.response.body && response.response.body !== "") {
+      return res
+        .status(400)
+        .json({ errors: ["missing body in payload response"] });
     }
 
-    await db.webhook.update(req.params.id, {
-      ...webhook,
-      status: status,
+    try {
+      const triggerTime = response.createdAt ?? Date.now();
+      let status: DBWebhook["status"] = { lastTriggeredAt: triggerTime };
+      if (
+        response.statusCode >= 300 ||
+        !response.statusCode ||
+        response.statusCode === 0
+      ) {
+        status = {
+          ...status,
+          lastFailure: {
+            error: errorMessage,
+            timestamp: triggerTime,
+            statusCode: response.statusCode,
+            response: response.response.body,
+          },
+        };
+      }
+      await db.webhook.updateStatus(webhook.id, status);
+    } catch (e) {
+      console.log(
+        `Unable to store status of webhook ${webhook.id} url: ${webhook.url}`
+      );
+    }
+
+    await db.webhookResponse.create({
+      id: uuid(),
+      webhookId: webhook.id,
+      createdAt: Date.now(),
+      statusCode: response.statusCode,
+      response: {
+        body: response.response.body,
+        status: response.statusCode,
+      },
     });
-
-    if (response) {
-      await db.webhookResponse.create({
-        ...response,
-        id: uuid(),
-      });
-    }
 
     res.status(204).end();
   }
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -281,7 +281,8 @@ app.post(
       );
     }
 
-    await db.webhookResponse.create({
+    // TODO : Change the response type and save the response making sure it's compatible object
+    /*await db.webhookResponse.create({
       id: uuid(),
       webhookId: webhook.id,
       createdAt: Date.now(),
@@ -290,7 +291,7 @@ app.post(
         body: response.response.body,
         status: response.statusCode,
       },
-    });
+    });*/
 
     res.status(204).end();
   }
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -8,7 +8,7 @@ import { makeNextHREF, parseFilters, parseOrder, FieldsMap } from "./helpers";
 import { db } from "../store";
 import sql from "sql-template-strings";
 import { UnprocessableEntityError, NotFoundError } from "../store/errors";
-import { WebhookResponse } from "../schema/types";
+import { WebhookResponse, WebhookStatusPayload } from "../schema/types";
 import { WithID } from "../store/types";
 
 function validateWebhookPayload(id, userId, createdAt, payload) {
@@ -241,26 +241,26 @@ app.post(
   async (req, res) => {
     const { id } = req.params;
     const webhook = await db.webhook.get(id);
-    if (!webhook) {
+    if (!webhook || webhook.deleted) {
       return res.status(404).json({ errors: ["not found"] });
     }
 
-    const doc = req.body.status;
-    const webhookResponse = req.body.response;
+    const { status, response } = req.body as WebhookStatusPayload;
 
-    if (!doc) {
+    if (!status) {
       return res.status(422).json({ errors: ["missing status in payload"] });
     }
 
     await db.webhook.update(req.params.id, {
       ...webhook,
-      status: doc,
+      status: status,
     });
 
-    if (webhookResponse) {
-      webhookResponse.id = uuid();
-      let w: WithID<WebhookResponse> = webhookResponse;
-      await db.webhookResponse.create(w);
+    if (response) {
+      await db.webhookResponse.create({
+        ...response,
+        id: uuid(),
+      });
     }
 
     res.status(204).end();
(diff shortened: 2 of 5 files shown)


packages/api/src/controllers/webhook.ts
```diff
@@ -237,6 +234,39 @@ app.patch(
...
 );
 
+app.post(
+  "/:id/status",
+  authorizer({ anyAdmin: true }),
+  validatePost("webhook-status-payload"),
+  async (req, res) => {
+    const { id } = req.params;
+    const webhook = await db.webhook.get(id);
+    if (!webhook) {
```
victorges: Gotta check if the webhook is deleted as well
```suggestion
    if (!webhook || webhook.deleted) {
```

victorges: Actually, maybe this is one of the few APIs we don't want to check that?

packages/api/src/controllers/webhook.ts
```diff
@@ -237,6 +234,39 @@ app.patch(
...
+      return res.status(404).json({ errors: ["not found"] });
+    }
+
+    const doc = req.body.status;
+    const webhookResponse = req.body.response;
```
victorges: **nit:** You can grab all the fields **with the proper typings!** like this:
```suggestion
    const { status, response} = req.body as WebhookStatusPayload;
```

You could still rename the destructured variables with sth like this `const { status: doc, response: webhookResponse}`, but I think the original names of the fields would be the clearest here anyway. `doc` kinda hides away what it is, would only use that if there's only one "payload" being handled in the function and `doc` is that.

packages/api/src/controllers/webhook.ts
```diff
@@ -237,6 +234,39 @@ app.patch(
...
+
+    await db.webhook.update(req.params.id, {
+      ...webhook,
+      status: doc,
+    });
+
+    if (webhookResponse) {
+      webhookResponse.id = uuid();
+      let w: WithID<WebhookResponse> = webhookResponse;
+      await db.webhookResponse.create(w);
```
victorges: typescript rescue
```suggestion
      await db.webhookResponse.create({
         ...webhookResponse,
         id: uuid()
      });
```

and if it still complains on the typing, can change last line to 
```ts
} as WithID<WebhookResponse>);
```

packages/api/src/controllers/webhook.test.js
```diff
@@ -190,6 +190,18 @@ describe("controllers/webhook", () => {
...
     });
 
+    it("update the status of a webhook", async () => {
+      const { id } = generatedWebhook;
+      const payload = {
+        status: {
+          lastTriggeredAt: Date.now(),
+          lastFailure: { timestamp: Date.now() },
+        },
+      };
```
victorges: You should try sending a response here as well to catch 1 issue that you likely haven't noticed 👀 

The `webhook-response` object has a required `eventId` because that's how it works for all webhook calls right now. I think we should make that field optional instead, just so the client doesn't have to generate some random ID just to fill that field.

packages/api/src/controllers/webhook.ts
```diff
@@ -237,6 +234,39 @@ app.patch(
...
+    }
+
+    const doc = req.body.status;
+    const webhookResponse = req.body.response;
+
+    if (!doc) {
+      return res.status(422).json({ errors: ["missing status in payload"] });
+    }
+
+    await db.webhook.update(req.params.id, {
```
victorges: There is a dedicated `updateStatus` function which should be used instead. Try taking a look on the `cannon.ts` code that does these updates to make sure you are doing the same thing here (or if there's anything you could reuse, like exposing some of the logic there as helpers):
https://github.com/livepeer/livepeer-com/blob/0f808b4b87eb72942eb098e8c3e4f13ea7699a90/packages/api/src/webhooks/cannon.ts#L446-L503

packages/api/src/controllers/webhook.ts
victorges: Hmmmmm, actually, thinking further about this, I think the clients shouldn't specify the `status` of the webhook. That because the `status` field is actually **computed** from the response:
https://github.com/livepeer/livepeer-com/blob/0f808b4b87eb72942eb098e8c3e4f13ea7699a90/packages/api/src/webhooks/cannon.ts#L456

So we would end up either missing the `lastFailure` field which is the most important for debuggability (and that's where the dashboards shows errors from), or the client would have to specify some redundant stuff in their request, with the response (which might be big 📈) going twice. So I am thinking we could do:
 - Receive only the `response` object in the request payload
 - Derive the `status` from it instead, likely reusing some of the same object in `cannon.ts` (or if you copy:
```ts
      const triggerTime = response.createdAt ?? Date.now()
      let status: DBWebhook["status"] = { lastTriggeredAt: triggerTime };
      if (response.statusCode >= 300 || !response.statusCode) {
        status = {
          ...status,
          lastFailure: {
            timestamp: triggerTime,
            statusCode: response.statusCode,
            response: response.response.body,
          },
        };
      }
      await this.db.webhook.updateStatus(response.webhookId, status);
```
)

victorges (approved):
Neat!!

packages/api/src/controllers/webhook.ts
gioelecerati: I guess this makes sense, but what about the cases where we don't get any response from the webhook? For example, when there is a connection reset or a failed DNS resolve, we actually don't have a status code and a response. Unless if the status code is missing then we consider it a failure and updated `LastFailure` accordingly?

gioelecerati: Ok I missed that it's actually handled! Sounds good!

victorges: I think we also have an `errorMessage` field in the webhook status to account for those non-HTTP failures. We could probably receive it from the caller as well, as a separate field apart from the response (just like on the status as well afaik). WDYT?

gioelecerati: Cool, makes sense! I am keeping also the `errorMessage` as a separate field. `response.statusCode` is a required field tho in the payload, so I either set up the `statusCode` as 0 when the response fails or I make it optional, wdyt?

victorges: You could have a request like this
```yaml
{
  errorMessage: foo,
  response: { status: 123, ... }
}
```

In which status(code) is a required field of response, but response is not a required field of that payload. So you can omit the whole response if you got some different error, but if you do send a response it must have a status code. WDYT?

packages/api/src/controllers/webhook.ts
```diff
@@ -241,28 +241,57 @@ app.post(
   async (req, res) => {
     const { id } = req.params;
     const webhook = await db.webhook.get(id);
-    if (!webhook) {
-      return res.status(404).json({ errors: ["not found"] });
+    if (!webhook || webhook.deleted) {
+      return res.status(404).json({ errors: ["webhook not found or deleted"] });
```
victorges, line 245: For the user, it's just a not found!
```suggestion
      return res.status(404).json({ errors: ["webhook not found"] });
```

packages/api/src/controllers/webhook.ts
```diff
@@ -241,28 +241,57 @@ app.post(
...
-    if (!doc) {
-      return res.status(422).json({ errors: ["missing status in payload"] });
+    if (!response || !response.response) {
+      return res.status(422).json({ errors: ["missing response in payload"] });
+    }
+    if (!response.response.body && response.response.body !== "") {
+      return res
+        .status(400)
+        .json({ errors: ["missing body in payload response"] });
     }
```
victorges: The response should be an optional field since the webhook call could have gotten an error (`errorMessage`) but not an actual HTTP response.

Apart from that, even if we wanted to have a required field here, this logic should be handled on the schema level, by making the field required with `required: [response]`. There are some more "semantic" checks that can't be declared too easily in a JSON schema, but otherwise we should not be doing "json schema" checks in our code as much as possible.

packages/api/src/controllers/webhook.ts
```diff
@@ -241,28 +241,57 @@ app.post(
...
-      let w: WithID<WebhookResponse> = webhookResponse;
-      await db.webhookResponse.create(w);
+    try {
+      const triggerTime = response.createdAt ?? Date.now();
+      let status: DBWebhook["status"] = { lastTriggeredAt: triggerTime };
+      if (
+        response.statusCode >= 300 ||
+        !response.statusCode ||
+        response.statusCode === 0
+      ) {
```
victorges, line 266: Response field should be optional since not always there will be a response, and `statusCode` should be obligatory (checked by JSON schema) and so we don't need to check if it's present here:
```suggestion
      if (
        errorMessage || 
        response?.statusCode >= 300
      ) {
```

packages/api/src/controllers/webhook.ts
```diff
@@ -241,28 +241,57 @@ app.post(
...
+        status = {
+          ...status,
+          lastFailure: {
+            error: errorMessage,
+            timestamp: triggerTime,
+            statusCode: response.statusCode,
+            response: response.response.body,
```
victorges, line 273:
```suggestion
            statusCode: response?.statusCode,
            response: response?.response.body,
```

packages/api/src/controllers/webhook.ts
```diff
@@ -241,28 +241,57 @@ app.post(
...
+    await db.webhookResponse.create({
+      id: uuid(),
+      webhookId: webhook.id,
+      createdAt: Date.now(),
+      statusCode: response.statusCode,
+      response: {
+        body: response.response.body,
+        status: response.statusCode,
+      },
+    });
```
victorges: Need to surround this with `if (response)`. It will not always be present and we shouldn't be writing empty/zeroed entries on the `webhook_response` if we're not saving the error message in there as well (we could save the error message 👀)

packages/api/src/schema/schema.yaml
```diff
@@ -696,8 +695,9 @@ components:
       type: object
       additionalProperties: false
       properties:
```
victorges: Right! On this object you would declare 
```suggestion
      additionalProperties: false
      required: [response]
      properties:
```

To make the response field obligatory and remove the need to check manually on your API handler code. BUT I don't think it should be required anyway, just explaining here what you could do on another case like this (or if you prefer to keep it as required)

packages/api/src/schema/schema.yaml
```diff
@@ -273,7 +273,6 @@ components:
       table: webhook_response
       required:
         - webhookId
-        - eventId
         - statusCode
```
victorges, line 276: Turns out the `statusCode` field **is** already required. No need to have those checks in the code :)

packages/api/src/controllers/webhook.ts
```diff
@@ -241,28 +241,57 @@ app.post(
...
     }
 
+    await db.webhookResponse.create({
+      id: uuid(),
+      webhookId: webhook.id,
+      createdAt: Date.now(),
+      statusCode: response.statusCode,
+      response: {
+        body: response.response.body,
+        status: response.statusCode,
```
victorges, line 292: There is also a lot more stuff in the response object. Please save at least the `response.headers` here as well but do go through the schema to figure out anything else we want to record. Ideally we should at least keep it consistent with the webhook cannon code, which right now it still isnt'.

packages/api/src/schema/schema.yaml
```diff
@@ -696,8 +695,9 @@ components:
       type: object
       additionalProperties: false
       properties:
-        status:
-          $ref: "#/components/schemas/webhook/properties/status"
+        errorMessage:
+          type: string
+          description: Error message if the webhook failed to process the event
         response:
           $ref: "#/components/schemas/webhook-response"
```
victorges, line 702: Hmm... Maybe here we actually want to send a `"#/components/schemas/webhook-response/properties/response"` instead? seems like this objet has a lot of other fields that we don't need while we only really use the `response.response` field anyway (there is a status code in there too). Would simplify the client and the server-side code. This is the most optional change.

victorges (approved):
LGTM again, but I think a few bugs would be fixed and it'd be a lot cleaner if the `response` field was not required in the payload of the new API. It is not always present and we shouldn't force it to be. Would naturally fix the bug of webhook responses by written even when there is no response data.

packages/api/src/controllers/webhook.ts
gioelecerati: I added deleted because it's never a user call but an admin one! But yeah, could remove it!

gioelecerati:
As per async discussion, for faster debugging of token gating the `webhookResponse` is not saved on update status for now, going to implement it with a compatible object in another pr.

packages/api/src/store/webhook-table.ts
```diff
@@ -29,7 +29,7 @@ export default class WebhookTable extends Table<DBWebhook> {
     const query = [sql`data->>'userId' = ${userId}`];
     if (streamId) {
       query.push(
-        sql`data->>'streamId' = ${streamId} OR data->>'streamId' IS NULL`
+        sql`(data->>'streamId' = ${streamId} OR data->>'streamId' IS NULL)`
```
victorges, line 32: Nice! At some point we should just do that inside that `find()` function that concats everything with an `AND`. Shouldn't be the consumer's responsibility to care about this.

packages/api/src/controllers/webhook.ts
victorges: FTR: We decided to simplify the first version and just comment out the `webhook_response` table update. Will do a follow-up to start saving the responses soon.

PR #1194, opened 2022-08-09
feat: added vod upload

### LINKED ISSUE #1024: VOD Dashboard UI: Creation Flow
https://www.figma.com/file/o0lkaZxamhUfG6eqSglhPO/Livepeer-VOD-Asset-Dashboard?node-id=51%3A1789

0xcadams:
Closes https://github.com/livepeer/studio/issues/1024


packages/www/hooks/use-api.tsx
```diff
@@ -840,6 +853,83 @@ const makeContext = (state: ApiState, setState) => {
...
+          ...state,
+          currentFileUploads: {
+            ...state.currentFileUploads,
+            [file.name]: { file, progress, error },
+          },
+        }));
+      };
+
+      const getTusUpload = (file: File, tusEndpoint?: string) =>
+        new tus.Upload(file, {
```
0xcadams, line 901: @victorges should I be using this endpoint? `tus` is mentioned in the docs, but might not be in production yet?

adamsoffer (changes requested):
- [x] Add a success state to the snackbar UI when the files are done uploading and make dismissable success-state
- [x] Remove "Source" column
- [x] Remove red active state inside drag and drag (looks like an error).
- [x] Rename "Import in progress" to "Upload in progress" upload

0xcadams:
https://images.zenhubusercontent.com/283575154/c1d0d380-8252-468a-b95b-0aac784afe18/screen_recording_2022_08_09_at_11_47_31_am.mov

packages/www/hooks/use-api.tsx
victorges: I think we should! Worst thing that could happen would be some errors, but I think it's good dogfooding to the new resumable uploads API to use it from the dashboard.

0xcadams: Sounds good!

adamsoffer (changes requested):
Lookin good! A few final todos:

- [x] Add a fixed width and ellipsis to longer names so they don't collide with the other row? image
- [x] Rename "Video on Demand Assets" to "Assets".
- [x] Add a checkmark next to each upload video when complete
image

- [x] Style warning banner 
image

packages/www/components/Dashboard/FileUpload/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+            css={{
+              borderRadius: "$3",
+              backgroundColor: "$red3",
+              mb: "$2",
+              p: "$2",
+            }}>
+            <Flex align="center">
+              <Box css={{ color: "$red11" }} as={ExclamationTriangleIcon} />
+              <Text css={{ fontWeight: 600, color: "$red11", ml: "$2" }}>
+                Do not close this page until upload is complete.
```
victorges, line 75: Would be cool if we eventually saved uploads on local storage or something like that so users are able to recover even if they accidentally closed the window (the code would need to recover the file somehow haha). Not a change for this PR just a comment.

adamsoffer (approved):
Looks great! Just one final minor suggestion. 
I noticed the download url gets cut off.
image

Could we add a max width to that link with an ellipsis and add a copy and download icon similar to what we do the sessions table?
image

packages/www/hooks/use-api.tsx
```diff
@@ -840,6 +854,95 @@ const makeContext = (state: ApiState, setState) => {
...
+            updateStateWithProgressOrError(file, 1, true);
+          },
+        });
+
+      for (const file of files) {
+        try {
+          updateStateWithProgressOrError(file, 0, false);
+
+          const uploadWithoutUrl = getTusUpload(file);
+          const previousUploads = await uploadWithoutUrl.findPreviousUploads();
```
victorges, line 924: I actually think you won't be able to find previous uploads without the URLs here, since the fingerprint key used to find the upload actually includes the `endpoint` opt in the upload: https://github.com/tus/tus-js-client/blob/6f723f07b016cd374d4fe65140a2794e8ca7ca56/lib/browser/fileSignature.js#L23

Did you manage to observe something different than that? Like uploads being restored in this initial flow.

So to get retry-able uploads even if the browser refreshes, I guess you'd need to manage a separate local storage entry with current uploads and their corresponding `tusEndpoints`. I don't think it's necessary to include that in this PR (especially since if you refresh you're gonna lose all the `File` references anyway), but if you do so lmk and I can help review it as well :)

victorges (approved):
LGTM! Thx for this

victorges:
Could we make this message about not closing the page more like a yellow/orange instead of a red? A couple of times using this feature my first thought noticing the red was that there had been some error with the upload.

Screen Shot 2022-08-16 at 21 09 23

iameli:
@0xcadams @victorges Just tested this and accidentally closed the tab. Could we get the "Are you sure?" confirmation while file uploads are active?

packages/www/components/Dashboard/FileUpload/index.tsx
0xcadams: Yeah absolutely. I actually think the [tus upload client](https://github.com/tus/tus-js-client/blob/master/docs/usage.md) does this for us already. It needs some testing, but it should resume the upload if you try to use the same file.

I just tried uploading and closing the tab while uploading and it saved the state to localstorage.

Screen Shot 2022-08-17 at 8 59 05 AM

However, it looks like the API shows the asset when the upload was cancelled.

Screen Shot 2022-08-17 at 8 59 51 AM

And then when resuming, the API shows the same asset duplicated once completed.

Screen Shot 2022-08-17 at 9 01 37 AM

We may need to file a bug for this

packages/www/hooks/use-api.tsx
0xcadams: Ahh I'm just seeing this comment. This makes sense, and explains the behavior I outlined in my other comment. I think we should file a bug for state management on the frontend to track `file->uploadEndpoint` and then reuse that upload endpoint if it exists. What do you think?

0xcadams:
> Looks great! Just one final minor suggestion. I noticed the download url gets cut off. 

This PR should be superseded by https://github.com/livepeer/studio/pull/1199 so this download URL was removed completely from the table (for the VOD detail changes)

> Could we make this message about not closing the page more like a yellow/orange instead of a red? A couple of times using this feature my first thought noticing the red was that there had been some error with the upload.

Yeah, I will make that change! Good point.

> Could we get the "Are you sure?" confirmation while file uploads are active?

Will add that

0xcadams:
Ported TODOs to the newer PR and closing this in favor of https://github.com/livepeer/studio/pull/1199#issuecomment-1218226972

packages/www/hooks/use-api.tsx
0xcadams: Added an issue here: https://github.com/livepeer/studio/issues/1203

victorges: Yeah nice, I agree! We'd need to store the `uploadEndpoint` and also the file name or hash somehow, so we only try to resume if it is the same file being uploaded (or even prompt the user like "hey you were uploading X, wanna continue? pick it on your filesystem again")

PR #1206, opened 2022-08-19
task: added retry logic for tasks

### LINKED ISSUE #1164: Add retry logic to upload and transcode errors

gioelecerati:

From the task-runner, a failing task can income as retryable or unretryable. This edit to the scheduler handle this information and retry the task if possible and if the attempts have not reached the max attempts allowed.  
The first retry is without delay, the other 2 retries have a delay of 30s and 45s before the next execution.

- Added the `attempts` field to the status of the task
- Made sure that these attempts are not overwritten in other updates
- Created a `retryTask` function in the scheduler and set the max_retries and delay coefficient variables
- Added a `step` field to the status of the task as per discord discussion (not used in this pr)

Fixes #1164 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1363,9 +1363,9 @@ components:
             errorMessage:
               type: string
               description: Error message if the task failed
-            attempts:
+            retries:
               type: number
-              description: Number of attemps done on the task
+              description: Number of retries done on the task
             step:
               type: string
               description: Step of the task processing
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1363,6 +1363,12 @@ components:
             errorMessage:
               type: string
               description: Error message if the task failed
+            attempts:
+              type: number
+              description: Number of attemps done on the task
+            step:
+              type: string
+              description: Step of the task processing
         output:
           type: object
           additionalProperties: false
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -20,6 +20,7 @@ const taskInfo = (task: Task): messages.TaskInfo => ({
 
 const MAX_TASK_RETRIES = 3;
 const TASK_RETRY_DELAY_COEFFICIENT = 15 * 1000;
+
 export class TaskScheduler {
   queue: Queue;
   running: boolean;
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -277,7 +277,7 @@ export class TaskScheduler {
   }
 
   async retryTask(task: WithID<Task>, event?: messages.TaskResult) {
-    let attempts = task.status.attempts + 1 || 1;
+    let attempts = task.status.attempts + 1 ?? 1;
     if (attempts <= MAX_TASK_RETRIES) {
       const status: Task["status"] = {
         ...task.status,
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -11,6 +11,7 @@ import {
 } from "../store/asset-table";
 import { RoutingKey } from "../store/queue";
 import { EventKey } from "../store/webhook-table";
+import { sleep } from "../util";
 
 const taskInfo = (task: Task): messages.TaskInfo => ({
   id: task.id,
@@ -18,8 +19,8 @@ const taskInfo = (task: Task): messages.TaskInfo => ({
   snapshot: task,
 });
 
-const MAX_RETRY_ATTEMPTS = 3;
-const TASK_RETRY_DELAY_COEFFICIENT = 15 * 1000;
+const MAX_RETRIES = 4;
+const TASK_RETRY_BASE_DELAY = 15 * 1000;
 
 export class TaskScheduler {
   queue: Queue;
@@ -82,9 +83,17 @@ export class TaskScheduler {
 
     // TODO: bundle all db updates in a single transaction
     if (event.error) {
-      if (!event.error.unretriable) {
-        await this.retryTask(task, event);
+      if (
+        !event.error.unretriable &&
+        (task.status.retries ?? 0) < MAX_RETRIES
+      ) {
+        await this.retryTask(task);
       } else {
+        if (task.status.retries) {
+          console.log(
+            `task retry process error: err=max retries reached taskId=${task.id}`
+          );
+        }
         await this.failTask(task, event.error.message);
       }
       console.log(
@@ -277,34 +286,23 @@ export class TaskScheduler {
     await this.updateTask(task, { status });
   }
 
-  async retryTask(task: WithID<Task>, event?: messages.TaskResult) {
-    let attempts = task.status.attempts + 1 ?? 1;
-    if (attempts <= MAX_RETRY_ATTEMPTS) {
-      const status: Task["status"] = {
-        ...task.status,
-        attempts: attempts,
-      };
+  async retryTask(task: WithID<Task>) {
+    let retries = (task.status.retries ?? 0) + 1;
+    const status: Task["status"] = {
+      ...task.status,
+      retries: retries,
+    };
 
-      await this.updateTask(task, { status });
-      task.status = status;
+    await this.updateTask(task, { status });
+    task.status = status;
 
-      if (attempts > 1) {
-        // No timeout at first retry, 15 seconds * attempt after that
-        let retryDelay = attempts * TASK_RETRY_DELAY_COEFFICIENT;
-        await new Promise((resolve) => setTimeout(resolve, retryDelay));
-      }
-
-      await this.enqueueTask(task);
-      return true;
-    } else {
-      console.log(
-        `task retry process error: err=max retries reached taskId=${task.id}`
-      );
-      if (event) {
-        await this.failTask(task, event.error.message);
-      }
-      return false;
+    if (retries > 1) {
+      // No timeout at first retry, 15 seconds * attempt after that
+      let retryDelay = retries * TASK_RETRY_BASE_DELAY;
+      await sleep(retryDelay);
     }
+
+    await this.enqueueTask(task);
   }
 
   async updateTask(task: Task, updates: Pick<Task, "status" | "output">) {
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -181,7 +181,11 @@ export class TaskScheduler {
     return true;
   }
 
-  private async failTask(task: Task, error: string, output?: Task["output"]) {
+  private async failTask(
+    task: WithID<Task>,
+    error: string,
+    output?: Task["output"]
+  ) {
     const status = {
       ...task.status,
       phase: "failed",
@@ -293,11 +297,9 @@ export class TaskScheduler {
       retries: retries,
     };
 
-    await this.updateTask(task, { status });
-    task.status = status;
+    task = await this.updateTask(task, { status });
 
     if (retries > 1) {
-      // No timeout at first retry, 15 seconds * attempt after that
       let retryDelay = retries * TASK_RETRY_BASE_DELAY;
       await sleep(retryDelay);
     }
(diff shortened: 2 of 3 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +277,36 @@ export class TaskScheduler {
...
+      const status: Task["status"] = {
+        ...task.status,
+        attempts: attempts,
+      };
+
+      await this.updateTask(task, { status });
+      task.status = status;
+
+      if (attempts > 1) {
+        // No timeout at first retry, 15 seconds * attempt after that
```
gioelecerati: This could increment exponentially instead of linearly, or in larger time frames, for example 0s 15s 1m

packages/api/src/task/scheduler.ts
```diff
@@ -167,11 +174,13 @@ export class TaskScheduler {
 
   private async failTask(task: Task, error: string, output?: Task["output"]) {
     const status = {
+      ...task.status,
```
gioelecerati, line 190: @victorges I added these on all the status updates to avoid loss of the attempts tracking, although I am not sure of all the implications. Is this correct from your POV or is there something that could cause other fields in the status to be overwritten or preserved when they shouldn't?

victorges: WDYT of doing something similar to this on the `updateTask` function?
https://github.com/livepeer/livepeer-com/blob/ae8791b43052ad4339f73389e73c5cd241c01e9b/packages/api/src/task/scheduler.ts#L327-L330

i.e. Callers don't need to worry about merging the previous status with the new status nor even of setting an `updatedAt`, as the update func will handle that.

packages/api/src/task/scheduler.ts
victorges: Hmm, I think linear is good for now for simplicity. Since we are starting with a short delay, we could also include one even on the first attempt WDYT? So it can be 15, 30, 45, 60...

packages/api/src/schema/schema.yaml
```diff
@@ -1363,6 +1363,12 @@ components:
             errorMessage:
               type: string
               description: Error message if the task failed
+            attempts:
+              type: number
+              description: Number of attemps done on the task
+            step:
```
victorges, line 1369: Nice, thanks! Would you mind updating the "progress" API as well to update the `step` of the task in the DB from the payload [1] received? That's how the `task-runner` would update it. 

Also, can you make sure we _clear_ that field on any non-`running` phase of the task? (i.e. we shouldn't keep the current `step` if we're changing the phase to `waiting` (retry), `completed`, etc)

[1] Also just noticed we don't do **any** payload validation on that API :harold_pain:. Would you mind creating a custom payload for it like the [`webhook-patch-payload`](https://github.com/livepeer/livepeer-com/blob/ae8791b43052ad4339f73389e73c5cd241c01e9b/packages/api/src/schema/schema.yaml#L697) and adding the `validatePost` middleware to it as well? 

This is all kind of unrelated to the retry of tasks itself, so feel free to revert this change from this PR and address it on a separate one instead (or even just create an issue for someone to pick up later).

packages/api/src/task/scheduler.ts
```diff
@@ -18,6 +18,9 @@ const taskInfo = (task: Task): messages.TaskInfo => ({
   snapshot: task,
 });
 
+const MAX_RETRY_ATTEMPTS = 3;
+const TASK_RETRY_DELAY_COEFFICIENT = 15 * 1000;
```
victorges: **nit:** WDYT of
```suggestion
const TASK_RETRY_BASE_DELAY = 15 * 1000;
```
(would imagine sth else from "coefficient")

packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +277,36 @@ export class TaskScheduler {
     await this.updateTask(task, { status });
   }
 
+  async retryTask(task: WithID<Task>, event?: messages.TaskResult) {
+    let attempts = task.status.attempts + 1 ?? 1;
```
victorges: You probably mean `(task.status.attempts ?? 0) + 1`

Rule of thumb: [never trust JavaScript will do what you imagine](https://www.destroyallsoftware.com/talks/wat). Boot a REPL shell and test out some expressions. 

Screen Shot 2022-08-22 at 17 57 09

packages/api/src/schema/schema.yaml
```diff
@@ -1363,6 +1363,12 @@ components:
             errorMessage:
               type: string
               description: Error message if the task failed
+            attempts:
+              type: number
+              description: Number of attemps done on the task
```
victorges: I think we should either name this `attempt` and always have it there and have them numbered starting from 1, to be read as an ordinal value ("first attempt", "second attempt", etc)

Or we can rename this to `retries` instead to have it mean "number of retries that were needed" or something like that, which then makes sense to start at `0` (and then not having it present on the first attempt). Kinda preferring this one, WDYT? (seems to be how you implemented, just need to rename the field)

packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +277,36 @@ export class TaskScheduler {
...
+        await new Promise((resolve) => setTimeout(resolve, retryDelay));
+      }
+
+      await this.enqueueTask(task);
+      return true;
+    } else {
+      console.log(
+        `task retry process error: err=max retries reached taskId=${task.id}`
+      );
+      if (event) {
```
victorges: Is there any case this function will be called without an event? 0o

packages/api/src/task/scheduler.ts
```diff
@@ -18,6 +18,9 @@ const taskInfo = (task: Task): messages.TaskInfo => ({
   snapshot: task,
 });
 
+const MAX_RETRY_ATTEMPTS = 3;
```
victorges: Just to align on the terms here... "Attempt" means a different thing than "retry". If the task executes 4 times, it's 4 attempts but only 3 retries (but 4 "tries"). "Retry attempt" just means "retry" here IMO and so it's unnecessary, but then it's confusing because the task field is only "attempt" which means a different thing.

I already suggested dropping the "attempt" term and just using "retry" instead, and so we can simplify this const to just `MAX_RETRIES` as well.

packages/api/src/task/scheduler.ts
```diff
@@ -79,8 +82,11 @@ export class TaskScheduler {
 
     // TODO: bundle all db updates in a single transaction
     if (event.error) {
-      await this.failTask(task, event.error.message);
-      // TODO: retry task
+      if (!event.error.unretriable) {
```
victorges: WDYT of moving the `MAX_ATTEMPTS` (actually `MAX_RETRIES`) check to here and simplifying the whole `retryTask` so it doesn't need to handle failing the task (and we only fail from here)?
```suggestion
      if (!event.error.unretriable && (task.status.retries ?? 0) < MAX_RETRIES) {
```

packages/api/src/task/scheduler.ts
victorges: Oh just remembered you saying the `attempts` fields was getting erased? Maybe it was never being set because of this bug 😄

packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +277,36 @@ export class TaskScheduler {
...
+  async retryTask(task: WithID<Task>, event?: messages.TaskResult) {
+    let attempts = task.status.attempts + 1 ?? 1;
+    if (attempts <= MAX_RETRY_ATTEMPTS) {
+      const status: Task["status"] = {
+        ...task.status,
+        attempts: attempts,
+      };
+
+      await this.updateTask(task, { status });
+      task.status = status;
```
victorges: You should avoid modifying objects in-place as it can lead to concurrency bugs (e.g. maybe `updateTask` is asynchronously sending the rabbitmq event for the task.udpated webhook, then you could be fiddling with the object in the middle of the serialization here).

So you should always give preference to creating new objects with updated fields (using spread syntax `task = { ...task, status }`) instead, or better yet, WDYT of having `updateTask` return the updated task? This will be especially useful if `updateTask` starts doing the merging of the `status` instead of just setting whatever the caller sends (I suggested that in the other comment you asked about the repeated `...task.status`).

Then this is just
```suggestion
      task = await this.updateTask(task, { status });
```

victorges: (this whole scheduler logic always reminds me that we need some SQL transactions here 🥲 )

packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +277,36 @@ export class TaskScheduler {
...
+
+      if (attempts > 1) {
+        // No timeout at first retry, 15 seconds * attempt after that
+        let retryDelay = attempts * TASK_RETRY_DELAY_COEFFICIENT;
+        await new Promise((resolve) => setTimeout(resolve, retryDelay));
```
victorges: https://github.com/livepeer/livepeer-com/blob/88236319556da6cf0c1c4f0dc9fa3b89d9c87a0c/packages/api/src/util.ts#L24
```suggestion
        await sleep(retryDelay);
```

packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +277,36 @@ export class TaskScheduler {
...
+      }
+
+      await this.enqueueTask(task);
+      return true;
```
victorges, line 298: I don't think these return values are used. If really not, should remove (`return false` below as well). For a moment I was confused thinking this was the "ack or nack the event" boolean value, but it is being completely ignored.

So the other rule of thumb here would be to not have "extraneous" inputs nor outputs to the code. It makes things look like they do other things. If we ever do need to know on the call-site if the task was really retried or not, we can add these return values back.

packages/api/src/task/scheduler.ts
victorges: Btw I actually loved the "sleep retry", I was biased having implemented the healthcheck logic on the task-runner through delayed AMQP events, and was thinking we'd need that complexity here as well. For such short delays it seems pretty much fine to just sleep on the event and retry tho. There might be a scalability problem if we have too many of these tasks retrying tho, each retry is effectively blocking 1 API consumer from processing other events, but I believe it works great for now.

victorges (approved):
LGTM!

packages/api/src/task/scheduler.ts
gioelecerati: Oh yeah very probable! Fixing it <3

packages/api/src/task/scheduler.ts
gioelecerati: Makes much sense, updated

packages/api/src/task/scheduler.ts
gioelecerati: Yeah, the task runner should receive the failing back and fire the retry right?

packages/api/src/schema/schema.yaml
gioelecerati: Will do it in a separate pr! It's just called by the task-runner tho, currently, right? So we can handled it after this is merged

packages/api/src/task/scheduler.ts
gioelecerati: :eyes: I figured it may be possible, but in any case I am going to remove this check with the failTask from this function and handle it on top as you suggested

packages/api/src/task/scheduler.ts
victorges: The task runner fires the failing, the API fires the retry, the task-runner will pick up the retry as just another task execution.

victorges (approved):
Nice!

victorges:
Only gotta fix the tests which are failing (maybe it was failing in master? try a rebase)

packages/api/src/task/scheduler.ts
```diff
@@ -165,13 +181,19 @@ export class TaskScheduler {
...
+    output?: Task["output"]
+  ) {
     const status = {
+      ...task.status,
       phase: "failed",
       updatedAt: Date.now(),
       errorMessage: error,
     } as const;
     await this.updateTask(task, {
+      ...task.status,
```
victorges, line 196: Just noticed this... Every failed task has a buggy schema now, with the status inlined in the task object hahah

Issue #1205, opened 2022-08-19
Fix parsing of Object Store URLs in the API


victorges:
[This](https://github.com/livepeer/go-tools/pull/1/files#diff-6167f81f9154eef1998710aea373c85e57ab3fc3f6d68f28134a367daf999bb3L238
) change in the `go-tools` Object Store library caused a backward
incompatible update in the format of Object Store URLs. Previously, they 
allowed 1 or 2 path segments, the first one meaning the region and the 
second one the bucket, or just the bucket if no region is provided. The
new version does not allow a region to be provided and only allow a bucket
to be defined.

[The logic in the API to parse those URLs](https://github.com/livepeer/livepeer-com/blob/6b86fb011c94e9c6dee5c9837110e0bdf7f08663/packages/api/src/controllers/helpers.ts#L128) was implemented as simple as
possible, and so we only added support to the exact format we used for the
VOD Object Store.

In order to update `task-runner` with the latest version of `go-tools` lib, we
will need to fix the OS URLs in the database so they don't contain the region.
But to be able to change that in the DB, we also need to change the API logic 
first to support parsing only 1 path segment, as it now requires 2.

This should be a simple fix but I'm writing EOD on a rush before a PTO cause
I can't get started on this right now. Worst case this is just a reminder for myself
next week.

victorges:
cc @cyberj0g what we were discussing on discord

cyberj0g:
Good catch. The root issue seem to be not fully abstracting away object store operations and not having test coverage of these parts. The fix is probably to add some methods to `go-tools`, e.g. signature-related, so users don't need to have any knowledge of URL structure and specific OS implementation. When refactoring `go-tools`, I didn't know the URLs are parsed somewhere else in the code. @victorges could you help me figure out the list of updates to `go-tools` interface?

PR #1228, opened 2022-08-23
api: FIx object store URL parsing

victorges:
Fixes the API logic for parsing Object Store URLs, currently used only for the 
VOD APIs for uploading files (both `direct` and `tus`).

The fix is necessary because the [Go library](https://github.com/livepeer/go-tools) for parsing these URLs changed
in an incompatible way for the Catalyst VOD work to flow through. Specifically, it does
not support having regions in the URLs anymore because it now supports specifying a path
prefix after the bucket name.

 - Create a helper function to parse these URLs so we don't need to keep repeating the logic
 - Update it to allow URLs to be specified without the region
 - Fix and add some tests

## -

 - [x] `yarn test`
 - [x] Deploy on staging, update OS URLs and ensure upload works

Implements #1205 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -629,7 +629,7 @@ export const setupTus = async (objectStoreId: string): Promise<void> => {
 
 async function createTusServer(objectStoreId: string) {
   const s3config = await getObjectStoreS3Config(objectStoreId);
-  const opts: tus.S3StoreOptions | S3ClientConfig = {
+  const opts: tus.S3StoreOptions & S3ClientConfig = {
     ...s3config,
     path: "/upload/tus",
     partSize: 8 * 1024 * 1024,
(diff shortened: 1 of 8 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -627,24 +628,12 @@ export const setupTus = async (objectStoreId: string): Promise<void> => {
...
-  const url = new URL(os.url);
-  const [_, vodRegion, vodBucket] = url.pathname.split("/");
-  let protocol = url.protocol;
-  if (protocol.includes("+")) {
-    protocol = protocol.split("+")[1];
-  }
-
+  const s3config = await getObjectStoreS3Config(objectStoreId);
   const opts: tus.S3StoreOptions | S3ClientConfig = {
+    ...s3config,
```
gioelecerati, line 633: ❤️

gioelecerati (commented):
For some reason tests are failing for me on `playback`, `asset` and `auth`

packages/api/src/controllers/playback.test.ts
```diff
@@ -45,7 +45,7 @@ describe("controllers/playback", () => {
     beforeEach(async () => {
       await db.objectStore.create({
         id: "mock_vod_store",
-        url: "http://user:password@localhost:8080/us-east-1/vod",
+        url: "s3+http://user:password@localhost:8080/us-east-1/vod",
```
gioelecerati, line 48: ooooh I see

gioelecerati (approved):
LGTM!

victorges:
All good.

PR #1233, opened 2022-08-26
validateAssetMeta merges any existing asset meta

### LINKED ISSUE #1229: Updating meta in VoD overrides existing fields
**Describe the bug (required)** A clear and concise description of what the bug
is.
When updating the metadata field for an asset, existing metadata fields get replaced by the updated ones.

**Expected behavior (required)** A clear and concise description of what you
expected to happen.

Expect to only update the metadata fields provided by the user and keep existing ones

**To Reproduce (required)** Steps to reproduce the behavior:

1. Create an asset
2. Update the asset with a new meta field using `PATCH`
3. Get the asset
4. Update the same asset with a different meta field using `PATCH` again
5. Get the asset and see the first meta field gone

**Device (required):**

- Device: MacBook Pro
- OS: macOS Monterey v 12.5
- Browser: Chrome
- Version: 104.0.5112.101 (Official Build) (arm64)

Shih-Yu:

Addressing the issue https://github.com/livepeer/studio/issues/1229

Change validateAssetMeta function to merge with any existing meta

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -47,10 +47,16 @@ const app = Router();
 
 const META_MAX_SIZE = 1024;
 
-function validateAssetMeta( meta: Record<string, string>, asset?:WithID<Asset> ) {
-  meta = _({ ...asset.meta, ...meta })
-            .omit(_.isNull)
-            .value();
+function validateAssetMeta(
+  meta: Record<string, string>,
+  asset?: WithID<Asset>
+) {
+  if (!meta) {
+    return meta;
+  }
+  meta = _({ ...asset?.meta, ...meta })
+    .omitBy(_.isNull)
+    .value();
   try {
     if (meta && JSON.stringify(meta).length > META_MAX_SIZE) {
       console.error(`provided meta exceeds max size of ${META_MAX_SIZE}`);
@@ -87,7 +93,6 @@ async function validateAssetPayload(
   defaultObjectStoreId: string,
   payload: NewAssetPayload
 ): Promise<WithID<Asset>> {
-  validateAssetMeta(payload.meta);
   if (payload.objectStoreId) {
     if (payload.objectStoreId !== defaultObjectStoreId) {
       // TODO: Allow assets Object Store to be changed at some point.
@@ -113,7 +118,7 @@ async function validateAssetPayload(
       updatedAt: createdAt,
     },
     name: payload.name,
-    meta: payload.meta,
+    meta: validateAssetMeta(payload.meta),
     objectStoreId: payload.objectStoreId || defaultObjectStoreId,
   };
 }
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -47,7 +47,10 @@ const app = Router();
 
 const META_MAX_SIZE = 1024;
 
-function validateAssetMeta(meta: Record<string, string>) {
+function validateAssetMeta( meta: Record<string, string>, asset?:WithID<Asset> ) {
```
victorges: Apparently the prettier is still not working on your env. Run `yarn prettier` manually from the project root instead.

packages/api/src/controllers/asset.ts
```diff
@@ -47,7 +47,10 @@ const app = Router();
...
+  meta = _({ ...asset.meta, ...meta })
```
victorges: The asset is not always available so this needs to be `?.`
```suggestion
  meta = _({ ...asset?.meta, ...meta })
```

packages/api/src/controllers/asset.ts
```diff
@@ -47,7 +47,10 @@ const app = Router();
...
+            .omit(_.isNull)
```
victorges: This does not "compile". You need to fix your typescript env and pay attention to the errors it shows. The correct function to call here is `omitBy` (but you also need to test it)
Screen Shot 2022-08-26 at 17 38 33

packages/api/src/controllers/asset.ts
```diff
@@ -816,12 +820,13 @@ app.patch(
...
     // update a specific asset
     const { id } = req.params;
-    const asset = await db.asset.get(id);
+    const asset = await db.asset.get( id );
     if (!asset || (asset.userId !== req.user.id && !req.user.admin)) {
       throw new NotFoundError(`asset not found`);
     } else if (asset.status.phase !== "ready") {
       throw new UnprocessableEntityError(`asset is not ready`);
     }
+    validateAssetMeta(meta, asset);
```
victorges: The return value of this function is being ignored. It is also not the only place where this is called. Anytime you change the behavior of a function you need to update everywhere it is used. So keep that in mind and look for the other usages in the code.

victorges (changes requested):
You should also add some unit tests for this feature as I believe you have not exercised the code to make sure it works (right now it does not). 

The tests would go around here (as separate `it`s): https://github.com/livepeer/livepeer-com/blob/845e074c9f755e876bb23db83b374901d4c68842/packages/api/src/controllers/asset.test.ts#L218

You should add some cases like: 
 - Allows adding `meta` to an asset
 - Allows adding fields to an existing `meta` in an `asset`
 - Allows removing a filed from the `meta` by setting it to `null`
 - Disallows any update that would result in a meta greater than that `META_MAX_SIZE` (incl if it only hits max size after the merge)

packages/api/src/controllers/asset.ts
victorges: Actually... This function is **already** being called in this API handler. You should call it only **once**. No reason to call it again. Remove this line and edit the original call instead.

victorges (approved):
LGTM!

PR #1240, opened 2022-08-30
api: Improve reliability of AMQP publishes

victorges:
This bumps the `amqp-connection-manager` lib version to get a couple
of new features and hopefully bug fixes.

We can now define a publish timeout, so we'll get an exception instead 
of hanging forever when a message publish fails.

We were also more than a major behind, so we'll hopefully get a couple of 
bug fixes and improvements as well.

 - Bump lib
 - Fix usage
 - Add default `publishTimeout`

## -

`yarn test`, staging

Many bug reports where tasks simply hang. This will hopefully turn them at least into an explicit error instead.

**Checklist:**

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -56,7 +56,7 @@
     "@types/cors": "^2.8.12",
     "ajv": "^6.10.0",
     "ajv-cli": "^3.1.0",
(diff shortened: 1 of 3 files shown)


gioelecerati (approved):
LGTM

PR #1244, opened 2022-09-02
api: Propagate progress field to asset

victorges:

Also add processing phase to assets so one doesn't have
to use the existence of the progress field to tell if it
has started or not.

 - Add `progress` field to assets
 - Add `processing` phase to assets

## -

`yarn test`

Unofficial issue discussed yesterday for SDK to unblock https://github.com/livepeer/livepeer.js/pull/27

**Checklist:**

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -16,8 +16,8 @@ const taskInfo = (task: Task): messages.TaskInfo => ({
   snapshot: task,
 });
 
-const MAX_RETRIES = 4;
-const TASK_RETRY_BASE_DELAY = 15 * 1000;
+const MAX_RETRIES = 2;
+const TASK_RETRY_BASE_DELAY = 30 * 1000;
 
 export class TaskScheduler {
   queue: Queue;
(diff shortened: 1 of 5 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -174,6 +177,8 @@ export class TaskScheduler {
       status: {
         ...task.status,
         phase: "completed",
+        progress: undefined,
+        step: undefined,
```
victorges: Need to figure out how to do this more cleanly instead of clearing the fields on every fn place we update task or asset status. Maybe some built-in logic in `updateTask` and `updateAsset`? Or always making the updates with all fields specified, instead of doing the `...` merge? Open to suggestions.

victorges: Thinking with myself, I think I'm gonna go with stopping the merge madness and just declare the full expected status on every update. Easier to reason about.

victorges: Done!

gioelecerati (approved):
With this pull request, I am unable to open the asset page on the dashboard in `dev`. It seems to be related to the `views` api failing tho. Apart of this, LGTM

PR #1258, opened 2022-09-08
Added id fetch from req.params to webhook PUT

micohasanen:
Adding the ID to the request body was not implied in the docs, so hopefully saving some trouble from future devs implementing webhooks!


0xcadams:
closing this because it is stale

PR #1260, opened 2022-09-09
Migrated from video.js to livepeer.js's player

### LINKED ISSUE #1227: Replace player in Livepeer Studio Dashboard with Livepeer Player

clacladev:
Migrated from video.js to livepeer.js's React VideoPlayer 

- Removed video.js
- Added livepeer.js and integrated StudioConfig provider
- Refactored the Player component to use livepeer.js's video player instead of video.js one

## -

Created a preview build with the changes and opened the an asset (this one requires admin privileges):
https://livepeer-studio-mdg4ql6nn-livepeer.vercel.app/dashboard/assets/6aa5e390-8724-4c5b-855b-4f2992cd1b40

Fixes #1227 

**Screenshots (optional):**

Screenshot 2022-09-09 at 14 32 40

**Checklist:**

- [ ] Test with live stream recordings

### DIFF
--- a/packages/www/pages/_app.tsx
+++ b/packages/www/pages/_app.tsx
@@ -16,6 +16,8 @@ import {
 } from "@livepeer/design-system";
 import { ThemeProvider } from "next-themes";
 import { DEFAULT_THEME } from "../lib/theme";
+import { createReactClient, LivepeerConfig } from "@livepeer/react";
+import { studioProvider } from "livepeer/providers/studio";
 
 const queryClient = new QueryClient();
 
@@ -54,6 +56,10 @@ Object.keys(themes).map(
   (key, _index) => (themeMap[themes[key].className] = themes[key].className)
 );
 
+const livepeerClient = createReactClient({
+  provider: studioProvider({ apiKey: "b05094e0-05b6-4260-b462-bae581fdcf13" }), // TODO: replace with env var
+});
+
 const App = ({ Component, pageProps }) => {
   globalStyles();
   return (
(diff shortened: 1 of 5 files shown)


packages/www/pages/_app.tsx
```diff
@@ -54,6 +56,10 @@ Object.keys(themes).map(
   (key, _index) => (themeMap[themes[key].className] = themes[key].className)
 );
 
+const livepeerClient = createReactClient({
+  provider: studioProvider({ apiKey: "b05094e0-05b6-4260-b462-bae581fdcf13" }), // TODO: replace with env var
```
victorges: Do we really need any custom API key here? The `playbackInfo` API is public, and even if it wasn't, we should not need any API key to use the player receiving only an HLS URL.

We can very likely just leave the default API key here in case the player just needs some `LivepeerProvider` to be present (ideally it should not). Worst case, if we ever do need to call APIs with `livepeer.js`, we should not use an API key but actually the same user JWT that is used on the dashboard (check `ApiProvider`). This would give access to the account the user is logged in, not a random fixed account that we insert an API key here (worse, we'd likely need to put an **admin** API key here and we definitely don't want to leak that anywhere, ever).

clacladev: Ok I see what you're saying. Let me check if I can successfully initialise the provider without the api key

clacladev: I confirm that we can initialise it without any api key and there are not warnings or errors. So I will follow that

adamsoffer (changes requested):
@clacladev small request: can we make sure the border bottom radius doesn't get cut off? Giving the video element a `display: flex` should fix.

CleanShot 2022-09-09 at 12 35 25@2x

packages/www/pages/_app.tsx
0xcadams: Yeah, we shouldn't need an API key here to just use the player. If we are planning on using any other hooks from livepeer.js, we'd need to allow the client to provide a JWT. This might be a good feature though!

adamsoffer: Yeah this would be a nice feature. I think it'd actually be worthwhile refactoring the dashboard frontend to use livepeer.js at some point.

adamsoffer:
I'm going to push this to prod. Can address my small comment in a separate PR.

clacladev:
@adamsoffer yes I see the small corner issue. I can address it in a small PR asap

PR #1271, opened 2022-09-12
api: Allow custom OSes to be used in VOD

victorges:
This is to make a few changes in the API to allow Storj to be used for VOD
with limited capabilities (direct and resumable uploads are currently broken).

Allow using custom OS on assets and recordings.

## -

Uploaded assets to Storj!!
https://www.loom.com/share/9dcdebc0f2dc463a9beab1ac03a8b4d1

Implements #1267 

**Screenshots (optional):**
https://www.loom.com/share/9dcdebc0f2dc463a9beab1ac03a8b4d1

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -35,6 +35,7 @@ import {
   ExportTaskParams,
   IpfsFileInfo,
   NewAssetPayload,
+  ObjectStore,
   Task,
 } from "../schema/types";
 import { WithID } from "../store/types";
@@ -42,6 +43,7 @@ import Queue from "../store/queue";
 import taskScheduler from "../task/scheduler";
 import { S3ClientConfig } from "@aws-sdk/client-s3";
 import os from "os";
+import { CliArgs } from "../parse-cli";
 
 const app = Router();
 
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -288,15 +310,15 @@ app.use(
     };
 
     if (Array.isArray(data)) {
-      return data.map(toExternalAsset);
+      return Promise.all(data.map(toExternalAsset));
     }
     if ("id" in data) {
       return toExternalAsset(data);
```
gioelecerati, line 316: Shouldn't also this be `await toExternalAsset(data)` ?

gioelecerati (approved):
LGTM

packages/api/src/controllers/asset.ts
victorges: On JS you can wrap promises like `Promise<Promise<Promise<string>>>` and they will still work the same. 
i.e. you can also return a promise from an async function, and that will still work for the caller. 

note1 : TypeScript would complain if this wasn't valid, so we have an extra guarantee there 😌 
note2: Maybe the "promise unwrap" only happens for `async` functions. Maybe if you actually have a variable that is a `Promise<Promise<string>>` you'd have to manually unwrap it into a `Promise<string>` somehow. IDK. Maybe it's just magic tho lol

victorges:
Closing this as we don't really want the `playbackRecordingObjectStoreId`. We will be migrating to catalyst where everything is under the same OS. We might need a part of this later (using OS public URL for playback for external OSes), but it's too small of a change to include this whole PR, we can refer back to this when we want to implement that.

PR #1275, opened 2022-09-14
api: Create access-control API for signing keys

### LINKED ISSUE #1284: Create signing-keys CRUD API
This involves actually generated the private/public key pair with a nice algorithm.

gioelecerati:

Creates a set of API for signing keys, as first milestone toward access control

- Create signing key schema
- POST create up to 10 signing keys, retrieving the private key if the keypair only on creation
- GET retrieve a list of your existing signing keys, with only the public key
- DELETE a signing key
- UPDATE a signing key, disabling it
- Create a EC P-256 keypair in `pem` format for JWT generations
- Tests

yarn test - postman

Fixes #1284 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/signing-key.test.ts
+++ b/packages/api/src/controllers/signing-key.test.ts
@@ -46,7 +46,8 @@ describe("controllers/signing-key", () => {
     let otherUserToken: string;
     let signingKey: WithID<SigningKey>;
     let samplePrivateKey: string;
-    let otherSigningKey: WithID<SigningKey>;
+    let decodedPublicKey: string;
+    let otherPublicKey: string;
 
     beforeEach(async () => {
       ({ client, adminUser, adminToken, nonAdminUser, nonAdminToken } =
@@ -58,9 +59,15 @@ describe("controllers/signing-key", () => {
       samplePrivateKey = created.privateKey;
       let res = await client.get(`/signing-key/${created.id}`);
       signingKey = await res.json();
+      // decoded public key is decoded b64 of signingKey.publicKey parsed as json
+      decodedPublicKey = Buffer.from(signingKey.publicKey, "base64").toString();
       created = await (await client.post("/signing-key")).json();
       res = await client.get(`/signing-key/${created.id}`);
-      otherSigningKey = await res.json();
+      let otherSigningKey = await res.json();
+      otherPublicKey = Buffer.from(
+        otherSigningKey.publicKey,
+        "base64"
+      ).toString();
     });
 
     it("should create a signing key and display the private key only on creation", async () => {
@@ -90,19 +97,17 @@ describe("controllers/signing-key", () => {
     it("should create a JWT using the private key and verify it with the public key", async () => {
       const expiration = Math.floor(Date.now() / 1000) + 1000;
       const payload: JwtPayload = {
-        sub: "4815162342",
+        sub: "b0dcxvwml48mxt2s",
+        action: "pull",
         name: "Satoshi Nakamoto",
-        admin: false,
-        signature: "0x4815162342",
         pub: signingKey.publicKey,
         exp: expiration,
+        video: "none",
       };
-      const token = jwt.sign(payload, samplePrivateKey, { algorithm: "RS256" });
-      const decoded = verifyJwt(token, signingKey.publicKey, {
-        complete: true,
-      });
-      expect(decoded.payload["exp"]).toEqual(expiration);
-      expect(() => jwt.verify(token, otherSigningKey.publicKey)).toThrow(
+      const token = jwt.sign(payload, samplePrivateKey, { algorithm: "ES256" });
+
+      expect(() => jwt.verify(token, decodedPublicKey)).not.toThrow();
+      expect(() => jwt.verify(token, otherPublicKey)).toThrow(
         JsonWebTokenError
       );
     });
--- a/packages/api/src/controllers/signing-key.test.ts
+++ b/packages/api/src/controllers/signing-key.test.ts
@@ -99,6 +99,27 @@ describe("controllers/signing-key", () => {
       expect(res.status).toBe(204);
     });
 
+    it("should allow disable and enable the signing key & change the name", async () => {
+      let res = await client.patch(`/signing-key/${signingKey.id}`, {
+        disabled: true,
+        name: "My test signing key 1",
+      });
+      expect(res.status).toBe(204);
+      res = await client.get(`/signing-key/${signingKey.id}`);
+      let updated = await res.json();
+      expect(updated.disabled).toBe(true);
+      expect(updated.name).toBe("My test signing key 1");
+      res = await client.patch(`/signing-key/${signingKey.id}`, {
+        disabled: false,
+        name: "My test signing key 2",
+      });
+      expect(res.status).toBe(204);
+      res = await client.get(`/signing-key/${signingKey.id}`);
+      updated = await res.json();
+      expect(updated.disabled).toBe(false);
+      expect(updated.name).toBe("My test signing key 2");
+    });
+
     it("shoud delete the signing key", async () => {
       let res = await client.delete(`/signing-key/${signingKey.id}`);
       expect(res.status).toBe(204);
--- a/packages/api/src/controllers/signing-key.test.ts
+++ b/packages/api/src/controllers/signing-key.test.ts
@@ -1,0 +1,80 @@
+import serverPromise, { TestServer } from "../test-server";
+import { TestClient, clearDatabase, setupUsers } from "../test-helpers";
+import { v4 as uuid } from "uuid";
+import { SigningKey, SigningKeyResponsePayload, User } from "../schema/types";
+import { db } from "../store";
+import { WithID } from "../store/types";
+
+// includes auth file tests
+
+let server: TestServer;
+let mockAdminUserInput: User;
+let mockNonAdminUserInput: User;
+
+// jest.setTimeout(70000)
+
+beforeAll(async () => {
+  server = await serverPromise;
+
+  mockAdminUserInput = {
+    email: "user_admin@gmail.com",
+    password: "x".repeat(64),
+  };
+
+  mockNonAdminUserInput = {
+    email: "user_non_admin@gmail.com",
+    password: "y".repeat(64),
+  };
+});
+
+afterEach(async () => {
+  await clearDatabase(server);
+});
+
+describe("controllers/multistream-target", () => {
+  describe("basic CRUD with JWT authorization", () => {
+    let client: TestClient;
+    let adminUser: User;
+    let adminToken: string;
+    let nonAdminUser: User;
+    let nonAdminToken: string;
+    let signingKey: WithID<SigningKey>;
+
+    beforeEach(async () => {
+      ({ client, adminUser, adminToken, nonAdminUser, nonAdminToken } =
+        await setupUsers(server, mockAdminUserInput, mockNonAdminUserInput));
+      client.jwtAuth = nonAdminToken;
+      let signingKey: WithID<SigningKey>;
+      let created: SigningKeyResponsePayload = await (
+        await client.post("/signing-key")
+      ).json();
+      let res = await client.get(`/signing-key/${created.publicKey.id}`);
+      signingKey = await res.json();
+    });
+
+    it("should create a signing key and display the private key only on creation", async () => {
+      const preCreationTime = Date.now();
+      let res = await client.post("/signing-key");
+      expect(res.status).toBe(201);
+      const created = (await res.json()) as SigningKeyResponsePayload;
+      expect(created.privateKey).toBeDefined();
+      expect(created.publicKey).toBeDefined();
+      expect(created.publicKey.id).toBeDefined();
+      expect(created.publicKey.createdAt).toBeGreaterThanOrEqual(
+        preCreationTime
+      );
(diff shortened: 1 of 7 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1523,6 +1525,42 @@ components:
...
+            Timestamp (in milliseconds) at which the signing-key was created
+          example: 1587667174725
+        userId:
+          type: string
+          readOnly: true
+          index: true
+          example: 78df0075-b5f3-4683-a618-1086faca35dc
+        publicKey:
+          type: string
+        deleted:
```
victorges, line 1577: Reminder to also create the `disabled` field which will be important for users to rotate keys safely.

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    let samplePrivateKey: string;
+    let otherSigningKey: WithID<SigningKey>;
+
+    beforeEach(async () => {
+      ({ client, adminUser, adminToken, nonAdminUser, nonAdminToken } =
+        await setupUsers(server, mockAdminUserInput, mockNonAdminUserInput));
+      client.jwtAuth = nonAdminToken;
+      let created: SigningKeyResponsePayload = await (
+        await client.post("/signing-key")
+      ).json();
```
victorges: **nit:** JS tip (optional ofc):
```suggestion
      let created: SigningKeyResponsePayload = await client.post("/signing-key").then(res => res.json());
```

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      samplePrivateKey = created.privateKey;
+      let res = await client.get(`/signing-key/${created.id}`);
+      signingKey = await res.json();
```
victorges: Doesn't the `post()` request return the signing-key on the payload already? If not, I think it should. Only difference would be that it has an additional privateKey.

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      const created = (await res.json()) as SigningKeyResponsePayload;
+      expect(created.privateKey).toBeDefined();
+      expect(created.publicKey).toBeDefined();
+      expect(created.id).toBeDefined();
+      expect(created.createdAt).toBeGreaterThanOrEqual(preCreationTime);
+      res = await client.get(`/signing-key/${created.id}`);
+      expect(res.status).toBe(200);
+      const getResponse = await res.json();
+      expect(getResponse.privateKey).toBeUndefined();
+      expect(getResponse.publicKey).toEqual(created.publicKey);
```
victorges: **nit:** jest tip, wdyt?:
```suggestion
      const created = (await res.json()) as SigningKeyResponsePayload;
      expect(created).toMatchObject({
          id: expect.any(String),
          privateKey: expect.any(String),
          publicKey: expect.any(String),
          createdAt: expect.any(Number),
        });
      expect(created.createdAt).toBeGreaterThanOrEqual(preCreationTime);
      res = await client.get(`/signing-key/${created.id}`);
      expect(res.status).toBe(200);
      const getResponse = await res.json();
      const { privateKey, ...withoutPvtk } = created;
      expect(getResponse).toEqual(withoutPvtk);
```

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      res = await client.delete(`/signing-key/${created.id}`);
+      expect(res.status).toBe(204);
+    });
+
+    it("should create a JWT using the private key and verify it with the public key", async () => {
+      const expiration = Math.floor(Date.now() / 1000) + 1000;
+      const payload: JwtPayload = {
+        sub: "4815162342",
+        name: "Satoshi Nakamoto",
```
victorges, line 102: Nice hahah

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      };
+      const token = jwt.sign(payload, samplePrivateKey, { algorithm: "RS256" });
+      const decoded = verifyJwt(token, signingKey.publicKey, {
+        complete: true,
+      });
+      expect(decoded.payload["exp"]).toEqual(expiration);
+      const failDecoded = verifyJwt(token, otherSigningKey.publicKey);
+      expect(failDecoded).toBeInstanceOf(JsonWebTokenError);
+      let res = await client.delete(`/signing-key/${signingKey.id}`);
+      expect(res.status).toBe(204);
```
victorges: You don't need to delete objects after every test! The `afterEach` function above already clears the entire temp test DB after each test :)

No problem keeping these tho, just letting you know

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        name: "My test signing key 2",
+      });
+      expect(res.status).toBe(204);
+      res = await client.get(`/signing-key/${signingKey.id}`);
+      updated = await res.json();
+      expect(updated.disabled).toBe(false);
+      expect(updated.name).toBe("My test signing key 2");
+    });
+
+    it("shoud delete the signing key", async () => {
```
victorges: typo
```suggestion
    it("should delete the signing key", async () => {
```

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      let res = await client.delete(`/signing-key/${signingKey.id}`);
+      expect(res.status).toBe(204);
+      res = await client.get(`/signing-key/${signingKey.id}`);
+      expect(res.status).toBe(404);
```
victorges: Can you check that the object is not fully deleted from the DB as well? You could do that by accessing the `db.signingKey` table directly here, and ensuring that the object _is_ there it just has a `deleted: true` property instead.

This would be good to make sure we don't lose these keys forever if users delete them.

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+const fieldsMap: FieldsMap = {
+  id: `signing_key.ID`,
+  name: { val: `signing_key.data->>'name'`, type: "full-text" },
+  deleted: { val: `signing_key.data->'deleted'`, type: "boolean" },
+  createdAt: { val: `signing_key.data->'createdAt'`, type: "int" },
+  userId: `signing_key.data->>'userId'`,
+};
+
+function generateSigningKeys() {
+  const keypair = generateKeyPairSync("rsa", {
```
victorges: Is there an `Async` version as well? If so, we should do that.

Sync functions block the whole runtime environment (oh node.js kkk) until they are done.

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+      type: "pkcs8",
+      format: "pem",
+    },
+  });
+  return keypair;
+}
+
+const app = Router();
+
+app.get("/", authorizer({}), async (req, res) => {
```
victorges: Maybe add a test that creates a signing key and lists all keys for this codecov comments to go away 😅 

A good test to do there is making sure users can't list keys from other users.

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+  if (!signingKey || signingKey.deleted) {
+    res.status(404);
+    return res.json({
+      errors: ["not found"],
+    });
+  }
+
+  if (req.user.admin !== true && req.user.id !== signingKey.userId) {
+    res.status(403);
+    return res.json({
```
victorges: Gosh this codecov thing is accurate!

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+      errors: ["user can only request information on their own signingKeys"],
```
victorges: WDYT making this just a `404` as well?

I have mixed feelings, and we are not consistent across all APIs, but since signing-keys are somewhat more sensitive data maybe we want to hide when it exists but belongs to another user.

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+    const id = uuid();
+    const keypair = generateSigningKeys();
+
+    if (!req.body.name || req.body.name === "") {
+      req.body.name = "Signing Key " + (output.length + 1);
+    }
+
+    var doc: WithID<SigningKey> = {
+      id,
+      name: req.body.name,
```
victorges: We should avoid mutating objects, especially the ones from the HTTP framework. This works as well and (IMO) is cleaner/clearer. WDYT?
```suggestion
    var doc: WithID<SigningKey> = {
      id,
      name: req.body.name || ("Signing Key " + (output.length + 1)),
```

Otherwise we could just have a helper variable in this function as well for the mutable data. Like
```ts
let name = req.body.name
if (!name) {
  name = "..."
}
// use name
```
(`!str` already checks both that the stream is defined and not empty! since the empty string is fals-y, it also works on the boolean operator `||` to fallback to a default)

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+      createdAt: Date.now(),
+      publicKey: keypair.publicKey,
+    };
+
+    await db.signingKey.create(doc);
+
+    var createdSigningKey: SigningKeyResponsePayload = {
+      ...doc,
+      privateKey: keypair.privateKey,
+    };
```
victorges: This is cool! Can spot very easily that the private key never goes anywhere but this response 😌

packages/api/src/controllers/signing-key.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    res.status(201);
+    res.json(createdSigningKey);
+  }
+);
+
+app.delete("/:id", authorizer({}), async (req, res) => {
+  const { id } = req.params;
+  const signingKey = await db.signingKey.get(id);
+  if (!signingKey) {
```
victorges: Gotta check if it's not already deleted as well

packages/api/src/schema/schema.yaml
```diff
@@ -1523,6 +1541,54 @@ components:
...
+        name:
+          type: string
+          description: Name of the signing key
+        createdAt:
+          type: number
+          readOnly: true
+          description:
+            Timestamp (in milliseconds) at which the signing-key was created
+          example: 1587667174725
+        lastUsedAt:
```
victorges, line 1564: This is not used yet right? Will be on the follow-up PR for the `/gating` or `/access-control` API or sth like that

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        signature: "0x4815162342",
+        exp: expiration,
+      };
+      const token = jwt.sign(payload, samplePrivateKey, { algorithm: "RS256" });
+      const decoded = verifyJwt(token, signingKey.publicKey, {
+        complete: true,
+      });
+      expect(decoded.payload["exp"]).toEqual(expiration);
+      const failDecoded = verifyJwt(token, otherSigningKey.publicKey);
+      expect(failDecoded).toBeInstanceOf(JsonWebTokenError);
```
victorges: **nit:** You don't need the `verifyJwt` helper if you use the jest-native check:
```suggestion
      expect(() => jwt.verify(token, publicKey)).toThrow(JsonWebTokenError);
```

I think this would be more self explanatory here, since it's not really a JS pattern to return errors instead of throwing them. Still optional change tho, only coding style in the end.

victorges (approved):
Awesome! Only some smaller nits or questions :)

packages/api/src/controllers/signing-key.test.ts
gioelecerati: Ahahahah this or Marty McFly, no other choices

packages/api/src/controllers/signing-key.ts
gioelecerati: Good point!

packages/api/src/schema/schema.yaml
gioelecerati: Yeah, it's just defined but not used yet!

packages/api/src/controllers/signing-key.test.ts
gioelecerati: Oh very nice!

packages/api/src/controllers/signing-key.ts
gioelecerati: Makes much more sense than mutating the req body, I added this

packages/api/src/controllers/signing-key.ts
gioelecerati: Sounds good!

packages/api/src/controllers/signing-key.test.ts
gioelecerati: Oh nice! Missed the `afterEach` function, thank you :)

packages/api/src/controllers/signing-key.test.ts
gioelecerati: Good point!

packages/api/src/controllers/signing-key.ts
gioelecerati: Done! kkk

PR #1276, opened 2022-09-14
api: Soft deletion of object store objects

victorges:
Differently from other APIs, we were not only soft deleting Object Stores when their
deletion was requested. This is bad because the references to ObjectStore objects are
made in other objects like assets and stream recordings, so we don't really want to lose
all data regarding those deleted OSes.

 - Migrate `object-store` controller to TypeScript
 - Starting using `markDeleted` instead of `delete` on object store for soft deletion behavior
 - Fix usage of object store by always checking `deleted` field (and `disabled`)

## -

`yarn test`

No.

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -858,6 +858,9 @@ components:
           description:
             If true then these object store will not be used for recording even
             if it is configured in the API command line.
+        deleted:
+          type: boolean
+          description: Set to true when the object store is deleted
         id:
           type: string
           example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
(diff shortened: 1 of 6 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -676,6 +676,19 @@ components:
         multistream:
           $ref: "#/components/schemas/stream/properties/multistream"
 
+    object-store-patch-payload:
+      type: object
+      additionalProperties: false
+      properties:
+        name:
+          $ref: "#/components/schemas/object-store/properties/name"
+        disabled:
```
gioelecerati, line 685: Nice, didn't know object stores and multistream targets had the `disabled` field, we could consider adding this as well to the signing keys for access control

gioelecerati (approved):
LGTM!

packages/api/src/schema/schema.yaml
victorges: Yes definitely!! We need to be able to disable keys before deleting them, I thought we forgot to include that in the requirements docs.

PR #1277, opened 2022-09-15
IPFS export date

### LINKED ISSUE #1253: Add upload date to IPFS integration component
Currently, the copy under "Decentralized Storage Providers" is static, which can be confusing if you've already stored an asset to IPFS. If an action was taken to pin an asset IPFS we should include an upload date so this is more clear.

clacladev:
Added the IPFS export date to the Asset page.

- Added `createdAt` property to the `asset.storage.ipfs` object in schemas
- If present, shows the formatted date in the IPFS storage box in the Asset page

Simulating locally the date being present.

Fixes #1253

**Screenshots (optional):**
Screenshot 2022-09-15 at 12 37 48

Screenshot 2022-09-15 at 12 37 56

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -999,7 +999,9 @@ components:
                 updatedAt:
                   type: number
                   readOnly: true
-                  description: Timestamp (in milliseconds) at which IPFS export task was updated
+                  description:
+                    Timestamp (in milliseconds) at which IPFS export task was
+                    updated
                   example: 1587667174725
             status:
               readOnly: true
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1211,6 +1211,11 @@ components:
           type: string
           readOnly: true
           description: URL to access file via HTTP through an IPFS gateway
+        createdAt:
+          type: number
+          readOnly: true
+          description: Timestamp (in milliseconds) at which IPFS export task was created
+          example: 1587667174725
 
     new-asset-payload:
       additionalProperties: false
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -996,6 +996,11 @@ components:
                 $ref: "#/components/schemas/ipfs-file-info/properties"
                 nftMetadata:
                   $ref: "#/components/schemas/ipfs-file-info"
+                updatedAt:
+                  type: number
+                  readOnly: true
+                  description: Timestamp (in milliseconds) at which IPFS export task was updated
+                  example: 1587667174725
             status:
               readOnly: true
               additionalProperties: false
@@ -1211,11 +1216,6 @@ components:
           type: string
           readOnly: true
           description: URL to access file via HTTP through an IPFS gateway
-        createdAt:
-          type: number
-          readOnly: true
-          description: Timestamp (in milliseconds) at which IPFS export task was created
-          example: 1587667174725
 
     new-asset-payload:
       additionalProperties: false
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -152,6 +152,7 @@ export class TaskScheduler {
               ipfs: {
                 spec: inputAsset.storage.ipfs.spec,
                 ...taskOutputToIpfsStorage(event.output.export.ipfs),
+                createdAt: task.createdAt,
               },
               status: {
                 phase: "ready",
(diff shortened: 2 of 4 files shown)


clacladev:
Tested in staging successfully
Screenshot 2022-09-15 at 16 17 52

clacladev:
@adamsoffer My implementation differs from the design in the how the date is formatted. In the end I went for the format that is already being used throughout the interface and in the same page. To keep consistency. In case it's not ok, let's discuss.

packages/api/src/schema/schema.yaml
```diff
@@ -1211,6 +1211,11 @@ components:
           type: string
           readOnly: true
           description: URL to access file via HTTP through an IPFS gateway
+        createdAt:
```
victorges: I think this field should be named `updatedAt` instead, since the `createdAt` field in the API schema right now is immutable, meaning when the object was originally created, not when they changed the last time.

In this case, the IPFS export especially of the `nftMetadata` can change multiple times. The "original file" export can also be changed, at least in terms of unpinning (`ipfs: false`) and re-pinning (`ipfs: true` again) it. 

So IMO it'd make more sense for this to be `updatedAt` instead. WDYT?

packages/api/src/task/scheduler.ts
```diff
@@ -152,6 +152,7 @@ export class TaskScheduler {
               ipfs: {
                 spec: inputAsset.storage.ipfs.spec,
                 ...taskOutputToIpfsStorage(event.output.export.ipfs),
+                createdAt: task.createdAt,
```
victorges: The `createdAt` will get the time when the `task` started, but not when it actually finished saving the file to IPFS. So we should use the task `updatedAt` here which is updated in the end of this function on line 169. We can create an `updatedAt` variable before this whole `switch` case adn then use it both here and in the update on 169.

victorges: Also, this field should be created inside `taskOutputToIpfsStorage`. It creates the `nftMetadata` field which should also have the `updatedAt` field set to the right value (not necessarily the same of the "root" IPFS export).

Another approach is not to have the field in the `nftMetadata`, and I'd be OK with that as well. But in this case you should add the field not to the `ipfs-file-info` schema, but to the `asset.storage.ipfs` schema instead. (Right below that line I linked you with the `$ref`).

packages/www/components/Dashboard/AssetDetails/AssetOverviewTab.tsx
```diff
@@ -145,6 +146,10 @@ const AssetOverviewTab = ({
               <Text size="3">IPFS</Text>
             </Flex>
 
+            {asset?.storage?.ipfs?.createdAt && (
+              <Text size="3" variant="gray">Uploaded on {moment.unix(asset?.storage?.ipfs?.createdAt / 1000).format("LLL")}</Text>
```
victorges: Cool :)

victorges (commented):
LGTM, but I think the field should be renamed to `updatedAt` and either also set on the `nftMetadata` or removed from its schema.

packages/api/src/schema/schema.yaml
clacladev: It makes sense. I used the `createdAt` because that was what I found in the task. But it makes sense to use `updatedAt` and generate it.

packages/api/src/task/scheduler.ts
clacladev: Agree on not having it inside the `nftMetadata` which seems more about the metadata per-se. So I will fix the schema.

I just disagree on defining `updatedAt = Date.now()` before the switch, given it's going to be used in different places that require the finishing of async functions. I would just fix my line with `updatedAt: Date.now()`. The same way it is currently used at 169. So both values are actually correct.

clacladev:
@victorges I made the updates to the name. I would ask you gently to review I did the right changes to the schemas

packages/api/src/task/scheduler.ts
victorges: That works as well! We just have tried using a consistent timestamp on DB updates that are conceptually a transaction (but aren't right now). All of the async operations here are DB updates (and maybe sending events, which conceptually also shouldn't change the timestamps)

When we do create transactions like that, all of these updates will effectively happen at the same time on the DB, so we used consistent timestamps since the beginning here. A side benefit is that you can also correlate the asset updated time with the task that finished at that time.

I'm fine with using the timestamp that we actually sent that specific update query tho. Just a different interpretation of the field which also makes sense.

victorges (approved):
LGTM!

packages/api/src/task/scheduler.ts
clacladev: Thanks for the explanation

PR #1286, opened 2022-09-16
api: Add playbackPolicy to assets & streams

### LINKED ISSUE #1289: Add playbackPolicy to playable objects (VOD&livestreams)
This is to make them configurable on whether they should need a signed JWT
or not and from what signing key (should we limit that?).

gioelecerati:

Add the ability to specify a playback policy to an Asset, making them `public`, `signed` or `private`
Comes with #1275

This is also applied to stream

Fixes #1289

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -649,7 +649,7 @@ components:
         meta:
           $ref: "#/components/schemas/asset/properties/meta"
         playbackPolicy:
-          $ref: "#/components/schemas/asset/properties/playbackPolicy"
+          $ref: "#/components/schemas/playback-policy"
         storage:
           additionalProperties: false
           properties:
@@ -733,6 +733,17 @@ components:
           description: Error message if the webhook failed to process the event
         response:
           $ref: "#/components/schemas/webhook-response"
+
+    playback-policy:
+      type: object
+      additionalProperties: false
+      properties:
+        enabled:
+          type: boolean
+          description: If the policy is enabled for this asset / stream
+        signingKeyId:
+          type: string
+          description: ID of the key to use for signing the playback token
 
     session:
       type: object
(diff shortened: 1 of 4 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -732,6 +738,15 @@ components:
         response:
           $ref: "#/components/schemas/webhook-response"
 
+    playback-policy:
```
victorges, line 741: WDYT of making this an object with a `type` field instead that has this `enum`? Then it could easily evolve soon to define an actual token-gating spec with some definition like
```json
"playbackPolicy": {
  "type": "tokenGated",
  "erc20": {
    "chain": "polygon",
    "contract": "0x123",
    "minBalance": 100
  }
}
```
or sth like that (token-gating schema is not the point, only having this be an object from the start).

## Devil's advocate 

The counter argument that I give to myself is that we can also not commit the same mistake of over-engineering these schemas and can just start simple instead. Maybe we'll never really implement token-gating fully on our side =P

We can always use a `oneOf` schema when we want to update this field to become an object instead, although that would still be a breaking change as it can explode client implementations out there, especially on static languages that try to parse it as a `string` and suddenly start getting an object.

So anyway, I'm not strongly opinionated on the object in order to KISS. Feel free to just ignore this idea but raising it here for consideration 😁

packages/api/src/schema/schema.yaml
```diff
@@ -964,6 +979,8 @@ components:
           readOnly: true
           example: https://livepeercdn.com/asset/ea03f37e-f861-4cdd-b495-0e60b6d753ad/index.m3u8
           description: URL for HLS playback
+        playbackPolicy:
+          $ref: "#/components/schemas/playback-policy"
```
victorges, line 985: `$` 🧐

victorges (approved):
LGTM!

packages/api/src/schema/schema.yaml
gioelecerati: wait what am I missing :eyes: isn't the $ right there?

packages/api/src/schema/schema.yaml
gioelecerati: So, are you proposing to make something like
```
"playbackPolicy":{
     "kind":"signed"
}
```
for now, where `kind` is the enum and we can expand later? Sounds good to me, on first iteration when I was putting the signingKey ids inside the policy it was actually like that. wdyt of the object like this?

victorges:
Sounds great!

victorges:
Yeah! Twas just a historical joke hahaha

victorges:
Turns out GitHub doesn't send the replies in the right conversation lol

packages/api/src/schema/schema.yaml
victorges: Looks great!

packages/api/src/schema/schema.yaml
victorges: Yeah it is, twas just a historical joke lol

PR #1291, opened 2022-09-19
www: Fix asset details page viewership components

victorges:

This is to fix the frontend for the viewership API, which broke after the last change,
but more importantly than that, shouldn't break the asset detail page anyhow.

 - Make views an optional component in the page, only if available
 - Fetch views by asset ID instead of playback ID

## -

Access viewership page and ensure it works again.

Fixes issue just found on deploy.

**Checklist:**


adamsoffer (approved):
:shipit:

PR #1295, opened 2022-09-22
Added Embed player in Asset and Stream page

### LINKED ISSUE #1282: Embeddable player in Dashboard
Feedback on On Demand. Context: https://app.zenhub.com/workspaces/studio-62e817e78dcba15ddb23d5ad/issues/livepeer/studio/1227

@adamsoffer can you update this ticket on any progress related to the embeddable player code in the dashboard? Curious how far we got.

clacladev:
- Added `Download` button below the player inside the Asset details page
- Added the `Share` menu, which then allows `Copy link` and `Embed` dialog, in Assets and Streams details pages

- I refactored and broken down the `/dashboard/assets/:id` and `/dashboard/streams/:id` page in multiple smaller components. This refactoring was needed to keep the complexity of these pages manageable
- I extracted the `ClipButton` which was implemented a 3 different times in the page in a reusable component. 

- I took screenshots before and after and they match
- I tested light and dark mode
- I tested the different copy buttons and checked the values
- I tested on both Chrome and Safari

⚠️ The asset and stream details page needs to be tested for some edge cases I may have missed by a second person ⚠️

Fixes #1282 

**Screenshots (optional):**
The design:
Screenshot 2022-09-21 at 13 11 13

The implementation:
Screenshot 2022-09-21 at 13 10 36

**Checklist:**
- [x] Confirm that the refactor ideas is ok
- [x] The asset and stream details page needs to be tested for some edge cases I may have missed by a second person
- [x] The Share, Copy and Embed icons are slightly different, because I used the ones included in our design system.
- [x] The position of the Popover, when pressing Share button, is below the video. In the designs it’s on top. Currently the Popover in the design system does not support the forcing of the position. If we want it, it requires extra work
- [x] The Copy Link button copies a link like [https://lvpr.tv?v=93363tya76n5paaf](https://lvpr.tv/?v=93363tya76n5paaf) . Is this the intended behaviour?
- [x] The stream page does not have a download button. Can we actually allow a stream download? How?
- [x] @adamsoffer can you confirm these points are ok?

### DIFF
--- a/packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
+++ b/packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
@@ -12,9 +12,9 @@ import {
 import { Asset } from "livepeer";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 
-const embedStringForAsset = (asset?: Asset) => `
+const embedStringForAsset = (playbackId: string) => `
 <iframe
-src="https://lvpr.tv?v=${asset?.playbackId}"
+src="https://lvpr.tv?v=${playbackId}"
 frameborder="0"
 allowfullscreen
 allow="autoplay; encrypted-media; picture-in-picture"
@@ -25,15 +25,15 @@ sandbox="allow-scripts">
 export type EmbedVideoDialog = {
   isOpen: boolean;
   onOpenChange: (isOpen: boolean) => void;
-  asset?: Asset;
+  playbackId: string;
 };
 
 const EmbedVideoDialog = ({
   isOpen,
   onOpenChange,
-  asset,
+  playbackId,
 }: EmbedVideoDialog) => {
-  const embedString = embedStringForAsset(asset);
+  const embedString = embedStringForAsset(playbackId);
   const [openSnackbar] = useSnackbar();
 
   const onCopy = () => {
--- a/packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
+++ b/packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
@@ -1,4 +1,14 @@
-import { Dialog, DialogContent, DialogTitle, Heading, Flex, Button, Text, TextField, useSnackbar } from "@livepeer/design-system";
+import {
+  Dialog,
+  DialogContent,
+  DialogTitle,
+  Heading,
+  Flex,
+  Button,
+  Text,
+  TextField,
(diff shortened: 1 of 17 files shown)


packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
```diff
@@ -0,0 +1,92 @@
...
+  Text,
+  TextField,
+  useSnackbar,
+} from "@livepeer/design-system";
+import { Asset } from "livepeer";
+import { CopyToClipboard } from "react-copy-to-clipboard";
+
+const embedStringForAsset = (playbackId: string) => `
+<iframe
+src="https://lvpr.tv?v=${playbackId}"
```
victorges: Regarding linking to the lvpr.tv player, I think the only real problem from it not using the livepeer.js player yet is that it will not publish any viewership metrics when used, which IMO is highly desired feature. We should update it ASAP to use the SDK player instead. 

@adamsoffer I mentioned today that I could update the player to use the SDK player, but after the day today I'm actually not sure when I'll have bandwidth to do that. It'd also involve setting up React from scratch on the project (hopefully a really minimal version of it) and ideally find a way to download/embedded it statically on the IPFS directory, not to have the IPFS website depending on external resources. Do you think anyone from UX team could pick up on that? I can provide all support on the existing stuff in the project, deploying on fleek, etc

victorges (commented):
Tested the preview and looks awesome! Only suggestion I have is to remove the `Download` button from the top right as well (not below the player) so we don't have the same thing in multiple places. WDYT?

clacladev:
@victorges Personally, I agree to remove the top-right Download button. It's redundant and also it's not the main CTA, so there's no good reason to have it highlighted IMO. @adamsoffer can I go ahead?

adamsoffer:
@clacladev agree let's remove the download button on the top right.

clacladev:
@adamsoffer @victorges Removed the redundant Download button. Whenever you can review it, then please just approve the PR so I can merge. Cheers 🙌

packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
adamsoffer: No problem! I can pick that up 👍

adamsoffer (approved):
Looks awesome. I just pushed a couple very minor fixes (removed the underline text decoration download button and fixed the iframe string format).

:shipit:

PR #1317, opened 2022-09-29
accessControl: Add /gate endpoint for Catalyst integration

### LINKED ISSUE #1285: Create API to validate access to a playback request
API is a little different than originally designed.

## Outdated description

There must also be a way for catalyst to verify that whatever object the
JWT is giving access to is  owned by the same user that owns the signing key.

This also includes defining what is the protocol for the JWT, like what
fields are required, optional, etc. Mux as an example: https://docs.mux.com/guides/video/secure-video-playback#4-generate-a-json-web-token-jwt

This should also update a `lastUsedAt` / `lastSeen` field in the signing keys.

### LINKED ISSUE #1288: Integrate with catalyst playback hook
Theoretically it should just work, but we know how these go. This means tracking development on the Catalyst side and ensuring the APIs are compatible, testing them together etc.

gioelecerati:

Creates a `/access-control/gate` endpoint for Catalyst integration

- Allow to pass a `stream` or an `asset` playbackId to the endpoint
- The response can be either `2XX` or `4XX` depending on the existence of the content, whether the publicKey is valid and owned by the same `User` that own the content

Condition | Status Code | Allowed
-- | -- | --
Content is not gated | 204 No Content | YES
Content is gated and JWT is valid | 204 No Content | YES
Content not found | 404 Not found | NO
Content is gated, Signing Key does not exist | 403 Forbidden | NO
Content is gated, signing key and content don’t share the same owner | 403 Forbidden | NO
Content is gated, signing key is disabled | 403 Forbidden | NO
Content does not have a playback Policy | 204 No Content | YES

- Tests for all the possible cases

yarn test

Fixes #1285 
Partially Fixes #1288 

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -100,12 +100,13 @@ describe("controllers/signing-key", () => {
     });
 
     it("should not allow playback on not existing streams or public keys", async () => {
+      client.jwtAuth = adminToken;
       const res = await client.post("/access-control/gate", {
-        stream: `video+0000000000`,
+        stream: `video+0000000000123`,
         pub: signingKey.publicKey,
         type: "jwt",
       });
-      expect(res.status).toBe(403);
+      expect(res.status).toBe(404);
       const res2 = await client.post("/access-control/gate", {
         stream: `video+${gatedPlaybackId}`,
         pub: "00000000000000",
@@ -115,6 +116,7 @@ describe("controllers/signing-key", () => {
     });
 
     it("should not allow playback if stream is gated an pub is missing", async () => {
+      client.jwtAuth = adminToken;
       const res2 = await client.post("/access-control/gate", {
         stream: `video+${gatedPlaybackId}`,
         type: "jwt",
@@ -123,6 +125,7 @@ describe("controllers/signing-key", () => {
     });
 
     it("should not allow playback if stream and public key does not share the same owner", async () => {
+      client.jwtAuth = adminToken;
       const otherStream = await client.post("/stream", {
         name: "test",
         playbackPolicy: {
@@ -170,6 +173,32 @@ describe("controllers/signing-key", () => {
         pub: signingKey.publicKey,
       });
       expect(res.status).toBe(204);
+    });
+
+    it("should not allow playback if signing key is deleted or disabled", async () => {
+      client.jwtAuth = nonAdminToken;
+      await client.patch(`/access-control/signing-key/${signingKey.id}`, {
+        disabled: true,
+      });
+      client.jwtAuth = adminToken;
+      const res = await client.post("/access-control/gate", {
+        stream: `video+${gatedPlaybackId}`,
+        type: "jwt",
+        pub: signingKey.publicKey,
+      });
+      expect(res.status).toBe(403);
+      client.jwtAuth = nonAdminToken;
+      await client.patch(`/access-control/signing-key/${signingKey.id}`, {
+        disabled: false,
+      });
+      await client.delete(`/access-control/signing-key/${signingKey.id}`);
+      client.jwtAuth = adminToken;
+      const res2 = await client.post("/access-control/gate", {
+        stream: `video+${gatedPlaybackId}`,
+        type: "jwt",
+        pub: signingKey.publicKey,
+      });
+      expect(res2.status).toBe(403);
     });
   });
 });
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -75,7 +75,7 @@ describe("controllers/signing-key", () => {
         pub: signingKey.publicKey,
         type: "jwt",
       });
-      expect(res2.status).toBe(200);
+      expect(res2.status).toBe(204);
     });
 
     it("should not allow playback on not existing streams or public keys", async () => {
@@ -122,7 +122,7 @@ describe("controllers/signing-key", () => {
         stream: `video+${publicPlaybackId}`,
         type: "jwt",
       });
-      expect(res.status).toBe(200);
+      expect(res.status).toBe(204);
     });
 
     it("should allow playback if playbackPolicy is not specified", async () => {
@@ -135,7 +135,7 @@ describe("controllers/signing-key", () => {
         type: "jwt",
         pub: signingKey.publicKey,
       });
-      expect(res.status).toBe(200);
+      expect(res.status).toBe(204);
     });
   });
 });
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -40,7 +40,6 @@ describe("controllers/signing-key", () => {
     let adminToken: string;
     let nonAdminUser: User;
     let nonAdminToken: string;
-    let otherUserToken: string;
     let signingKey: WithID<SigningKey>;
     let gatedPlaybackId: string;
     let publicPlaybackId: string;
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -141,13 +141,24 @@ describe("controllers/signing-key", () => {
       expect(res.status).toBe(403);
     });
 
-    it("should allow playback on public playbackId", async () => {
+    it("should allow playback on public playbackId with and without a public key provided", async () => {
       client.jwtAuth = adminToken;
       const res = await client.post("/access-control/gate", {
         stream: `video+${publicPlaybackId}`,
         type: "jwt",
       });
       expect(res.status).toBe(204);
+      client.jwtAuth = nonAdminToken;
+      const stream = await client.post("/stream", {
+        name: "test",
+      });
+      client.jwtAuth = adminToken;
+      const streamPlaybackId = (await stream.json()).playbackId;
+      const res2 = await client.post("/access-control/gate", {
+        stream: `video+${streamPlaybackId}`,
+        type: "jwt",
+      });
+      expect(res2.status).toBe(204);
     });
 
     it("should allow playback if playbackPolicy is not specified", async () => {
@@ -173,6 +184,24 @@ describe("controllers/signing-key", () => {
         pub: signingKey.publicKey,
       });
       expect(res.status).toBe(204);
+    });
+
+    it("should not allow playback if user is suspended", async () => {
+      client.jwtAuth = adminToken;
+      const res = await client.post("/access-control/gate", {
+        stream: `video+${gatedAsset.playbackId}`,
+        type: "jwt",
+        pub: signingKey.publicKey,
+      });
+      expect(res.status).toBe(204);
+      await db.user.update(gatedAsset.userId, { suspended: true });
+      const res2 = await client.post("/access-control/gate", {
+        stream: `video+${gatedAsset.playbackId}`,
+        type: "jwt",
+        pub: signingKey.publicKey,
+      });
+      expect(res2.status).toBe(403);
+      await db.user.update(gatedAsset.userId, { suspended: false });
     });
 
     it("should not allow playback if signing key is deleted or disabled", async () => {
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -58,7 +58,7 @@ describe("controllers/signing-key", () => {
       const res2 = await client.post("/stream", {
         name: "test",
         playbackPolicy: {
-          type: "signed",
+          type: "jwt",
         },
       });
       const stream = await res2.json();
@@ -78,7 +78,7 @@ describe("controllers/signing-key", () => {
         createdAt: Date.now(),
         playbackId: await generateUniquePlaybackId(id),
         playbackPolicy: {
-          type: "signed",
+          type: "jwt",
         },
         status: {
           phase: "ready",
@@ -129,7 +129,7 @@ describe("controllers/signing-key", () => {
       const otherStream = await client.post("/stream", {
         name: "test",
         playbackPolicy: {
-          type: "signed",
+          type: "jwt",
         },
       });
       const otherStreamPlaybackId = (await otherStream.json()).playbackId;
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/access-control.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      expect(res2.status).toBe(204);
+    });
+
+    it("should not allow playback on not existing streams or public keys", async () => {
+      const res = await client.post("/access-control/gate", {
+        stream: `video+0000000000`,
+        pub: signingKey.publicKey,
+        type: "jwt",
+      });
+      expect(res.status).toBe(403);
```
victorges: Should these be 404s instead?

packages/api/src/controllers/access-control.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    it("should not allow playback if stream is gated an pub is missing", async () => {
+      const res2 = await client.post("/access-control/gate", {
+        stream: `video+${gatedPlaybackId}`,
+        type: "jwt",
+      });
+      expect(res2.status).toBe(403);
+    });
+
+    it("should not allow playback if stream and public key does not share the same owner", async () => {
```
victorges, line 126: nice test

packages/api/src/controllers/access-control.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      client.jwtAuth = nonAdminToken;
+      const stream = await client.post("/stream", {
+        name: "test",
+      });
+      client.jwtAuth = adminToken;
+      const streamPlaybackId = (await stream.json()).playbackId;
+      const res = await client.post("/access-control/gate", {
+        stream: `video+${streamPlaybackId}`,
+        type: "jwt",
+        pub: signingKey.publicKey,
```
victorges, line 173: Is this required on this case? Might wanna ensure it works if it is not specified as well, like the previous test

packages/api/src/controllers/access-control.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      });
+      const otherStreamPlaybackId = (await otherStream.json()).playbackId;
+      const res = await client.post("/access-control/gate", {
+        stream: `video+${otherStreamPlaybackId}`,
+        pub: signingKey.publicKey,
+        type: "jwt",
+      });
+      expect(res.status).toBe(403);
+    });
+
```
victorges, line 142: Can you add some additional test cases?
 - Does not allow playback if signing key is disabled
 - Does not allow playback if signing key is deleted (or doesn't even exist)
 - Does not allow playback if the `playbackPolicy` is something we don't understand (this is necessary for bumping the version for example. maybe we find something we don't know and we should fail-closed, not open!)

And maybe we should consider implementing support for not allowing playback if the stream or the user are suspended. WDYT?

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+  authorizer({ anyAdmin: true }),
+  validatePost("access-control-gate-payload"),
+  async (req, res) => {
+    const playbackId = req.body.stream.replace("video+", "");
+    const content =
+      (await db.stream.getByPlaybackId(playbackId)) ||
+      (await db.asset.getByPlaybackId(playbackId));
+
+    if (!content) {
+      throw new ForbiddenError("Content not found");
```
victorges: Yeah, here, maybe it should be a 404

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+    }
+
+    if (content.deleted) {
+      throw new ForbiddenError("Stream is deleted");
```
victorges: We should not differentiate between "not found" and "deleted", so I think the above check should just be `!content || content.deleted` and we return the same error on both.

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+    }
+
+    const playbackPolicy = content.playbackPolicy;
+
+    if (playbackPolicy) {
+      if (playbackPolicy.type === "public") {
```
victorges: **nit:** Could create a variable for the policy type here that already defaults to whatever we want as the default if it's not set. Then IMO the code paths would be clearer, without a 3rd code path a dozen lines below for when the policy is not set. WDYT? Sth like
```suggestion
    const playPolicyType = content.playbackPolicy?.type ?? "public";
    if (playPolicyType === "public") {
       ...
    } else if (playPolicyType === "signed") {
       ...
    } else {
       throw new BadRequestError(`unknown policy type: ${playPolicyType}`)
    }
```

The if/else chain could also be a switch case if you like those (sometimes I do xD)

victorges: Also note that the final `else` (which would be the `switch` `default` case) is absolutely necessary so we don't allow playback of any playback policy that the current version does not understand. 

So even if you don't refactor into the temp variable I suggested, which is an optional change, you must add explicit handling of an unknown value here, instead of just falling back to allowing playback which is happening below if `playbackPolicy.type` doesn't match any of the cases here.

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+        res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
```
victorges: This can be set on every response, before any `if` branch. We don't need to repeat it on every case

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+
+      if (playbackPolicy.type === "signed") {
+        if (!req.body.pub) {
+          throw new ForbiddenError("Stream is gated and requires a public key");
+        }
+
+        const query = [];
+        query.push(sql`signing_key.data->>'publicKey' = ${req.body.pub}`);
+        const [signingKeyOutput] = await db.signingKey.find(query, {
+          limit: 1,
```
victorges: Can we do `limit: 2` and also make sure there are no signing keys with the same public key? Just a special case below of `output.length > 1` as well. We definitely want to know if we ever have a bug where this happens, and should not allow the repeated signing keys to be used anymore as they are likely compromised.

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+        });
+
+        if (signingKeyOutput.length == 0) {
+          throw new ForbiddenError(
```
victorges: This should be a 404 as well wdyt?

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +14,66 @@ const app = Router();
...
+        }
+
+        if (signingKey.disabled || signingKey.deleted) {
+          throw new ForbiddenError("The public key is disabled or deleted");
+        }
+      }
+    }
+    res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
+    res.status(204);
+    return res.end();
```
victorges: I think it is dangerous to structure the code in a way that anything that passes through will end up allowing the playback to happen. If you don't like the refactoring suggestion I added above, WDYT of at least moving this to inside the `playbackPolicy.type === "signed"` branch above, and changing this "default response" to something like:? 
```ts
throw new InternalServerError("unreachable code")
```

victorges (approved):
LGTM, but I have some specific suggestions to make the code a little more maintainable specifically regarding security. Overall, we should default to (gate) "closed" instead of "open" in unexpected code paths or inputs.

packages/api/src/controllers/access-control.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      client.jwtAuth = nonAdminToken;
+      let created: SigningKeyResponsePayload = await client
+        .post("/access-control/signing-key")
+        .then((res) => res.json());
+      let res = await client.get(`/access-control/signing-key/${created.id}`);
+      signingKey = await res.json();
+      const res2 = await client.post("/stream", {
+        name: "test",
+        playbackPolicy: {
+          type: "signed",
```
victorges: Ah, one more thing... Can we change this playback policy type to `jwt`? It will be confusing to add other "signature" methods later (like web3 signature) but have only the JWT playback policy being called "signed".

Can be made on a separate PR as it's unrelated to this new API, but IMO it'd be really desirable.

packages/api/src/controllers/access-control.test.ts
gioelecerati: Could be, in the spec doc and the discussion with Redouane we defined that the possible Status Codes are `200` for allowed playback (became 204) and `403` for disallowed playback, but we could just do 2XX and 4XX so we can be more specific!

packages/api/src/controllers/access-control.ts
gioelecerati: Good point!

packages/api/src/controllers/access-control.test.ts
gioelecerati: Sounds good, added!

about not allowing playback if the stream or the user are suspended, wouldn't the stream be interrupted anyway so impossible to playback anyway?

packages/api/src/controllers/access-control.ts
gioelecerati: Makes sense! I was using a switch case first, but the check on the string was failing as it was using `strict` and I wasn't able to get true on `playbackPolicy.type === "signed"` :eyes: 

Meanwhile I did it with the if else, putting the final else for unknown policy types as you suggested!

packages/api/src/controllers/access-control.ts
gioelecerati: Good point!

packages/api/src/controllers/access-control.ts
gioelecerati: Even for missing signing keys? I would say a `403` with `The content is gated but the corresponding Signing Key was not found` would be more precise, as I would expect `404` to be for a missing content

packages/api/src/controllers/access-control.ts
gioelecerati: Moved into the signed case!

packages/api/src/controllers/access-control.test.ts
gioelecerati: Cool, changing it to `jwt`

packages/api/src/controllers/access-control.test.ts
gioelecerati: WDYM? This is creating a stream without a `playbackPolicy` specified and assuring that `204` is returned

victorges: But it's calling the `/gate` API with a public key. I mean that the playback should work even without any public key on the payload, instead of specifying a random one (e.g. there could be a bug in the API where it requires **some** JWT in the `/gate` API, even if it doesn't validate anything about it)

packages/api/src/controllers/access-control.test.ts
victorges: This is another case of layered security [1]. Just because there is _theoretically_ another system that is going to do the blocking, it doesn't mean it's detrimental to add more redundant layers here that would make it more effective. Like that idea of listing 2 elements when only 1 is expected and ensuring that only 1 is returned. Just because we expect something to behave in a certain way does not mean that there aren't any bugs or incidents :)

In fact, the current system for suspending streams and users is very much fallible, and the API for it is broken **right now**. We only have an idea of how this is gonna be better in the future, with events about the streams sent to catalyst etc, but it is only an idea right now. 

By adding these checks here, we will "immediately" (after the `/gate` cache expires) block access to all viewers watching the stream. Even if the stream suspension fails, which it very often does, the viewers will still get blocked and then the stream become useless for the pirates.

[1] The original term is for like more hardcore security, but I think it does apply here: https://blog.totalprosource.com/what-is-layered-security-how-does-it-defend-your-network

packages/api/src/controllers/access-control.ts
victorges: Cool! Yeah that switch issue is very weird, IDK what's up. An if/else chain looks great and sometimes better than switch anyway which have several idiosyncracies (like that "temporal something" we had the other day)

packages/api/src/controllers/access-control.ts
victorges: Makes sense!

packages/api/src/controllers/access-control.ts
victorges: Cool

victorges (approved):
re-LGTM!
(havent checked the code again but the replies to comment make full sense)

packages/api/src/controllers/access-control.test.ts
gioelecerati: Oh I get it, yeah, there is a specific test for that `"should allow playback on public playbackId"` which has no public key specified. I will add another point to that where both the publicKey is missing in the payload and the playbackPolicy in the content

packages/api/src/controllers/access-control.test.ts
gioelecerati: Cool! Thanks for the write up and the info, it makes sense, although eventually, if the user is not actually suspended on Mist and he is still able to stream, he would just need to make the stream public again to keep it streaming and playback normally.

I added the check and a test for the suspended users!

victorges: The playback should be blocked for suspended users/streams even if the given content is public. Isn't this new API called even for public assets/streams?

packages/api/src/controllers/access-control.ts
```diff
@@ -8,4 +19,77 @@ const app = Router();
...
+        );
+      }
+
+      if (signingKey.disabled || signingKey.deleted) {
+        throw new ForbiddenError("The public key is disabled or deleted");
+      }
+
+      const user = await db.user.get(content.userId);
+
+      if (user.suspended) {
```
victorges, line 79: This should not be here. It is common logic for any kind of playback policy. If the user is suspended, playback should be invariably blocked.

So this should go to before any branch on the playback policy, right after fetching the content itself.

Also, you should also check `|| content.suspended` which is available for stream objects. Given assets don't have that field, you might need to do `|| ("suspended" in content && content.suspended)` for TypeScript happiness.

PR #1329, opened 2022-10-05
api: add isTestUser flag

iameli:
Useful for filtering test streams out of Prometheus data and not much else.

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -798,7 +798,7 @@ app.post("/", authorizer({}), validatePost("stream"), async (req, res) => {
   const createdAt = Date.now();
   const streamKey = await generateUniqueStreamKey(id);
   let playbackId = await generateUniquePlaybackId(id, [streamKey]);
-  if (req.user.isTestUser === true) {
+  if (req.user.isTestUser) {
     playbackId += "-test";
   }
 
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -797,7 +797,10 @@ app.post("/", authorizer({}), validatePost("stream"), async (req, res) => {
   const id = uuid();
   const createdAt = Date.now();
   const streamKey = await generateUniqueStreamKey(id);
-  const playbackId = await generateUniquePlaybackId(id, [streamKey]);
+  let playbackId = await generateUniquePlaybackId(id, [streamKey]);
+  if (req.user.isTestUser === true) {
+    playbackId += "-test";
+  }
 
   let objectStoreId;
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -2063,6 +2063,10 @@ components:
           example: 03AGdBq24blDQQCyt9BSYjS0YZ3-tZapaQ5xO2-oykG_4dS_91TzH9qOziOwXbff_QiDOsXmWKxqtOrmOEx7qjijNhbexBzvNjM6HbbhXcji1rKcxxQRsL-t3SbycwZuNXos9OjnrYmfV9F3zsy1o73ia5Wlw_Zcf1jwbm3n-yQ2epifcMHXJhsUrqmUe6a1J1mPaD2heHvCOS3vYm0rWijdMl2E4bXQf75BiWvbLKxCWv5ZF8279nCMcGyJBIw1kno-A0x5KTOlEVxV369nz12lkvRfOtRABKB49jMTygAc2BKh9GBlAYGO6PBPoIk7_BCo7I_hsHoJeT4c80ucOPraZby0QJM5jcCQcqVCYrcaPKPiUMcIEixIhB5FAfDs88uYqDWMcgZdMRKCvmxEC3ONDVDtq0nLMc4rniuC5sQTz1E4D6SR_Adik
           description: Recaptcha v3 Token
           writeOnly: true
+        isTestUser:
+          type: boolean
+          example: false
+          description: User creates test streams and playbackIds should be flagged as such
```
victorges: Is the idea to set these directly on the DB? I'm fine with that actually, but could you also add an `internal` boolean field here? I created that so @ecmulli could filter out internal Livepeer users from the usage queries (not always we use our @livepeer.org emails lol).

Since you're adding this here, might as well add that other one.

packages/api/src/controllers/stream.ts
```diff
@@ -797,7 +797,10 @@ app.post("/", authorizer({}), validatePost("stream"), async (req, res) => {
   const id = uuid();
   const createdAt = Date.now();
   const streamKey = await generateUniqueStreamKey(id);
-  const playbackId = await generateUniquePlaybackId(id, [streamKey]);
+  let playbackId = await generateUniquePlaybackId(id, [streamKey]);
+  if (req.user.isTestUser === true) {
```
victorges: 🦆 just this already works:
```suggestion
  if (req.user.isTestUser) {
```

victorges (approved):
LGTM

packages/api/src/schema/schema.yaml
iameli: Done!

PR #1332, opened 2022-10-06
fix: fix crash in monster

0xcadams:
Fixes crash in monster


0xcadams:
Screen Shot 2022-10-06 at 4 33 24 PM

packages/www/layouts/streamDetail.tsx
```diff
@@ -71,7 +71,7 @@ const StreamDetail = ({
           ingest.sort((a, b) => a.base.localeCompare(b.base));
         }
         if (ingest.length > 0) {
```
victorges: Should this have ? as well? If ingest is null I guess it will explode here first

victorges (approved):
LGTM

PR #1334, opened 2022-10-07
access-control: make /gate public

gioelecerati:

Make the gate api public

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -21,7 +21,6 @@ app.use("/access-control", accessControl);
 
 accessControl.post(
   "/gate",
-  authorizer({}),
   validatePost("access-control-gate-payload"),
   async (req, res) => {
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -21,7 +21,7 @@ app.use("/access-control", accessControl);
 
 accessControl.post(
   "/gate",
-  authorizer({ anyAdmin: true }),
+  authorizer({}),
```
victorges: You should remove this middleware as the API should be public. There is no authorization.

gioelecerati: Didn't get we wanted it full public, still thought with a token required. Sounds good and removed

victorges: Cool!

victorges (approved):
LGTM

PR #1350, opened 2022-10-19
Fixes the Delete and Update webhook dialogs

### LINKED ISSUE #1340: UI Issues when configuring webhooks
2 Issues:
1. Edit webhook modal obstructed by page
2. Cannot select events when creating webhook

----

1. Edit webhook modal obstructed by page:

2. https://www.loom.com/share/c2b8476cb4f343158381c4b2a6277e84

clacladev:
Fixes the Delete and Update webhook dialogs, which were displayed behind the body of the content. Fixed the Event selector inside the Create/Update web hook dialog.

- Displaying the Delete and Update webhook dialogs in front of the content
- Refactored the page breaking it down into smaller more manageable components
- Fixed the Events Select to make it work on all browsers (it was working only on Chrome!)
- Fixed the position of the Events dropdown to not overlap the Select box

Tested on different browser and dark/light mode.

Fixes #1340

**Screenshots (optional):**
Screenshot 2022-10-19 at 12 25 38

Screenshot 2022-10-19 at 12 25 43

Events Select all 3 major browsers:
Screenshot 2022-10-20 at 10 47 22
Screenshot 2022-10-20 at 10 47 52
Screenshot 2022-10-20 at 10 48 20


adamsoffer (changes requested):
Looks good! I just noticed a few cosmetic issues related to webhooks that we may as well tackle in this issue as well.

1. Can we change this border from `$primary7` to `$neutral6`
CleanShot 2022-10-20 at 12 19 57@2x

2. Can we fix this overflow text? Can use an ellipsis...
CleanShot 2022-10-20 at 12 20 19@2x

3. Can we fix this `z-index` on dialog overlays so that it covers the sidebar too?
CleanShot 2022-10-20 at 12 18 30@2x

clacladev:
@adamsoffer I fixed the border color and the long urls overlapping. Good catch.
Screenshot 2022-10-20 at 11 35 37

Screenshot 2022-10-20 at 11 56 45


I would suggest that I fix the z-index issue in another PR, given that turns out **ALL** popups in the app have the same issue. Also, it turns out that **ALL** popups on smaller screens break.
Should I create a new issue and then work on it on a PR, or should I just work on a new PR?

Screenshot 2022-10-20 at 12 00 52

Screenshot 2022-10-20 at 12 01 09

adamsoffer:
> I would suggest that I fix the z-index issue in another PR, given that turns out ALL popups in the app have the same issue. Also, it turns out that ALL popups on smaller screens break.
Should I create a new issue and then work on it on a PR, or should I just work on a new PR?

Okay yeah in that case let's create a separate issue + PR to fix the dialog. Sounds good 👍

adamsoffer (changes requested):
LGTM! Can we just resolve that merge conflict? Then good to merge.

PR #1354, opened 2022-10-21
api+www: Delete asset meta field

victorges:

Update: we decided to delete the meta field altogether.

We have had plenty of confusion around this field already because of the unfortunate
name that makes it look like it means the NFT metadata exported to IPFS, when it is
just for adding user info to the API object on Studio for filtering etc.

Since we added a little more to that whole confusion now, with [a whole new field in the SDK](https://livepeerjs.org/react/asset/useUpdateAsset#return-value)
for setting the metadata, instead of the existing `storage.ipfs.nftMetadata`, I decided to
fix the part I'm responsible for and already had a quick fix in mind. Hence this PR.

Also made the changes in a partially backward compatible way, by not removing any 
existing `meta` field and copying them to the new `tags` field instead. So we will keep
compatibility on reads, but any write will need to use `tags` instead of `meta`. I opted for
this approach because we have less than 10 assets from real users that are actually using
this right now, and all of them look like just testing the feature (and most of them also
interpreting it wrong as NFT metadata, as we imply in the docs and dashboard):
image

Notice that deploying this also depends on:
 - [ ] Fix any documentation about the `meta` field
 - [ ] Update the SDK to use the new `tags` field instead of `meta` (or just remove it from there for now)

An alternative to this might actually be deleting this field altogether, either only from the dashboard 
(as the current UI was misguided by thinking the `meta` were NFT metadata) or both from the API
and dashboard, since not many users are depending on it anyway. We can discuss about it.

 - Rename `meta` to `tags` in API 
 - Add compatibility logic to keep `meta` untouched and just copy it to the `tags` in asset objects returned by the API
 - Update dashboard asset detail view to refer only to tags, not metadata
 - Update asset edit dialog the same way, and fix the example

## -

 - Play around with tags on the dashboard 
 - Call an existing asset with `meta` and make sure `tags` also show up with the same value

The `meta` confusion. Not the whole NFT metadata confusion anymore tho 🙃 

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -55,32 +55,6 @@ function shouldUseCatalyst({ query, user, config }: Request) {
   return 100 * Math.random() < config.vodCatalystPipelineRolloutPercent;
 }
 
-function validateAssetTags(
-  tags: Record<string, string>,
-  asset?: WithID<Asset>
-) {
-  if (!tags) {
-    return tags;
-  }
-  tags = _({ ...asset?.tags, ...tags })
-    .omitBy(_.isNull)
-    .value();
-  try {
-    if (tags && JSON.stringify(tags).length > TAGS_MAX_SIZE) {
-      console.error(`provided tags exceeds max size of ${TAGS_MAX_SIZE}`);
-      throw new UnprocessableEntityError(
-        `the provided tags exceeds max size of ${TAGS_MAX_SIZE} characters`
-      );
-    }
-  } catch (e) {
-    console.error(`couldn't parse the provided tags=${tags}`);
-    throw new UnprocessableEntityError(
-      `the provided tags are not in a valid json format`
-    );
-  }
-  return tags;
-}
-
 function cleanAssetTracks(asset: WithID<Asset>) {
   return !asset.videoSpec
     ? asset
@@ -127,7 +101,6 @@ async function validateAssetPayload(
     },
     name: payload.name,
     playbackPolicy: payload.playbackPolicy,
-    tags: validateAssetTags(payload.tags),
     objectStoreId: payload.objectStoreId || defaultObjectStoreId,
   };
 }
@@ -295,8 +268,6 @@ app.use(
     let toExternalAsset = (a: WithID<Asset>) => {
       a = withPlaybackUrls(ingest, a);
       a = assetWithIpfsUrls(a);
-      // TODO: Remove compatibiltiy logic (tags used to be called meta)
-      a = { ...a, tags: a.tags || (a as any).meta };
       if (req.user.admin) {
         return a;
       }
@@ -807,7 +778,6 @@ app.patch(
     // these are the only updateable fields
     let {
       name,
-      tags,
       playbackPolicy,
       storage: storageInput,
     } = req.body as AssetPatchPayload;
@@ -832,15 +802,12 @@ app.patch(
       throw new UnprocessableEntityError(`asset is not ready`);
     }
 
-    tags = validateAssetTags(tags, asset);
-
     if (storage) {
       storage = await reconcileAssetStorage(req, asset, storage);
     }
 
     await req.taskScheduler.updateAsset(asset, {
       name,
-      tags,
       storage,
       playbackPolicy,
     });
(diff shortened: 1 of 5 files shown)


0xcadams:
As a user, I wouldn't expect "tags" to be key value pairs in an API, so I think this may need to evolve a bit. I don't think the existing data, from that screenshot, reflects "tags".

I personally think we should deprecate this field entirely (AKA make it read-only for now and remove in the future), communicate that to our users, and then re-create metadata in a thoughtful way (with that TBD based on discussions). What do you think?

victorges:
> I personally think we should deprecate this field entirely (AKA make it read-only for now and remove in the future), communicate that to our users, and then re-create metadata in a thoughtful way (with that TBD based on discussions). What do you think?

Yeah I'm kinda leaning on that direction now. This feature was mostly copied from Mux as we assumed it would be a good idea at the time (cc @gioelecerati), but I think it's getting in conflicts with more strategic features we want like the NFT metadata support. Let's phase this out for now and focus on the NFT metadata, and we can bring it back if we realize our users actually need it.

@0xcadams what would you suggest in the dashboard? Just remove all the code related to this? Or maybe just comment out the component for NFT metadata on the asset details page, but keep all the code around for when we do the NFT metadata work in the short-term? (on the API it's worth deleting, we don't know even know if/when we'll want this `meta`/`tags` feature).

gioelecerati:
> > I personally think we should deprecate this field entirely (AKA make it read-only for now and remove in the future), communicate that to our users, and then re-create metadata in a thoughtful way (with that TBD based on discussions). What do you think?
> 
> Yeah I'm kinda leaning on that direction now. This feature was mostly copied from Mux as we assumed it would be a good idea at the time (cc @gioelecerati), but I think it's getting in conflicts with more strategic features we want like the NFT metadata support. Let's phase this out for now and focus on the NFT metadata, and we can bring it back if we realize our users actually need it.
> 
> @0xcadams what would you suggest in the dashboard? Just remove all the code related to this? Or maybe just comment out the component for NFT metadata on the asset details page, but keep all the code around for when we do the NFT metadata work in the short-term? (on the API it's worth deleting, we don't know even know if/when we'll want this `meta`/`tags` feature).

Agreed, we should phase out `meta` now. At time we thought it would have been useful to let user make their own categorization and tags, but now we are anyway in a situation where we prefer the developer to have their own logic in their own database / application, so it makes even less sense to keep it.

0xcadams:
> > I personally think we should deprecate this field entirely (AKA make it read-only for now and remove in the future), communicate that to our users, and then re-create metadata in a thoughtful way (with that TBD based on discussions). What do you think?
> 
> Yeah I'm kinda leaning on that direction now. This feature was mostly copied from Mux as we assumed it would be a good idea at the time (cc @gioelecerati), but I think it's getting in conflicts with more strategic features we want like the NFT metadata support. Let's phase this out for now and focus on the NFT metadata, and we can bring it back if we realize our users actually need it.
> 
> @0xcadams what would you suggest in the dashboard? Just remove all the code related to this? Or maybe just comment out the component for NFT metadata on the asset details page, but keep all the code around for when we do the NFT metadata work in the short-term? (on the API it's worth deleting, we don't know even know if/when we'll want this `meta`/`tags` feature).

Yeah I think removing the "edit metadata" button makes sense for now. I'd opt for keeping the `metadata` section in the frontend and just making it read-only from the UI. And that being read from the "nft metadata" which currently exists. But could see the argument for removing entirely from the frontend.

victorges:
@0xcadams OK I've removed all the UI regarding metadata from the dashboard. I've did it through commenting out some relevant parts of the code that set the UI elements, but I kept most of the logic there to reuse later when we want to bring this back. Let me know if you'd rather just delete the code, and we can refer back to this PR if we want to recover any part of it when we implement it properly for NFT metadata.

On keeping a read-only display: I'd rather not display the NFT metadata for now either, since we also don't display the CID part of it, which is the most important. I also have my doubts whether we should display the metadata that is on the API or the real metadata that is actually saved on IPFS, i.e. actually reading the file from the gateway URL. It could be confusing to display what's in the API. Since we need to rethink this whole story anyway, I've opted to just removing it for now and we can bring it back the right way later.

IOW this is now ready for review as per deleting `meta` from Studio ✨ Could you give it a go?

0xcadams (approved):
LGTM

PR #1360, opened 2022-10-22
api: Fix recording webhooks

victorges:
As an exercise of understanding the current recording pipeline as we'll need
to revamp it soon for the catalyst integreation (and since a key customer was
suffering with that), I found some easy improvements we could make to the 
recording webhooks in order to make them more useful.

The main changes revolve towards sending the session object with the recording
both on the `started` and `ready` webhooks. For that, it also required changing the
point in the code where we send the `recording.started` (but the `ready` had to stay
where it was).

 - Move the `recording.ready` hook from `/setactive` to `POST /stream/:id` (`MistProcLivepeer`), where we create the user session object
 - Add `session` object to the webhook payload both `started` and `ready`
 - Make sure that session object is the user session, not the final "child stream" from MistProcLivepeer. Relevant only for the `recording.ready` hook.

## -

 - Deploy to staging
 - Make a recorded livestream
 - Check webhooks are received with the session in payload

Fixes customer issue.

**Screenshots (optional):**
https://webhook.site/#!/d5fd9d61-383c-4891-8d10-0e29886fef50/c587140c-a5ae-4b22-a576-7d35f0553ff8/1

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/session.ts
+++ b/packages/api/src/controllers/session.ts
@@ -129,11 +129,8 @@ app.get("/", authorizer({}), async (req, res, next) => {
     res.links({ next: makeNextHREF(req, newCursor) });
   }
   let sessions = output.map((session) =>
-    withRecordingFields(ingest, session, !!forceUrl)
+    toExternalSession(session, ingest, req.user.admin)
   );
-  if (!req.user.admin) {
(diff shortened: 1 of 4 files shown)


iameli (approved):
LGTM that nothing's obviously off, not intimately familiar with this anymore

PR #1365, opened 2022-10-25
api: Add source field to assets

victorges:
This is to create a new useful metadata in asset objects which is some object representing
where they come from / how they were created.

The idea for this is to solve an immediate user need for relating their VOD assets with the livestream
sessions of their users. We previously had no official way for that, and had workarounds in our 
dashboard that (I didn't even remember but also) we should not have our users depend on 
(regex on playback url of the asset).

It also opens room for an API proposed by @yondonfu for supporting automatic import + playback by CID
directly from IPFS (with no need to manually re-store the asset on IPFS).

 - Create `source` field on the asset which could have several types and args
 - Set the `source` field on every asset creation today.

## -

Create an asset in the various different ways and make sure source field is set appropriately.

No. Customer support.

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -74,6 +74,7 @@ describe("controllers/signing-key", () => {
       gatedAsset = await db.asset.create({
         id,
         name: "test-storage",
+        source: { type: "directUpload" },
         createdAt: Date.now(),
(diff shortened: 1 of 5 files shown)


0xcadams (approved):
LGTM

gioelecerati (approved):
Very cool, LGTM!

PR #1367, opened 2022-10-25
Fetch playback URL for imported asset using source ID

yondonfu:

Supersedes https://github.com/livepeer/studio/pull/1364 in favor of building off of https://github.com/livepeer/studio/pull/1365 which introduces a `source` field to the `Asset` entity.

This PR allows a user to specify a `source` field in the request body for a `/api/asset/import` request that looks like:

```
{
    "url": ...,
    "type": ...,
    "id": ...
}
```

For now, `type` must be `url`.

The user is responsible for specifying a value for the `id` field. For example, the value could be an IPFS CID or Arweave tx ID.

The `/playback` endpoint can then do a lookup for assets based on `source.id` which is useful when the asset is imported from an IPFS gateway and you want to plug in a CID to playback the asset. Or when the asset is imported from an Arweave gateway and you want to plug in a tx ID to playback the asset.

See commit history.

Updated unit tests.

Fixes https://github.com/livepeer/studio/issues/1362.

TODO: Test in an e2e environment and make sure /playback returns the correct playback URL for a source ID.

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -485,11 +485,17 @@ const uploadWithUrlHandler: RequestHandler = async (req, res) => {
     req.user.id,
     Date.now(),
     req.config.vodObjectStoreId,
-    req.body
+    req.body,
+    req.body.source
   );
-  if (!req.body.url) {
+
+  const url =
+    req.body.url ?? (req.body.source ? req.body.source.url : undefined);
+  if (!url) {
     return res.status(422).json({
-      errors: [`Must provide a "url" field for the asset contents`],
+      errors: [
+        `Must provide a "url" or "source.url" field for the asset contents`,
+      ],
     });
   }
 
@@ -499,7 +505,7 @@ const uploadWithUrlHandler: RequestHandler = async (req, res) => {
     taskType,
     {
       [taskType]: {
-        url: req.body.url,
+        url,
       },
     },
     undefined,
--- a/packages/api/src/store/asset-table.ts
+++ b/packages/api/src/store/asset-table.ts
@@ -53,4 +53,19 @@ export default class AssetTable extends Table<WithID<Asset>> {
     }
     return assets[0];
   }
+
+  async getBySourceId(id: string): Promise<WithID<Asset>> {
+    const query = [
+      sql`asset.data->'source'->>'id' = ${id}`,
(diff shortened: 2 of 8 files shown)


packages/api/src/store/asset-table.ts
```diff
@@ -53,4 +53,19 @@ export default class AssetTable extends Table<WithID<Asset>> {
...
+      sql`asset.data->>'deleted' IS NULL`,
+    ];
+    const [assets] = await this.find(query, {
+      limit: 2,
+      order: "coalesce((asset.data->'createdAt')::bigint, 0) ASC",
+    });
+    if (!assets || assets.length < 1) {
+      return null;
+    }
+    return assets[0];
```
0xcadams, line 69: Just calling out one unintended side-effect that may happen here - if two users create an asset with the same source CID, the first asset will have all views registered under it. I don't think this is a blocking issue, but something we'll need to iterate on.

packages/api/src/controllers/asset.ts
```diff
@@ -485,11 +485,17 @@ const uploadWithUrlHandler: RequestHandler = async (req, res) => {
...
     Date.now(),
     req.config.vodObjectStoreId,
-    req.body
+    req.body,
+    req.body.source
   );
-  if (!req.body.url) {
+
+  const url =
+    req.body.url ?? (req.body.source ? req.body.source.url : undefined);
```
0xcadams, line 493:
```suggestion
    req?.body?.url ?? req?.body?.source?.url;
```

0xcadams (approved):

packages/api/src/store/asset-table.ts
yondonfu: Yeah good call out!

yondonfu:
After a conversation in Discord with @victorges, I'm closing this PR in favor of https://github.com/livepeer/studio/pull/1370 which should be simpler approach and also ensures that the CIDs used to lookup playback URLs of imported IPFS assets actually match the asset data imported. Hope this is the last PR that I close for this issue :)

Cherry picked https://github.com/livepeer/studio/pull/1367/commits/7f0c972b2199d15b27a9b33490cede349ca22413 into https://github.com/livepeer/studio/pull/1370 as well.

PR #1368, opened 2022-10-26
Implemented Signing Keys

### LINKED ISSUE #1345: Implement signing key view
https://www.figma.com/file/o0lkaZxamhUfG6eqSglhPO/Livepeer-Studio-Dashboard?node-id=845%3A4956

clacladev:
Added Signing Keys panel in the Developer section.

Added:
- Signing Keys list
- Create new key dialog
- Delete keys
- General refactoring

I tested with the different themes, Safari and Chrome.

Fixes #1345

**Screenshots (optional):**
Screenshot 2022-10-26 at 14 50 21
Screenshot 2022-10-26 at 14 34 46
Screenshot 2022-10-26 at 14 35 00
Screenshot 2022-10-26 at 14 35 07

### DIFF
--- a/packages/www/components/Dashboard/AssetDetails/AssetDetailsBox.tsx
+++ b/packages/www/components/Dashboard/AssetDetails/AssetDetailsBox.tsx
@@ -4,7 +4,7 @@ import numeral from "numeral";
 import RelativeTime from "../RelativeTime";
 import ShowURL from "../ShowURL";
 import { useMemo } from "react";
-import ClipButton from "../ClipButton";
+import ClipButton from "../Clipping/ClipButton";
 
 const Cell = ({ children, css = {} }) => {
   return (
--- a/packages/www/components/Dashboard/AssetDetails/AssetOverviewTab.tsx
+++ b/packages/www/components/Dashboard/AssetDetails/AssetOverviewTab.tsx
@@ -19,7 +19,7 @@ import DatabaseIcon from "../../../public/img/icons/database.svg";
 import { useApi } from "hooks";
 import Spinner from "../Spinner";
 import moment from "moment";
-import ClipButton from "../ClipButton";
+import ClipButton from "../Clipping/ClipButton";
 
 const StyledIpfsIcon = styled(IpfsIcon, {
   color: "$gray",
--- a/packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
+++ b/packages/www/components/Dashboard/AssetDetails/EmbedVideoDialog.tsx
@@ -9,7 +9,6 @@ import {
   TextField,
   useSnackbar,
 } from "@livepeer/design-system";
-import { Asset } from "livepeer";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 
 const embedStringForAsset = (playbackId: string) =>
--- a/packages/www/components/Dashboard/AssetsTable/helpers.tsx
+++ b/packages/www/components/Dashboard/AssetsTable/helpers.tsx
@@ -99,7 +99,7 @@ export const rowsPageFromState = async (
         name: {
           id: asset.id,
           href: `/dashboard/assets/${asset.id}`,
-          assetName: asset.name,
+          name: asset.name,
           isStatusFailed,
           errorMessage,
         },
--- a/packages/www/components/Dashboard/Clipping/ClipButton.tsx
+++ b/packages/www/components/Dashboard/Clipping/ClipButton.tsx
@@ -1,0 +1,83 @@
+import {
+  useSnackbar,
+  HoverCardRoot,
+  HoverCardTrigger,
+  Flex,
+  Box,
+  HoverCardContent,
+  Text,
+} from "@livepeer/design-system";
+import { CopyIcon } from "@radix-ui/react-icons";
+import { useState, useEffect } from "react";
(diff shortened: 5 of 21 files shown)


clacladev:
@adamsoffer I need to confirm a few things for the Empty state:

I am going to use the same design as the Empty API Keys screen
- The text in the empty state should be Signing keys allow you to add access control to streams and assets, is it ok?
- The Learn More link in the empty state, where should it point to? https://docs.livepeer.studio/category/api as for the Empty API Key screen?

victorges:
My cents on the questions:
> The text in the empty state should be Signing keys allow you to add access control to streams and assets, is it ok?

Looks good, I'd just use `playback policies` instead of `access control`, as that's what we've been using latency to refer to the feature and what users will set in their streams/assets. Maybe sth like:
"Signing keys allow you to use playback policies with your streams to restrict access to them"

> The Learn More link in the empty state, where should it point to? https://docs.livepeer.studio/category/api as for the Empty API Key screen?

I believe it should point to one of the docs that are being created for playback policies. Maybe this:?
https://github.com/livepeer/livepeer-studio-docs/pull/224/files#diff-94beae3c2ab5015501c8b00a1a49192f6525672694e4db2685bd516f186d44b4

We don't have a "guide" yet, so I guess we can point to that API reference for now which has some intro about the signing keys etc.

victorges:
Testing the UI, got some ideas, in order of priority: 
 - The signing keys have a `lastSeen` field, equivalent to the ones for API keys, which should be displayed in the table much in the same way. We went out of our way on the API to enable this, as it was a requirement in the feature, and it seems like a relatively small effort to include it in the UI (can probably reuse the API key stuff? idk). So I think we should include this here. I can help you get an API key with a non-empty `lastSeen` value for testing, lmk if you want that or can mock on the frontend for testing. The times are in the same format as all others, as a unix milliseconds timestamp (default for JS as well).
 - It would be great to have the 'disable' support for the signing keys. You can `PATCH` them with a `disabled: true` to make them unusable (signatures invalid), but still not delete them yet. This is especially useful for migrating between keys, as you don't want to completely delete your old one and then not be able to rollback if you committed any mistakes. It could be done the same way as the deletion, as a button on the top when you select some keys, I'm just not sure how to display that info on the table. Maybe just a new column? Or reuse the `Last Used` column from the above point and show `disabled` when that's the case. Or maybe some special element on the name of the API key when it is disabled, mvp just a _(disabled)_ on the end.
 - When creating a signing key, it was not obvious for me that the private/public keys could be copied (e.g. on your screenshot). Maybe some users will end up missing that if we don't show any hint in the UI that they can copy it until they hover over the text component? Could be one of those "2 overlapped squares copy icon" somewhere in the box, as a hint, maybe on the bottom right or something. Not a must if you think users will generally end up finding the copy functionality before getting frustrated, in the end if they try to select the text to copy they will see the "copy to clipboard" text show up anyway.

WDYT?

victorges:
Oh another thing, WDYT of making the public keys copiable from the table? Currently, the existing signing keys are displayed on the table and the public key is necessarily cropped as it won't fit on any screen. Maybe it could be click-to-copiable so one could get the full value and compare with whatever they are looking at with more security.

adamsoffer:
@victorges - thanks for the feedback. @clacladev I'll followup here with an updated design that include Victor's suggestions.

victorges:
Oh, just noticed that the page title is showing `API Keys - Livepeer Studio` instead of `Singing Keys` (or just `Livepeer Studio`). Like on the name of the tab in the browser.

Sorry about making "loose" comments like this, but I don't know where this is being set in the code so couldn't send it on a specific line.

clacladev:
Thanks @victorges for testing. I am fixing a the issues you pointed out.

Regarding the updated API docs, is it going to go live today/tomorrow? In that case I would wait to merge this PR, so I can use the latest docs url. Otherwise I will create a ticket to update the url in the future, when it gets live.

@adamsoffer Regarding the improvements, not the fixes, that Victor suggested. Last seen should be easy enough to add. Disabling a key too. The question becomes is that a priority? If so with an updated design I am good to go.

Regarding copiable public key, technically you can double click on the public key and copy. Even if it shows the `...` it will copy the entire key. I don't know if that's something we want to make more obvious or if it's not a priority.

Making the public/private fields more intuitive to copy with the icon, is something we can do. It's probably a better UX, it just depends on the overall priorities.

adamsoffer:
> @adamsoffer Regarding the improvements, not the fixes, that Victor suggested. Last seen should be easy enough to add. Disabling a key too. The question becomes is that a priority? If so with an updated design I am good to go.

I'm okay with with pushing those other improvements in another PR. 

> The text in the empty state should be Signing keys allow you to add access control to streams and assets, is it ok?

LGTM

> The Learn More link in the empty state, where should it point to? https://docs.livepeer.studio/category/api as for the Empty API Key screen?

The api reference for creating a signing key will eventually live at `docs.livepeer.studio/reference/api/create-signing-keys`  once merged so maybe we can link there.

adamsoffer:
CleanShot 2022-10-27 at 19 47 15@2x

This might be getting addressed in another issue, but can we fix the spacing of this table? Can use the figma design as a reference.

clacladev:
@adamsoffer all agreed changes are implemented. Tests passing. This PR is ready for last review

adamsoffer (approved):

PR #1372, opened 2022-10-26
api: Support ar:// source URL lookups

yondonfu:

Follow up to https://github.com/livepeer/studio/pull/1370 adding support for looking up ar:// source URL assets.

Depends on https://github.com/livepeer/task-runner/pull/77.

See commit history.

Updated unit tests.

N/A

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/playback.ts
+++ b/packages/api/src/controllers/playback.ts
@@ -44,11 +44,12 @@ const newPlaybackInfo = (
 const getAssetPlaybackUrl = async (
   ingest: string,
   id: string,
-  cid: boolean
+  isPlaybackAlias: boolean
 ) => {
-  const asset = cid
+  const asset = isPlaybackAlias
     ? (await db.asset.getByIpfsCid(id)) ??
-      (await db.asset.getBySourceURL("ipfs://" + id))
+      (await db.asset.getBySourceURL("ipfs://" + id)) ??
+      (await db.asset.getBySourceURL("ar://" + id))
     : await db.asset.getByPlaybackId(id);
   if (!asset || asset.deleted) {
     return null;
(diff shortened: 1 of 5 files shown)


victorges (approved):
LGTM

packages/api/src/controllers/playback.ts
```diff
@@ -44,11 +44,12 @@ const newPlaybackInfo = (
 const getAssetPlaybackUrl = async (
   ingest: string,
   id: string,
-  cid: boolean
+  isPlaybackAlias: boolean
```
yondonfu: Figured that this might be a more descriptive name since the boolean doesn't necessarily indicate that the id is a CID anymore as it could be a CID or a Arweave tx ID. Or, more generally it feels like the id could be a "playback alias". Welcome feedback/suggestions on this though.

victorges: Yeah I'm thinking about that too. Maybe we should remove the ternary condition on [the calling function](https://github.com/livepeer/livepeer-com/blob/bb6093c6760beda6947f91d38b4b43847d8e4a1b/packages/api/src/controllers/playback.ts#L83-L85) and simply make the "by playback ID" the first possibility in this function.

The only caveat from that is exactly the reason for the `if` check below in this function: `getByPlaybackId` is a function that may return a `deleted` asset, as it doesn't include that filter on the query it makes for the asset. It seems like every user of that function already treats an asset with `deleted: true` the same way it would treat a `null` asset anyway. So feel free to:
 - update `getByPlaybackId` to also have an `asset.data->>'deleted' IS NULL` condition
 - Remove this boolean entirely from this function
 - Make this function itself query by playback ID before it queries for anything else, which we can now easily chain with `??`

An alternative would be keeping this function handling the `.deleted` value from the response of the `getPlaybackId` func. That would be OK as well if you won't wanna touch any SQL anymore. I think the other approach would be cleaner tho so that's my soft preference.

yondonfu: The proposal SGTM - seems much cleaner 🥳

PR #1438, opened 2022-11-03
Removed the footer from the Admin layout to improve UX

clacladev:
Removed the footer from the Admin layout to improve UX.
It was big and moving while the page loaded, being very intrusive. It requested visual attention. Just removing it makes a better UX.

**Screenshots (optional):**

**Before**
Screenshot 2022-11-03 at 15 52 54

**After**
Screenshot 2022-11-03 at 15 53 00


adamsoffer (approved):
Thanks. Good stop gap solution until we move these admin view into the new dashboard UI.

PR #1441, opened 2022-11-04
Allow the upload of any file as VOD file

clacladev:
Allow the upload of any file as VOD file

⚠️DO NOT MERGE YET PLEASE⚠️


victorges (approved):
LGTM

packages/www/components/Dashboard/AssetsTable/CreateAssetDialog.tsx
```diff
@@ -17,6 +17,8 @@ import { useDropzone } from "react-dropzone";
 import { Cross2Icon } from "@radix-ui/react-icons";
 import omit from "lodash.omit";
 
+const acceptedMimeTypes = [];
```
victorges: If you're cautious about accidentally deploying this to prod, WDYT of doing it like this?
```suggestion
const acceptedMimeTypes = isStaging() ? [] : ["video/mp4"];
```

We can still revert this change later, but won't risk having this in prod.

PR #1457, opened 2022-11-13
api/queue: Add support for a global CloudAMQP rabbit setup

victorges:

This adds support for a global RabbitMQ setup by allowing to split tasks events into different regions.

Notice that webhooks are still global, but I think that's OK. Only recordings might get processed in a
different region from the original stream.

 - Receive exchange name from CLI args
 - Use it to create the exchange and queues

## -

Deployed to staging and checked CloudAMQP.

Fixes all my rabbit nightmares.

**Checklist:**


victorges (commented):
self-review LGTM cause I can't wait to deploy this

PR #1458, opened 2022-11-13
www/asset: Fix "processing" progress update

victorges:

This introduces a little delay after finishing upload so that there's time for the
task to start executing. This will guarantee that we will show a swift update from 
"uploading 100%" to "processing x%" instead of taking up to 10 seconds depending
on the refresh interval.

Create a `sleep` helper and use it to delay state invalidation.

## -

Upload on the vercel preview and check described behavior.

No only my "OCD".

**Checklist:**


0xcadams (approved):
LGTM

PR #1466, opened 2022-11-15
www: Fix asset list page not loading

### LINKED ISSUE #1388: Asset page is slow to load on Dashboard
Asset list on dashboard (.studio/dashboard/assets) is slow to load, as compared to other pages.

victorges:
I found out we had more than just the account-with-a-lot-of-tasks problem,
it was actually an unnecessary big "list" query that was being done for the 
tasks in the account. That didn't include any `limit` so was looking for 200
objects in the DB for no reason.

To fix, added a limit to that query of up to 20 items. We should eventually fix
the API not to have such big defaults as well.

On the way also noticed we were doing 2 potentially slow queries in sequence,
so parallelized them as well. This should make the loading more snappy even 
on the E2E test accounts.

 - Add limit to tasks list query
 - Parallelize fetches

## -

Access dashboard with E2E account and check if it loads in a reasonable time.

Fixes https://github.com/livepeer/studio/issues/1388

**Checklist:**

### DIFF
--- a/packages/www/components/Dashboard/AssetsTable/helpers.tsx
+++ b/packages/www/components/Dashboard/AssetsTable/helpers.tsx
@@ -1,5 +1,5 @@
 import { Asset } from "livepeer";
-import { Task } from "@livepeer.studio/api";
+import { Asset as ApiAsset, Task } from "@livepeer.studio/api";
 import { Box } from "@livepeer/design-system";
 import { FileUpload } from "hooks/use-api/types";
 import { get } from "lodash";
@@ -13,6 +13,9 @@ import { stringSort, dateSort } from "../Table/sorts";
 import { RowsPageFromStateResult, SortTypeArgs } from "../Table/types";
 import { State } from "../Table";
 import TableEmptyState from "../Table/components/TableEmptyState";
+import { useApi } from "hooks";
+
+type ApiClient = ReturnType<typeof useApi>;
 
 const liveStreamHosts = ["livepeercdn.com", "cdn.livepeer.com"];
 
@@ -71,8 +74,8 @@ export const makeColumns = () => [
 export const rowsPageFromState = async (
   state: State<AssetsTableData>,
   userId: string,
-  getAssets: Function,
-  getTasks: Function,
+  getAssets: ApiClient["getAssets"],
+  getTasks: ApiClient["getTasks"],
   onDeleteAsset: Function
 ): Promise<RowsPageFromStateResult<AssetsTableData>> => {
   const assetsPromise = getAssets(userId, {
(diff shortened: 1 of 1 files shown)


packages/www/components/Dashboard/AssetsTable/helpers.tsx
```diff
@@ -75,15 +75,18 @@ export const rowsPageFromState = async (
...
     cursor: state.cursor,
     order: state.order,
     count: true,
   });
-
-  const [tasks] = await getTasks(userId);
+  const tasksPromise = getTasks(userId, {
+    limit: state.pageSize.toString(),
+  });
+  const [assets, nextCursor, count] = await assetsPromise;
```
0xcadams: Could we do a Promise.all here? I think these will still execute serially

0xcadams (changes requested):

packages/www/components/Dashboard/AssetsTable/helpers.tsx
clacladev: Yes these will still execute serially. As you suggested `... = await Promise.all([assetsPromise, tasksPromise)` will make them be fetched in parallel.

victorges: Hmm are you sure? As far as I (thought) I knew, `await` doesn't have relation with when the awaited task executes, only when you yield the execution back to the event loop for it do execute anything that is scheduled. If it wasn't like that, how could we get those "uncaught exception" errors when no one is awaiting the promise? If the exception happens it means the code executes even with no one awaiting.

So in this case, we'd _start_ 2 jobs, enqueing them for execution, and once we `await` the first one (or anything else) we only stop executing the current function, but anything that is on the event loop will be executed, not only the one we awaited for. Then the 2 jobs are free to execute and do so in parallel (not concurrently).

This behavior is also what I observed when I tested this. The 2 requests are made at the same time, it's not serial (try it out!).

Am I missing something here? I can do an isolated test when I get to the PC tomorrow,  but on these online fiddles I tried they don't let you do async stuff.

clacladev: I am pretty sure that the execution in that scope is suspended when the `await` is hit, until the promise is resolved. Then the next line is executed.

https://programmingwithswift.com/run-async-await-in-parallel-with-javascript/

0xcadams: Yes you're right, those promises will start executing without awaiting, I've never seen this pattern though and I don't think it's clear that this is intended to run in parallel. Why not use promise all here?

victorges: Mostly because of typing issues which will get pretty annoying with those complex return types. Each promise has its own type so the combined type will be just a union of all the types, but maybe an `as const` will help here. Worst case I can do a separate`await` all and then get each result separately.
update: wasnt really a problem

Coming from C# it's pretty clear to me that these run in parallel, but I understand it can be confusing in the single threaded environment of JS. So I can definitely follow the most common pattern here. Maybe error handling will be better as well with that.

victorges: Implemented it here!
https://github.com/livepeer/studio/pull/1466/commits/f26695dddfb274e894371f4aec3e05a7e77fd89f

0xcadams (approved):
LGTM!

PR #1493, opened 2022-11-22
access-control: gate: more logs to debug access control gate

gioelecerati:

Expand logs

yarn test

**Screenshots (optional):**

**Checklist:**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -31,16 +31,8 @@ accessControl.post(
       (await db.stream.getByPlaybackId(playbackId)) ||
       (await db.asset.getByPlaybackId(playbackId));
 
-    let contentLog;
-    let userLog;
-    let signingKeyLog;
-
     if (!content || content.deleted) {
-      try {
-        contentLog = JSON.stringify(content);
-      } catch (e) {
-        contentLog = content;
-      }
+      const contentLog = JSON.stringify(JSON.stringify(content));
 
       console.log(`
         access-control: gate: content not found for playbackId ${playbackId}, disallowing playback, content=${contentLog}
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -31,18 +31,35 @@ accessControl.post(
...
+    let contentLog;
+    let userLog;
+    let signingKeyLog;
+
     if (!content || content.deleted) {
+      try {
+        contentLog = JSON.stringify(content);
+      } catch (e) {
+        contentLog = content;
+      }
```
victorges:
```suggestion
    const contentLog = JSON.stringify(content);
    const userLog = JSON.stringify(user);
    const signingKeyLog = JSON.stringify(signingKey);
```

There's no reasonable scenario where stringify will throw, and when it throws we don't want to catch it anyway (e.g. OOM), especially for an object that just came from our DB. So we can simplify this logic and avoid so many branches.

victorges: While at it, we could also escape the log strings so that loki is able to parse the fields properly, WDYT? The easiest way that I know of is to stringify twice, which will encode the stringified string as a JSON string, i.e. escaped
```suggestion
    const contentLog = JSON.stringify(JSON.stringify(content));
    const userLog = JSON.stringify(JSON.stringify(user));
    const signingKeyLog = JSON.stringify(JSON.stringify(signingKey));
```

packages/api/src/controllers/access-control.ts
```diff
@@ -31,18 +31,35 @@ accessControl.post(
...
     }
 
     const user = await db.user.get(content.userId);
 
     if (user.suspended || ("suspended" in content && content.suspended)) {
+      try {
+        contentLog = JSON.stringify(content);
+        userLog = JSON.stringify(user);
+      } catch (e) {
+        contentLog = content;
```
victorges: If you keep the try/catch, forgot the user here

packages/api/src/controllers/access-control.ts
```diff
@@ -31,18 +31,35 @@ accessControl.post(
...
+      }
+
       console.log(`
-        access-control: disallowing access for contentId=${content.id}, user ${user.id} is suspended
+        access-control: gate: disallowing access for contentId=${content.id}, user ${user.id} is suspended, content=${contentLog}, user=${userLog}
```
victorges: What about the playback ID in all logs?

victorges (approved):
LGTM

packages/api/src/controllers/access-control.ts
gioelecerati: Right, for some reason I thought in typescript JSON.stringify(undefined) would throw. Much better, gonna remove all the try catches!

packages/api/src/controllers/access-control.ts
gioelecerati: haven't added it because I was logging the full content, but yeah, better add it anyway so if the content is undefined for some reason we catch it

PR #1495, opened 2022-11-22
api: Fix race condition on task schedule

victorges:
We were updating the DB only after sending the Rabbit event, which was good not to update the task in case of failures to send the event, but could lead to race conditions in case the event was processed before we could get back and update the DB.

Couldn't think of an ideal solution without getting too complex, so decided to just move the DB update to before the event. Now tasks will go to waiting before we know if the event scheduling actually worked (but only when it is intended to happen).

Also added error handling to the publish, failing the task if so.


gioelecerati (approved):
LGTM

PR #1524, opened 2022-12-04
api: Create strongConsistency option in task&asset APIs

victorges:

This is necessary for task-runner as it requests the objects almost 
immediately after they are creatd. Sometimes rabbit is quicker than the DB.

 - Create `strongConsistency` query string which, when set to true, avoids using the replica to read the objects

## -

`yarn test`

No.

**Checklist:**


gioelecerati (approved):
LGTM! Of course a rabbit is faster

PR #1536, opened 2022-12-07
api: Implement concurrent tasks rate-limiting for VOD

victorges:
This is to implement a couple rate-limiting protections on the VOD pipeline against exaggerated traffic
from specific users. There are 3 main and independent strategies for rate-limiting implemented here.

The strategies are:
1.   A maximum number of tasks that the pipeline can run for the same user concurrently.
  - This works by querying for tasks in the `running` or "retrying" state, and returning 429 on the initial request made by `task-runner` to start executing the task. Task runner learned how to process that here https://github.com/livepeer/task-runner/pull/95 and will backoff the message on the delayed queue for a minute, before trying to start again
2. A maximum number of tasks that can be enqueued for the same user. This includes the tasks that are currently executing.
  - Differently from the previous strategy, this doesn't just backoff the execution of the task, but it returns an explicit error to the user when they try to create more tasks. 
  - The point here is to protect the delayed queue which is used in the first strategy. We delay events for 1 minute at each step, so there is a number of tasks being delayed at which we would not be able to process even the quick-delayed-tasks in a timely manner. Assuming a latency of 100ms to process each message and the same 15x concurrency on a region/queue, this would mean that `60*1000/100*15=9000` messages could create a problem if we didn't have this rate limiting, which seems pretty possible.
  - Based on the numbers above, setup a limit of 100 tasks (by default, everything is configurable) enqueued for each user.
3. A deduplication strategy to avoid processing the same file multiple times. Yeah this might be annoying for testing, but will protect us in production from possible bugs in client implementations or our own (like the old livepeer.js bug that published created of assets for the same URL). The neat thing here is that we don't return an error: we just return the already running/duplicate task back.

## -

`yarn test`, transcoded dozens of files in staging, tried uploading 200 files at once (got rate limited), observed how execution is now happening 5 tasks at a time, pretty orderly, and tried publishing duplicate assets and ensured same asset/task was returned.

I think so but idk where it is. It's a blocker for MediaCovnert launch on 12/16.

**Checklist:**

### DIFF
--- a/packages/api/src/store/asset-table.ts
+++ b/packages/api/src/store/asset-table.ts
@@ -82,7 +82,7 @@ export default class AssetTable extends Table<WithID<Asset>> {
       sql`asset.data->'source'->>'type' = 'url'`,
       sql`asset.data->>'deleted' IS NULL`,
       sql`asset.data->'status'->>'phase' IN ('waiting', 'pending')`,
-      sql`asset.data->'userId' = ${userId}`,
+      sql`asset.data->>'userId' = ${userId}`,
       sql`asset.data->'source'->>'url' = ${url}`,
       sql`coalesce((asset.data->'createdAt')::bigint, 0) > ${createdAfter}`,
     ];
--- a/packages/api/src/store/asset-table.ts
+++ b/packages/api/src/store/asset-table.ts
@@ -3,6 +3,8 @@ import { Asset, Task } from "../schema/types";
 import Table from "./table";
 import { QueryOptions, WithID } from "./types";
 
+const DUPLICATE_ASSETS_THRESHOLD = 15 * 60 * 1000; // 15 mins
+
 export const taskOutputToIpfsStorage = (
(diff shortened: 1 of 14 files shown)


packages/api/src/store/asset-table.ts
```diff
@@ -72,4 +73,21 @@ export default class AssetTable extends Table<WithID<Asset>> {
     }
     return assets[0];
   }
+
+  async findDuplicateUrlUpload(
+    url: string,
+    userId: string
+  ): Promise<WithID<Asset>> {
+    const createdAfter = Date.now() - DUPLICATE_ASSETS_THRESHOLD;
```
thomshutt, line 81: Good idea, I think DUPLICATE_ASSETS_THRESHOLD saves us from most weird states (something ends up permanently in `waiting` / `processing` etc.)

thomshutt (approved):

PR #1574, opened 2023-01-02
Fix returning S3 credentials from /api/transcode

leszko:
Redact S3 credentials while calling `/api/transcode`.

### DIFF
--- a/packages/api/src/controllers/transcode.ts
+++ b/packages/api/src/controllers/transcode.ts
@@ -1,7 +1,11 @@
 import { authorizer, validatePost } from "../middleware";
 import { Router } from "express";
 import { TranscodePayload } from "../schema/types";
-import { toObjectStoreUrl, ObjectStoreStorage } from "./helpers";
+import {
+  toObjectStoreUrl,
+  ObjectStoreStorage,
+  taskWithoutCredentials,
+} from "./helpers";
 import { ensureQueueCapacity } from "../task/scheduler";
 
 const app = Router();
@@ -44,7 +48,7 @@ app.post(
       null,
       req.user.id
     );
-    res.json(task);
+    res.json(taskWithoutCredentials(task));
   }
 );
 
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -12,7 +12,8 @@ import { S3StoreOptions as TusS3Opts } from "tus-node-server";
 import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 
 import { db } from "../store";
-import { ObjectStore } from "../schema/types";
+import { WithID } from "../store/types";
+import { ObjectStore, Task } from "../schema/types";
 
 const ITERATIONS = 10000;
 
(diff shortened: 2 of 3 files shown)


packages/api/src/controllers/transcode.ts
```diff
@@ -44,7 +48,7 @@ app.post(
       null,
       req.user.id
     );
-    res.json(task);
+    res.json(taskWithoutCredentials(task));
```
victorges, line 51: The pattern that we have established on the API is to just delete the "write-only" fields from responses. This should be done by:
 - Tagging the field `writeOnly` in the schema: https://github.com/livepeer/livepeer-com/blob/3171e199ee4887ed5c80943a48162ba2bc70b19b/packages/api/src/schema/schema.yaml#L948
  - Call `db.<table>.cleanWriteOnlyResponse` on the object: https://github.com/livepeer/livepeer-com/blob/3171e199ee4887ed5c80943a48162ba2bc70b19b/packages/api/src/controllers/task.ts#L105
  
 Notice that we use the `mung` middleware above so we don't have to repeat the clean-up logic on every API handler. Feel free to just inline a call to `db.tas.cleanWriteOnlyResponse` here though.

victorges: As this is not a behavioral regression tho, the same helper used before was just moved to being used here as well, I think it's fine as is if you prefer the current way. Would be good to have the consistency across APIs tho.

victorges (approved):
Seems like a less important filtering since we were only returning the credentials as a response to the same request that created the task with the same credentials (right?). 

Either way LGTM.

packages/api/src/controllers/transcode.ts
leszko: I think the `writeOnly` pattern is not applicable here, since we actually do want to return the `url` field (just without credentials). We could introduce a new label in the schema (like `redact` or `containsPassword`), but I believe it's too much overhead right now, so I'm merging it as it is.

Saying that, we can think about some more global redact mechanism for studio. It's actually related to the credentials issue I described here (and would love to get your feedback on it as well): https://www.notion.so/livepeer/S3-Credentials-in-logs-0f038d4ea5a54d24b65634fa6dd10267

victorges: I think overall it's not that important from an interface perspective to return the URL field, as the user has just made a request with that as the input, so they have all the data necessary for the URL, and they are not going to use the internal Object Store URL for anything anyway (we even abstracted it on their input). In fact, we **do not** want users depending on this URLs publicly or we'll lose the ability to iterate on them freely as an internal abstraction.

This is also almost the exact same case as the other places where we do write-only filtering, where it could even be useful to return a part of the data to the user (e.g. return the multistream URLs only removing the secret stream key), but we opt for a simple but consistent approach of removing those fields from the response instead. So a write-only field does seem appropriate to me here as a similar use case to the others we already have. I do agree that a "redact/containsPassword" special element would be overkill though, especially as there's no universal way of determining where the secrets would be in the URLs, so makes sense to keep this ad-hoc implementations instead.

PR #1590, opened 2023-01-05
access-control: add lit jwt verify api & playbackPolicies

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -115,6 +115,9 @@ async function validateAssetPayload(
     }
   }
 
+  // Validate playbackPolicy on creation to generate resourceId & check if unifiedAccessControlConditions is present when using lit_signing_condition
+  payload.playbackPolicy = validateAssetPlaybackPolicy(payload.playbackPolicy);
+
   return {
     id,
     playbackId,
@@ -133,6 +136,22 @@ async function validateAssetPayload(
     playbackPolicy: payload.playbackPolicy,
     objectStoreId: payload.objectStoreId || defaultObjectStoreId,
   };
+}
+
+function validateAssetPlaybackPolicy(playbackPolicy: Asset["playbackPolicy"]) {
+  if (playbackPolicy?.type == "lit_signing_condition") {
+    if (!playbackPolicy.unifiedAccessControlConditions) {
+      throw new UnprocessableEntityError(
+        `playbackPolicy.unifiedAccessControlConditions is required when using lit_signing_condition`
+      );
+    }
+  }
+
+  if (!playbackPolicy?.resourceId) {
+    // TODO: Generate resourceId if not present
+  }
+
+  return playbackPolicy;
 }
 
 async function getActiveObjectStore(id: string) {
@@ -650,7 +669,6 @@ const transcodeAssetHandler: RequestHandler = async (req, res) => {
   outputAsset.sourceAssetId = inputAsset.sourceAssetId ?? inputAsset.id;
 
   await ensureQueueCapacity(req.config, req.user.id);
-
   outputAsset = await createAsset(outputAsset, req.queue);
 
   const task = await req.taskScheduler.scheduleTask(
@@ -723,7 +741,6 @@ app.post(
     // as direct uploads will also need a lot of effort from callers to upload
     // the files, so the risk is much smaller.
     await ensureQueueCapacity(req.config, req.user.id);
-
     asset = await createAsset(asset, req.queue);
 
     const taskType = shouldUseCatalyst(req) ? "upload" : "import";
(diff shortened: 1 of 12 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -135,6 +138,25 @@ async function validateAssetPayload(
   };
 }
 
+function validateAssetPlaybackPolicy(playbackPolicy: Asset["playbackPolicy"]) {
+  if (playbackPolicy?.type == "lit_signing_condition") {
+    if (!playbackPolicy.unifiedAccessControlConditions) {
+      throw new UnprocessableEntityError(
+        `playbackPolicy.unifiedAccessControlConditions is required when using lit_signing_condition`
+      );
+    }
```
victorges, line 163: This can be enforced on the schema through a `oneOf` definition of the playback policy. Then it's a required field only for this type of policy. Certainly not a required change tho, only throwing the idea.

packages/api/src/controllers/asset.ts
```diff
@@ -135,6 +138,25 @@ async function validateAssetPayload(
...
+    if (!playbackPolicy?.resourceId) {
+      // TODO: Generate resourceId if not present
```
victorges: my curr favorite
```suggestion
      return {
        "baseUrl": "playback.livepeer.studio",
        "path": `/gate/${asset.playbackId}`,
          "orgId": "livepeer",
          "role": "",
          "extraData": `createdAt=${asset.createdAt}`
      }
```

packages/api/src/controllers/asset.ts
```diff
@@ -135,6 +138,25 @@ async function validateAssetPayload(
...
+    }
+  }
+  if (playbackPolicy?.type == "jwt") {
```
victorges, line 174: Perhaps check that it's not one of the supported ones instead
```suggestion
  if (!["public", "lit_signing_conditions"].includes(playbackPolicy?.type)) {
```

packages/api/src/controllers/asset.ts
```diff
@@ -135,6 +138,25 @@ async function validateAssetPayload(
...
+    throw new BadRequestError(
+      `playbackPolicy.type jwt is not supported for vod. Please use lit_signing_condition instead.`
```
victorges, line 176: Perhaps check that it's not one of the supported ones instead
```suggestion
  if (!["public", "lit_signing_conditions"].includes(playbackPolicy?.type)) {
    throw new BadRequestError(
      `playbackPolicy.type "${playbackPolicy?.type}" is not supported for on-demand. Please use "public" or "lit_signing_condition" instead.`
```

packages/api/src/schema/schema.yaml
```diff
@@ -770,7 +770,28 @@ components:
       properties:
         type:
           type: string
-          enum: [public, jwt]
+          enum: [public, jwt, lit_signing_condition]
+        unifiedAccessControlConditions:
+          type: array
+          description: LIT Unified Access Control Conditions
```
victorges, line 798: I never know if it's LIT or Lit after all. AFAIK is not an acronym is it?

victorges (commented):
looking good

packages/api/src/controllers/asset.ts
```diff
@@ -931,6 +951,8 @@ app.patch(
       storage = await reconcileAssetStorage(req, asset, storage);
     }
 
+    playbackPolicy = validateAssetPlaybackPolicy(playbackPolicy);
```
victorges: Actually asset playback policies **cannot** be changed. We need to disallow changing the type from `public` to `lit_signing_condition` (and vice versa) because that would require moving the asset files between object stores. We also cannot allow changing the `resourceId` of the asset as that would give control of the account owner over the creator resource. 

The right logic here is to disallow the changes mentioned above, and only allow updating the `unifiedAccessControlConditions` field instead.

packages/api/src/controllers/asset.ts
```diff
@@ -146,15 +150,25 @@ async function validateAssetPayload(
   };
 }
 
-function validateAssetPlaybackPolicy(playbackPolicy: Asset["playbackPolicy"]) {
+function validateAssetPlaybackPolicy(
+  playbackPolicy: Asset["playbackPolicy"],
+  playbackId?: string,
+  createdAt?: number
```
gioelecerati: I don't like passing the params, I would have passed the asset itself but in the validateAssetPayload I have no asset yet, will make it in a better way later

victorges: Hmm yeah, let's refrain from refactoring this too much until we decide we are going through with the feature, not just the poc

victorges (approved):

PR #1597, opened 2023-01-07
packages/api: Add version information to prometheus metrics

hjpotter92:


.github/workflows/api.yaml
```diff
@@ -1,88 +0,0 @@
...
-          DOCKER_USER: ${{ secrets.DOCKER_USER }}
-        run: |
-          # docker login
-          docker version
-          docker login -u $DOCKER_USER -p $DOCKER_PASS
-
-      - name: docker build & push
-        env:
-          DOCKER_TAGS: ${{ steps.tags.outputs.tags }}
-          VERSION: ${{ steps.tags.outputs.version }}
```
victorges, line 79: I wonder if we should remove the scripts/logic in the `package.json` to use these vars. If we're not using it anymore, might be preferred.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,98 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        build:
+          - path: packages/www
+            image: livepeerci/www
+            package: www
+          - path: packages/api
+            image: livepeerci/api
+            package: api
```
victorges, line 34: Oh nice!

packages/api/package.json
```diff
@@ -15,7 +15,7 @@
     "docker:build": "docker build $(printf ' -t livepeerci/api:%s' ${DOCKER_TAGS:-master}) --build-arg GITHUB_SHA --build-arg VERSION -f Dockerfile ../..",
     "docker:push": "for TAG in ${DOCKER_TAGS:-dev}; do docker push livepeerci/api:${TAG}; done",
     "docker": "run-s docker:build docker:push",
-    "postgres:start": "docker run --rm -d --name postgres -e POSTGRES_HOST_AUTH_METHOD=trust -p 5432:5432 -v \"$(pwd)/data/postgres:/var/lib/postgresql/data\" postgres",
+    "postgres:start": "docker run --rm -d --name postgres -e POSTGRES_HOST_AUTH_METHOD=trust -p 5432:5432 -v \"$(pwd)/data/postgres:/var/lib/postgresql/data\" postgres:14-alpine",
```
victorges, line 16: I think the tests on the API can be pretty flaky depending on these versions. Did you check if they (`yarn test`) work on the local dev env if running this postgres version?

packages/api/src/index.ts
```diff
@@ -109,6 +120,20 @@ export default async function makeApp(params: CliArgs) {
...
+      name: "livepeer_api_versions",
+      help: "Versions used by livepeer api node",
+      labelNames: ["nodeversion", "apiversion", "os", "arch"],
+    });
+    version.set(
+      {
+        nodeversion: process.version,
+        arch: process.arch,
+        os: process.platform,
+        apiversion: getGitHash(),
```
victorges: I wonder if this should be just a global field in all metrics instead, like configured on the Prometheus client. Then we could also compare any metric based on the app version. It should not increase the cardinality, but not sure if it'd increase storage use significantly (victoria compresses the hell out of it tho, maybe not).

Also, I'd name this just `version`, as `www` doesn't really exist in the API server context. WDYt?

packages/www/Dockerfile
```diff
@@ -1,12 +1,13 @@
-# To be run from the root of the repository so that we can pick up other packages
+FROM	node:14
 
-FROM node:14
+WORKDIR	/app
+
+COPY	package.json yarn.lock lerna.json	./
+
+COPY	.	.
+
```
victorges, line 8: Not a change here, but shouldn't the `yarn install` below go between these 2 copies? Otherwise there's no point in copying the `yarn` files first AFAIK

victorges (approved):
LGTM!

packages/api/src/index.ts
```diff
@@ -109,6 +120,20 @@ export default async function makeApp(params: CliArgs) {
   app.use(router);
 
   if (listen) {
+    let version = new Gauge({
+      name: "livepeer_api_versions",
```
thomshutt: I think from Eli's comment in https://discord.com/channels/423160867534929930/993331529495351357/1060584233086759024 that we're going to standardise on a metric called "version" with "version" and "app" as labels

.github/workflows/api.yaml
hjpotter92: cleaned up the scripts from `package.json` entirely.

packages/api/package.json
hjpotter92: postgres (without tags) starts pgsql 15. the sql scripts in project are incompatible w/ 15, but work fine with 14. so i left the tag as 14. tests seem to pass anyway so not sure 🤷🏾

packages/www/Dockerfile
hjpotter92: that's what i thought too. but wasn't sure how lerna packages work here...

packages/api/src/index.ts
iameli: Yeah, that's probably a good idea, but go-livepeer already uses `livepeerversion` so eh.

packages/www/Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+FROM	node:14
 
-FROM node:14
+WORKDIR	/app
+
+COPY	package.json yarn.lock lerna.json	./
+
+RUN	yarn install --frozen-lockfile
+
+COPY	.	.
```
iameli, line 7: This will dump a developer's `node_modules` directory into any locally-built Docker images. Need a `.dockerignore` if we're going this route

packages/www/package.json
```diff
@@ -11,8 +11,6 @@
     "postbuild": "next-sitemap",
     "type-check": "tsc --pretty --noEmit",
     "docs-positions": "node ./generate-docs-positions.js",
-    "docker:build": "docker build $(printf ' -t livepeerci/www:%s' ${DOCKER_TAGS:-master}) -f Dockerfile ../..",
```
iameli, line 14: You can get rid of push maybe, but I've used this command to build the image locally in the past

packages/www/Dockerfile
victorges: AFAICT nothing changed in this logic apart from changing `ADD` to `COPY`

hjpotter92: i added a .dockerignore to keep docker build context small when dev builds images locally

PR #1611, opened 2023-01-12
api: playback: added optional auth on CID playbackInfo requests

gioelecerati:

As described [here](https://www.notion.so/Duplicate-IPFS-CID-35ed12800dd24c62be7d386024b0b548), the occurrence of duplicate CID in assets across users could cause the player or other clients to retrieve only the first user uploading a certain file. This pr is to allow (for-now)optional authorization on the `playback` endpoint to retrieve the right playback info without breaking the current Livepeer player or other clients using the unauthorized endpoint.

This is currently not using the authorizer middleware as it requires authentication and this an optional one. In case the optional auth will remain, could be considered to edit the AuthzParams in auth.ts to allow this kind of flow

yarn test

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/playback.ts
```diff
@@ -44,11 +45,12 @@ const newPlaybackInfo = (
...
-  id: string
+  id: string,
+  user?: User
 ) => {
   const asset =
     (await db.asset.getByPlaybackId(id)) ??
-    (await db.asset.getByIpfsCid(id)) ??
+    (await db.asset.getByIpfsCid(id, user)) ??
     (await db.asset.getBySourceURL("ipfs://" + id)) ??
     (await db.asset.getBySourceURL("ar://" + id));
```
victorges: These are also cross-account decentralized ID queries and also need an user-preference logic.

packages/api/src/controllers/playback.ts
```diff
@@ -44,11 +45,12 @@ const newPlaybackInfo = (
...
   config: Request["config"],
   ingest: string,
-  id: string
+  id: string,
+  user?: User
 ) => {
   const asset =
     (await db.asset.getByPlaybackId(id)) ??
-    (await db.asset.getByIpfsCid(id)) ??
+    (await db.asset.getByIpfsCid(id, user)) ??
```
victorges, line 57: A flaw here is that if I pass my API key in the request I will get a 404 if the asset exists but is not owned by me. That asset that exists, even though is not in my account, is public, so IMO we should return it to anyone be them authenticated or not. In the future, we will support some kind of "public view" of an asset which is rate limited as well, which is what other users and anonymous callers would see. For now, we should just return the existing asset anyway.

So this logic needs to change to to use the asset from the user, if the user is authenticated and the asset exists, otherwise to return whatever playback ID already exists for that ID. This could be done either on the SQL query, maybe ordering based on the result being the ID we're looking for or not? Or maybe explicitly here, by doing separate queries first with the user then one without, in case of not found. We've already got a big list of queries here though and I'd not like to add 3 more now, so ideally we'd solve this on the SQL level.

WDYT?

victorges (changes requested):
Apart from the specific suggestions, could you also add some unit tests for the new behavior under `playback.test.ts`?

packages/api/src/controllers/playback.ts
gioelecerati: Mmm yeah in theory I would agree (even on long term) if the playbackUrl we provide when you pass the API key is a generic one and not one of another user. If we think this pr as a first step to let Studio track on demand usage, I would say the most important thing is to never provide playbackUrls of other users.

It could be a good idea now to do like this, and as a next step think about a system to generate a global separate playbackUrl for public assets in public view, that could be rate limited on the future or that can have a separate global usage.

Does it make sense for you?

packages/api/src/controllers/playback.ts
gioelecerati: You're right! added

packages/api/src/controllers/playback.ts
victorges: That would be a breaking change though, and I'm not sure what implications it would have. It also seems weird to me for an unauthenticated request to have more access than an authenticated one, incentivizing users to just use an unauthenticated player instead.

The authentication here is only in order to give preference to the users own assets when reading here, but it should not give them less content on their apps.

IMO we should not break this global playback URLs contract here, and continue allowing querying for any playback URL, as those should not be private. i.e. playback IDs should be all you need to build (or fetch) a playback URL
The work for the future would be giving support to these public rate limited playback URLs, but for now we should keep the current functionality working.

gioelecerati: Yeah, of course I agree we shouldn't introduce a breaking change on this.
For now, as you suggested, I added that when authenticated it returns the first available playbackUrl and added some tests for these new conditions

packages/api/src/store/asset-table.ts
```diff
@@ -42,34 +42,56 @@ export default class AssetTable extends Table<WithID<Asset>> {
...
     });
     if (!assets || assets.length < 1) {
-      return null;
+      [assets] = await this.find(query, {
+        limit: 2,
+        order,
+      });
+      if (!assets || assets.length < 1) {
+        return null;
+      }
```
victorges, line 66: Nice!

victorges (approved):
LGTM

packages/api/src/controllers/playback.ts
```diff
@@ -49,13 +49,14 @@ const newPlaybackInfo = (
 const getAssetPlaybackUrl = async (
   config: Request["config"],
   ingest: string,
-  id: string
+  id: string,
+  user?: User
```
victorges, line 53: Turns out this argument was never set on the call-site. I think this never worked and IDK how the tests didn't catch it 😱

PR #1612, opened 2023-01-12
api: Make CID search check source URL as well

victorges:

Fixes CID search on the asset list API (`/asset?cid=<x>`) by also searching the
source URL of assets, consistently with the playback info endpoint.

 - Create `sourceUrl` field filter in asset list (it's already indexed as well so that's safe)
 - Create a custom processing logic for the `cid` query string in the list API
 - Make that custom processing logic with an `OR` condition checking either pinned CID or source URL

Sending to staging to check if CID-imported assets are also found, not only CID-exported.

Fixes an issue lenstube was having for using this.

**Checklist**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - "master"
+      - "dev"
     tags:
       - "v*"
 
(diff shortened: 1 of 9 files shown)


0xcadams (approved):
LGTM (as long as tests pass!)

PR #1613, opened 2023-01-13
Removed duplicated initial fetcher calls in Tables

### LINKED ISSUE #1559: Optimization of asset list on dashboard
On the VOD asset page we currently run 2 queries during page load: 
1- grab the first 20 assets
2- count how many assets there are total

Both of these block rendering of the asset list.

At high numbers of assets, this second query causes a noticeable delay in page render. The purpose of the second query is to render the number of total assets -- this does not need to block page load, and can appear after the page has loaded.

clacladev:
Removed duplicated initial fetcher calls in Tables

- Small refactor to pass the default `initialSortBy` when initialising the table and its state. Now since the state does not change during the first rendering of the table, there is only one request to the api, not two as before
- Applied the same changes to all the tables in studio

Fixes #1559


0xcadams (approved):
LGTM

PR #1615, opened 2023-01-17
api: Create experiment abstraction for closed beta features

victorges:

This is to create a new abstraction called "experiment" to allow us to iterate on
private experiments with specific customers without risking having a feature widely
available to all the users.

This was originally designed here: https://www.notion.so/livepeer/Feature-Experimentation-Playbook-f80717040b254c1ca3591169bcf6611f

 - Create new experiment schema/table
 - Create support for GIN indexes to use in that table
 - Create new experiment API CRUD
 - Create specific endpoints for the more useful experiment actions (checking, adding/removing users etc)

`yarn test`, also putting on staging to test the internal api

Implements #1617 

**Checklist**


packages/api/src/controllers/access-control.ts
```diff
@@ -12,14 +11,11 @@ import {
...
-const accessControl = Router();
-
 const app = Router();
 
-accessControl.use("/signing-key", signingKeyApp);
-app.use("/access-control", accessControl);
+app.use("/signing-key", signingKeyApp);
 
-accessControl.post(
+app.post(
```
gioelecerati, line 18: Thank you for standardizing this

packages/api/src/schema/schema.yaml
```diff
@@ -2309,6 +2329,54 @@ components:
...
+
+    experiment:
+      type: object
+      table: experiment
+      additionalProperties: false
+      required: [name, userIds]
+      properties:
+        type:
+          type: string
+          enum: [experiment]
```
gioelecerati: Is this enum needed? Are there gonna be ever other types in the experiment table?

packages/api/src/schema/schema.yaml
```diff
@@ -2309,6 +2329,54 @@ components:
...
+          index: true
+          description:
+            User ID of the creator of the experiment. Only used for auditing as
+            experiment APIs is only accessible to admins, which can all
+            experiments anyway.
+        createdAt:
+          type: number
+          readOnly: true
+          description:
+            Timestamp (in milliseconds) at which the signing-key was created
```
gioelecerati: Left `signing-key` into the description :D

packages/api/src/controllers/experiment.ts
```diff
@@ -0,0 +1,225 @@
...
+      throw new ForbiddenError("user is not an experiment subject");
+    }
+    return next();
+  };
+
+for (const [experiment, api] of Object.entries(experimentApis)) {
+  app.use(`/api/${experiment}`, experimentSubjectsOnly(experiment), api);
+}
+
+app.get("/check/:experiment", authorizer({}), async (req, res) => {
```
gioelecerati, line 49: Maybe we want to add some `experiment.test.js` at least for this only nonAdmin function

packages/api/src/controllers/experiment.ts
```diff
@@ -0,0 +1,215 @@
...
+const app = Router();
+
+const experimentSubjectsOnly =
+  (experiment: string) => async (req, res, next) => {
+    await ensureExperimentSubject(experiment, req.user?.id);
+    return next();
+  };
+
+for (const [experiment, api] of Object.entries(experimentApis)) {
+  app.use(`/api/${experiment}`, experimentSubjectsOnly(experiment), api);
```
victorges: I'm unhappy with the final paths of the experimental APIs. e.g.: `/api/experiment/api/lit-signing-condition/verify-lit-jwt`

I think the double `/api` looks weird but don't have many other ideas. Do you think `/x` would be too bad? i.e. `/api/experiment/x/lit-signing-condition/verify-lit-jwt`

Or otherwise I take any other ideas y'all might have!

gioelecerati: Double `/api` definitely looks weird. Also `/x/` looks weird to me. What about using a number instead? Like `/1/` or `/1.0/` acting like an api manager version, so it's something that developers are used to encounter?

victorges: Hmm yeah, numbers would not be too bad but it'd seem pretty artificial to me, just to have some dummy value that we never update haha

packages/api/src/schema/schema.yaml
victorges: I followed the pattern from some old schemas here, but I think it's not necessary anymore! Probably only used when we had a single table for everything in some NoSQL databases

packages/api/src/schema/schema.yaml
victorges: oops

packages/api/src/controllers/experiment.ts
victorges: Hmmm what about an explicitly dummy `-`?
`/api/experiment/-/lit-signing-condition/verify-lit-jwt`

packages/api/src/controllers/experiment.ts
victorges: Good point! Added some tests to it.

packages/api/src/controllers/experiment.ts
victorges: Gone with `-` :)

PR #1632, opened 2023-01-24
api: Expose UCAN id

### LINKED ISSUE #364: Create DID API in Livepeer Studio
Add new endpoint here: https://github.com/livepeer/studio/tree/master/packages/api/src/controllers

The output format as specified in the Tech Design Doc.

red-0ne:

Add `/api/did` endpoint to expose Livepeer's DID key got from UCAN_DID` env variable

Fixes https://github.com/livepeer/catalyst/issues/364

### DIFF
--- a/packages/api/src/controllers/did.js
+++ b/packages/api/src/controllers/did.js
@@ -2,8 +2,14 @@ import Router from "express/lib/router";
 
 const app = Router();
 
-app.get("/", (_, res) => {
-  res.json({ did: process.env.UCAN_DID || "" });
+app.get("/", (req, res) => {
+  const did = req.config.did;
+  if (did) {
+    res.json({ did });
+  } else {
+    res.status(501);
+    res.json({ errors: ["DID key not configured"] });
+  }
 });
 
 export default app;
--- a/packages/api/src/controllers/did.js
+++ b/packages/api/src/controllers/did.js
@@ -1,0 +1,9 @@
+import Router from "express/lib/router";
+
+const app = Router();
+
+app.get("/", (_, res) => {
+  res.json({ did: process.env.UCAN_DID || "" });
+});
+
+export default app;
--- a/packages/api/src/controllers/index.ts
+++ b/packages/api/src/controllers/index.ts
@@ -22,6 +22,7 @@ import region from "./region";
 import session from "./session";
 import cdnData from "./cdn-data";
 import playback from "./playback";
+import did from "./did";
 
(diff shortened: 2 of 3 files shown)


packages/api/src/controllers/did.js
```diff
@@ -0,0 +1,9 @@
+import Router from "express/lib/router";
+
+const app = Router();
+
+app.get("/", (_, res) => {
+  res.json({ did: process.env.UCAN_DID || "" });
```
leszko: 1. I think the env variable name is not correct. In your PR https://github.com/livepeer/livepeer-infra/pull/1163, you defined it as `LP_API_DID`, here you use `UCAN_DID`. Could you double-check?

Also, please check how the environment variables are resolved in Livepeer Studio. I see that the `LP_API_` prefix is parsed somehow as flags, so please check if we should pass it somehow via flags to be consistent with other env variables.

2. I think that if the variable is not defined we should not return an empty string, but rather the specific HTTP Error code, I think `501` in this case.

3. I'd consider some simple validation if the env variable is in the correct format `did:key:...`.

leszko (changes requested):
Thanks for the PR @red-0ne 

Added a few comments. PTAL. Saying that, I checked and it works fine 😃

packages/api/src/controllers/did.js
victorges: You should actually read that env from the parsed `req.config.ucanDid` instead! With the corresponding flag declaration in `parse-cli.ts` file. Those can also come from env vars (if you're curious check yargs usage), and in fact all of them do in our production deployments. Envs are also transformed from LP_API_SNAKE_CASE to lowerCamel

packages/api/src/store/experiment-table.ts
```diff
@@ -7,8 +7,12 @@ import Table from "./table";
 import { WithID } from "./types";
 
 export async function isExperimentSubject(experiment: string, userId: string) {
-  const { audienceUserIds } = await db.experiment.getByNameOrId(experiment);
-  return audienceUserIds?.includes(userId) ?? false;
+  try {
```
leszko: Is this change related to your PR?

leszko (approved):
Added one comment. Other than that, LGTM 👍

packages/api/src/store/experiment-table.ts
red-0ne: It came from a rebase from master.

PR #1638, opened 2023-01-31
api: Allow calling setActive only to bump stream's lastSeen field

victorges:
Just needed to add some checks to make sure we some things (like hooks) when the isActive value doesn't change.

Unblocks: https://github.com/livepeer/stream-tester/pull/259


victorges (commented):
self-LGTM after adding tests to cover all scenarios

PR #1641, opened 2023-02-01
playback: added static MP4 renditions

### LINKED ISSUE #130: Catalyst <-> Studio integration for Short form videos
After a sync with @emranemran , some work is needed to be done on Studio Task Runner and Catalyst for short form videos, specifically:

- add `AutoMP4s: bool` in `api/vod` in both the request studio & the catalyst-api struct
- Add `Mp4Outputs` to TranscodeStatusMessage from Catalyst to Studio
- Potentially, before populating the struct of TranscodeStatusMessage check on the storage if the mp4 is there (may require more time)
- `api/transcode` endpoint need to recognize the AutoMP4s

gioelecerati:

Fixes https://github.com/livepeer/task-runner/issues/130
Added static renditions to Playback Info API to support incoming short form mp4s from Catalyst

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -183,7 +183,7 @@ export function getStaticPlaybackInfo(
     .filter((f) => f.type === "static_transcoded_mp4")
     .map((f) => ({
       playbackUrl: pathJoin(os.publicUrl, asset.playbackId, f.path),
-      size: asset.size,
+      size: f.size ?? 0,
     }));
 }
 
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -176,7 +176,7 @@ export function getStaticPlaybackUrls(
 ): string[] {
   let files = asset.files;
   if (files) {
-    files = files.filter((f) => f.type === "static_file");
+    files = files.filter((f) => f.type === "static_transcoded_mp4");
   } else {
     return [];
   }
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -169,14 +169,22 @@ async function getActiveObjectStore(id: string) {
   return os;
 }
 
-export function getStaticPlaybackUrls(
+export type StaticPlaybackInfo = {
+  playbackUrl: string;
+  size: number;
+};
+
+export function getStaticPlaybackInfo(
   {}: Request["config"],
   asset: WithID<Asset>,
   os: ObjectStore
-): string[] {
+): StaticPlaybackInfo[] {
   return (asset.files ?? [])
     .filter((f) => f.type === "static_transcoded_mp4")
-    .map((f) => pathJoin(os.publicUrl, asset.playbackId, f.path));
+    .map((f) => ({
+      playbackUrl: pathJoin(os.publicUrl, asset.playbackId, f.path),
+      size: asset.size,
+    }));
 }
 
 export function getPlaybackUrl(
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -181,9 +181,6 @@ export function getStaticPlaybackUrls(
     return [];
   }
   return files?.map((f) => {
-    if (os.id !== vodCatalystObjectStoreId) {
-      return pathJoin(os.publicUrl, asset.playbackId, f.path);
-    }
     return pathJoin(os.publicUrl, asset.playbackId, f.path);
   });
 }
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -170,19 +170,13 @@ async function getActiveObjectStore(id: string) {
 }
 
 export function getStaticPlaybackUrls(
-  { vodCatalystObjectStoreId }: Request["config"],
+  {}: Request["config"],
   asset: WithID<Asset>,
   os: ObjectStore
 ): string[] {
-  let files = asset.files;
-  if (files) {
-    files = files.filter((f) => f.type === "static_transcoded_mp4");
-  } else {
-    return [];
-  }
-  return files?.map((f) => {
-    return pathJoin(os.publicUrl, asset.playbackId, f.path);
-  });
+  return (asset.files ?? [])
+    .filter((f) => f.type === "static_transcoded_mp4")
+    .map((f) => pathJoin(os.publicUrl, asset.playbackId, f.path));
 }
 
 export function getPlaybackUrl(
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -178,7 +178,6 @@ export type StaticPlaybackInfo = {
 };
 
 export function getStaticPlaybackInfo(
-  {}: Request["config"],
   asset: WithID<Asset>,
   os: ObjectStore
 ): StaticPlaybackInfo[] {
@@ -186,11 +185,12 @@ export function getStaticPlaybackInfo(
     .filter((f) => f.type === "static_transcoded_mp4")
     .map((f) => ({
       playbackUrl: pathJoin(os.publicUrl, asset.playbackId, f.path),
-      size: f.spec?.size ?? null,
-      height: f.spec?.height ?? null,
-      width: f.spec?.width ?? null,
-      bitrate: f.spec?.bitrate ?? null,
-    }));
+      size: f.spec?.size,
+      height: f.spec?.height,
+      width: f.spec?.width,
+      bitrate: f.spec?.bitrate,
+    }))
+    .sort((a, b) => b.bitrate - a.bitrate);
 }
 
 export function getPlaybackUrl(
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -169,6 +169,25 @@ async function getActiveObjectStore(id: string) {
   return os;
 }
 
+export function getStaticPlaybackUrls(
+  { vodCatalystObjectStoreId }: Request["config"],
+  asset: WithID<Asset>,
+  os: ObjectStore
+): string[] {
+  let files = asset.files;
+  if (files) {
+    files = files.filter((f) => f.type === "static_file");
+  } else {
+    return [];
+  }
+  return files?.map((f) => {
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -169,6 +169,25 @@ async function getActiveObjectStore(id: string) {
...
+  asset: WithID<Asset>,
+  os: ObjectStore
+): string[] {
+  let files = asset.files;
+  if (files) {
+    files = files.filter((f) => f.type === "static_file");
+  } else {
+    return [];
+  }
+  return files?.map((f) => {
```
victorges: **nit**: can be a single functional concatenation here:
```suggestion
  return (asset.files ?? [])
      .filter((f) => f.type === "static_file")
      .map((f) => {
        ...
      });
```

packages/api/src/controllers/asset.ts
```diff
@@ -169,6 +169,25 @@ async function getActiveObjectStore(id: string) {
...
+    if (os.id !== vodCatalystObjectStoreId) {
+      return pathJoin(os.publicUrl, asset.playbackId, f.path);
+    }
```
victorges: Let's skip this check now, I think we'll just stay with the OS public URL for long.

packages/api/src/controllers/playback.ts
```diff
@@ -19,33 +22,56 @@ type PlaybackInfo = {
     playbackPolicy?: Asset["playbackPolicy"] | Stream["playbackPolicy"];
     source: {
       // the only supported format is HLS for now
-      hrn: "HLS (TS)";
-      type: "html5/application/vnd.apple.mpegurl";
+      hrn: "HLS (TS)" | "MP4";
+      type: "html5/application/vnd.apple.mpegurl" | "html5/video/mp4";
```
victorges, line 26: Can you check with Jaron if these would be the values used by mist as well?

packages/api/src/controllers/playback.ts
```diff
@@ -19,33 +22,56 @@ type PlaybackInfo = {
...
+          hrn: "HLS (TS)",
+          type: "html5/application/vnd.apple.mpegurl",
+          url: hlsUrl,
+        },
+      ],
+    },
+  };
+  if (staticFilesPlaybackUrls && staticFilesPlaybackUrls.length > 0) {
+    for (let staticUrl of staticFilesPlaybackUrls) {
+      const rendition = staticUrl.split("/").pop().replace(".mp4", "");
```
victorges: Let's have this as an explicit field in the `files[]` object please! hahahah

Also, I think a resolution or file size would be better here. Idk we should have logic playing arbitrary rendition names like `720p60f` or sth like that 😆

packages/api/src/controllers/playback.ts
gioelecerati: Confirmed as `"html5/video/mp4"` by Jaron! On `ws` instead would be `ws/video/mp4`

packages/api/src/controllers/playback.ts
gioelecerati: Ahahah yep. cc'ing @emranemran @mjh1 on why we need the probing in the output transcoded file. Currently I have just the size, I think the player could use that but wouldn't be able to know exactly which resolution it's using

mjh1: sure, i'll open a ticket for adding width and height fields. we can get this from the mediaconvert job status so wouldn't necessarily need to probe the output.

packages/api/src/controllers/playback.ts
```diff
@@ -19,33 +22,56 @@ type PlaybackInfo = {
     playbackPolicy?: Asset["playbackPolicy"] | Stream["playbackPolicy"];
     source: {
       // the only supported format is HLS for now
-      hrn: "HLS (TS)";
-      type: "html5/application/vnd.apple.mpegurl";
+      hrn: "HLS (TS)" | "MP4";
+      type: "html5/application/vnd.apple.mpegurl" | "html5/video/mp4";
       url: string;
+      rendition?: string;
```
0xcadams: Just for my understanding - what does this field correspond to?

gioelecerati: This is supposed to be the rendition of the specific output, for example `360p`, however we are changing it today to `size`, as we don't have the probe data of the file yet coming back from catalyst. After we get it, we can provide additional metadata such as `bitrate` for the player

packages/api/src/controllers/playback.ts
emranemran: Makes sense. 
@mjh1: filed this for you to just use mc job status for now to unblock studio team: https://github.com/livepeer/catalyst-api/issues/428

And i can take this one to add ffprobe logic to generate more metadata on output files: https://github.com/livepeer/catalyst-api/issues/429

packages/api/src/controllers/playback.ts
0xcadams: Yeah, IMO size is a bit arbitrary to the player, since it'll have no concept of how size corresponds to the actual video dimensions and we won't be able to reliably get network speed. Ideally resolution details are included in the response long-term, since we're choosing a singular mp4 resolution to render since it's not streaming.

gioelecerati: I totally agree. Currently you can understand the resolution from the name (e.g. https://asset-cdn.lp-playback.monster/hls/e02eyumal0sigr24/static360p0.mp4), but I opened this on the api client https://github.com/livepeer/go-api-client/pull/27 which can be expanded later with the additional probe data when ready. For now only the size is included.

victorges: @0xcadams We can also easily include an approximate `bitrate` here, by dividing this `size` with the video duration. Does that help?

packages/api/src/controllers/asset.ts
```diff
@@ -169,6 +169,30 @@ async function getActiveObjectStore(id: string) {
   return os;
 }
 
+export type StaticPlaybackInfo = {
+  playbackUrl: string;
+  size: number;
+  height: number;
+  width: number;
+  bitrate: number;
+};
```
victorges, line 178: At some point we should create proper schemas for these types, especially to get playback info API documented in there as well

packages/api/src/controllers/asset.ts
```diff
@@ -169,6 +169,30 @@ async function getActiveObjectStore(id: string) {
...
+
+export function getStaticPlaybackInfo(
+  {}: Request["config"],
```
victorges: Should probably remove this unused arg.

packages/api/src/controllers/asset.ts
```diff
@@ -169,6 +169,30 @@ async function getActiveObjectStore(id: string) {
...
+): StaticPlaybackInfo[] {
+  return (asset.files ?? [])
+    .filter((f) => f.type === "static_transcoded_mp4")
+    .map((f) => ({
+      playbackUrl: pathJoin(os.publicUrl, asset.playbackId, f.path),
+      size: f.spec?.size ?? null,
+      height: f.spec?.height ?? null,
+      width: f.spec?.width ?? null,
+      bitrate: f.spec?.bitrate ?? null,
+    }));
```
victorges: WDYT of sorting these for a consistent response? For descending bitrate order:
```suggestion
    })).sort((a, b) => b.bitrate - a.bitrate);
```

(apparently `null == 0` for this, no need even to coalesce)

packages/api/src/controllers/playback.ts
```diff
@@ -19,33 +23,53 @@ type PlaybackInfo = {
     playbackPolicy?: Asset["playbackPolicy"] | Stream["playbackPolicy"];
     source: {
       // the only supported format is HLS for now
```
victorges: Remove this comment

packages/api/src/controllers/playback.ts
```diff
@@ -19,33 +23,53 @@ type PlaybackInfo = {
...
-      hrn: "HLS (TS)";
-      type: "html5/application/vnd.apple.mpegurl";
+      hrn: "HLS (TS)" | "MP4";
+      type: "html5/application/vnd.apple.mpegurl" | "html5/video/mp4";
       url: string;
+      size?: number;
+      width?: number;
+      height?: number;
+      bitrate?: number;
```
victorges, line 31: **nit:** Can also make the params clear through the type with a `|`. e.g.
```suggestion
      hrn: "HLS (TS)";
      type: "html5/application/vnd.apple.mpegurl";
      url: string;
    } | {
      hrn: "MP4";
      type: "html5/video/mp4";
      url: string;
      size: number;
      width: number;
      height: number;
      bitrate: number;
```

packages/api/src/controllers/asset.ts
```diff
@@ -169,6 +169,30 @@ async function getActiveObjectStore(id: string) {
...
+  os: ObjectStore
+): StaticPlaybackInfo[] {
+  return (asset.files ?? [])
+    .filter((f) => f.type === "static_transcoded_mp4")
+    .map((f) => ({
+      playbackUrl: pathJoin(os.publicUrl, asset.playbackId, f.path),
+      size: f.spec?.size ?? null,
+      height: f.spec?.height ?? null,
+      width: f.spec?.width ?? null,
+      bitrate: f.spec?.bitrate ?? null,
```
victorges: Are these values optional? If they are only not present during the intermediate versions we made during testing, I think we should remove this coalescing logic here. It will only make ourselves wonder in 6 months "omg are these not always present?" hahaha

victorges (approved):
Nice!

packages/api/src/controllers/asset.ts
gioelecerati: They are not optional, just being afraid while on staging that something wasn't coming in! Removed :)

packages/api/src/controllers/playback.ts
gioelecerati: Removed!

packages/api/src/controllers/asset.ts
gioelecerati: Cool :)

packages/api/src/controllers/asset.ts
gioelecerati: Let's leave this and the `|` in the type in the other comment for later?

PR #1644, opened 2023-02-06
Add convertion from base64 to base64url

leszko:
Convert the web3.storage proof from base64 to base64url

I believe that we should accept the base64 format from the user because it's more-widely used, plus the user does not need to be aware that internally we use some `w3://` URL.

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -10,6 +10,7 @@ import { createHmac } from "crypto";
 import { S3Client, PutObjectCommand, S3ClientConfig } from "@aws-sdk/client-s3";
 import { S3StoreOptions as TusS3Opts } from "tus-node-server";
 import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
+import base64url from "base64url";
 
 import { WithID } from "../store/types";
 import { ObjectStore, Task } from "../schema/types";
@@ -159,7 +160,7 @@ export function toWeb3StorageUrl(storage: Web3StoreStorage): string {
   if (!storage.credentials || !storage.credentials.proof) {
     throw new Error("undefined property 'credentials.proof'");
   }
-  return `w3s://${storage.credentials.proof}@/`;
+  return `w3s://${base64url.fromBase64(storage.credentials.proof)}@/`;
 }
 
 export function toObjectStoreUrl(storage: ObjectStoreStorage): string {
--- a/packages/api/src/controllers/helpers.test.ts
+++ b/packages/api/src/controllers/helpers.test.ts
@@ -190,11 +190,11 @@ describe("convert w3 storage to object store URL", () => {
       type: "web3.storage",
       credentials: {
         proof:
-          "EaJlcm9vdHOAZ3ZlcnNpb24BmgIBcRIg2uxHpcPYSWNtifMKFkPC7IEDvFDCxCd3ADViv0coV7SnYXNYRO2hA0AnblHEW38s3lSlcwaDjPn",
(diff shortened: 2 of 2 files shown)


packages/api/src/controllers/helpers.ts
```diff
@@ -159,7 +160,7 @@ export function toWeb3StorageUrl(storage: Web3StoreStorage): string {
   if (!storage.credentials || !storage.credentials.proof) {
     throw new Error("undefined property 'credentials.proof'");
   }
-  return `w3s://${storage.credentials.proof}@/`;
+  return `w3s://${base64url.fromBase64(storage.credentials.proof)}@/`;
```
victorges, line 163: Don't we want users to specify the `credentials.proof` field in `base64url` as well? Could work better inside JSON as well and just keep it e2e consistent to avoid confusion. Or is there any downside in taking the input as `base64url`?

victorges (approved):
LGTM either way, but I think being consistent on the public interface, taking things in `base64url` would be preferrable

packages/api/src/controllers/helpers.ts
leszko: The only downside I see is the UX. I think there is great CMD for all OS, so users would need to convert it somehow on their own.

To give you an example, here's how users will need to generate the proof on their side.

```
$ w3 delegation create did:key:z6MkgfrQdK62xvXmrysMBvaBHTZL3VvFqH87NVdLD2x4yAa9 | base64
```

the CMD `base64` is available everywhere, on every OS. With `base64url`, they will probably need to install the CMD with NPM: https://www.npmjs.com/package/base64url

Wdyt? @victorges

victorges: Oh I see. I wasn't aware that it would be harder for users to encode in `base64url`. I'm fine with keeping this as is then, tho I am still a little uneasy with the inconsistency between the interfaces.

The only idea I had was to support _both_ encodings here. So the interface is officially `base64url` as well, but we offer a syntax sugar in the API that allows you to specify it as a regular `base64` and we convert it internally for you. This way it's not inconsistent, just a more flexible interface. From what I checked, we could do this by just attempting to decode with `base64url` then if there's an error, with `base64` instead (then re-encode as url).

All this would be some extra complexity here only cause I ill-advisedly suggested to use `base64url` encoding on the `drivers`. Given this new context, I'd be fine switching back to `base64+url encoding` instead, but I guess we'd need at least 2 new PRs for that (`go-tools` then `catalyst-api`). In the end, we are going to deal with the double encoding ourselves anyway, and I think it would be consistent with the other drivers as well (Google S3 credentials are in base64 and they need to be URL-encoded as well).

But either way, in summary, no required changes here. I'm a little unhappy with the inconsistency but don't want to create more work. So I'm fine with either option:
 - Leave it as is, `base64url` internally and `base64` externally
 - Add support for `base64url` here _as well_, but still keeping raw `base64` support for ergonomy
 - Changing the `drivers` implementation to use `base64`+URL encoding

leszko: I hear you Victor, I don't like the inconsistency neither. I like your idea of accepting both `base64url` and `base64`. Let me add it.

leszko: I actually just discovered that it already worked as expected, so just added a unit test case.

The story is that if you perform `base64url.fromBase()` on a string already base64url-encoded, then it has no effect. That's because base64url-encoding is actually just base64+url-escape, so it's just replacing a few characters.

victorges: Oh that's amazing! All good then 😁

I think base64url is different from URL encoding tho. It replaces the special characters from b64 (I think / and +) with URL-compatible ones (_ and -), instead of percent-encoding them. It also removes any padding with = (not sure what it replaces it with)

But if this helper function literally just replaces the special characters and trims the trailing = chars, i guess it will be a noop if it's already b64url yeah! Cool :)

PR #1645, opened 2023-02-06
api: Allow providing decryption keys to decrypt uploaded assets

victorges:
This is in order to support uploading files encrypted with Lit directly, normally stored on Bundlr.

Since the focus is on Lot, we also have an `algorithm` field which can only be AES-256-CBC for now
which is what's used for Lit. It is surprisingly cumbersome to do crypto in Go tho, so adding more 
algorithms here might not be as cheap as just calling a different function.

 - Add field in task, asset and new asset payload
 - Pass the field around internally and omit it from responses
 - Make sure we delete the field in the end of a task as well
 - Improve credential cleanup logic

Send lit and manually encrypted files to check it worked.

Yes of a bundlr integration. Missing explicit issue.

**Checklist**

### DIFF
--- a/packages/api/src/controllers/task.ts
+++ b/packages/api/src/controllers/task.ts
@@ -69,17 +69,28 @@ function taskWithIpfsUrls(
   });
 }
 
-function taskWithoutCredentials(task: WithID<Task>): WithID<Task> {
-  if (task?.type !== "transcode-file") {
-    return task;
+export function taskParamsWithoutCredentials(
+  type: Task["type"],
+  params: Task["params"]
+): Task["params"] {
+  const result = _.cloneDeep(params);
+  switch (type) {
+    case "transcode-file":
+      result["transcode-file"].input.url = deleteCredentials(
+        params["transcode-file"].input.url
+      );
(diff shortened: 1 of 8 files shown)


packages/api/src/controllers/task.ts
```diff
@@ -69,6 +69,30 @@ function taskWithIpfsUrls(
...
+        params["transcode-file"].input.url
+      );
+      result["transcode-file"].storage.url = deleteCredentials(
+        params["transcode-file"].storage.url
+      );
+      break;
+    case "upload":
+      const encryption = params.upload?.encryption;
+      if (encryption) {
+        result.upload.encryption = { ...encryption, key: "***" };
```
gioelecerati, line 89: I am not getting this key `***` what's stands for. Is it redacted when retrieving?

gioelecerati (approved):
Looking good to me

packages/api/src/controllers/task.ts
victorges: yeah! we redact it even on the DB as well after the task is done, similarly to transcode api

victorges (commented):
self-review LGTM

PR #1651, opened 2023-02-14
asset: Delete from object storage task

gioelecerati:
… task status del update

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/task.ts
+++ b/packages/api/src/controllers/task.ts
@@ -299,6 +299,15 @@ app.post("/:id/status", authorizer({ anyAdmin: true }), async (req, res) => {
     }
   }
 
+  if (task.type == "delete" && task.status.phase == "completed") {
+    await req.taskScheduler.updateAsset(task.inputAssetId, {
+      status: {
+        phase: "deleted",
+        updatedAt: Date.now(),
+      },
+    });
+  }
+
   res.status(200);
   res.json({ id, status });
 });
--- a/packages/api/src/controllers/task.ts
+++ b/packages/api/src/controllers/task.ts
@@ -285,41 +285,35 @@ app.post("/:id/status", authorizer({ anyAdmin: true }), async (req, res) => {
   }
   if (task.inputAssetId) {
     const asset = await db.asset.get(task.inputAssetId);
-    if (task.type === "export") {
-      if (task.id === asset?.storage?.status?.tasks.pending) {
-        await req.taskScheduler.updateAsset(asset.id, {
-          storage: {
-            ...asset.storage,
+    switch (task.type) {
+      case "export":
+        if (task.id === asset?.storage?.status?.tasks.pending) {
+          await req.taskScheduler.updateAsset(asset, {
+            storage: {
+              ...asset.storage,
+              status: {
+                phase: "processing",
+                progress: doc.progress,
+                tasks: asset.storage.status.tasks,
+              },
+            },
+          });
+        }
+        break;
+      case "delete":
+        await req.taskScheduler.updateAsset(
+          asset,
+          {
             status: {
-              phase: "processing",
+              phase: "deleting",
               progress: doc.progress,
-              tasks: asset.storage.status.tasks,
+              updatedAt: Date.now(),
             },
           },
-        });
-      }
-    } else if (task.type === "delete") {
-      await req.taskScheduler.updateAsset(
-        asset.id,
-        {
-          status: {
-            phase: "deleting",
-            progress: doc.progress,
-            updatedAt: Date.now(),
-          },
-        },
-        { allowedPhases: ["deleting"] }
-      );
+          { allowedPhases: ["deleting"] }
+        );
+        break;
     }
-  }
-
-  if (task.type == "delete" && task.status.phase == "completed") {
-    await req.taskScheduler.updateAsset(task.inputAssetId, {
-      status: {
-        phase: "deleted",
-        updatedAt: Date.now(),
-      },
-    });
   }
 
   res.status(200);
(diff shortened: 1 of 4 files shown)


packages/api/src/controllers/task.ts
```diff
@@ -285,20 +285,43 @@ app.post("/:id/status", authorizer({ anyAdmin: true }), async (req, res) => {
   }
   if (task.inputAssetId) {
     const asset = await db.asset.get(task.inputAssetId);
-    if (task.id === asset?.storage?.status?.tasks.pending) {
-      await req.taskScheduler.updateAsset(asset.id, {
-        storage: {
-          ...asset.storage,
+    if (task.type === "export") {
```
victorges: **nit:** WDYT of a `switch`

packages/api/src/controllers/task.ts
```diff
@@ -285,20 +285,43 @@ app.post("/:id/status", authorizer({ anyAdmin: true }), async (req, res) => {
...
+              phase: "processing",
+              progress: doc.progress,
+              tasks: asset.storage.status.tasks,
+            },
+          },
+        });
+      }
+    } else if (task.type === "delete") {
+      await req.taskScheduler.updateAsset(
+        asset.id,
```
victorges: **nit:** Since you already have the asset object here, you can just pass it as the argument instead of `asset.id` and avoid a redundant query on the DB inside `updateAsset`. Same for the other update above (it was already like that, my bad)

packages/api/src/controllers/task.ts
```diff
@@ -285,20 +285,43 @@ app.post("/:id/status", authorizer({ anyAdmin: true }), async (req, res) => {
...
 
+  if (task.type == "delete" && task.status.phase == "completed") {
+    await req.taskScheduler.updateAsset(task.inputAssetId, {
+      status: {
+        phase: "deleted",
+        updatedAt: Date.now(),
+      },
+    });
+  }
+
```
victorges: This must be removed. Terminal events are not processed here, only on scheduler.
```suggestion
```

packages/api/src/schema/schema.yaml
```diff
@@ -1219,6 +1219,8 @@ components:
                 - processing
                 - ready
                 - failed
+                - deleted
+                - deleting
```
victorges: **nit:** ordering temporally
```suggestion
                - deleting
                - deleted
```

packages/api/src/schema/schema.yaml
```diff
@@ -1674,6 +1677,14 @@ components:
               $ref: "#/components/schemas/task/properties/params/properties/upload"
             export:
               $ref: "#/components/schemas/export-task-params"
+            delete:
+              type: object
+              additionalProperties: false
+              description: Parameters for the delete task
+              properties:
+                id:
```
victorges, line 1685: lmao 😆

packages/api/src/schema/schema.yaml
```diff
@@ -1793,6 +1804,14 @@ components:
...
             import:
               $ref: "#/components/schemas/task/properties/output/properties/upload"
+            delete:
+              type: object
+              additionalProperties: false
+              description: Output of the delete task
+              properties:
+                objectStoreId:
+                  type: string
+                  description: ID of the deleted asset
```
victorges: This field should be removed. Object store ID is already in the asset, and the description is wrongly copied from the other field.

If you really want to have _some_ return here, WDYT of returning the number of files and cumulative size of the files deleted?

packages/api/src/schema/schema.yaml
```diff
@@ -1674,6 +1677,14 @@ components:
...
             export:
               $ref: "#/components/schemas/export-task-params"
+            delete:
+              type: object
+              additionalProperties: false
+              description: Parameters for the delete task
+              properties:
+                id:
+                  type: string
+                  description: ID of the asset to delete
```
victorges: Can you expand this description explaining that this field is not necessary and not used, kept here only to avoid an empty object?

packages/api/src/task/scheduler.ts
```diff
@@ -138,6 +138,30 @@ export class TaskScheduler {
           },
         });
         break;
+      case "delete":
```
victorges, line 141: You should add some handling for the failure case as well. I'm not sure myself what to do in that case, especially since the asset will probably end up in a partially deleted state which would be tricky. We can just noop and leave the asset hanging on the `deleting` phase, but maybe there's something better that we can do. WDYT?

packages/api/src/task/scheduler.ts
```diff
@@ -409,20 +433,23 @@ export class TaskScheduler {
...
-    await this.queue.publishWebhook("events.asset.deleted", {
-      type: "webhook_event",
-      id: uuid(),
-      timestamp: Date.now(),
-      event: "asset.deleted",
-      userId: asset.userId,
-      payload: {
-        asset: {
+
+    await db.asset.update(asset.id, {
```
victorges: The order is important here, this should only be done after the task has been scheduled. So invert this statement with the `scheduleTask` one.

packages/api/src/task/scheduler.ts
```diff
@@ -409,20 +433,23 @@ export class TaskScheduler {
...
-    await db.asset.markDeleted(asset.id);
-    await this.queue.publishWebhook("events.asset.deleted", {
-      type: "webhook_event",
-      id: uuid(),
-      timestamp: Date.now(),
-      event: "asset.deleted",
-      userId: asset.userId,
-      payload: {
-        asset: {
+
```
victorges, line 436: You should probably check if the asset doesn't already have `phase: deleted`, in which case we should either return an error like a 404. Better yet, this function is only ever called with a full `Asset`, never with a `string`. We should remove the `string` parameter here and make all the checks we need on the controller side, together with the place we check the asset belongs to the user that is actually calling a delete on it. There we should check:
 - that the asset is not already `phase: deleted`, in which case we should return a 404
 - otherwise, if the asset has `deleted: true`, only allow "redeleting" if user is an admin (so we can use this to cleanup old assets)

victorges (changes requested):
There is 1 change that must be removed which is the erroneous handling of `completed` state on the `/status` API. The others are mostly opinions, but some held strongly

PR #1652, opened 2023-02-15
Revert 'return orchestrators pending activation for Confluence upgrade'

leszko:
After the Arbitrum migration, we should move to the logic as before, meaning that only active Os can accept transcoding.

Related Discord discussion: https://discord.com/channels/423160867534929930/750796877762527262/1074970658544558092

This PR is an exact revert of this: https://github.com/livepeer/studio/pull/911

Tested locally.


victorges (approved):
I don't have the full context of this but the code changes make sense.

adamsoffer (approved):
LGTM

PR #1659, opened 2023-02-24
playback: return first created Asset by CID both in storage & import

gioelecerati:
The playback API was looking for CIDs both into the storage field and into the source import URL, in this order.
However, this was returning the first created asset with storage.ipfs.cid, as that was the first check.

With this pr, they are both checked and compared to retrieve the asset actually created first in both cases.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/store/asset-table.ts
+++ b/packages/api/src/store/asset-table.ts
@@ -2,6 +2,7 @@ import sql from "sql-template-strings";
 import { Asset, Task, User } from "../schema/types";
 import Table from "./table";
 import { QueryOptions, WithID } from "./types";
+import { db } from "../store";
 
 const DUPLICATE_ASSETS_THRESHOLD = 15 * 60 * 1000; // 15 mins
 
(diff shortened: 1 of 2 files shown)


packages/api/src/store/asset-table.ts
```diff
@@ -43,6 +44,17 @@ export default class AssetTable extends Table<WithID<Asset>> {
   }
 
   async getByIpfsCid(cid: string, user?: User): Promise<WithID<Asset>> {
+    let assetByStorageCid = await db.asset.getByStorageCid(cid, user);
+    let assetBySourceUrl = await db.asset.getBySourceURL("ipfs://" + cid, user);
```
victorges, line 48: I think we should make this as a single round trip to the DB by making the condition on line `59` as an OR of either `storage->ipfs->>cid` or `source->>'url'`. We can probably keep the condition for only ready assets as well, no need to nest that condition only for the source URL check. WDYT?

packages/api/src/store/asset-table.ts
```diff
@@ -43,6 +44,17 @@ export default class AssetTable extends Table<WithID<Asset>> {
   }
 
   async getByIpfsCid(cid: string, user?: User): Promise<WithID<Asset>> {
```
victorges, line 46: Please check other uses of this function and make sure they still make sense (e.g. not making a separate redundant query for the source URL as well). Always a good practice when changing the behavior of a function, and this one specifically I **think** was used from the asset controller as well.

PR #1661, opened 2023-03-01
asset: static mp4 flag

gioelecerati:

Allow generate mp4 flag
refer to https://github.com/livepeer/catalyst-api/pull/518
https://github.com/livepeer/task-runner/pull/161

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1044,6 +1044,10 @@ components:
           example: ea03f37e-f861-4cdd-b495-0e60b6d753ad
           index: true
           description: Used to form recording URL for HLS playback
+        generateMp4s:
+          type: boolean
+          writeOnly: true
+          description: Whether to generate MP4s for the asset.
         files:
           type: array
           writeOnly: true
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1044,6 +1044,10 @@ components:
           example: ea03f37e-f861-4cdd-b495-0e60b6d753ad
           index: true
           description: Used to form recording URL for HLS playback
+        generateMp4s:
```
victorges: I think an "action" field like this makes sense in the task objects, but not in the assets which are a description of an entity, not of an action. 

In this case, WDYT of using a descriptive field instead of an action here? I like `staticMp4s` instead, WDYT? 

FTR mux uses a `mp4_support: standard` field instead (not even a boolean). I think we're fine with a bool for now.

victorges (approved):
LGTM but the field name is a bit weird to me being part of a "descriptive" entity, not one that represents an action (like the task)

packages/api/src/schema/schema.yaml
gioelecerati: Renamed to `staticMp4` on the api parameter, and on task runner I am sending `forceMp4` in the output location when it's true!

packages/api/src/schema/schema.yaml
```diff
@@ -1044,6 +1044,10 @@ components:
           example: ea03f37e-f861-4cdd-b495-0e60b6d753ad
           index: true
           description: Used to form recording URL for HLS playback
+        staticMp4:
```
victorges, line 1047: @iameli Speak now or forever hold your peace 🕊️

victorges (approved):
re-LGTM!

PR #1662, opened 2023-03-02
api: Improve session lifecycle handling

victorges:

This is a bit massive change, but it completely rewrites the way that we handle livestream
sessions and their corresponding recordings on the API. 

The explanation here started to get too big so I wrote a doc instead: 
https://lvpr.link/live-rec-fact

 - Made the "session merging" fields private and deprecated
 - Removed all "write" usage of them, keeping only the "read" usage for backward compatibility on existing streams
 - Added a logic to re-use previous transcoding sessions if they are recent and haven't been used (crash-loop protection)
 - Started writing the `isActive` field on child streams ("transcoding sessions") from `stream-info-service`
 - Used that `isActive` field to trigger recording webhooks and generation of VOD assets from them
 - Created new `none` state on the `recordingStatus` to differentiate in progress from failed recordings
 - Make sure to send `recording.ready` and create VOD assets when the `isActive` hadn't been updated by mapic

 - Record a livestream
 - Make sure the session shows up on the stream page
 - Make sure it has the correct length
 - Make sure it has recording link and that it works
 - Make sure a VOD asset is generated from the recording
 - Kill `MistProcLivepeer` mid-stream
 - Make sure a new session shows up and all recordings are available in the end

Implements API-7

**Screenshots (optional)**
image
This `n/a` is so much better than all of them being `In progress` as before :feelsgood:
We can also later change the frontend to only display the sessions that actually have something in them (but lets wait for catalyst recordings)

**Checklist**


packages/api/src/webhooks/cannon.ts
```diff
@@ -502,18 +489,44 @@ export default class WebhookCannon {
...
+      await sleep(5000 + (lastSeen - activeThreshold));
+      return this.handleRecordingReadyChecks(sessionId, mp4Url, true);
+    }
+
+    if (typeof isActive === "boolean" && !isActive) {
+      throw new UnprocessableEntityError("Session recording already handled");
+    }
+    await this.db.stream.update(sessionId, { isActive: false });
+
+    await this.recordingToVodAsset(session, mp4Url);
```
iameli: Is there an argument to be made for doing these in the opposite order in case `recordingToVodAsset` fails for whatever reason?

iameli (approved):
LGTM to the extent I still remember how this recording workflow works 🤔

packages/api/src/store/stream-table.ts
```diff
@@ -41,13 +41,36 @@ export interface StreamStats {
   transcodedSegmentsDuration?: number;
 }
 
-export interface DBStreamFields extends StreamStats {
-  previousStats?: StreamStats;
```
gioelecerati, line 45: Something got broken into the tests because of this field, although I cannot spot the reason, it just got moved.

packages/api/src/controllers/session.ts
```diff
@@ -134,97 +134,6 @@ app.get("/", authorizer({}), async (req, res, next) => {
   res.json(sessions);
 });
 
-app.get("/migrate2", authorizer({ anyAdmin: true }), async (req, res, next) => {
```
gioelecerati, line 137: Why these two endpoints got removed? What they were used for and by what?

packages/api/src/controllers/stream.ts
```diff
@@ -956,60 +921,99 @@ app.put(
...
+  // backward compat with child streams that didn't have isActive field
+  if (sessions.length === 0) {
+    const last = await db.stream.getLastSession(id);
+    if (last && last.isActive === undefined) {
+      sessions = [last];
+    }
+  }
+
+  for (const session of sessions) {
+    try {
```
gioelecerati: nit: `session.id` could be put in a const here to be used in both the try block and the error message
```suggestion
    const sessionId = session.id;
    try {
```

packages/api/src/controllers/stream.ts
```diff
@@ -956,60 +921,99 @@ app.put(
...
+
+  // backward compat with child streams that didn't have isActive field
+  if (sessions.length === 0) {
+    const last = await db.stream.getLastSession(id);
+    if (last && last.isActive === undefined) {
+      sessions = [last];
+    }
+  }
+
+  for (const session of sessions) {
```
gioelecerati, line 1006: nit: could be breakdown in two separate smaller functions for clarity
```suggestion
  for (const session of sessions) {
    try {
      await processSession(session, queue, ingest);
    } catch (err) {
      logger.error(
        `Error processing session session_id=${session.id} err=`,
        err
      );
    }
  }
}

async function processSession(
  session: DBStream,
  queue: Queue,
  ingest: string
) {
```

gioelecerati (approved):
LGTM with comments

packages/api/src/controllers/session.ts
victorges: They were for migrating super old database schemas into old database schemas. They were probably called to exhaustion the first time they were created and were forgotten to be deleted. We can always look back for this for inspiration in case we ever want to create more migration flows like this (e.g. to migrate storage schema in assets, like we did recently last year).

packages/api/src/controllers/stream.ts
victorges: Not sure I like it that much as in the end it will just mean 1 less character to write for the same thing 😆, and I like the clarity of `session.id` which explicitly tells where it comes from

packages/api/src/store/stream-table.ts
victorges: Fixed!

packages/api/src/controllers/stream.ts
victorges: As we are only talking about 7 lines of logic, I think it's kinda overkill to extract this helper, ChatGPT does some of those every now and then... I'm gonna do it anyway in case you also think it's more readable, not only copied from GPT 😄

victorges: Done, but gave it a better name than a generic "processSession" that would only obfuscate what is done in that piece of the code. Perhaps the whole logic from `triggerSessionRecordingHooks` is more easily understandable now only by skimming through the code.

packages/api/src/webhooks/cannon.ts
victorges: I'm having to do the DB update first in order to avoid generating duplicate assets for the same recording. An alternative would be creating a unique index on the `source.sessionId` field but that would seem like overkill TBH. We need to sort out our transaction game here.

PR #1663, opened 2023-03-04
api: Implement IPFS storage on upload

victorges:
Basically copied the patch logic to upload and that's it!

 - Move schema to `new-asset-payload` and reference it from patch
 - Extract helper to convert the `storageInput` request to the asset state 
 - Set `storage` field on the created asset
 - profit! `task-runner` will already copy the asset to IPFS [here](https://github.com/livepeer/task-runner/blob/main/task/upload.go#L437)

Actually haven't, only a PoC and it's pretty late. 

Should be pretty easy tho just call any `/upload` endpoint with `storage.ipfs = true` for example.

Implements API-17

**Checklist**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -654,21 +654,7 @@ components:
         playbackPolicy:
           $ref: "#/components/schemas/playback-policy"
         storage:
-          additionalProperties: false
-          properties:
-            ipfs:
-              oneOf:
-                - type: object
-                  additionalProperties: false
-                  properties:
-                    spec:
-                      oneOf:
-                        - type: "null"
-                        - $ref: "#/components/schemas/asset/properties/storage/properties/ipfs/properties/spec"
-                - type: [boolean, "null"]
-                  description:
-                    Set to true to make default export to IPFS. False or null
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1423,6 +1409,22 @@ components:
...
+          additionalProperties: false
+          properties:
+            ipfs:
+              oneOf:
+                - type: object
+                  additionalProperties: false
+                  properties:
+                    spec:
+                      oneOf:
+                        - type: "null"
```
gioelecerati, line 1422: Just out of curiosity, I saw it was already like that, but Is this how the null type is defined here? Is it required to be a string in quotation marks?

gioelecerati (approved):
LGTM

packages/api/src/schema/schema.yaml
victorges: Hmm, actually for YAML it's just a string, but I think `null` is a special keyword since it also exists in JSON. If we used `null` here, this object would have a `{"type": null}` field (no type on JSON) instead of a `{"type": "null"}` (`null` type on JSON schema). [ref](https://stackoverflow.com/a/21121267/1560458)

PR #1664, opened 2023-03-08
Add targetSegmentSizeSecs to transcode API

thomshutt:
- Add target_segment_size_secs to transcode API
- snake_case

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1633,9 +1633,10 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/ffmpeg-profile"
-        target_segment_size_secs:
+        targetSegmentSizeSecs:
           type: number
-          description: How many seconds the duration of each output segment should be
+          description:
+            How many seconds the duration of each output segment should be
         catalystPipelineStrategy:
           $ref: "#/components/schemas/task/properties/params/properties/upload/properties/catalystPipelineStrategy"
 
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1633,6 +1633,9 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/ffmpeg-profile"
+        target_segment_size_secs:
```
victorges: We should use `camelCase` in the API for consistency. I'm kinda convinced on underscore for the future, but we have only camel field names today so we should keep that. 

(it is a mess for "enums" here tho, we have all of them – camel, kebab, snake... lol)

packages/api/src/schema/schema.yaml
```diff
@@ -1775,6 +1778,9 @@ components:
                   type: array
                   items:
                     $ref: "#/components/schemas/ffmpeg-profile"
+                target_segment_size_secs:
```
victorges: Same for camel 🐫

victorges (approved):
LGTM but 🐍➡️🐫

packages/api/src/schema/schema.yaml
thomshutt: Sounds good, I wasn't sure which to go with from looking at the existing fields - will change now

victorges:
Oh also, the build/tests are failing for some reason. Please fix those before merging

PR #1665, opened 2023-03-09
access-control: added new playbackPolicy & webhook trigger

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -197,7 +197,7 @@ async function fireGateWebhook(
     // TODO: remember to handle redirects
     statusCode = resp.status;
 
-    if (resp.status <= 200 || resp.status > 400) {
+    if (resp.status <= 200 || resp.status >= 300) {
       console.error(
         `access-control: gate: webhook=${webhook.id} got response status != 2XX statusCode=${resp.status}`
       );
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -141,29 +141,28 @@ app.post(
           content.playbackPolicy,
           req.body.accessKey
         );
-        switch (statusCode) {
-          case 200:
-            res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
-            res.status(204);
+        if (statusCode >= 200 && statusCode < 300) {
+          res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
+          res.status(204);
+          return res.end();
+        } else if (statusCode === 0) {
+          console.log(`
+            access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} failed, disallowing playback
+          `);
+          throw new BadGatewayError(
+            "Content is gated and corresponding webhook failed"
+          );
+        } else {
+          console.log(`
+            access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} returned status code ${statusCode}, disallowing playback
+          `);
+          if (statusCode >= 400 && statusCode < 500) {
+            res.status(statusCode);
             return res.end();
-          case 0:
-            console.log(`
-              access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} failed, disallowing playback
-            `);
-            throw new BadGatewayError(
-              "Content is gated and corresponding webhook failed"
-            );
-          default:
-            console.log(`
-              access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} returned status code ${statusCode}, disallowing playback
-            `);
-            if (statusCode >= 400 && statusCode < 500) {
-              res.status(statusCode);
-              return res.end();
-            }
-            throw new BadGatewayError(
-              "Content is gated and corresponding webhook failed"
-            );
+          }
+          throw new BadGatewayError(
+            "Content is gated and corresponding webhook failed"
+          );
         }
       default:
         console.log(`
@@ -192,6 +191,7 @@ async function fireGateWebhook(
     body: JSON.stringify({
       context: plabackPolicy.webhookContext,
       accessKey: accessKey,
+      timestamp: timestamp,
     }),
   };
 
@@ -202,13 +202,11 @@ async function fireGateWebhook(
   const startTime = process.hrtime();
   let resp: Response;
   let errorMessage: string;
-  let responseBody: string;
   let statusCode: number;
 
   try {
     console.log(`webhook ${webhook.id} firing`);
     resp = await fetchWithTimeout(webhook.url, params);
-    responseBody = await resp.text();
     statusCode = resp.status;
 
     if (resp.status >= 200 && resp.status < 300) {
@@ -231,7 +229,7 @@ async function fireGateWebhook(
       timestamp,
       statusCode,
       errorMessage,
-      responseBody
+      undefined
     );
     return statusCode;
   }
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -17,7 +17,6 @@ import { PlaybackPolicy } from "../schema/types";
 import { signatureHeaders, storeTriggerStatus } from "../webhooks/cannon";
 import { Response } from "node-fetch";
 import { fetchWithTimeout } from "../util";
-import { error } from "winston";
 
 const WEBHOOK_TIMEOUT = 5 * 1000;
 const app = Router();
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -10,7 +10,13 @@ import {
   BadRequestError,
 } from "../store/errors";
 import tracking from "../middleware/tracking";
+import { DBWebhook } from "../store/webhook-table";
+import { PlaybackPolicy } from "../schema/types";
+import { addSignatureHeader } from "../webhooks/cannon";
+import { Response } from "node-fetch";
+import { fetchWithTimeout } from "../util";
 
+const WEBHOOK_TIMEOUT = 5 * 1000;
 const app = Router();
 
 app.use("/signing-key", signingKeyApp);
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+          `);
+          throw new ForbiddenError("The public key is disabled or deleted");
+        }
+
+        tracking.recordSigningKeyValidation(signingKey.id);
+        res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
+        res.status(204);
+        return res.end();
+      case "webhook":
+        if (!req.body.accessKey) {
```
victorges: Should also check the `req.body.type` field.

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+          throw new ForbiddenError(
+            "Content is gated and requires an access key"
+          );
+        }
+        const webhook = await db.webhook.get(content.playbackPolicy.webhookId);
+        if (!webhook) {
+          console.log(`
+            access-control: gate: content with playbackId ${playbackId} is gated but corresponding webhook not found for webhookId ${content.playbackPolicy.webhookId}, disallowing playback
+          `);
+          throw new ForbiddenError(
```
victorges: This should probably be an `InternalServerError`

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+          webhook,
+          content.playbackPolicy,
+          req.body.accessKey
         );
-      }
-
-      if (signingKeyOutput.length > 1) {
-        let collisionKeys = JSON.stringify(signingKeyOutput);
+        switch (statusCode) {
+          case 200:
```
victorges: We should allow any 2xx status code, not only 200. In that case it's easier to do a "range if" instead of a switch here (unless you wanna list out 100 cases 😆)

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+            res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
+            res.status(204);
+            return res.end();
+          case 0:
```
victorges: How can a 0 status code happen? 0o

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+              access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} failed, disallowing playback
+            `);
+            throw new BadGatewayError(
+              "Content is gated and corresponding webhook failed"
+            );
+          default:
+            console.log(`
+              access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} returned status code ${statusCode}, disallowing playback
+            `);
+            if (statusCode >= 400 && statusCode < 500) {
```
victorges: Yeah let's drop the `switch` and just do some range ifs like this. It's currently a switch of 1 case or default haha

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+          case 0:
+            console.log(`
+              access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} failed, disallowing playback
+            `);
+            throw new BadGatewayError(
+              "Content is gated and corresponding webhook failed"
+            );
+          default:
+            console.log(`
+              access-control: gate: content with playbackId ${playbackId} is gated but webhook ${webhook.id} returned status code ${statusCode}, disallowing playback
```
victorges: Can you change all of these logs to be structured? Loki is not able to parse "free form text" like this and it is troublesome to make any use out of them on grafana. We use something similar to this format: https://brandur.org/logfmt

e.g. This could be `access-control: gate: content gated but webhook returned bad status code, disallowing playback playbackId=${playbackId} webhookId=${webhook.id} statusCode=${statusCode}`
This is only 1 example, all logs (at least the new ones) should be structured

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+            `);
+            if (statusCode >= 400 && statusCode < 500) {
+              res.status(statusCode);
```
victorges: We should not proxy the status code (sorry I may have have said that at some point). This should be a `400` or `403`.

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+              return res.end();
+            }
+            throw new BadGatewayError(
+              "Content is gated and corresponding webhook failed"
+            );
+        }
+      default:
```
victorges, line 154: I was confused to what this referred to with all the switches haha. One more reason to drop the inner switch and keep only the outer one on playback policy type.

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+  let timestamp = Date.now();
+  let params = {
+    method: "POST",
+    headers: {
+      "content-type": "application/json",
+      "user-agent": "livepeer.studio",
+    },
+    timeout: WEBHOOK_TIMEOUT,
+    body: JSON.stringify({
+      context: plabackPolicy.webhookContext,
```
victorges, line 176: WDYT of adding the timestamp in the payload as well for consistency?

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+  const startTime = process.hrtime();
+  let resp: Response;
+  let errorMessage: string;
+  let responseBody: string;
+  let statusCode: number;
+
+  try {
+    console.log(`webhook ${webhook.id} firing`);
+    resp = await fetchWithTimeout(webhook.url, params);
+    responseBody = await resp.text();
```
victorges: We should not read or store this, one can probably exhaust our DB by returning large bodies here. The response payload should be completely ignored, we should only check the status code.

Also check if we need to "close" the response in some way if we ignore the body. Not sure how that disposal works in node 🤔

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+    statusCode = resp.status;
+
+    if (resp.status >= 200 && resp.status < 300) {
+      // 2xx requests are cool. all is good
+      console.log(`webhook ${webhook.id} fired successfully`);
```
victorges: Do we really need these logs? They seem very debugg-y and this API might get a **lot** of traffic. Perhaps only a metric should be enough?

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+
+    console.error(
+      `webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`
+    );
+  } catch (e) {
+    console.log(`
+      webhook ${webhook.id} failed to fire: ${e.message}
+    `);
+    errorMessage = e.message;
+    statusCode = 0;
```
victorges, line 207: Can we just throw an exception from here? I think returning a "`0` status code" is very confusing. Simply `throw new BadGateway(...)`

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,184 @@ app.post(
...
+  } finally {
+    await this.storeTriggerStatus(
+      webhook,
+      timestamp,
+      statusCode,
+      errorMessage,
+      responseBody
+    );
+    return statusCode;
```
victorges: This return should be either inside the `try` or after the `finally`, you should almost never return from a `finally` block. Not only the data flow can get confusing, it can also lead to really unexpected results.
e.g. Any exception thrown in the try block will be eaten by this. If in the future we change line 227 to throw an exception instead, it wouldn't work and likely an `undefined` status code would be returned instead lol

packages/api/src/schema/schema.yaml
```diff
@@ -762,6 +762,9 @@ components:
         pub:
           type: string
           description: Base64 of Public key used for JWT verification
+        accessKey:
```
victorges, line 741: Missing changes:
 - Creating an `accessKey` enum value for the `type` field. Notice this is the "type" **for catalyst**, which is why it should NOT be `webhook`. Catalyst has no idea what the access key will be used for.
 - Creating a `playbackId` field (instead of `stream` which doesn't make sense anymore)

Ideally, we should have this object as `oneOf` to have clear & properly validated schemas here. Otherwise you should add at least 1 more check in the code which is checking the `req.body.type` as well. I made a separate comment about that in the code, but I prefer the `oneOf` approach TBH, we shouldn't be writing any schema validation code ourselves, our logic is already complex enough kkkk.

packages/api/src/schema/schema.yaml
```diff
@@ -793,6 +796,7 @@ components:
             - public
             - jwt
             - lit_signing_condition # only for lit-signing-condition experiment
```
victorges, line 774: Omg a snake kebab on the same line

packages/api/src/schema/schema.yaml
```diff
@@ -814,6 +818,13 @@ components:
               type: string
             extraData:
               type: string
+        webhookId:
```
victorges, line 797: It would be nice if this was a `oneOf` as well. Even better than the gate payload since it is actual user input. We shouldn't allow users to specify a `webhookId` if their type is `jwt`, nor the lit stuff when the type is webhook, etc etc.

As an example of `oneOf` check the `asset.source`, it's pretty much the same use of that: https://github.com/livepeer/livepeer-com/blob/62b2672ee600f55fe5df5c2bf5a2532ef10963ed/packages/api/src/schema/schema.yaml#L1095

packages/api/src/webhooks/cannon.ts
```diff
@@ -546,3 +546,13 @@ export default class WebhookCannon {
     );
   }
 }
+
+export async function addSignatureHeader(
+  params: any,
```
victorges: This should not be `any`, there is a [very specific type for this](https://github.com/livepeer/livepeer-com/blob/88236319556da6cf0c1c4f0dc9fa3b89d9c87a0c/packages/api/src/util.ts#L2). We should never use `any` unless there is no choice (e.g. third-party lib, too annoying or actually irrelevant type, idk)

packages/api/src/webhooks/cannon.ts
```diff
@@ -546,3 +546,13 @@ export default class WebhookCannon {
...
+  webhook: DBWebhook,
+  timestamp: number
+) {
+  let signature = sign(params.body, webhook.sharedSecret);
+  params.headers[SIGNATURE_HEADER] = `t=${timestamp},v1=${signature}`;
+  return params;
+}
```
victorges: **niiit**: I think this function receives and returns more than it needs to, which makes it harder to know what it does from the call-site. The interface issues I see:
 - it has no reason to be async
 - it could even change the request timeout if it wanted to lol, it gets the full request params
 - one also can't tell what information it needs. looks like it needs the whole webhook spec and request params when it only needs 1 field from each 
 - it looks like it returns a new params object, as if it were thread-safe for example, but it actually mutates the object in-place and then always returns the same thing. callers should never need to re-assign their local variables but every caller does because of that. Also misleading IMO.

WDYT of?
```suggestion
export function signatureHeaders(
  payload: any,
  sharedSecret: string,
  timestamp: number
) {
  if (!sharedSecret) return {};

  let signature = sign(payload, sharedSecret);
  return {[SIGNATURE_HEADER]: `t=${timestamp},v1=${signature}`};
}
```
added a suggestion on the call-site to show how it'd look like.

packages/api/src/webhooks/cannon.ts
```diff
@@ -392,8 +393,7 @@ export default class WebhookCannon {
 
       // sign payload if there is a webhook secret
       if (webhook.sharedSecret) {
-        let signature = sign(params.body, webhook.sharedSecret);
-        params.headers[SIGNATURE_HEADER] = `t=${timestamp},v1=${signature}`;
+        params = await addSignatureHeader(params, webhook, timestamp);
       }
```
victorges: Related to the suggestion on the helper below, this would look like:
```suggestion
        const sigHeaders = signatureHeaders(params.body, webhook.sharedSecret, timestamp);
        params.headers = { ...params.headers, ...sigHeaders};
```

orr this could even be inlined on the original `headers` object definition! Like 
```ts
    ...signatureHeaders(payload, webhook.sharedSecret, timestamp),
}
```
in the end of the object (would only need to define the `payload` separately first)

packages/api/src/controllers/access-control.ts
gioelecerati: Actually about status codes in the response I was going to use the status codes defined here 
https://www.notion.so/livepeer/Signing-Keys-for-Access-Control-8879f96a5aaf409eb4042acc281f5caa?pvs=4#6b07279a3e234715ad991b1e58dffb6e

So in this case it would be `403 Forbidden` because it's the same as when we don't find the SigningKey, but I am open to change both to InternalServerError or NotFound

packages/api/src/controllers/access-control.ts
gioelecerati: I am specifically setting the status code at 0 when I am not able to receive it (e.g. timeouts), as it's usually done by xmlhttprequest, but I can change it and make it undefined instead in that case!

gioelecerati: https://fetch.spec.whatwg.org/#concept-network-error

packages/api/src/controllers/access-control.ts
gioelecerati: RIght! yeah, I initially thought to support only 200 and 204, but changed my mind and I was going to drop the switch either way, this was just a draft :) changing it

packages/api/src/controllers/access-control.ts
gioelecerati: All these logs in this endpoint are indeed coming from a pr where we added a big amount of them when investigating random gating happening on public streams. we can decrease them!

packages/api/src/controllers/access-control.ts
victorges: Not a strong opinion, feel free to keep 403! I didn't remember that original doc/decision.

Just to explain what I meant: This is an invalid state in our own database altogether which our code let happen! The user should never be able to create an object (asset) that references something (webhook) that does not exist. So it's not an user error but our own and hence the 500. The webhook might've been deleted, but in this case it'd have a `deleted: true` anyway and not not exist here completely.

Speaking of that, we might want to check the `deleted` field of the webhook here 👀 (on that case a 403 makes sense tho, so lets just stick to that for all "cant find webhook" cases)

packages/api/src/controllers/access-control.ts
victorges: Was not aware this was a draft! The PR is not created as a draft.

packages/api/src/controllers/access-control.ts
victorges: Hm, not a fan of returning a dummy value on an exceptional case, that's what exceptions are for and both provide more info (error msg) + avoid the need of understanding how such dummy values are created (which requires reading the sub-function code or docs, which we don't usually write). 

But that's too low-level and doesn't matter much anyway. If you're more comfortable following that pattern, just keep it!

packages/api/src/controllers/access-control.ts
victorges: Ah true! Yeah, let's worst case keep them commented out so we can bring them back easily, but we don't want that much verbosity in prod.

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,182 @@ app.post(
...
+      `webhook ${webhook.id} didn't get 200 back! response status: ${resp.status}`
+    );
+  } catch (e) {
+    console.log(`
+      webhook ${webhook.id} failed to fire: ${e.message}
+    `);
+    errorMessage = e.message;
+    statusCode = 0;
+  } finally {
+    await this.storeTriggerStatus(
```
victorges: Actually, I don't think `this` is defined here 👀 

Surprised how TypeScript let this pass. You probably need to create a helper to re-use the logic or reference the cannon here.

packages/api/src/controllers/access-control.ts
```diff
@@ -50,73 +57,182 @@ app.post(
...
+        if (!req.body.accessKey) {
+          console.log(`
+            access-control: gate: no accessKey provided for playbackId ${playbackId}, disallowing playback
+          `);
+          throw new ForbiddenError(
+            "Content is gated and requires an access key"
+          );
+        }
+        const webhook = await db.webhook.get(content.playbackPolicy.webhookId);
+        if (!webhook) {
```
victorges, line 123: Also! We must check if the webhook can listen to the event `playback.accessControl` (or whatever). I know you haven't created it yet, but commenting here as a reminder.

packages/api/src/controllers/access-control.ts
gioelecerati: Sounds good! let's make it InternalServerError

packages/api/src/controllers/access-control.ts
gioelecerati: Oh right sorry! Btw, now checking for all 2xx status codes without with an if else :)

packages/api/src/controllers/access-control.ts
gioelecerati: I am keeping the 0 status code in the ts logic and returning 502 Bad Gate(way) to Catalyst, FYI

packages/api/src/controllers/access-control.ts
gioelecerati: Right! Makes sense, will make these changes and comment out or remove most of the logs to be less verbose

packages/api/src/controllers/access-control.ts
gioelecerati: Yeah I noticed that too, I took it from the cannon and ts doesn't complain. I will recreate the helper

packages/api/src/controllers/access-control.ts
victorges: Sounds good!

packages/api/src/controllers/access-control.ts
```diff
@@ -40,83 +44,181 @@ app.post(
...
-
-      if (signingKeyOutput.length == 0) {
-        console.log(`
-          access-control: gate: content with playbackId ${playbackId} is gated but corresponding public key not found for key ${req.body.pub}, disallowing playback
-        `);
-        throw new ForbiddenError(
-          "Content is gated and corresponding public key not found"
+    switch (playbackPolicyType) {
+      case "public":
+        res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
```
victorges: How about we always set this header for the response? Meaning do it in the beginning of the controller. If we reply with a 5xx status code (like network error idk), clients will still avoid caching it anyway as they cache only 2xx and 4xx. ([catalyst only caches 2xx](https://github.com/livepeer/catalyst/blob/main/cmd/catalyst-node/handlers/access-control/access-control.go#L220))

Will avoid having paths where we forget to set it in the code tho wanted it and getting /gate api bursts from invalid playback URLs.

packages/api/src/controllers/access-control.ts
victorges: This is still needed 👀 

And thinking about it, maybe we should allow fetching an account webhook of the `playback.accessControl` type, WDYT? So it defaults to idk the oldest active webhook for that event type, if you don't specify it in the `playbackPolicy` 🤔

packages/api/src/controllers/access-control.ts
```diff
@@ -40,83 +44,181 @@ app.post(
...
-
-      if (signingKeyOutput.length > 1) {
-        let collisionKeys = JSON.stringify(signingKeyOutput);
-        console.log(`
-          access-control: gate: content contentId ${content.id} with playbackId=${playbackId} is gated but multiple (${signingKeyOutput.length}) public keys found for key ${req.body.pub}, disallowing playback, colliding keys=${collisionKeys}
-        `);
+        if (statusCode >= 200 && statusCode < 300) {
+          res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
+          res.status(204);
+          return res.end();
```
victorges: This logic could also be shared with the `public` playback policy by going in the end of the function instead. So it's like each `case` here should validate the access conditions and throw an error if it fails, otherwise let it continue the handler and allow playback. WDYT?

packages/api/src/controllers/access-control.ts
```diff
@@ -40,83 +44,181 @@ app.post(
...
+  let resp: Response;
+  let errorMessage: string;
+  let statusCode: number;
+
+  try {
+    resp = await fetchWithTimeout(webhook.url, params);
+    // TODO: remember to handle redirects
+    statusCode = resp.status;
+
+    if (resp.status <= 200 || resp.status > 400) {
```
victorges: From the log below I think you meant
```suggestion
    if (resp.status < 200 || resp.status >= 300) {
```

packages/api/src/controllers/access-control.ts
```diff
@@ -40,83 +44,181 @@ app.post(
...
+  } finally {
+    await storeTriggerStatus(
+      webhook,
+      timestamp,
+      statusCode,
+      errorMessage,
+      undefined
+    );
+    if (resp) {
+      // Dispose of the response body
```
victorges, line 217: noice

packages/api/src/controllers/asset.ts
```diff
@@ -61,7 +61,10 @@ function defaultObjectStoreId(
   if (isOldPipeline) {
     return config.vodObjectStoreId;
   }
-  if (body.playbackPolicy?.type === "lit_signing_condition") {
+  if (
+    body.playbackPolicy?.type === "lit_signing_condition" ||
+    body.playbackPolicy?.type === "webhook"
+  ) {
```
victorges: Hmmmmmmm perhaps this should be 
```suggestion
const policyType = body.playbackPolicy?.type;
if (policyType && policyType !== "public") {
```

WDYT??

packages/api/src/controllers/asset.ts
```diff
@@ -1027,6 +1030,15 @@ app.patch(
       }
     }
 
+    if (
+      (playbackPolicy && asset.playbackPolicy?.type === "webhook") ||
+      playbackPolicy?.type === "webhook"
```
victorges, line 1066: Perhaps we actually need a helper like `isPrivatePlaybackPolicy` which does the `!== public` check, and use it here and in the other place where we choose the OS ID.

Here, the logic we should check is
```suggestion
      isPrivatePlaybackPolicy(playbackPolicy) !== isPrivatePlaybackPolicy(asset.playbackPolicy)
```
WDYT?

packages/api/src/webhooks/cannon.ts
```diff
@@ -546,3 +538,41 @@ export default class WebhookCannon {
...
+    }
+    await db.webhook.updateStatus(webhook.id, status);
+  } catch (e) {
+    console.log(
+      `Unable to store status of webhook ${webhook.id} url: ${webhook.url}`
+    );
+  }
+}
+
+export function signatureHeaders(
```
victorges, line 590: Neat!

victorges (approved):
LGTM! Mostly nits

packages/api/src/controllers/access-control.ts
victorges: Correction: Catalyst client caches anything, not only 2xx requests. Misunderstood that first return value. Not sure about this change anymore, but I'd still be more comfortable caching an error for a while than never caching bad input responses... 🤔 
Or we can update catalyst to only cache 2xx and 4xx and have the simpler logic here.

packages/api/src/controllers/asset.ts
gioelecerati: Makes sense, I arrived late on this and now I don't want to touch too much, can open a separate PR

victorges: Yeah do it. This is not only a nit, the functionality is incomplete (its ignoring jwt policies for example)

PR #1667, opened 2023-03-10
api: Add scheduledAt field to tasks to allow proper time-to-ready metric

victorges:
This adds a `scheduledAt` field to tasks so we can tell how long an asset
actually took to upload then how long it actually took to be processed.

upload time = task->scheduledAt - task->createdAt 
time to ready = task->status->updatedAt - task->scheduledAt

On the way, also added an `uploading` phase to assets cause we need that for too long already.

 - Add `scheduledAt` field to tasks and udpate them from scheduler
 - Add `uploading` phase to assets and update it from controller
 - Rename scheduler task scheduling functions

in `yarn test`s we trust

Implements API-31

**Checklist**

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -301,18 +301,21 @@ export class TaskScheduler {
     return task;
   }
 
-  async enqueueTask(task: WithID<Task>) {
-    const status: Task["status"] = {
-      phase: "waiting",
-      updatedAt: Date.now(),
-      retries: task.status.retries,
-    };
-    await this.updateTask(task, { status });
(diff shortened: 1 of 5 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -250,7 +250,7 @@ export class TaskScheduler {
     }
   }
 
-  async scheduleTask(
+  async spawnAndScheduleTask(
```
gioelecerati: nit: What do you think of renaming this function `createAndScheduleTask` which better describes the purpose of it?

gioelecerati (approved):
LGTM!

PR #1668, opened 2023-03-14
webhook: created playback.accessControl event & return playbackPolicy in playback request

gioelecerati:

This pull request adds a new webhook event called playback.accessControl, which is triggered when the playback policy type is not public.
Also, return the playbackPolicy with the playback info
To be updated accordingly with #1665

- Update the event list in packages/api/src/schema/schema.yaml to include the new event playback.accessControl
- Add the new event to the eventOptions array in packages/www/components/Dashboard/WebhookDialogs/CreateEditDialog.tsx
- Return playbackPolicy from playback endpoint

yarn test

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -1,6 +1,6 @@
 import signingKeyApp from "./signing-key";
 import { validatePost } from "../middleware";
-import { Router } from "express";
+import { Request, Router } from "express";
 import _ from "lodash";
 import { db } from "../store";
 import sql from "sql-template-strings";
@@ -10,6 +10,10 @@ import {
   BadRequestError,
 } from "../store/errors";
 import tracking from "../middleware/tracking";
+import { v4 as uuid } from "uuid";
+import { DBStream } from "../store/stream-table";
+import { Asset } from "../schema/types";
+import { WithID } from "../store/types";
 
 const app = Router();
 
@@ -49,6 +53,11 @@ app.post(
     }
 
     const playbackPolicyType = content.playbackPolicy?.type ?? "public";
+
+    if (playbackPolicyType !== "public") {
+      // TODO: move it into specific policy conditions
+      fireEventWebhook(req, content);
+    }
 
     if (playbackPolicyType === "public") {
       res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
@@ -118,5 +127,20 @@ app.post(
     }
   }
 );
+
+async function fireEventWebhook(
+  req: Request,
+  content: DBStream | WithID<Asset>
+) {
+  const queue = req.queue;
+  await queue.publishWebhook(`events.playback.accessControl`, {
+    type: "webhook_event",
+    id: uuid(),
+    timestamp: Date.now(),
+    streamId: content.id,
+    event: "playback.accessControl",
+    userId: content.userId,
+  });
+}
 
 export default app;
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -1,6 +1,6 @@
 import signingKeyApp from "./signing-key";
 import { validatePost } from "../middleware";
-import { Request, Router } from "express";
+import { Router } from "express";
 import _ from "lodash";
 import { db } from "../store";
 import sql from "sql-template-strings";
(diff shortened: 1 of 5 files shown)


gioelecerati:
`playback.user.new` event needs to get removed

packages/api/src/schema/schema.yaml
```diff
@@ -213,6 +213,7 @@ components:
               - multistream.error
               - multistream.disconnected
               - playback.user.new
+              - playback.accessControl
```
0xcadams, line 216: I noticed that the other event types are named with all lowercase - is kebab case preferred here? Purely stylistic choice, so no strong opinion either way.

packages/api/src/controllers/access-control.ts
```diff
@@ -119,4 +128,19 @@ app.post(
   }
 );
 
+async function fireEventWebhook(
+  req: Request,
+  content: DBStream | WithID<Asset>
+) {
+  const queue = req.queue;
+  await queue.publishWebhook(`events.playback.accessControl`, {
```
0xcadams: I noticed that in the scoping doc we mention adding a context object - is this something we plan on adding in a future iteration?
Screenshot 2023-03-14 at 9 27 05 AM

packages/api/src/schema/schema.yaml
gioelecerati: Yeah, probably worth it too make it lowercase here as well, so `playback.accesscontrol` , I agree!

packages/api/src/controllers/access-control.ts
gioelecerati: This is a different thing, the screen from the notion doc is referring to the webhook that Studio is going to call during the `gate` check (https://github.com/livepeer/studio/pull/1665), this pr is referring instead to user configured webhooks events that notify access control triggers

gioelecerati: I am removing this part of the code in any case. No need to add a notification apart of normal gate call. The only webhook request is happening in `fireGateWebhook` in the other pr

cc @0xcadams

packages/api/src/schema/schema.yaml
gioelecerati: After discussing with Victor, we are keeping the casing here instead of the lower case, if it's fine for you @0xcadams 
would be strange to have a name-space like `access.control` and it's the only one not lowercase because it's the first one with 2 words (user.new is going to be removed soon either way)

0xcadams: Sounds great!

packages/api/src/controllers/playback.ts
```diff
@@ -89,16 +89,13 @@ const getAssetPlaybackUrl = async (
...
-    asset.userId
-  );
+
   return !playbackUrl
     ? null
     : {
         staticFilesPlaybackInfo,
         playbackUrl,
-        playbackPolicy: inExperiment ? asset.playbackPolicy : null,
+        playbackPolicy: asset.playbackPolicy || null,
```
victorges, line 98: This `||` is unnecessary. IMO it's better to not include the field if it doesn't exist instead of specifying it as `null` (my bad on including it as `null` in the first place)
```suggestion
        playbackPolicy: asset.playbackPolicy,
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -90,6 +90,10 @@ export default class WebhookCannon {
   }
 
   async processWebhookEvent(event: messages.WebhookEvent): Promise<boolean> {
+    if (event.event === "playback.accessControl") {
+      // Cannot fire this event as a webhook, this is specific to access control and fired there
```
victorges, line 111: Adding some more comment+runtime log info:
```suggestion
      // Cannot fire this webhook asynchronously. This is specific to access control and fired on
      // the /access-control/gate API controller synchronously
      console.error("Received unexpected access control event in async webhook cannon. id=${event.id} userId=${event.userId}")
```

victorges (approved):
LGTM

0xcadams:
Can we add the webhook ID to the UI? Currently I see this, which I wouldn't be able to pull an ID from without knowing it's in the URL

Screenshot 2023-03-28 at 10 17 00 AM

PR #1673, opened 2023-03-30
Add support for MP4 in Transcode API

leszko:


packages/api/src/schema/schema.yaml
```diff
@@ -1784,6 +1794,15 @@ components:
                           type: string
                           description: Path for the HLS output
                           example: /samplevideo/hls
+                    mp4:
```
victorges, line 1797: What's the difference between the 2 `outputs` fields you added this?

victorges (approved):
LGTM! only a question comment

packages/api/src/schema/schema.yaml
leszko: One is for HLS output (so if you specify it, you'll get the HLS output video), one is for MP4 (so if you specify it, you'll get your video transcoded as mp4 files).

But, their structure is the same.

thomshutt (approved):

PR #1675, opened 2023-03-31
api/asset: Fix playback policy validation

victorges:
followups from https://github.com/livepeer/studio/pull/1674/files


gioelecerati (approved):
LGTM

PR #1684, opened 2023-04-05
updated the api schema for readme.io integration

suhailkakar:
This PR cleans up the schema file and also making it more detailed for better devex.

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1,45 +1,62 @@
 openapi: 3.0.1
 info:
-  title: Livepeer Studio API Reference
-  version: 0.0.1-alpha.0
+  title: Livepeer API Reference
+  description: >-
+    This API reference provides documentation for the Livepeer Studio API, which
+    is a RESTful API that allows users to manage and interact with Livepeer
+    Studio, the gateway to Livepeer network.
+  version: 1.0.0
+  termsOfService: "https://livepeer.studio/terms-of-service"
+
+servers:
+  - url: https://livepeer.studio/api
 
 tags:
-  - name: stream
-    description: Stream Endpoints
-  - name: object-store
-    description: Object Stores Endpoints
-  - name: multistream-target
-    description: Multistream Target Endpoints
-  - name: user
-    description: User Endpoints
-  - name: api-token
-    description: Api Token Endpoints
-  - name: login
-    description: User login
-  - name: verification
-    description: User Email Verification
-  - name: password-reset
-    description: User Password Reset
-  - name: make-admin
-    description: Changes user admin status
-  - name: webhook
-    description: Webhooks endpoints
-  - name: asset
-    description: Asset Endpoints
-  - name: task
-    description: Task Endpoints
-  - name: signing-key
-    description: Signing Key Endpoints
-  - name: webhook-response
-    description: webhook client responses
-  - name: verify-email
-    description: Resend User Email Verification
+  - name: Stream Endpoints
+    description: "Endpoints for creating, managing and deleting streams."
+  - name: Object Stores Endpoints
+    description: >-
+      Endpoints for managing object stores, such as Amazon S3 and Google Cloud
+      Storage.
+  - name: Multistream Target Endpoints
+    description: >-
+      Endpoints for managing multistream targets, which allow users to
+      distribute their streams to multiple destinations.
+  - name: User Endpoints
+    description: >-
+      Endpoints for managing users, such as creating new users and retrieving
+      user information.
+  - name: API Token Endpoints
+    description: >-
+      Endpoints for managing API tokens, which are used to authenticate requests
+      to the Livepeer Studio API.
+  - name: User Login Endpoints
+    description: Endpoints for user authentication and login.
+  - name: User Email Verification Endpoints
+    description: Endpoints for verifying a user's email address.
+  - name: User Password Reset Endpoints
+    description: Endpoints for resetting a user's password.
+  - name: Make Admin Endpoints
+    description: Endpoints for changing a user's admin status.
+  - name: Webhook Endpoints
+    description: Endpoints for managing webhooks and their subscriptions.
+  - name: Asset Endpoints
+    description: Endpoints for managing the video assets.
+  - name: Task Endpoints
+    description:
+      "Endpoints for managing background tasks, such as video transcoding."
+  - name: Signing Key Endpoints
+    description: Endpoints for managing signing keys.
+  - name: Webhook Response Endpoints
+    description: Endpoints for handling responses from webhook clients.
+  - name: Verify Email Endpoints
+    description: Endpoints for resending email verification requests to users.
 
 components:
   schemas:
     ffmpeg-profile:
       type: object
-      description: LMPS ffmpeg profile
+      description: LMPS ffmpeg profile for video encoding
       additionalProperties: false
       required:
         - width
@@ -50,80 +67,106 @@ components:
       properties:
         width:
           type: integer
+          description: The width of the encoded video in pixels.
           minimum: 128
+          example: 1920
         name:
           type: string
+          description: A descriptive name for the video profile.
           minLength: 1
           maxLength: 500
-          example: "720p"
+          example: 1080p
         height:
           type: integer
+          description: The height of the encoded video in pixels.
           minimum: 128
+          example: 1080
         bitrate:
           type: integer
+          description:
+            The target bitrate for the encoded video in bits per second.
           minimum: 400
+          example: 4000000
         fps:
           type: integer
-          minimum: 0
+          description: The target frames per second for the encoded video.
+          minimum: 1
+          example: 30
         fpsDen:
           type: integer
+          description: >-
+            The denominator of the target frames per second. Used in conjunction
+            with `fps` to specify the exact frame rate as a fraction.
           minimum: 1
+          example: 1
         gop:
           type: string
+          description: >-
+            The GOP (Group of Pictures) structure for the encoded video. This is
+            a string in the format of "M/N", where M is the number of frames in
+            a GOP and N is the distance in frames between two I-frames (also
+            known as keyframes).
+          example: 25/250
         profile:
           type: string
+          description: The video codec profile to use for encoding the video.
           enum:
             - H264Baseline
             - H264Main
             - H264High
             - H264ConstrainedHigh
+            - HEVCMain
+            - HEVCHigh
+            - VP8
+            - VP9
+          example: H264High
         encoder:
           type: string
+          description: The video codec to use for encoding the video.
           enum:
             - h264
             - hevc
             - vp8
             - vp9
+          example: h264
     cdn-data-row:
       type: object
       description:
-        Fields names taken from log format description
-        https://support.highwinds.com/hc/en-us/articles/360052181571-Raw-Log-Access-through-GCS
+        Fields taken from raw log access data provided by StackPath CDN.
       additionalProperties: false
       required:
         - unique_client_ips
         - total_sc_bytes
-      # oneOf: # allows validation, but generates too generic type definition
-      #   - required:
-      #       - stream_id
-      #   - required:
-      #       - playback_id
       properties:
         stream_id:
           type: string
-          description: Session UUID. Used in recordings URLs.
+          description:
+            A unique session UUID that can be used to retrieve recording URLs.
         playback_id:
           type: string
-          description: Playback ID.
+          description:
+            A playback ID that can be used to retrieve recording URLs.
         unique_client_ips:
           type: number
-          description:
-            Number of unique IP addresses of clients. Not accurate because
-            StackPath zeroes last part of the IP address.
+          description: >-
+            The number of unique client IP addresses. Note that StackPath zeroes
+            out the last part of the IP address for privacy reasons.
         total_filesize:
           type: number
-          description: Size of the asset being delivered.
+          description: The size of the asset being delivered.
         total_cs_bytes:
           type: number
           description: The total size of the request header.
         total_sc_bytes:
           type: number
-          description: Total bytes in the response to the client.
+          description: The total size of the response to the client.
         count:
           type: number
-          description: Number of parsed log lines
+          description: The number of parsed log lines.
     cdn-data-payload:
       type: array
+      description:
+        An array of log data entries obtained from StackPath CDN logs.
       items:
         type: object
         additionalProperties: false
@@ -135,10 +178,15 @@ components:
         properties:
           region:
             type: string
+            description: The region where the log data was obtained from.
           file_name:
             type: string
+            description: The name of the file containing the log data.
           date:
             type: number
+            description: >-
+              The date when the log data was obtained (in Unix timestamp
+              format).
           data:
             type: array
             minItems: 1
@@ -146,7 +194,9 @@ components:
               $ref: "#/components/schemas/cdn-data-row"
     cdn-usage-last:
       type: object
-      table: cdn_usage_last
+      description: >-
+        This schema represents a record of the last CDN usage for a specific
+        region and file name.
       required:
         - region
         - fileName
@@ -156,16 +206,16 @@ components:
           type: string
           readOnly: true
           example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+          description: The unique identifier for this record.
         fileName:
           type: number
-          index: true
+          description: The unique identifier of the file.
         region:
           type: string
-          index: true
           unique: true
+          description: The region where the file is stored.
     webhook:
       type: object
-      table: webhook
       required:
         - name
         - url
@@ -173,34 +223,29 @@ components:
       properties:
         id:
           type: string
-          readOnly: true
+          format: uuid
+          description: The unique identifier for the webhook
           example: de7818e7-610a-4057-8f6f-b785dc1e6f88
         kind:
           type: string
-          readOnly: true
+          description: 'The kind of object, always set to "webhook"'
           example: webhook
         name:
           type: string
+          description: The name of the webhook
+          example: my-webhook
         userId:
           type: string
-          readOnly: true
-          index: true
+          format: uuid
+          description: The unique identifier of the user who created the webhook
         createdAt:
-          type: number
-          readOnly: true
+          type: integer
           description:
-            Timestamp (in milliseconds) at which stream object was created
+            The timestamp (in milliseconds) at which the webhook was created
           example: 1587667174725
-        event:
-          writeOnly: true
-          deprecated: true
-          description:
-            "@deprecated Non-persisted field. To be used on creation API only."
-          $ref: "#/components/schemas/webhook/properties/events/items"
         events:
           type: array
-          index: true
-          indexType: gin
+          description: The events that will trigger the webhook
           items:
             type: string
             enum:
@@ -213,7 +258,6 @@ components:
               - multistream.error
               - multistream.disconnected
               - playback.user.new
-              - playback.accessControl
               - asset.created
               - asset.updated
               - asset.failed
@@ -226,59 +270,58 @@ components:
         url:
           type: string
           format: uri
-          pattern: ^http(s)?://
+          pattern: "^http(s)?://"
+          description: The URL to which the webhook will send its payloads
+          example: "https://example.com/webhook"
         deleted:
           type: boolean
+          description: A flag indicating whether the webhook has been deleted
           default: false
         sharedSecret:
           type: string
+          description: The shared secret used to sign the webhook payload
           writeOnly: true
-          description: shared secret used to sign the webhook payload
         streamId:
           type: string
-          description: streamId of the stream on which the webhook is applied
+          format: uuid
+          description:
+            The unique identifier of the stream on which the webhook is applied
         status:
           type: object
-          description: status of webhook
-          readOnly: true
+          description: The status of the webhook
           properties:
             lastFailure:
               type: object
-              description: failure timestamp and error message with status code
-              readOnly: true
+              description: "The details of the last failure, if any"
               properties:
                 timestamp:
-                  type: number
-                  readOnly: true
+                  type: integer
+                  description: >-
+                    The timestamp (in milliseconds) at which the last failure
+                    occurred
                   example: 1587667174725
-                  description:
-                    Timestamp (in milliseconds) at which the webhook last failed
                 error:
                   type: string
-                  readOnly: true
-                  description: Webhook failure error message
-                  example: "Error message"
+                  description:
+                    The error message associated with the last failure
+                  example: Error message
                 response:
                   type: string
-                  readOnly: true
-                  description: Webhook failure response
-                  example: "Response body"
+                  description:
+                    The response body associated with the last failure
+                  example: Response body
                 statusCode:
-                  type: number
-                  readOnly: true
-                  description: Webhook failure status code
+                  type: integer
+                  description: The status code associated with the last failure
                   example: 500
             lastTriggeredAt:
-              type: number
-              readyOnly: true
-              description:
-                Timestamp (in milliseconds) at which the webhook last was
+              type: integer
+              description: >-
+                The timestamp (in milliseconds) at which the webhook was last
                 triggered
               example: 1587667174725
-
     webhook-response:
       type: object
-      table: webhook_response
       required:
         - webhookId
         - statusCode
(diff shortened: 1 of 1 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1,45 +1,62 @@
 openapi: 3.0.1
 info:
-  title: Livepeer Studio API Reference
-  version: 0.0.1-alpha.0
+  title: Livepeer API Reference
+  description: >-
+    This API reference provides documentation for the Livepeer Studio API, which
```
adamsoffer, line 5: "for the Livepeer ~Studio~ API"

packages/api/src/schema/schema.yaml
victorges, line 1: This file has both public API and private DB schemas, and there are several internal fields here that should not be exposed on the API reference. In the current form I haven't seen any change in this PR that solves this problem. 

I have already talked about some ideas about this back in our original call, like having a separate schema file or maybe tagging all the internal fields here and omitting them from the generated reference. The first idea (separate schema) might work better for the readme.io workflow that I understood, which reads a committed file from github directly (right?). Is there any readme.io feature that could help? (e.g. some filter to not include fields with a specific param like `internal: true`)

WDYT? Do you need any more context to carry on that fix? We can do a follow-up call if you need a refresher, now with more context on the whole schema! It might have been too much information back then.

We can also take this as a first step of the whole task. This PR alone improves the documentation for a bunch of fields by a lot! The current output cannot become our public API reference tho as it contains internal and possibly sensitive stuff. We can merge this as is as an improvement and work on the API reference separation right next, before releasing any reference docs. Separating would be better to facilitate review as well.

packages/api/src/schema/schema.yaml
```diff
@@ -1,45 +1,62 @@
...
+    description: Endpoints for handling responses from webhook clients.
+  - name: Verify Email Endpoints
+    description: Endpoints for resending email verification requests to users.
 
 components:
   schemas:
     ffmpeg-profile:
       type: object
-      description: LMPS ffmpeg profile
+      description: LMPS ffmpeg profile for video encoding
```
victorges, line 59:
```suggestion
      description: Livepeer profile for video encoding
```

packages/api/src/schema/schema.yaml
```diff
@@ -50,80 +67,106 @@ components:
...
           minimum: 1
+          example: 1
         gop:
           type: string
+          description: >-
+            The GOP (Group of Pictures) structure for the encoded video. This is
+            a string in the format of "M/N", where M is the number of frames in
+            a GOP and N is the distance in frames between two I-frames (also
+            known as keyframes).
+          example: 25/250
```
victorges, line 109: This is not the right format. GOPs are specified as just a duration string instead like `2s`

packages/api/src/schema/schema.yaml
```diff
@@ -50,80 +67,106 @@ components:
...
+          description: The video codec profile to use for encoding the video.
           enum:
             - H264Baseline
             - H264Main
             - H264High
             - H264ConstrainedHigh
+            - HEVCMain
+            - HEVCHigh
+            - VP8
+            - VP9
```
victorges, line 121: These are output profiles/codecs and we do not support these as output AFAIK. 
Weird that we have them below on `codecs` as well. Should probably remove all of them.

packages/api/src/schema/schema.yaml
```diff
@@ -50,80 +67,106 @@ components:
...
             - hevc
             - vp8
             - vp9
+          example: h264
     cdn-data-row:
       type: object
       description:
-        Fields names taken from log format description
-        https://support.highwinds.com/hc/en-us/articles/360052181571-Raw-Log-Access-through-GCS
+        Fields taken from raw log access data provided by StackPath CDN.
```
victorges, line 135: This is an example of an internal object that should not be exposed on the API reference. Should be fine keeping that internal link here as a documentation helper (in case it still works)

packages/api/src/schema/schema.yaml
```diff
@@ -50,80 +67,106 @@ components:
...
         stream_id:
           type: string
-          description: Session UUID. Used in recordings URLs.
+          description:
+            A unique session UUID that can be used to retrieve recording URLs.
         playback_id:
           type: string
-          description: Playback ID.
+          description:
+            A playback ID that can be used to retrieve recording URLs.
```
victorges, line 148:
```suggestion
            A playback ID that can be used to retrieve playback URLs.
```

packages/api/src/schema/schema.yaml
```diff
@@ -50,80 +67,106 @@ components:
...
           description: The total size of the request header.
         total_sc_bytes:
           type: number
-          description: Total bytes in the response to the client.
+          description: The total size of the response to the client.
         count:
           type: number
-          description: Number of parsed log lines
+          description: The number of parsed log lines.
     cdn-data-payload:
```
victorges, line 166: Same as internal object

packages/api/src/schema/schema.yaml
```diff
@@ -156,51 +206,46 @@ components:
           type: string
           readOnly: true
           example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+          description: The unique identifier for this record.
         fileName:
           type: number
-          index: true
```
victorges, line 161: You absolutely cannot remove our custom properties on the fields like this one or `unique`, or `internal` etc. This is what guarantees that an index will be generated for that field in the database. This is also why we need that separate DB schema (or some other solution) that I told you about.

packages/api/src/schema/schema.yaml
```diff
@@ -156,51 +206,46 @@ components:
...
       properties:
         id:
           type: string
-          readOnly: true
+          format: uuid
+          description: The unique identifier for the webhook
           example: de7818e7-610a-4057-8f6f-b785dc1e6f88
         kind:
           type: string
-          readOnly: true
```
victorges, line 180: All `readOnly` properties should also stay there and is what guarantees the field won't be included in request paylods that use it. Probably why the stream schema is so polluted on the current api reference that was generated. But there are probably other fields that should be `readOnly` and don't have it yet either, I can help with those once we get there.

packages/api/src/schema/schema.yaml
```diff
@@ -156,51 +206,46 @@ components:
...
       type: object
-      table: webhook
       required:
         - name
         - url
       additionalProperties: false
       properties:
         id:
           type: string
-          readOnly: true
```
victorges, line 176: Cannot remove.

packages/api/src/schema/schema.yaml
```diff
@@ -156,51 +206,46 @@ components:
...
+          description: 'The kind of object, always set to "webhook"'
           example: webhook
         name:
           type: string
+          description: The name of the webhook
+          example: my-webhook
         userId:
           type: string
-          readOnly: true
-          index: true
```
victorges, line 187: Cannot remove. Undo the removal of all the `readOnly`, `writeOnly`, `index`, `indexType`, `unique` etc...

packages/api/src/schema/schema.yaml
```diff
@@ -156,51 +206,46 @@ components:
...
+          description: The unique identifier of the user who created the webhook
         createdAt:
-          type: number
-          readOnly: true
+          type: integer
           description:
-            Timestamp (in milliseconds) at which stream object was created
+            The timestamp (in milliseconds) at which the webhook was created
           example: 1587667174725
-        event:
```
victorges, line 194: Uhm, it might be enough time to undo this change, yeah, but we should also remove the code that handled it! If you don't want to touch that code now, let's leave it in the schema for now.

packages/api/src/schema/schema.yaml
```diff
@@ -226,111 +270,115 @@ components:
         url:
           type: string
           format: uri
-          pattern: ^http(s)?://
+          pattern: "^http(s)?://"
+          description: The URL to which the webhook will send its payloads
+          example: "https://example.com/webhook"
```
victorges, line 275: WDYT of putting a wehook.site example?

packages/api/src/schema/schema.yaml
```diff
@@ -226,111 +270,115 @@ components:
...
         deleted:
           type: boolean
+          description: A flag indicating whether the webhook has been deleted
           default: false
         sharedSecret:
           type: string
+          description: The shared secret used to sign the webhook payload
```
victorges, line 282: Would be good to reference the webhooks doc here for details on how the shared secret is used (if we do have those docs, maybe we dont 😬 )

packages/api/src/schema/schema.yaml
```diff
@@ -354,33 +402,27 @@ components:
                 type: string
               probability:
                 type: number
-
     stream:
       type: object
-      table: stream
```
victorges, line 360: Cannot remove `table` either

packages/api/src/schema/schema.yaml
```diff
@@ -415,46 +457,63 @@ components:
         isActive:
           type: boolean
           description: If currently active
```
victorges, line 459: I think this one could be improved a bit wdyt? maybe just an expanded
```suggestion
          description: Whether this stream is currently active
```

packages/api/src/schema/schema.yaml
```diff
@@ -415,46 +457,63 @@ components:
...
+        partialSession:
+          description:
+            Indicates that this is not final object of `user's` session
+          type: boolean
+        previousSessions:
+          description:
+            Ids of the previous sessions which are part of `user's` session
+          type: array
+          items:
+            type: string
```
victorges, line 483: These are internal fields.

packages/api/src/schema/schema.yaml
```diff
@@ -415,46 +457,63 @@ components:
...
+            - name: 480p0
+              fps: 0
+              bitrate: 1600000
+              width: 854
+              height: 480
+            - name: 720p0
+              fps: 0
+              bitrate: 3000000
+              width: 1280
+              height: 720
```
victorges, line 516: Does this show nicely as a JSON on readme.io? The good thing about the previous format is that it is exactly what we set in the code: https://github.com/livepeer/livepeer-com/blob/8ca070d9333149c1dd7a00f609d730b0003df6ab/packages/api/src/controllers/stream.ts#L51

Perhaps make it a full JSON object instead so it's more consistent (i.e. instead of an array with a `-` item for each element, an actual `[ <elm> ]` comma-separated syntax, as in the code

packages/api/src/schema/schema.yaml
```diff
@@ -549,6 +611,9 @@ components:
               type: object
             streamNameGroups:
               type: array
+              items: #test
+                type: object
```
victorges, line 615: Forgot this? 👀

packages/api/src/schema/schema.yaml
```diff
@@ -549,6 +611,9 @@ components:
...
+
             sourceInfo:
```
victorges, line 617: Do not bother about this whole `wowza` object (or any references to that). That is unused today. It should not be included in any docs.

packages/api/src/schema/schema.yaml
```diff
@@ -646,7 +708,6 @@ components:
           $ref: "#/components/schemas/playback-policy"
         storage:
           $ref: "#/components/schemas/new-asset-payload/properties/storage"
-
```
victorges, line 649: Can we keep the whitespace between objects? Generally hard to parse through this huge file without any separation between the higher level entities. There are some components in the beginning that dont have spacing today, we can add to those for consistency instead of removing from the ones that already have it.

packages/api/src/schema/schema.yaml
```diff
@@ -672,25 +732,22 @@ components:
...
           description:
             Name of template to send to the users regarding the suspension.
-          enum: [copyright]
-
+          enum:
+            - copyright
     multistream-target-patch-payload:
       $ref: "#/components/schemas/multistream-target"
-      required: []
-      table: null
```
victorges, line 692: Also cannot be removed.

packages/api/src/schema/schema.yaml
```diff
@@ -724,25 +779,23 @@ components:
...
-      required: [type, stream]
+      required:
+        - type
+        - stream
       properties:
         type:
           type: string
-          enum: [jwt, accessKey]
+          enum:
+            - jwt
```
victorges, line 792: Missing accessKey

packages/api/src/schema/schema.yaml
```diff
@@ -724,25 +779,23 @@ components:
...
         stream:
           type: string
           description: Stream ID to which this gate applies
         pub:
           type: string
           description: Base64 of Public key used for JWT verification
-        accessKey:
-          type: string
-          description: Access key used for access-control verification
```
victorges, line 744: Should not be removed

packages/api/src/schema/schema.yaml
```diff
@@ -759,21 +812,20 @@ components:
...
+      required:
+        - type
       properties:
         type:
           type: string
           enum:
             - public
             - jwt
-            - lit_signing_condition # only for lit-signing-condition experiment
-            - webhook
```
victorges, line 776: Also should not be removed. I think you had a couple rebase problems here.

packages/api/src/schema/schema.yaml
```diff
@@ -795,41 +852,28 @@ components:
               type: string
             extraData:
               type: string
-        webhookId:
-          type: string
-          description: ID of the webhook to use for playback policy
-        webhookContext:
-          type: object
-          description: User-defined webhook context
-          additionalProperties: true
```
victorges, line 804: Rebase problem++

packages/api/src/schema/schema.yaml
```diff
@@ -858,42 +902,35 @@ components:
...
+          description: >-
             Should this stream be recorded? Uses default settings. For more
             customization, create and configure an object store.
           type: boolean
           example: false
         recordingStatus:
           type: string
-          readonly: true
           description: Status of the recording process of this stream session.
-          enum: [waiting, ready, none]
```
victorges, line 889: More rebase trouble

packages/api/src/schema/schema.yaml
```diff
@@ -858,42 +902,35 @@ components:
           type: number
           example: 2
           description: Rate at which transcodedBytes increases (bytes/second)
-        broadcasterHost:
```
victorges, line 861: Rebase trouble

packages/api/src/schema/schema.yaml
```diff
@@ -908,7 +945,8 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/ffmpeg-profile"
-
+        lastSessionId:
```
victorges, line 948: What's this new field for?

Edit: Actually this is probably more rebase trouble? 😅

packages/api/src/schema/schema.yaml
```diff
@@ -1083,21 +1109,18 @@ components:
...
-          readOnly: true
-          example: https://livepeercdn.com/asset/ea03f37e-f861-4cdd-b495-0e60b6d753ad/index.m3u8
+          example: >-
+            https://livepeercdn.com/asset/ea03f37e-f861-4cdd-b495-0e60b6d753ad/index.m3u8
           description: URL for HLS playback
         downloadUrl:
           type: string
-          readOnly: true
-          example: https://livepeercdn.com/asset/eaw4nk06ts2d0mzb/video
+          example: "https://livepeercdn.com/asset/eaw4nk06ts2d0mzb/video"
```
victorges, line 1117: WDYT of updating these examples with a recent URL? Those are using some old formats. Feel free to pick the URLs from any sample video uploaded anywhere

packages/api/src/schema/schema.yaml
```diff
@@ -1110,44 +1133,54 @@ components:
...
+                    - recording
                 sessionId:
                   type: string
                   description:
                     ID of the session from which this asset was created
             - additionalProperties: false
-              required: [type, inputAssetId]
+              required:
+                - type
+                - inputAssetId
```
victorges, line 1175: Hmm... Did you actually run the schema through some kind of prettier? I see a lot of non-functional styling changes that didn't really seem necessary... it's been a long review 😅 

It's generally better to at least separate such things in their own commits that can be reviewed separately.

packages/api/src/schema/schema.yaml
```diff
@@ -1164,32 +1197,34 @@ components:
...
                         will embed the Livepeer Player on the NFT while 'file'
                         will reference only the immutable MP4 files.
                     nftMetadata:
                       type: object
-                      description:
+                      description: >-
                         Additional data to add to the NFT metadata exported to
                         IPFS. Will be deep merged with the default metadata
                         exported.
-                $ref: "#/components/schemas/ipfs-file-info/properties"
```
victorges, line 1179: This cannot be removed.

packages/api/src/schema/schema.yaml
```diff
@@ -1428,46 +1454,44 @@ components:
               oneOf:
                 - type: object
                   additionalProperties: false
-                  properties:
-                    spec:
-                      oneOf:
-                        - type: "null"
-                        - $ref: "#/components/schemas/asset/properties/storage/properties/ipfs/properties/spec"
-                - type: [boolean, "null"]
-                  description:
```
victorges, line 1437: Cannot be removed.

packages/api/src/schema/schema.yaml
```diff
@@ -1641,18 +1668,6 @@ components:
                   type: string
                   description: Path for the HLS output
                   example: /samplevideo/hls
-            mp4:
```
victorges, line 1644: Rebase trouble.

packages/api/src/schema/schema.yaml
```diff
@@ -1662,58 +1677,43 @@ components:
...
           example: 66E2161C-7670-4D05-B71D-DA2D6979556F
         type:
           type: string
           description: Type of the task
           enum:
             - upload
-            - import # Import is the deprecated pipeline, upload is on Catalyst
+            - import
             - export
-            - transcode # Transcode is the deprecated pipeline, used for deprecated `/api/asset/transcode` and `/api/asset/:assetId/transcode` endpoints
```
victorges, line 1691: Are those comments invalid? Saw others been removed as well, I thought it was valid yaml syntax.

packages/api/src/schema/schema.yaml
```diff
@@ -1662,58 +1677,43 @@ components:
...
+            - transcode
             - transcode-file
         createdAt:
           type: number
-          readOnly: true
           description: Timestamp (in milliseconds) at which task was created
           example: 1587667174725
-        scheduledAt:
```
victorges, line 1698: Rebase...

packages/api/src/schema/schema.yaml
```diff
@@ -1806,27 +1806,18 @@ components:
                           type: string
                           description: Path for the HLS output
                           example: /samplevideo/hls
-                    mp4:
```
victorges, line 1809: Rebase...

packages/api/src/schema/schema.yaml
```diff
@@ -2056,31 +2043,31 @@ components:
...
       type: object
-      required: [publicKey, privateKey]
+      required:
+        - publicKey
+        - privateKey
       properties:
-        $ref: "#/components/schemas/signing-key/properties"
+        signing-key:
+          type: object
+          $ref: "#/components/schemas/signing-key"
```
victorges, line 2054: This is not the same thing as before.

packages/api/src/schema/schema.yaml
```diff
@@ -2485,69 +2451,54 @@ components:
...
-      table: usage
       additionalProperties: false
       properties:
         id:
           type: string
           description: day associated with usage
-          example: 2021-01-30
-          unique: true
-          index: true
+          example: 2021-01-30T00:00:00.000Z
```
victorges, line 2496: Is this the actual format these are provided in?

packages/api/src/schema/schema.yaml
```diff
@@ -2567,90 +2518,148 @@ components:
 
 paths:
   /stream:
-    tags:
-      - stream
```
victorges, line 2571: Hmm what are the tags used for? I see you renamed all of them but then the references were removed.

Additionally, is there perhaps a way to group the API references on readme.io based on tags, instead of the path? I think it'd look nicer and we can separate for example Live from VOD from other specific stuff (like "user").

packages/api/src/schema/schema.yaml
```diff
@@ -2567,90 +2518,148 @@ components:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
```
victorges, line 2552: Could we have like just 1 entry like "error response"? We don't return only 401 or 500 from these APIs, so I think we can both simplify the docs as well as be more extensive. Could specify the status code as "4xx/5xx" or "> 400" here, or something like that. That should also replace that "default" response listed in these. WDYT?

packages/api/src/schema/schema.yaml
```diff
@@ -2567,90 +2518,148 @@ components:
 
 paths:
```
victorges, line 2519: As chatted on Discord, we are both missing a lot of relevant APIs here as well as listing APIs that are not relevant to users. I can help come up with the exact list of APIs that should be documented here once we get there.

packages/api/src/schema/schema.yaml
```diff
@@ -3022,17 +3086,26 @@ paths:
               schema:
                 $ref: "#/components/schemas/error"
 
-  /user/{id}:
+  "/user/{id}":
     patch:
```
victorges, line 3090: This API doesn't exist. IDK why it was added here lol

victorges (changes requested):
Thanks for all the documentation on those fields! Apart from that, this changest pretty much broke the API schema in several ways, which is why tests are also failing. Check the specific comments below.

packages/api/src/schema/schema.yaml
suhailkakar: Yes, that make sense. I have already tried different params like internal, x-hidden but non of them worked on the readme.io ui. 

I have reached out to their team and currently the only way to hide the API references is using the Readme dashboard. 

unnamed

packages/api/src/schema/schema.yaml
suhailkakar: I believe it does exist: https://github.com/livepeer/studio/blob/master/packages/api/src/schema/schema.yaml#L3025

packages/api/src/schema/schema.yaml
suhailkakar: Not exactly sure, but i didn't find any other asset APIs in the schema: https://github.com/livepeer/studio/blob/d6c4db23b0965495b0f637699c8022098122e00f/packages/api/src/schema/schema.yaml#L3198-L3218

There is only a `get`, is the others are on separate file?

packages/api/src/schema/schema.yaml
suhailkakar: Sounds great!

packages/api/src/schema/schema.yaml
suhailkakar: Yes, The tags are not really need & doesn't show in the readme.io ui. Do we need them in the schema.yaml? (if yes, will add them back)

suhailkakar:
@victorges Thanks for the review, yeah that all make sense! Let's have a follow-up call regarding these (will dm you on discord)

packages/api/src/schema/schema.yaml
victorges: I think we don't need them! Feel free to delete them

packages/api/src/schema/schema.yaml
victorges: I think the biggest confusion is that you are assuming this schema is the source of truth! It is actually not. The only section of this schema that has been maintained is the `components` section, while the `paths` were updated on a best-effort basis but not really kept in sync until now.

As for the APIs, the source of truth are the API controllers code. They are all in the `src/controllers` folder and you can generally find the APIs easily there, declared with express routers. The schemas here are either incomplete sometimes, or totally hallucinating hahaha. In case of assets/VOD, it is missing a bunch of APIs like request-upload etc. In case of users, it's listing APIs that don't really exist, like that other comment I made. 

Just to be clear, this is a tech debt issue on this current schema, and not a problem with this PR. But part of the work in this API reference project is cleaning up that tech debt and documenting here the actual APIs that we want in our API reference. This is the context-dependent part of the work that I mentioned to you I can help on the low-level. Let's just sort the bigger picture first (like how to separate the API&DB schemas).

packages/api/src/schema/schema.yaml
victorges: Please refer to my other comment. The schema is not the source of truth, only the code is.

And clarifying my original comment: 
With "IDK why it was added here lol" I actually meant idk why it had been added to this schema in the first place (not a change from this PR). Maybe the API existed at some point and was removed, IDK, but on the controller the only patch API I can find is [`PATCH /user/:id/suspended`](https://github.com/livepeer/livepeer-com/blob/cb8d298f5c1c6220c46fab07fbe39158070c2408/packages/api/src/controllers/user.ts#L409) which is also an admin-only API anyway, so should not be documented here.

PR #1686, opened 2023-04-10
api: Create deletedAt field on assets

victorges:

This is to create a new deletedAt field on assets for us to use in our own billing and monitoring.

It will be updated once the asset is deleted, and we will also send the corresponding webhooks
from that operation.

 - Delete assets via a SQL update instead of the existing markDeleted helper
 - Create deletedAt field and updated it on deletion as well
 - Clean-up VOD webhook payloads by cleaning up task/asset objects as well

✅ `yarn test`
 - put on staging and check deletedAt field gets created

Implements DAT-80

**Checklist**


packages/api/src/controllers/asset.ts
```diff
@@ -441,41 +442,47 @@ function parseUploadUrl(
...
 }
 
+export async function toExternalAsset(
+  a: WithID<Asset>,
+  config: CliArgs,
+  details = false,
+  isAdmin = false
+) {
+  const { ipfsGatewayUrl, ingest: ingestsConfig } = config;
+  const ingest = ingestsConfig[0]?.base;
```
leszko, line 455: I'm not sure I understand this line, why do you take `[0]` from the ingest configs. Could you maybe add a comment with the explanation?

leszko (approved):
LGTM. Added one super nit comment.

One more general comment, I think this PR both adds new functionality (`deletedAt`), but also does some refactoring. I think it'd be better to split it into a separate refactoring PR in the future. Simpler to review (and rollback if needed).

packages/api/src/controllers/asset.ts
victorges: So this is actually not new logic, it's just copied from all the other places we use that `getIngest` function. As you noticed, I had to refactor that config in this PR in order to have access to the ingests from the `scheduler` (which doesn't have a reference to a request).

The only known source for this is existing code that used the ingest config: https://github.com/livepeer/livepeer-com/blob/0fdf4239faab869958998eb0f0d782ded98d6004/packages/api/src/controllers/session.ts#L121-L126
I don't know why the config was made an array instead of a slice tho, but every code I saw just gets the first element, and we also only ever deploy with 1 element in the array. I added a comment explaining that here.

PR #1692, opened 2023-04-19
Implement Recording V2

emranemran:

Change Recording from using go-livepeer (aka. Recording V1) to using Mist (aka Recording V2).

Related design doc: [Livestream Recordings Refactor](https://www.notion.so/Livestream-Recordings-Refactor-2023-03-06-cdcf96ce84444d04be4003e097e65264)

- Add env variable `LP_API_RECORDING_SOURCE_URL` to define the public bucket URL where Mist stores recordings
- Add `recordingSessionId` field to the `session` DB object and fill it with `sessionID` (`uuid`) passed from Mist
- Use Catalyst VOD transcoding to output the recording asset
- Resolve recording URL (both HLS and MP4) from asset instead of statically defined `source.mp4` and `index.m3u8`
- Change UI to read MP4 from `mp4Url` (rather than converting HLS URL)

**Additional comments**
- We now need to make 2 DB calls to fetch each HLS/MP4 recording path; this may have some performance implications, but I suggest to refactor/improve it in a separate PR
- There should be also some additional changes to clean up the Recording V1 code, but I think we should do it after the deployment to prod (Added ticket for it [here](https://linear.app/livepeer/issue/VID-147/studio-clean-up-recording-v1))

**TODO before merging this PR**
- [x] Clean up code & unit tests
- [x] Fix issue with session re-usage
- [x] Fix issue in Mist to storage Manifest file in the correct directory
- [ ] HLS Manifest input in Catalyst

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/stream.ts
```diff
@@ -193,13 +193,38 @@ export function getPlaybackUrl(ingest: string, stream: DBStream) {
...
+
+  // For transcoded recordings, we store only one MP4 output file
+  const type = mp4 ? "static_transcoded_mp4" : "catalyst_hls_manifest";
+  const path = asset.files?.find((f) => f.type === type)?.path;
+  if (!path) {
+    // Recording renditions not yet ready
+    return;
+  }
+
+  return pathJoin(os.publicUrl, asset.playbackId, path);
```
victorges: You should use [this](https://github.com/livepeer/livepeer-com/blob/ad8d909a045b536ffa7cc15955e828e6e4e6aa5d/packages/api/src/controllers/asset.ts#L227) function instead which is already exported from the asset controller. To avoid fetching the object store, you can even jut use `withPlaybackUrls` (same file, also exported) which will do that for you. It's also better cause it already checks the `asset.status.phase` field as well. Then your logic here can be `mp4 ? asset.playbackUrl : asset.downloadUrl` 😌 

You might want to rename the function on `import` so it's not just a generic "get/with playback URL" in this file which does not work for streams. [Example](https://github.com/livepeer/livepeer-com/blob/486b847193ebabc92caadc4b10db63edae533e67/packages/api/src/controllers/playback.ts#L8)

packages/api/src/controllers/stream.ts
```diff
@@ -389,23 +414,25 @@ export function getRecordingFields(
...
-    : {
-        recordingStatus,
-        recordingUrl: getRecordingUrl(ingest, session),
-        mp4Url: getRecordingUrl(ingest, session, true),
-      };
+  if (!isReady && !forceUrl) {
+    return { recordingStatus };
+  }
+  const recordingUrl = await getRecordingUrl(ingest, session);
+  const mp4Url = await getRecordingUrl(ingest, session, true);
```
victorges: Could we get both fields in a single call? Otherwise, we'll be going 4 times to the DB to fetch the OS and the asset twice each.

packages/api/src/controllers/stream.ts
```diff
@@ -696,27 +725,8 @@ app.post(
       return res.json({ errors: ["not found"] });
     }
 
+    const recordingSessionId = req.query.sessionId?.toString();
```
victorges: WDYT of just calling this `sessionId` everywhere? (also creating the session object with this as the ID, if it's an UUID in the same format as ours)

packages/api/src/controllers/stream.ts
```diff
@@ -770,6 +781,7 @@ app.post(
       profiles: childStream.profiles,
       record,
       recordObjectStoreId,
+      recordingSessionId,
```
victorges: Yeah, IMO this should be just "session ID" and it should be used as the `id` field above. We should not use the same "child stream" ID as the ID of the session amymore (especially since they're 1-N now, not 1-1). WDYT?

packages/api/src/controllers/stream.ts
```diff
@@ -770,6 +781,7 @@ app.post(
...
       recordingStatus: record ? "waiting" : undefined,
     };
     await db.session.create(session);
```
victorges: I think you're missing some logic to check if the session doesn't already exist before creating a new one.

packages/api/src/controllers/stream.ts
```diff
@@ -1085,8 +1097,8 @@ function publishDelayedRecordingReadyHook(
       userId: session.userId,
       sessionId: session.id,
       payload: {
-        recordingUrl: getRecordingUrl(ingest, session),
-        mp4Url: getRecordingUrl(ingest, session, true),
+        recordingUrl: await getRecordingUrl(ingest, session),
+        mp4Url: await getRecordingUrl(ingest, session, true),
```
victorges: I believe this will not work anymore, since `getRecordingUrl` will return `undefined` when the recording session hasn't been created yet and it will only be created on the handling of this webhook.

packages/api/src/parse-cli.ts
```diff
@@ -316,6 +316,11 @@ export default function parseCli(argv?: string | readonly string[]) {
           "base stream name to be used in wildcard-based routing scheme.",
         type: "string",
       },
+      "recording-source-url": {
+        describe:
+          "base stream name to be used in wildcard-based routing scheme.",
+        type: "string",
+      },
```
victorges: WDYT of doing this similarly to other VOD object stores? Meaning we create a specific object store in the API and then this field is only a `recordingSourceObjectStoreId`, with the actual url/publicUrl being fetched from the DB whenever necessary. We will only be using the public URL in the API for now, but I think we may evolve this in different directions (e.g. catalyst asking the "write" URL from Studio) and the consistency is good IMO.

I can help creating the object store object if you wanna follow this approach! Gimme the URL/publicURL pair and I can give you the ID.

packages/api/src/parse-cli.ts
```diff
@@ -316,6 +316,11 @@ export default function parseCli(argv?: string | readonly string[]) {
           "base stream name to be used in wildcard-based routing scheme.",
         type: "string",
       },
+      "recording-source-url": {
+        describe:
+          "base stream name to be used in wildcard-based routing scheme.",
```
victorges: The description is outdated

packages/api/src/schema/schema.yaml
```diff
@@ -486,6 +486,10 @@ components:
           example: D8321C3E-B29C-45EB-A1BB-A623D8BE0F65
           description:
             ID of object store where to which this stream was recorded
+        recordingSessionId:
+          type: string
```
victorges, line 490: Are we going to be querying by these fields? If so, you probably want to add an `index: true` directive to this (or even `unique: true` where that's relevant (e.g. sessions table))

packages/api/src/controllers/stream.ts
```diff
@@ -1069,12 +1081,12 @@ function publishRecordingStartedHook(
 /**
  * We don't actually send the webhook here, but schedule an event after a timeout.
  */
-function publishDelayedRecordingReadyHook(
+async function publishDelayedRecordingReadyHook(
```
victorges, line 1114: Do we still want to delay this event, or may as well send it immediately? That assuming that this will be called with an official "end" event from Mist, not the current `/setactive` which is fallible (if that's the case, let's keep the delay)

victorges: Also something else: you might want to take a look on the caller of this function. It might need to change as well (e.g. not sending one event for each child stream, fetching sessions based on `recordingSessionId`/`sessionId` field and not just assuming there is 1 session with the same ID of the child stream (1-1).

packages/api/src/webhooks/cannon.ts
```diff
@@ -552,11 +550,15 @@ export default class WebhookCannon {
...
     await taskScheduler.createAndScheduleTask(
-      "import",
+      "upload",
       {
-        import: {
-          url: mp4RecordingUrl,
-          recordedSessionId: session.id,
+        upload: {
+          url: pathJoin(
+            this.recordingSourceUrl,
```
victorges: So are we not getting a "recording ended" (conceptually) event from Catalyst anymore? If we do, I think it'd make more sense for this source URL to come on that payload as well. If not, this is fine.

packages/www/components/Dashboard/StreamDetails/SessionsTable/helpers.tsx
```diff
@@ -138,7 +138,7 @@ export const rowsPageFromState = async (
             ) : (
               <Box>—</Box>
             ),
-          mp4Url: stream.recordingUrl ? stream.recordingUrl : undefined,
+          mp4Url: stream.mp4Url ? stream.mp4Url : stream.recordingUrl,
```
victorges: **nit:** by definition this renders the same result
```suggestion
          mp4Url: stream.mp4Url ?? stream.recordingUrl,
```

victorges: Althoooough, do we want to show an HLS URL on the download button? I think we probably just don't show anything
```suggestion
          mp4Url: stream.mp4Url,
```

packages/www/components/Dashboard/StreamSessionsTable/helpers.tsx
```diff
@@ -153,7 +153,7 @@ export const rowsPageFromState = async (
             ) : (
               <Box>—</Box>
             ),
-          mp4Url: stream.recordingUrl ? stream.recordingUrl : undefined,
+          mp4Url: stream.mp4Url ? stream.mp4Url : stream.recordingUrl,
```
victorges: Same

victorges (changes requested):
Looking good overall! Most of the comments are only suggestions, but some are required for the whole thing to work. 

Specifically, the most important one is about not always creating the session object. There should be some logic on that hook to check if the session already exists and skip that if it's the case. Or, if we use the `recordingSessionId` as the `session.id`, you can just try creating it and handle the primary key error from the DB if it already exists. 

Speaking of that, the other high level point is removing the "recording session id" concept and just calling it "session id". This will make even more sense if we use it as the `session.id` as well. Just a suggestion tho, don't need to do it especially if the rename in the code feels painful now.

packages/api/src/controllers/stream.ts
leszko: Updated in https://github.com/livepeer/studio/pull/1692/commits/44d4cf4509b4b3c7a0c4cd326b74e7ba11aecaf2

Note that I had to change how `getDownloadUrl()` works, but I think it's a better behavior. It first tried to fetch mp4 rendition file and only if it does not exist, then it returns the source file. @victorges Let me know if it's ok in your opinion.

packages/api/src/webhooks/cannon.ts
leszko: No, we're not calling "recording ended" from Catalyst.

packages/api/src/controllers/stream.ts
leszko: Updated in a10bb97278aa3e333629469accb329d4b888cf5f

packages/api/src/controllers/stream.ts
leszko: I think we still need to keep the relation 1-1 between child stream and sessions because we make this assumption in a lot of code parts by passing child stream as a session, e.g. [here](https://github.com/livepeer/studio/blob/cbdbc0d335d132c5b6610c17d60e4fac687e9a1b/packages/api/src/controllers/stream.ts#L453) and [here](https://github.com/livepeer/studio/blob/cbdbc0d335d132c5b6610c17d60e4fac687e9a1b/packages/api/src/controllers/stream.ts#L593).

In general, I think we could change this relation to 1-N and reuse sessions (while not reusing child streams), but I think we would need to make some bigger refactoring/change. So, I'd leave it as a separate PR. 

Wdyt? Or am I missing something?

packages/api/src/controllers/stream.ts
leszko: I think we should name it `sessionId` only if it's used as `id` of the `session` entity. We could actually use it as as `session.id` (and child `stream.id`), but I'm not sure we can do it, because I see this comment in the code:

```javascript
    // The first four letters of our playback id are the shard key.
    const id = stream.playbackId.slice(0, 4) + uuid().slice(4);
```

Then, I guess we cannot we want to keep sharding with the playback id. Wdyt?

packages/api/src/controllers/stream.ts
leszko: My understanding is that we still keep the relation of 1-1 between the child stream and session. Then, I think we'll always create a session here. Isn't that the case?

packages/api/src/controllers/stream.ts
leszko: Yeah, I think you're right. But I guess we don't need these recording fields in the `recording.ready` event. Removed here: d9cd17dba40435187e79d62690731a761c2abdba

packages/api/src/parse-cli.ts
leszko: Updated in be2a1f64d7ccf309584a69035295d88989833208

leszko: Updated in be2a1f64d7ccf309584a69035295d88989833208

packages/api/src/parse-cli.ts
leszko: Updated in be2a1f64d7ccf309584a69035295d88989833208

packages/api/src/schema/schema.yaml
leszko: Added index here https://github.com/livepeer/studio/pull/1692/commits/918b9f73b5a7fcc663a27eea37fd6748dd5cdc27

I'm not sure about `unique`, didn't add it for now. My reasoning is that, even though I expect them to be unique, it's not a requirement from the studio perspective.

packages/api/src/controllers/stream.ts
leszko: Both these points are discussed in different comments.

1. There will be no "official end` event, so I think we should leave it as it is.
2. I think we'll still have 1-1 relation between child streams and sessions.

packages/www/components/Dashboard/StreamDetails/SessionsTable/helpers.tsx
leszko: I think we need to keep it that way for the sake of backwards compatibility. This `mp4Url` is passed to `makeMP4Url()` and then mp4 url is created from hlsUrl.

leszko (commented):
@victorges I addressed all your review comments. Could you take a look again? 🙏

packages/api/src/controllers/stream.ts
victorges: Hm yeah, we have that logic for sharding. It is only really necessary for the actual "transcoding sessions" (child streams) though [1], but have no real use on the sessions! So we must keep the sharding logic for the child streams, but are free to update the sessions as we want! I'd say they should have the `session.id = sessionId` and probably not even have a `playbackId`.

> (and child stream.id)

Note that we cannot use the `sessionId` for the child stream since there can be multiple child streams for the same `sessionId`. The child stream should stay almost the same with this new logic, not changing anything apart from the new field that points to the session ID. The sessions that are free to change as it makes the most sense for Recording V2.

[1] We have special routing configs for the `/live/{sessionId}` API, which routes streams that start with a specific character to a fixed broadcaster instance. I believe this was made to optimize resources on broadcasters etc, which is still valid until we move to catalyst-local broadcasters.

Edit: Whole comment based on the assumption we're doing 1-N, which is what I thought was the agreed upon design. Some of it not relevant if we go 1-1.

packages/api/src/controllers/stream.ts
victorges: I thought this was the main requirement for this change in Studio API! To have a consistent session object that corresponds to the "Mist session" (and thus, recording). If we don't make this change, there will be other places that will break, like for example:
 - now creating N assets from 1 recording cause we have potentially multiple sessions per "mist session"
 - showing multiple sessions on the API and dashboard that now have the same recording (or worse, the duplicated ones from above) 

This 1-1 relationship hasn't existed [for more than a month](https://github.com/livepeer/studio/pull/1662) [1] and it was only updated to be like that to simplify creating the right 1-N relationship that we want or rec v2 (it was 1-N before but in a broken way).

So IMO that change to a 1-N relationship is exactly what we should be doing in this PR! Otherwise we're gonna have to solve a lot of other problems to go in a direction that we don't really wanna go.

[1] And again, did that while thinking I was doing the recordings v2 myself. I'm sorry you're having to pick up all this context mid-refactor 😅

packages/api/src/controllers/stream.ts
victorges: I had the exact other understanding as explained in the other comment and on that Discord thread. We should have 1-N relationship between sessions and child streams of the same session.

packages/api/src/controllers/stream.ts
victorges: I'm gonna check with @emranemran, I haven't seen this decision of keeping child streams and sessions 1-1 and I don't know how that's gonna work on the API/dashboard.

packages/www/components/Dashboard/StreamDetails/SessionsTable/helpers.tsx
victorges: Ahh I see! Hadn't got that part. Sounds good then 😄

packages/api/src/parse-cli.ts
victorges: Cool! Have you already created the corresponding object stores? Ideally they should be in the `infraservice@livepeer.org` account, creds in 1pwd. Or you can send me the URLs and I can create it

packages/api/src/controllers/asset.ts
```diff
@@ -263,6 +263,13 @@ function getDownloadUrl(
   asset: WithID<Asset>,
   os: ObjectStore
 ): string {
+  // If mp4 rendition file exists, then use it as download url
+  const staticPlaybackInfos = getStaticPlaybackInfo(asset, os);
+  if (staticPlaybackInfos.length > 0) {
+    return staticPlaybackInfos[0].playbackUrl;
```
victorges: I don't think we have a guarantee that this array will be ordered. This should probably pick the rendition with the highest quality instead of just returning the first one.

packages/api/src/controllers/asset.ts
```diff
@@ -263,6 +263,13 @@ function getDownloadUrl(
...
+  }
```
victorges: This will change the behavior of the download URL. Instead of returning the original source file as provided by the user, which is what might be saved on IPFS for example and have a strict content integrity requirement, it will start returning an altered version of it. This could lead to unexpected content mismatches like hashes/CID not matching etc. I can see why we want to use this for recordings though, where there is no source file.

So for now, WDYT of just adding an optional boolean to this function like `useStaticRenditions` or something like that, that is only set from the recordings flow? Or perhaps only if `asset.source.type === 'recording'`, which might be safer so this also works on the asset API as well.

On the long term, I think Catalyst should start creating a static rendition with the same quality as the source and save it on the same place as we copy the source file (`/<playbackId>/video`, or probably add an extension there). This is probably more complex tho, especially since the copy is not even done on Catalyst today (still on `task-runner`)

packages/api/src/controllers/stream.ts
```diff
@@ -193,19 +195,22 @@ export function getPlaybackUrl(ingest: string, stream: DBStream) {
   return pathJoin(ingest, `hls`, stream.playbackId, `index.m3u8`);
 }
 
-async function getRecordingUrl(
+async function getRecordingUrls(
```
victorges, line 198: Nice!

packages/api/src/controllers/stream.ts
```diff
@@ -1097,10 +1123,8 @@ async function publishDelayedRecordingReadyHook(
       userId: session.userId,
       sessionId: session.id,
       payload: {
-        recordingUrl: await getRecordingUrl(ingest, session),
-        mp4Url: await getRecordingUrl(ingest, session, true),
```
victorges: Can you add those from the cannon instead so we don't break compatibility?

packages/api/src/schema/schema.yaml
```diff
@@ -908,6 +909,7 @@ components:
           type: string
           example: 39bf51c4-38eb-4e71-8d29-36da22035dc2
           description: UUID of Recording Session created by Mist
+          index: true
```
victorges: Oh just realized one thing... We have turned off the automatic index creation on most tables but the VOD ones as they were locking the DB. I think it still makes sense to have the index declared here for the future, but we might need to create these indexes manually before deploying. 

Can you make sure to ping me once we merge this PR / plan on deploy to prod so I can help with that? A table scan on the `streams` table could be pretty catastrophic 🥲

victorges (commented):
Changes make sense, made some extra comments.

I still think we should move to 1-N, though I'm less confident we can still do it this week given the time constraint and, as you mentioned, the complexity of that. I don't think the current version will work well though, especially with the multiple assets being created for the same recording, one for each transcoding session (child stream) created... Let's try to sync tomorrow.

If we move to 1-N, there is also 1 missing change here: update [`stream-info-service`](https://github.com/livepeer/livepeer-com/blob/0fdf4239faab869958998eb0f0d782ded98d6004/packages/api/src/app/stream-info/stream-info-app.ts#L366) to update the session given the right ID. It should be as simple as:
 - Check all the usages of the `db.session` table
 - Instead of using the child stream ID (`si.stream.id`) use the session ID instead (`si.stream.sessionId`)
 - Probably better to keep backward compat as `si.stream.sessionId ?? si.stream.id`

emranemran:
> I still think we should move to 1-N, though I'm less confident we can still do it this week given the time constraint and, as you mentioned, the complexity of that. I don't think the current version will work well though, especially with the multiple assets being created for the same recording, one for each transcoding session (child stream) created... Let's try to sync tomorrow.

Yeah, I too thought we were planning to go to 1-N between session and child session. The discord discussion might have gotten lost in translation unfortunately. Since we extended our deadline, we can try to change it up - let's sync up during your day tomorrow to get on the same page.

PS. we're all learning together on how much of a beast this refactor project is turning out to be so lot's of great lessons for us to learn from during a post-launch review (cc @thomshutt).

emranemran:
In addition, I think some of the confusing bits might be around our use of "sessions" in multiple places. 

* 1-1 relation between mist session to studio session
* 1-N relation between studio session to child stream

packages/api/src/parse-cli.ts
leszko: Not yet, but I'll create them (or ask for your help) before merging this PR.

packages/api/src/controllers/asset.ts
leszko: They are sorted by bitrate, so it should be ok. Plus, we'll generate only one MP4 rendition for recordings. [Here's](https://linear.app/livepeer/issue/VID-127/catalyst-api-generate-mp4-for-only-highest-quality-rendition) the related ticket.

packages/api/src/controllers/asset.ts
leszko: Ok, let's add this flag for now. It's a little hacky, but it's probably the simplest to do right now. Added here: cc9162f4b4972545974648195354f9f5895e5b87

packages/api/src/schema/schema.yaml
leszko: 👍

packages/api/src/controllers/stream.ts
leszko: Isn't `cannon` only for new recordings, which will always work with Recording V2? Not sure I understand why we need backward-compatibility for this part.

leszko (commented):
Added come comments.

packages/api/src/controllers/stream.ts
victorges: Oh, the cannon is not used only for recordings. The recordings logic there is actually a "plugin" on top of its main responsibility: firing webhooks. So this is the `recording.ready` webhook, which fires 1 minute after the stream is done, which can let a user's backend know that they can do sth with the recording. 

Ideally, we would also change these webhooks to only call the `recording.ready` one the recording asset is created, since that's how users will really interact with it. I think we can do a low effort now, by keeping the `recording.ready` when the asset is _created_ (started processing) and then tell users to wait for the proper VOD webhooks to know when it's actually available.

So this `payload` here is actually the **webhook payload**. I was thinking that we could omit the URLs here, since they're not ready yet, but add them from that `recording.ready` special handling on the webhook cannon. Maybe also an `assetId` field? Theoretically the URL won't work immediately when we send the webhook, but I think it's still better than removing the field that some users might be depending on. WDYT?

victorges:
OK sharing some context on other changes that might come from the shift to 1-N:

### [stream-info-service](https://github.com/livepeer/livepeer-com/blob/0fdf4239faab869958998eb0f0d782ded98d6004/packages/api/src/app/stream-info/stream-info-app.ts#L366)

That code currently updates the child stream, session and parent stream with the usage metrics. With this new relationship between child stream and sessions, instead of fetching a session with the same ID as the child stream, we should fetch it from the `sessionId` field instead. Repeating the steps I sent on the other comment:
 - Check all the usages of the db.session table
 - Instead of using the child stream ID (si.stream.id) use the session ID instead (si.stream.sessionId)
 - Probably better to keep backward compat as si.stream.sessionId ?? si.stream.id
 
 ### [recording.ready event](https://github.com/livepeer/livepeer-com/blob/8ca070d9333149c1dd7a00f609d730b0003df6ab/packages/api/src/controllers/stream.ts#L1002)
That function is called with the `/setactive = false` call made from mapic when the stream ends (that one from the CONN_CLOSE trigger), or to "reconcile" sessions that may have never got that hook (which is why there's all that `isStreamStale` logic). Check the call-chain to understand more of that context.

Anyway, here we also need to change the logic that assumes that the session has the same ID as the child stream, to something that fetches the session from the .sessionId field instead. We might want to keep _some_ backward compatibility (doing the old logic if there's no `sessionId` present), but it shouldn't be too hard hopefully. Feel free to delete that current backward compatibility logic there, I created on the last refactor and it can be safely removed by now. Just so we don't need 2 separate backward compats logic. 

This logic is closely tied to that other one that you found in [the webhook cannon](https://github.com/livepeer/livepeer-com/blob/486b847193ebabc92caadc4b10db63edae533e67/packages/api/src/webhooks/cannon.ts#L493). Both need to be updated consistently with each other (and ideally with backward compat, but not a big issue if we lose some recordings on the deploy. we don't have much recording reliability today anyway).

Some caveats:
 - There's [some logic](https://github.com/livepeer/livepeer-com/blob/8ca070d9333149c1dd7a00f609d730b0003df6ab/packages/api/src/controllers/stream.ts#L1049) in there that assumes the child stream ID is the same as the session as well, and assumes there is a 1-1 relationship. (sorry, in that code the "child stream" is named session and the actual "session" is named "userSession" 😅, feel free to rename)
 - There's also some handling of an `isActive` field in the child streams. [It is used as a flag](https://github.com/livepeer/livepeer-com/blob/486b847193ebabc92caadc4b10db63edae533e67/packages/api/src/webhooks/cannon.ts#L520-L527) to avoid creating duplicate assets for the same session recording. 

On our new flow, we probably want to change that whole logic to be more based on the sessions than on child streams. The only reason it was based on child streams before was because the "transcoding sessions" (child streams) were where the recordings happened. Now we're moving to a "mist session" (session) based recording, so we should change that logic accordingly. That `isActive` field on child streams was literally created on the last refactor, feel free to replace it with one on the `session` instead and have a similar logic. 


### [the webhook cannon](https://github.com/livepeer/livepeer-com/blob/486b847193ebabc92caadc4b10db63edae533e67/packages/api/src/webhooks/cannon.ts#L493)

Already mentioned on the above section, but that's the other place where I know we will need to make changes.

### [`withRecordingFields` usage](https://github.com/livepeer/studio/blob/cbdbc0d335d132c5b6610c17d60e4fac687e9a1b/packages/api/src/controllers/stream.ts#L453) (you mentioned them yourself)

I think we can update that logic to fetch the session recording ID instead, but I wouldn't fret too much about it. I don't think it will be a big issue if we just stopped returning those URLs for the child streams. We only recommend our users to use the `/session` endpoints anyway, and I hope that's what they've been using 😟

leszko:
@victorges I think I addressed all parts related to the 1-N change. Btw. thanks for the hints about the code parts.

I think the most difficult is this `isActive` to avoid asset duplication for the given session. I removed all related code and used `session.id` as `asset.id`. It may be considered a hack, but on the other hand, I think this solves the issue of duplicates in the right way. I think the previous solution was actually prone to race conditions. If we don't want the asset and session to share the id, the other option is to create a new field like `asset.recordingSessionId` and make it `unique`. In theory, we could make `asset.source.sessionId` as `unique`, but I guess we already have some duplicated entries in prod. Let me know your thoughts.

Other than that, I think it works fine.

leszko (commented):
@victorges Please look into this PR again.

packages/api/src/controllers/asset.ts
```diff
@@ -267,8 +267,16 @@ function getDownloadUrl(
   { vodObjectStoreId }: Request["config"],
   ingest: string,
   asset: WithID<Asset>,
-  os: ObjectStore
+  os: ObjectStore,
+  useStaticRenditions: boolean
 ): string {
+  if (useStaticRenditions) {
```
victorges: Sorry I know I suggested this myself, but I think we need to make this conditional on the `asset` object only, otherwise when this function is executed through the asset API flow, it will not return the correct download URL for the recording-based asset.

So I think we need sth like:
```suggestion
  if (asset.source?.type === "recording" && !asset.playbackRecodingId) {
```
This will match only the assets that come from the new recordings v2 flow (which went through the `import` flow and always had that `playbackRecordingId` field set). Then we also don't need the boolean arg anymore. WDYT?

packages/api/src/controllers/stream.test.ts
```diff
@@ -1254,120 +1255,150 @@ describe("controllers/stream", () => {
...
+      expect(sessions[0].id).toEqual(sessionId);
       expect(sessions[0].transcodedSegments).toEqual(4);
-      expect(sessions[0].createdAt).toEqual(sess1.createdAt);
 
-      // create second session
-      res = await client.post(`/stream/${parent.id}/stream`, {
-        ...smallStream,
-        name: "sess2",
-      });
+      // create second stream re-using the same session
```
victorges, line 1325: Nice :)

packages/api/src/controllers/stream.ts
```diff
@@ -193,13 +195,42 @@ export function getPlaybackUrl(ingest: string, stream: DBStream) {
...
-  return pathJoin(
+async function getRecordingUrls(
+  config: CliArgs,
+  ingest: string,
+  session: DBSession
+): Promise<{ recordingUrl: string; mp4Url: string }> {
+  const asset = await db.asset.getBySessionId(session.id);
+  if (!asset || ["waiting", "processing"].includes(asset.status?.phase)) {
+    // Recording processing in progress
+    return;
```
victorges: It's probably more future-proof to check for the single phase where we expect a recording to be available
```suggestion
  if (asset?.status?.phase !== "ready") {
    // Recording processing in progress
    return;
```

packages/api/src/controllers/stream.ts
```diff
@@ -193,13 +195,42 @@ export function getPlaybackUrl(ingest: string, stream: DBStream) {
...
+  }
+
+  if (asset.id != session.id) {
```
victorges: Neat :)

**nit:**
```suggestion
  if (asset.id !== session.id) {
```

Also, another way to check if it's an old pipeline asset is through the `asset.playbackRecordingId` field. I think it might be more specific so slightly preferrable. Really soft opinion tho, it's fine like this.

packages/api/src/controllers/stream.ts
```diff
@@ -442,32 +486,39 @@ app.get("/:parentId/sessions", authorizer({}), async (req, res) => {
     }
   }
 
-  let [sessions] = await db.stream.find(query, {
+  let [sessions] = await db.session.find(query, {
```
victorges, line 482: Nice I like this

packages/api/src/controllers/stream.ts
```diff
@@ -442,32 +486,39 @@ app.get("/:parentId/sessions", authorizer({}), async (req, res) => {
...
-        ...combinedStats,
-        createdAt: session.userSessionCreatedAt || session.createdAt,
-      };
-    }
-    return session;
-  });
+      if (!raw) {
+        if (session.previousSessions && session.previousSessions.length) {
+          session.id = session.previousSessions[0]; // return id of the first session object so
+          // user always see same id for the 'user' session
```
victorges, line 508: Hmm, the only problem here is that this logic depends on some deprecated fields on the old child stream objects that were used as sessions and recordings.

I think we need some further compatibility logic here. Specifically, here we should probably identify that the `session` is from the old pipeline and treat it the same way as before, by fetching the `db.stream` child stream instead. I'm 99% sure that we have a guarantee that every `db.session` from the old pipeline had a corresponding `db.stream` object with the same ID, so here we just need to fetch it using the same ID. So the full thing looks like:
 - check if it is a new or old session type. If we don't have that yet, we should probably create a field for that, maybe even just a `version` wdyt? Would prefer to use something implicit though to avoid handling an additional version field.
 - if it's a new session, call `withRecordingFields` on it and return (this whole `raw` logic is only valid on old old pipeline, when we merged child streams by timestamp)
 - if it's an old session, fetch the child stream and go through this old logic of `raw`. Also call `withRecordingFields` but I'd even just fake that it is a session with an `as any` just to simplify things here (if the code works).

WDYT? I wrote a lot but I think the actual change is small, just trying to explain it.

packages/api/src/controllers/stream.ts
```diff
@@ -734,6 +763,7 @@ app.post(
       objectStoreId: stream.objectStoreId,
       record,
       recordObjectStoreId,
+      sessionId,
```
victorges, line 767: I wonder if we should create the session before the child stream just not to risk creating a reference to something that doesn't exist.

packages/api/src/controllers/stream.ts
```diff
@@ -1062,19 +1102,20 @@ function publishRecordingStartedHook(
...
-    payload: { session: toExternalSession(session, ingest) },
+    payload: { session: await toExternalSession(config, session, ingest) },
   });
 }
 
 /**
  * We don't actually send the webhook here, but schedule an event after a timeout.
  */
-function publishDelayedRecordingReadyHook(
+async function publishDelayedRecordingReadyHook(
```
victorges, line 1114: How much work do you think it would be to make the `recording.ready` hook only fire when the recording asset is created? Might end up with some convoluted dependencies between VOD and webhook flows, but maybe not (maybe that dep already exists after all). I don't think it needs to be included in this PR, but I think it would be a good followup to keep a sensible `recording.ready` webhook (that means what it says). WDYT?

packages/api/src/webhooks/cannon.ts
```diff
@@ -514,23 +515,19 @@ export default class WebhookCannon {
...
     }
 
-    const res = await this.db.stream.update(
-      [sql`id = ${sessionId}`, sql`data->>'isActive' != 'false'`],
-      { isActive: false },
-      { throwIfEmpty: false }
-    );
-    if (res.rowCount < 1) {
+    const res = await this.db.asset.get(sessionId);
+    if (res) {
```
victorges, line 522: Can you make sure to gracefully handle duplicate primary key errors from the `createAsset` function? Otherwise we may still be subject to race conditions here

victorges (approved):
LGTM!! 🎉 

tho the tests are still failing

packages/api/src/controllers/asset.ts
leszko: Updated in 290b6c98cd9cc1edfd2ec7686f137b7119199bd9

packages/api/src/controllers/stream.ts
leszko: I think we would also need to check `failed`, so it's actually the same.

packages/api/src/controllers/stream.ts
leszko: I think, we can't use `playbackRecordingId` here, because it's set in task-runner only if the recording succeeded. For failed assets we don't have `playbackRecordingId`.

packages/api/src/controllers/stream.ts
leszko: I think it's ok. We first create a session and then create child stream which has a reference to the session. So we should never have a child stream with a reference to the session that does not exist.

packages/api/src/webhooks/cannon.ts
leszko: Updated in dbf928ac0af75d35bbb264284e3e5a8075cc3ead

packages/api/src/controllers/stream.ts
leszko: Good idea. Let's do it as a follow-up. Created a linear ticket: https://linear.app/livepeer/issue/VID-153/studio-trigger-webhook-only-if-recording-is-processed

packages/api/src/controllers/stream.ts
leszko: Ahh, good point. I think then I need to add this `version` property, because otherwise it'd be very inefficient to check if we should use `stream` or `session`.

Updated in 4b8f4234fd3fd78675b4f97da1b2e6a3ba6df415

leszko (commented):
Addressed all the comments.

leszko (approved):

PR #1693, opened 2023-04-19
Set up an endpoint for meeting rooms

mjh1:

Provides a new endpoint to be able to create meeting rooms for multi-participant live streams.

Ran the app locally with `yarn dev` and tested with different requests to the new endpoint. There is also a unit test provided.

**Checklist**

### DIFF
--- a/packages/api/src/controllers/room.test.ts
+++ b/packages/api/src/controllers/room.test.ts
@@ -1,0 +1,69 @@
+import serverPromise, { TestServer } from "../test-server";
+import { TestClient, clearDatabase, setupUsers } from "../test-helpers";
+import { v4 as uuid } from "uuid";
+import { User } from "../schema/types";
+import { db } from "../store";
+import { RoomServiceClient } from "livekit-server-sdk";
+import { CreateOptions } from "livekit-server-sdk/dist/RoomServiceClient";
+jest.mock("livekit-server-sdk");
+const MockedRoomServiceClient =
+  RoomServiceClient as jest.Mock<RoomServiceClient>;
+
+let server: TestServer;
+let mockAdminUserInput: User;
+let mockNonAdminUserInput: User;
+
+beforeAll(async () => {
+  server = await serverPromise;
+
+  mockAdminUserInput = {
+    email: "user_admin@gmail.com",
+    password: "x".repeat(64),
+  };
+
+  mockNonAdminUserInput = {
+    email: "user_non_admin@gmail.com",
+    password: "y".repeat(64),
+  };
+});
+
+afterEach(async () => {
+  await clearDatabase(server);
+});
+
+describe("controllers/room", () => {
+  let client: TestClient;
+  let adminUser: User;
+  let adminApiKey: string;
+  let nonAdminUser: User;
+  let nonAdminToken: string;
+
+  const createStub = jest.fn((options: CreateOptions) => {
+    return Promise.resolve();
+  });
+  beforeEach(async () => {
+    ({ client, adminUser, adminApiKey, nonAdminUser, nonAdminToken } =
+      await setupUsers(server, mockAdminUserInput, mockNonAdminUserInput));
+    client.jwtAuth = nonAdminToken;
+
+    MockedRoomServiceClient.mockImplementation(() => {
+      return {
+        createRoom: createStub,
+      };
+    });
+  });
+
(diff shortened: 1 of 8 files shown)


packages/api/src/controllers/room.ts
```diff
@@ -0,0 +1,34 @@
...
+const app = Router();
+
+app.post("/", authorizer({}), async (req, res) => {
+  if (req.config.livekitHost == "") {
+    res.status(400);
+    return res.send("not enabled");
+  }
+
+  const id = uuid();
+  const svc = new RoomServiceClient(
```
victorges, line 15: Is it ok to create one of these on every request? If there are definite gains in having a single instance lmk and we can work sth out, like putting it on the `req` object as well

packages/api/src/controllers/room.ts
```diff
@@ -0,0 +1,34 @@
...
+    req.config.livekitApiKey,
+    req.config.livekitSecret
+  );
+  const opts = {
+    name: id,
+    // timeout in seconds
+    emptyTimeout: 30 * 60,
+    maxParticipants: 2,
+  };
+  const room = await svc.createRoom(opts);
```
victorges, line 26: Do we not need to create a stream anymore? So there will be a separate call to have the room be livestreamed, is that it?

packages/api/src/schema/schema.yaml
```diff
@@ -3289,3 +3289,14 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/error"
+
+  /room:
+    post:
+      description: create a new room
+      responses:
```
victorges, line 3296: Since you're adding it here, mind documenting the success respnose schema? can just do an inline schema here with an `id` field etc

packages/api/src/controllers/room.test.ts
```diff
@@ -0,0 +1,62 @@
+import serverPromise, { TestServer } from "../test-server";
+import { TestClient, clearDatabase, setupUsers } from "../test-helpers";
+import { User } from "../schema/types";
+import { RoomServiceClient } from "livekit-server-sdk";
+jest.mock("livekit-server-sdk");
+const MockedRoomServiceClient =
+  RoomServiceClient as jest.Mock<RoomServiceClient>;
```
victorges, line 7: Maybe these should be the first lines in this file, just in case anyone changes the impl to use the mocked stuff statically (or this is really magic and I don't know how it works)

packages/api/src/controllers/room.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      await setupUsers(server, mockAdminUserInput, mockNonAdminUserInput));
+    client.jwtAuth = nonAdminToken;
+  });
+
+  describe("room creation", () => {
+    it("should create a room", async () => {
+      let roomServiceClient = new RoomServiceClient("");
+      const mockCreateRoom = jest.spyOn(roomServiceClient, "createRoom");
+      MockedRoomServiceClient.mockReturnValueOnce(roomServiceClient);
+      mockCreateRoom.mockReturnValueOnce(Promise.resolve(undefined));
```
victorges, line 49: Cool! Didn't even know jest had that mocking builtin

victorges (approved):
LGTM!

packages/api/src/controllers/room.ts
mjh1: Yeah I wondered that too, when I asked livekit they said it should be ok:
> either way works! it's pretty inexpensive to create each time

Sound ok to you or should we press further?

victorges: Sounds good for now!

packages/api/src/controllers/room.ts
mjh1: That's right, since Hunter mentioned that not every room will have a live stream i've kept it separate.

packages/api/src/controllers/room.test.ts
mjh1: What do you mean by using the mocked stuff statically?

victorges: e.g.: if we moved the livekit client creation to the "file root" or server creation level, not creating one on every request

That could run before the mocks are setup and get the wrong result. Not a huge priority since it works now though, it's fine as is

PR #1694, opened 2023-04-19
Added webrtc playback endpoints

0xcadams:
Fixes https://linear.app/livepeer/issue/DX-156/add-new-endpoints-to-the-ui-for-playback


packages/api/src/controllers/playback.ts
```diff
@@ -66,7 +71,13 @@ function newPlaybackInfo(
     type: "html5/application/vnd.apple.mpegurl",
     url: hlsUrl,
   });
-
+  if (webRtcUrl) {
+    playbackInfo.meta.source.push({
+      hrn: "WebRTC (VP8)",
+      type: "html5/video/vp8",
```
victorges: Does this mean that Mist will be transcoding everything to VP8? 😮

victorges (approved):

packages/api/src/controllers/playback.ts
0xcadams: I am not sure what the `hrn` should be here. What would you include instead?

packages/api/src/controllers/playback.ts
```diff
@@ -1,7 +1,8 @@
 import { Request, Router } from "express";
 import { db } from "../store";
 import {
-  getPlaybackUrl as streamPlaybackUrl,
+  getHLSPlaybackUrl as getHLSPlaybackUrl,
+  getWebRTCPlaybackUrl as getWebRTCPlaybackUrl,
```
victorges: Do we still need the rename?

packages/api/src/controllers/playback.ts
```diff
@@ -35,8 +36,11 @@ type PlaybackInfo = {
     live?: 0 | 1;
     playbackPolicy?: Asset["playbackPolicy"] | Stream["playbackPolicy"];
     source: {
-      hrn: "HLS (TS)" | "MP4";
-      type: "html5/application/vnd.apple.mpegurl" | "html5/video/mp4";
+      hrn: "HLS (TS)" | "MP4" | "WebRTC (VP8)";
+      type:
+        | "html5/application/vnd.apple.mpegurl"
+        | "html5/video/mp4"
+        | "html5/video/vp8";
```
victorges: Wish there was something more specific we could put here, but couldn't find anything myself

packages/api/src/controllers/playback.ts
victorges: Oh yeah I had just made that same comment in the other file. I don't know what else we could set for the `hrn`, seems like we're restricted to MIME types.

The confusion about VP8 is probably something else tho, assuming that you got these values like `WebRTC (VP8)` from Mist. Curious about how that video ends up in VP8 codec if it goes in/out of go-livepeer as H264/5

victorges (approved):
relgtm!

packages/api/src/controllers/playback.ts
0xcadams: I just changed this - does H264 make more sense? I checked and this is what mist is using

PR #1697, opened 2023-04-24
Update default configuration of stream-info-service

leszko:
With the default catalyst setup `localhost` does not work, but `127.0.0.1` 🤔 

Let's just change it to simplify the local setup.


victorges (approved):
:shipit:

PR #1699, opened 2023-04-24
Save tracking info for rooms

mjh1:

I've created a database table to be able to keep track of rooms. I've implemented a webhook receiver so that we can store information like which participants have joined and what tracks they have published.

- Modified existing unit test to cover new functionality.
- Tested webhooks locally to observe the expected updated info in the room table.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/room.ts
+++ b/packages/api/src/controllers/room.ts
@@ -1,7 +1,11 @@
-import { Router } from "express";
-import { RoomServiceClient } from "livekit-server-sdk";
+import express, { Router } from "express";
+import { RoomServiceClient, WebhookReceiver } from "livekit-server-sdk";
 import { v4 as uuid } from "uuid";
 import { authorizer } from "../middleware";
+import { db } from "../store";
+import { ForbiddenError } from "../store/errors";
+import { Room } from "../schema/types";
+import { WithID } from "../store/types";
 
 const app = Router();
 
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/room.ts
```diff
@@ -17,18 +21,119 @@ app.post("/", authorizer({}), async (req, res) => {
...
+  const room = await svc.createRoom({
     name: id,
     // timeout in seconds
-    emptyTimeout: 30 * 60,
-    maxParticipants: 2,
-  };
-  const room = await svc.createRoom(opts);
+    emptyTimeout: 15 * 60,
+    maxParticipants: 10,
+    metadata: `{"userId": "${req.user.id}"}`,
```
victorges: Nice to have the metadata! As a nit, you could write a syntax-safe JSON here as well:
```suggestion
    metadata: JSON.stringify({userId: req.user.id}),
```

packages/api/src/controllers/room.ts
```diff
@@ -17,18 +21,119 @@ app.post("/", authorizer({}), async (req, res) => {
...
+  for (const key in room.participants) {
+    room.participants[key].tracksPublished = undefined;
+  }
+  room.events = undefined;
+  return room;
+}
+
+// Implement a webhook handler to receive webhooks from Livekit to update our state with room and participant details.
+app.use("/webhook", express.raw({ type: "*/*" }));
+app.post("/webhook", async (req, res) => {
```
victorges: **nit:** I think this would have the same effect (unless you want the middleware on other methods not only POST)
```suggestion
app.post("/webhook", express.raw({ type: "*/*" }), async (req, res) => {
```

packages/api/src/controllers/room.ts
```diff
@@ -17,18 +21,119 @@ app.post("/", authorizer({}), async (req, res) => {
...
+  const receiver = new WebhookReceiver(
+    req.config.livekitApiKey,
+    req.config.livekitSecret
+  );
+  console.log("livekit webhook auth", req.get("Authorization"));
```
victorges: Should probably remove/comment this log before deploy

packages/api/src/controllers/room.ts
```diff
@@ -17,18 +21,119 @@ app.post("/", authorizer({}), async (req, res) => {
...
+    case "participant_left":
+      const roomId = event.room.name;
+      const svc = new RoomServiceClient(
+        req.config.livekitHost,
+        req.config.livekitApiKey,
+        req.config.livekitSecret
+      );
+      const participants = await svc.listParticipants(roomId);
+
+      room = await db.room.get(roomId);
```
victorges: Maybe double check this room exists and return a 4xx error if it doesn't (or anything that LiveKit won't retry)

packages/api/src/schema/schema.yaml
```diff
@@ -3294,6 +3358,25 @@ paths:
     post:
       description: create a new room
       responses:
+        200:
+          description: Success
+          content:
+            application/json:
+              $ref: "#/components/schemas/create-room-response"
```
victorges: If it's only for $refing it here, I'd say we can just inline it (and remove the top-level component)
```suggestion
              additionalProperties: false
              properties:
                id:
                  type: string
```

packages/api/src/schema/schema.yaml
```diff
@@ -2572,6 +2572,70 @@ components:
...
+      table: rooms
+      required:
+        - id
+        - participants
+        - events
+      properties:
+        id:
+          type: string
+          unique: true
+          index: true
```
victorges: Actually the `id` field is the only exception! We don't need any index here.
```suggestion
```

In the database, we have a separate column for the ID which is also the table primary key, and when querying by it we always check that table instead of the `id` in the JSON column. The index here would be redundant/unused.

packages/api/src/schema/schema.yaml
```diff
@@ -2572,6 +2572,70 @@ components:
...
+          readOnly: true
+        userId:
+          type: string
+          readOnly: true
+          example: 66E2161C-7670-4D05-B71D-DA2D6979556F
+          index: true
```
victorges, line 2591: You need to add `room` to this list here: https://github.com/livepeer/livepeer-com/blob/a326c2f31bfd7638390ba703bcf7f1ca6a7cad1e/packages/api/src/store/table.ts#L352

Otherwise the indexes won't really be created. This is a tech debt we've been carrying since the time we had some issues with indexes on the `stream` table which is huge and gets a lot of updates.

packages/api/src/schema/schema.yaml
```diff
@@ -2572,6 +2572,70 @@ components:
           type: number
           example: 200000
 
+    create-room-response:
+      additionalProperties: false
+      properties:
+        id:
+          type: string
+
+    room:
```
victorges, line 2575: Would you mind adding documentation (`description`) to the user-facing fields in this object?

packages/api/src/controllers/room.ts
```diff
@@ -17,18 +21,119 @@ app.post("/", authorizer({}), async (req, res) => {
...
+});
+
+function toExternalRoom(room: Room, isAdmin = false) {
+  if (isAdmin) {
+    return room;
+  }
+  room.userId = undefined;
+  for (const key in room.participants) {
+    room.participants[key].tracksPublished = undefined;
+  }
```
victorges, line 70: Maybe consider hiding the entire `participants` field. Do users really need access to all that information? Might be easier to start by using it only internally for tracking/billing and consider exposing only parts of it that users may need.

packages/api/src/controllers/room.ts
```diff
@@ -17,18 +21,119 @@ app.post("/", authorizer({}), async (req, res) => {
...
+            mimeType: track.mimeType,
+            timestamp: new Date().getTime(),
+          };
+        }
+      }
+      await db.room.replace(room);
+      break;
+    case "room_started":
+    case "room_finished":
+      room = await db.room.get(event.room.name);
```
victorges: Same for checking if the room exists. Maybe we want this step outside the `switch`?

victorges (approved):
LGTM

packages/api/src/controllers/room.ts
mjh1: I think it will be helpful to give them at least the IDs and names of the participants so that they would easily be able to see who has joined and remove people if needed (api coming in next PR). I will tidy up the remaining fields too in the next PR.

PR #1700, opened 2023-04-25
api: Add `creatorId` field to assets

victorges:

This adds the `creatorId` field to assets, which right now is an unverified value
provided by the caller. We store it in an object with a `type` and `value` for 
future-proofness.

 - Add `creatorId` field to `asset` as an object
 - Add `creatorId` field to `new-asset-payload` as the same object or also just a string for ergonomy
 - Map the field to the asset object on creation

`yarn test`

Implements DAT-65

**Checklist**

### DIFF
--- a/packages/api/src/schema/schema.yaml
+++ b/packages/api/src/schema/schema.yaml
@@ -1438,6 +1438,12 @@ components:
           example: true
         playbackPolicy:
           $ref: "#/components/schemas/playback-policy"
+        creatorId:
+          oneOf:
+            - $ref: "#/components/schemas/asset/properties/creatorId"
+            - type: string
+              description:
+                Developer-managed (unverified) creator ID of the asset
         storage:
           additionalProperties: false
           properties:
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/schema.yaml
```diff
@@ -1426,6 +1438,12 @@ components:
           example: true
         playbackPolicy:
           $ref: "#/components/schemas/playback-policy"
+        creatorId:
+          oneOf:
+            - $ref: "#/components/schemas/asset/properties/creatorId"
+            - type: string
+              description:
+                Developer-managed (unverified) creator ID of the asset
```
victorges: Wonder if we need this helper syntax [1] or if it would be better to have developers explicitly specify the `"unverified"` string just so it's on their face lol

[1] (this means that when creating an asset one could not only specify `creatorId: {type:unverified, value: foo}` but also just `creatorId: foo`)

0xcadams: Yeah, I like the helper syntax. Especially since we only support `unverified` currently, it seems clunky to require that to be passed in.

0xcadams (approved):
🚢

PR #1705, opened 2023-04-27
Add remaining room APIs

mjh1:

Add in remaining room APIs for the multiparticipant feature

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/room.test.ts
+++ b/packages/api/src/controllers/room.test.ts
@@ -1,10 +1,11 @@
 import serverPromise, { TestServer } from "../test-server";
 import { TestClient, clearDatabase, setupUsers } from "../test-helpers";
 import { User } from "../schema/types";
-import { RoomServiceClient } from "livekit-server-sdk";
+import { EgressClient, RoomServiceClient } from "livekit-server-sdk";
 jest.mock("livekit-server-sdk");
 const MockedRoomServiceClient =
   RoomServiceClient as jest.Mock<RoomServiceClient>;
+const MockedEgressClient = EgressClient as jest.Mock<EgressClient>;
 
 let server: TestServer;
 let mockAdminUserInput: User;
@@ -34,22 +35,24 @@ describe("controllers/room", () => {
   let adminApiKey: string;
   let nonAdminUser: User;
   let nonAdminToken: string;
+  let roomServiceClient = new RoomServiceClient("");
+  let mockCreateRoom: jest.SpyInstance;
+  let egressClient = new EgressClient("");
 
   beforeEach(async () => {
     ({ client, adminUser, adminApiKey, nonAdminUser, nonAdminToken } =
       await setupUsers(server, mockAdminUserInput, mockNonAdminUserInput));
     client.jwtAuth = nonAdminToken;
+
+    MockedRoomServiceClient.mockReturnValue(roomServiceClient);
+    mockCreateRoom = jest.spyOn(roomServiceClient, "createRoom");
+    mockCreateRoom.mockReturnValueOnce(Promise.resolve(undefined));
+    MockedEgressClient.mockReturnValueOnce(egressClient);
   });
 
   describe("room creation", () => {
-    it("should create a room", async () => {
-      let roomServiceClient = new RoomServiceClient("");
-      const mockCreateRoom = jest.spyOn(roomServiceClient, "createRoom");
-      MockedRoomServiceClient.mockReturnValueOnce(roomServiceClient);
-      mockCreateRoom.mockReturnValueOnce(Promise.resolve(undefined));
-
+    it("should create and delete rooms", async () => {
       let res = await client.post(`/room`);
-
       expect(res.status).toBe(201);
       const roomRes = await res.json();
       expect(roomRes.id).toBeDefined();
@@ -64,6 +67,60 @@ describe("controllers/room", () => {
       const getRoomRes = await res.json();
       expect(getRoomRes.id).toBe(roomRes.id);
       expect(getRoomRes.userId).toBeUndefined();
+
+      const mockDeleteRoom = jest.spyOn(roomServiceClient, "deleteRoom");
+      res = await client.delete(`/room/${roomRes.id}`);
+      expect(res.status).toBe(204);
+      expect(mockDeleteRoom).toHaveBeenCalledTimes(1);
+      expect(mockDeleteRoom.mock.calls[0][0]).toBe(roomRes.id);
+
+      // should no longer exist
+      res = await client.get(`/room/${roomRes.id}`);
+      expect(res.status).toBe(404);
+    });
+
+    it("should start and stop egress", async () => {
+      let res = await client.post(`/room`);
+      expect(res.status).toBe(201);
+      const roomRes = await res.json();
+      expect(roomRes.id).toBeDefined();
+
+      let mockStartEgress = jest.spyOn(
+        egressClient,
+        "startRoomCompositeEgress"
+      );
+      mockStartEgress.mockReturnValueOnce(
+        Promise.resolve({
+          egressId: "egress-id",
+        })
+      );
+      let mockStopEgress = jest.spyOn(egressClient, "stopEgress");
+      mockStopEgress.mockReturnValueOnce(Promise.resolve(undefined));
+
+      res = await client.delete(`/room/${roomRes.id}/egress`);
+      expect(res.status).toBe(400);
+
+      res = await client.post(`/room/${roomRes.id}/egress`, {
+        rtmpURL: "rtmp",
+      });
+      expect(res.status).toBe(204);
+
+      // already started so should 400
+      res = await client.post(`/room/${roomRes.id}/egress`, {
+        rtmpURL: "rtmp",
+      });
+      expect(res.status).toBe(400);
+
+      res = await client.delete(`/room/${roomRes.id}/egress`);
+      expect(res.status).toBe(204);
+
+      // already stopped so should 400
+      res = await client.delete(`/room/${roomRes.id}/egress`);
+      expect(res.status).toBe(400);
+
+      expect(mockStartEgress).toHaveBeenCalledTimes(1);
+      expect(mockStopEgress).toHaveBeenCalledTimes(1);
+      expect(mockStopEgress.mock.calls[0][0]).toBe("egress-id");
     });
   });
 });
--- a/packages/api/src/controllers/room.test.ts
+++ b/packages/api/src/controllers/room.test.ts
@@ -35,19 +35,22 @@ describe("controllers/room", () => {
   let adminApiKey: string;
   let nonAdminUser: User;
   let nonAdminToken: string;
-  let roomServiceClient = new RoomServiceClient("");
+  let roomServiceClient;
   let mockCreateRoom: jest.SpyInstance;
-  let egressClient = new EgressClient("");
+  let egressClient;
 
   beforeEach(async () => {
     ({ client, adminUser, adminApiKey, nonAdminUser, nonAdminToken } =
       await setupUsers(server, mockAdminUserInput, mockNonAdminUserInput));
     client.jwtAuth = nonAdminToken;
 
+    roomServiceClient = new RoomServiceClient("");
+    egressClient = new EgressClient("");
+
     MockedRoomServiceClient.mockReturnValue(roomServiceClient);
     mockCreateRoom = jest.spyOn(roomServiceClient, "createRoom");
     mockCreateRoom.mockReturnValueOnce(Promise.resolve(undefined));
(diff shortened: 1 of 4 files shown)


packages/api/src/controllers/room.test.ts
```diff
@@ -34,22 +35,24 @@ describe("controllers/room", () => {
   let adminApiKey: string;
   let nonAdminUser: User;
   let nonAdminToken: string;
+  let roomServiceClient = new RoomServiceClient("");
+  let mockCreateRoom: jest.SpyInstance;
+  let egressClient = new EgressClient("");
```
victorges: To keep tests independent it might be better to only declare the variables here, but initialize them on `beforeEach`

packages/api/src/controllers/room.test.ts
```diff
@@ -64,6 +67,60 @@ describe("controllers/room", () => {
...
+        })
+      );
+      let mockStopEgress = jest.spyOn(egressClient, "stopEgress");
+      mockStopEgress.mockReturnValueOnce(Promise.resolve(undefined));
+
+      res = await client.delete(`/room/${roomRes.id}/egress`);
+      expect(res.status).toBe(400);
+
+      res = await client.post(`/room/${roomRes.id}/egress`, {
+        rtmpURL: "rtmp",
```
victorges: Could we allow at least sending a stream ID here, so it's not something alien to the rest of our API? (if you do so, make sure to check the stream belongs to the same user as the room / as the caller)

That should also be the recommended way for egress configuration, with an optional way of specifying a "third-party" stream. wdyt?

packages/api/src/controllers/room.test.ts
```diff
@@ -64,6 +67,60 @@ describe("controllers/room", () => {
...
+      });
+      expect(res.status).toBe(204);
+
+      // already started so should 400
+      res = await client.post(`/room/${roomRes.id}/egress`, {
+        rtmpURL: "rtmp",
+      });
+      expect(res.status).toBe(400);
```
victorges, line 148: Should it allow multiple (different) egresses to be added?

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
   }
-  room.userId = undefined;
-  for (const key in room.participants) {
-    room.participants[key].tracksPublished = undefined;
+  const roomForResp = {
+    id: room.id,
+    updatedAt: room.updatedAt,
+    createdAt: room.createdAt,
+    participants: room.participants,
+  };
```
victorges, line 86: Are this all keys in the object? If so, this helper syntax might be your friend
```suggestion
  const roomForResp = { ...room };
```

If you are omitting some fields then nevermind me.

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
   }
-  room.events = undefined;
-  return room;
+  return roomForResp;
 }
 
+app.delete("/:roomId", authorizer({}), async (req, res) => {
+  const room = await db.room.get(req.params.roomId);
+
+  if (!req.user.admin && req.user.id !== room.userId) {
```
victorges: Should check `deleted` field as well perhaps?

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+  const svc = new RoomServiceClient(
+    req.config.livekitHost,
+    req.config.livekitApiKey,
+    req.config.livekitSecret
+  );
+  await svc.deleteRoom(req.params.roomId);
+
+  room.deleted = true;
+  room.deletedAt = new Date().getTime();
+  await db.room.replace(room);
```
victorges: I think an isolated update might be cleaner/safer in general than full replaces (and maybe even cheaper, given this room object can get pretty big). i.e. wdytof
```suggestion
  await db.room.update(room.id, { deleted: true, deletedAt: Date.now() });
```

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+});
+
+app.post(
+  "/:roomId/egress",
+  authorizer({}),
+  validatePost("room-egress-payload"),
+  async (req, res) => {
+    const room = await db.room.get(req.params.roomId);
+
+    if (!req.user.admin && req.user.id !== room.userId) {
```
victorges: Need to check the `deleted` field as well

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+        layout: "speaker-dark",
+        encodingOptions: {
+          keyFrameInterval: 2,
+        },
+      }
+    );
+    console.log("egress started", info);
+    room.egressId = info.egressId;
+    room.updatedAt = new Date().getTime();
+    await db.room.replace(room);
```
victorges: Same regarding preference for `update` instead of full `replace`. These are just optional suggestions tho, will avoid making the same comment in other places below.

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+    }
+
+    const egressClient = new EgressClient(
+      req.config.livekitHost,
+      req.config.livekitApiKey,
+      req.config.livekitSecret
+    );
+    const output: StreamOutput = {
+      protocol: StreamProtocol.RTMP,
+      urls: [req.body.rtmpURL],
```
victorges: Yeah I think this URL should come from an existing `stream` object, being specified as a `streamId` instead. Optionally we may want to allow specifying custom URLs, but IMO we should integrate this with the rest of our API.

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+    room.updatedAt = new Date().getTime();
+    await db.room.replace(room);
+    res.status(204).end();
+  }
+);
+
+app.delete("/:roomId/egress", authorizer({}), async (req, res) => {
+  const room = await db.room.get(req.params.roomId);
+
+  if (!req.user.admin && req.user.id !== room.userId) {
```
victorges: Same for `deleted` (maybe different error message (maybe even create a local helper in this file like `getExistingRoom` lol, I always hate these repetitions in these controllers))

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+  const info = await egressClient.stopEgress(room.egressId);
+  console.log("egress stopped", info);
+  room.egressId = undefined;
+  room.updatedAt = new Date().getTime();
+  await db.room.replace(room);
+  res.status(204).end();
+});
+
+app.post(
+  "/:roomId/user",
```
victorges, line 178: Can you add tests for this?

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+  authorizer({}),
+  validatePost("room-user-payload"),
+  async (req, res) => {
+    const room = await db.room.get(req.params.roomId);
+
+    if (!req.user.admin && req.user.id !== room.userId) {
```
victorges: Same for `deleted`

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+      }
+    );
+    at.addGrant({ roomJoin: true, room: req.params.roomId });
+    const token = at.toJwt();
+
+    res.status(201);
+    res.json({
+      id: id,
+      joinUrl:
+        "https://meet.livekit.io/custom?liveKitUrl=" +
```
victorges: Are we hosting a custom version of this ourselves at some point? Might wanna make this domain configurable already if so.

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+        "&token=" +
+        token,
+    });
+  }
+);
+
+app.delete("/:roomId/user/:userId", authorizer({}), async (req, res) => {
+  const room = await db.room.get(req.params.roomId);
+
+  if (!req.user.admin && req.user.id !== room.userId) {
```
victorges: Same for deleted

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+    res.status(201);
+    res.json({
+      id: id,
+      joinUrl:
+        "https://meet.livekit.io/custom?liveKitUrl=" +
+        req.config.livekitHost +
+        "&token=" +
+        token,
+    });
+  }
```
victorges, line 208: Don't we need to edit the `participants` field in the room here? Or is it only updated on the webhooks?

packages/api/src/controllers/room.ts
```diff
@@ -64,14 +72,163 @@ function toExternalRoom(room: Room, isAdmin = false) {
...
+);
+
+app.delete("/:roomId/user/:userId", authorizer({}), async (req, res) => {
```
victorges: **nit:** Would use `:participantId` for that path param not to be confused with the `userId` that has another meaning in these API objects (the owner of the room, in this case)

packages/api/src/controllers/room.ts
```diff
@@ -96,7 +253,7 @@ app.post("/webhook", express.raw({ type: "*/*" }), async (req, res) => {
     throw new BadRequestError(`no room name on event`);
   }
   const room = await db.room.get(roomId);
-  if (!room) {
+  if (!room || room.deleted) {
     throw new NotFoundError(`room not found`);
   }
```
victorges: Hmm, in case of webhooks maybe we want to keep tracking info that we get after the room was deleted, so possibly not check the `deleted` field here. WDYT?

packages/api/src/controllers/room.ts
```diff
@@ -37,6 +44,7 @@ app.post("/", authorizer({}), async (req, res) => {
   await db.room.create({
     id: id,
     userId: req.user.id,
+    createdAt: new Date().getTime(),
```
victorges: **nit:** `Date.now()` is a short-form for that

packages/api/src/schema/schema.yaml
```diff
@@ -3369,7 +3407,14 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/error"
+  /room/{room}:
     get:
+      parameters:
+        - name: room
```
victorges: Maybe `roomId` or just `id` on these?

victorges (approved):
LGTM with nits, the only a little stronger preference is for the egress API to receive a stream ID instead (or primarily) of a raw RTMP URL. Still not a required change tho.

packages/api/src/controllers/room.test.ts
mjh1: We don't need that for the current design so for now at least I think we should keep it simple and just allow one.

packages/api/src/controllers/room.ts
mjh1: We're omitting internal stuff like userId.

packages/api/src/controllers/room.ts
mjh1: oh nice yeah didn't realise we could do that 👍

packages/api/src/controllers/room.ts
mjh1: Yeah that makes sense, and tbh I think if the room doesn't exist I'd want livekit to retry sending, from the testing I've done I don't think this will be a lot of failures but I'm just thinking of when the room starts up there will be a brief window when we may not have created the db item yet but it has been created on livekit's side. So I'll change this to a 5xx rather than 404.

packages/api/src/controllers/room.ts
mjh1: I was thinking we only really need to track who has actually joined the room (since we only start being charged at that point). But I guess it might be useful to keep track of when and what tokens have been generated, for debugging. WDYT?

packages/api/src/controllers/room.test.ts
mjh1: Yeah you're right, it should definitely just accept a streamID, and the third party stream would be something to consider later on down the line, for now we've only discussed streaming to livepeer.

PR #1708, opened 2023-04-27
api: Default to HTTPS protocol for frontend links

victorges:

We were creating http URLs that break on Brave. Don't know when x-forwarded-proto
stopped being set but I think the default should be HTTP anyway.

http->https

Create an account and make sure email verification is HTTPS

DX-184

**Checklist**


adamsoffer (approved):
LGTM

PR #1709, opened 2023-04-28
chore: bump lerna

iameli:
i was running into problems with an outdated node-gyp with a transitive lerna dependency. and python 3.11. looks like this fixes it!

https://github.com/nodejs/node-gyp/issues/2219


0xcadams:
lerna's gotta go, this is stale and we will rework lerna

PR #1714, opened 2023-05-03
api/auth: Add capacity to authorize via playback ID

victorges:
This is to increment the `/api/auth` API, used by external services that need to
build on top of the API authorization, to also support authorizing access to a 
given playback ID. This playback ID can not only be the canonical playback ID of
assets/streams, but can also be the "aliased" IDs like a dStorage ID.

This will be used in the viewership v2 API when it is accessed with a playback ID filter.

 - Refactor "playback info" implementation to have a separate "get resource by playback ID" function
 - Add `x-livepeer-playback-id` header support to auth API
 - Parallelize the authorization checks just in case 
   - (also need to optimize the `getResourceByPlaybackId` at some point)

`yarn test`

Implements API-50

**Checklist**

### DIFF
--- a/packages/api/src/controllers/auth.ts
+++ b/packages/api/src/controllers/auth.ts
@@ -2,12 +2,12 @@
  * Special controller for forwarding all incoming requests to a geolocated API region
  */
 
-import { Request, Router } from "express";
+import { Request, Response, Router } from "express";
 import { authorizer } from "../middleware";
 import { db } from "../store";
 import { ForbiddenError, NotFoundError } from "../store/errors";
 import Table from "../store/table";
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/auth.ts
```diff
@@ -16,30 +17,51 @@ type UserOwnedObj = { id: string; deleted?: boolean; userId?: string };
 async function checkUserOwned(
   req: Request,
   headerName: string,
-  table: Table<UserOwnedObj>
+  getter: (id: string) => Promise<UserOwnedObj>
```
leszko, line 20: Nice abstraction!

leszko (approved):
LGTM 👍

PR #1715, opened 2023-05-04
encryption: public key endpoint & encrypted key in upload request

gioelecerati:

Currently left as experiment

- Added new endpoint /public-key to retrieve the public key.
- Added encryptedKey field in the encryption schema and related validation.
- Updated the asset.ts to validate base64 encoding for the encrypted key.
- Added the isValidBase64 function in helpers.ts to check if a string is valid base64.
- Updated parse-cli.ts to add the catalystBaseUrl option.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/app-router.ts
+++ b/packages/api/src/app-router.ts
@@ -56,7 +56,6 @@ export default async function makeApp(params: CliArgs) {
     postgresUrl,
     postgresReplicaUrl,
     frontendDomain = "livepeer.studio",
-    catalystBaseUrl = "http://localhost:7979",
     supportAddr,
     sendgridTemplateId,
     sendgridApiKey,
@@ -163,7 +162,6 @@ export default async function makeApp(params: CliArgs) {
     req.queue = queue;
     req.taskScheduler = taskScheduler;
     req.stripe = stripe;
-    req.catalystBaseUrl = catalystBaseUrl;
     next();
   });
   app.use(
--- a/packages/api/src/app-router.ts
+++ b/packages/api/src/app-router.ts
@@ -56,6 +56,7 @@ export default async function makeApp(params: CliArgs) {
     postgresUrl,
     postgresReplicaUrl,
     frontendDomain = "livepeer.studio",
+    catalystBaseUrl = "http://localhost:7979",
     supportAddr,
     sendgridTemplateId,
     sendgridApiKey,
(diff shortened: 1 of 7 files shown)


packages/api/src/app-router.ts
```diff
@@ -159,6 +160,7 @@ export default async function makeApp(params: CliArgs) {
     req.queue = queue;
     req.taskScheduler = taskScheduler;
     req.stripe = stripe;
+    req.catalystBaseUrl = catalystBaseUrl;
```
victorges: **nit:** I think this request object is a bit messy today, we lazily use it to store singletons, configs, etc. So I'd generally prefer to avoid adding more stuff to it. That might be just my opinion tho.

Not sure if you noticed, but we already have the `req.config` field from which we can access absolutely any CLI parameter. So WDYT of using `req.config.catalystBaseUrl` instead of creating a new request extension field?

packages/api/src/controllers/access-control.ts
```diff
@@ -159,66 +222,32 @@ app.post(
...
-async function fireGateWebhook(
-  webhook: DBWebhook,
-  plabackPolicy: PlaybackPolicy,
-  accessKey: string
-) {
-  let timestamp = Date.now();
-  let params = {
-    method: "POST",
+app.get("/public-key", async (req, res) => {
+  let catalystBaseUrl = req.catalystBaseUrl;
```
victorges: Yeah this can be
```suggestion
  const { catalystBaseUrl } = req.config;
```

packages/api/src/controllers/access-control.ts
```diff
@@ -159,66 +222,32 @@ app.post(
...
-    );
-    if (resp) {
-      // Dispose of the response body
-      await resp.text();
-    }
-  }
-  return statusCode;
-}
+  await fetch(url, options)
+    .then((response) => {
```
victorges, line 237: optional style nit: Can you use `async/await` pattern instead of `then` chain? IMO it's more readable as an iterative sequence than with all the callbacks, but you may disagree as well.

packages/api/src/schema/schema.yaml
```diff
@@ -1453,8 +1453,13 @@ components:
         encryption:
           type: object
           additionalProperties: false
-          required: [key]
```
victorges: This should be `required: [encryptedKey]` instead, since you're checking it in the code (then you don't need to check it in the code).

We should not keep support for the old experiment (no external user has access to it and we're not using it). So please remove the old `key` field and that experiment checking logic. This can be on a follow-up, but the new logic should work even without the experiment (right now I think you're still checking the experiment any time `encryption` is present)

victorges (approved):
LGTM

victorges (approved):
Nice. Just need to fix the CI.

PR #1717, opened 2023-05-05
api/playback: Stop returning assets from other users on playback info

victorges:
This is to stop returning assets from other users on the playback info endpoint.
This is a breaking change so we're doing it with a cut-off date logic instead
of breaking existing assets that might be used cross-user today.

 - Add cut-off date to search for cross-user assets
 - Allow lvpr.tv to access any asset to continue to have CID support for embeddable player 
 - Add a log when we return old assets to a different user, to evaluate fixing this

`yarn test` with the new tests

Fixes API-45

**Checklist**

### DIFF
--- a/packages/api/src/controllers/playback.ts
+++ b/packages/api/src/controllers/playback.ts
@@ -26,7 +26,7 @@ import { NotFoundError, UnprocessableEntityError } from "../store/errors";
  */
 export const CROSS_USER_ASSETS_CUTOFF_DATE = new Date(2023, 5, 10).getTime();
 
-var embeddablePlayerOrigin = /^https:\/\/(.+\.)?lvpr.tv\/$/;
+var embeddablePlayerOrigin = /^https:\/\/(.+\.)?lvpr.tv$/;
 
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/playback.ts
```diff
@@ -184,7 +213,10 @@ app.get("/:id", async (req, res) => {
   const ingest = ingests[0].base;
 
   let { id } = req.params;
-  const info = await getPlaybackInfo(req, ingest, id);
+  const isEmbeddablePlayer = embeddablePlayerOrigin.test(
+    req.headers["origin"] ?? ""
```
gioelecerati, line 217: This could be spoofed on applications outside of the browser to do a cross user query on CIDs, but we probably don't mind

gioelecerati (approved):
LGTM

packages/api/src/controllers/playback.ts
victorges: Yeah it's more about billing the right user than protecting any kind of sensitive info.

PR #1719, opened 2023-05-08
Trigger recording.ready webhook when vod is already processed

leszko:

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/stream.ts
```diff
@@ -1195,13 +1195,13 @@ async function publishDelayedRecordingReadyHook(
   ingest: string
 ) {
   return await queue.delayedPublishWebhook(
-    "events.recording.ready",
+    "events.recording.waiting",
```
victorges, line 1201: Love it!

packages/api/src/webhooks/cannon.ts
```diff
@@ -114,12 +113,12 @@ export default class WebhookCannon {
       return true;
     }
 
-    if (event === "recording.ready" && sessionId) {
+    if (event === "recording.waiting" && sessionId) {
       try {
         await this.handleRecordingReadyChecks(sessionId);
```
victorges: Can you rename this helper and anything inside that references recording.ready

packages/api/src/task/scheduler.ts
```diff
@@ -400,6 +401,30 @@ export class TaskScheduler {
         },
       });
     }
+    if (task.status.phase === "completed" && task.outputAssetId) {
```
victorges: We should avoid adding more business logic to this `updateTask` helper, as it should handle mostly the "storage update" and any consequential things like webhooks (tho this is a webhook haha). We probably did a poor job in keeping that true here over time, but in this specific case we're handling the completion phase of a task, which has a better place in the code today.

So the right place for this logic would be inside `processTaskEvent` instead, specifically under the [`"upload"` task case](https://github.com/livepeer/livepeer-com/blob/ad8d909a045b536ffa7cc15955e828e6e4e6aa5d/packages/api/src/task/scheduler.ts#L124-L126) which is what we use for recordings afaik. There it would also have more guarantee that it is processing an actual completion event for the task, and not any update that could trigger accidental/duplicate webhooks here.

Would you mind moving this to `processTaskEvent`?

victorges (approved):
The logic looks solid, only a nit about where I think this new logic should go. We're overloading that `updateTask` function too much by now and I think there's a more appropriate place for handling the completion event.

victorges (approved):
The logic looks solid, only a nit about where I think this new logic should go. We're overloading that `updateTask` function too much by now and I think there's a more appropriate place for handling the completion event.

packages/api/src/task/scheduler.ts
leszko: Updated

victorges (approved):
re-LGTM!

victorges (approved):
re-LGTM!

PR #1720, opened 2023-05-08
api/schema: Fix API schema after API reference refactor

victorges:

On the API reference refactor we created new schema files but forgot to delete
the old one, so it was able to merge cleanly even though it had conflicts with master.

This was caused by a couple other conflicting changes to the schema. Specifically:
 - stream health
 - recordings v2
 - creator ID
 - multi-participant (room abstraction&endpoints)

 - Added all those lost updates
 - Went through those moving private stuff to db-schema
 - ✅ 

 - Compared old and new generated schemas throughout all the [dev] commits here
 - `yarn test`

No.

**Checklist**


packages/api/src/compile-schemas.js
```diff
@@ -1,6 +1,6 @@
 import Ajv from "ajv";
 import pack from "ajv-pack";
-import { safeLoad as parseYaml } from "js-yaml";
+import { safeLoad as parseYaml, safeDump as serializeYaml } from "js-yaml";
```
gioelecerati, line 3: Is `safeDump` supposed to be imported?

gioelecerati (approved):
LGTM

packages/api/src/compile-schemas.js
victorges: Yeah it's used to write the merged yaml now!

PR #1721, opened 2023-05-09
Clean up Recording V1

leszko:
**This PR should be merged after Recording V2 is in prod for at least couple of weeks**.

This PR is a PR with cleanup for recording V1 and should not include any functional changes.

After we're moved to Recording V2 for good, we should clean up the code and infra a little bit. Otherwise, the code is very confusing. The work will include:
1. This PR to clean up the studio code
2. Send PR to `livepeer-infra` to stop using env variables related to Recording V1 + stop broadcasters to perform recording
3. Remove env variables related to Recording V1 in Vault: 
  - `LP_RECORD_OBJECT_STORE_ID_EU_CENTRAL`
  - `LP_RECORD_OBJECT_STORE_ID_US_WEST`
  - `LP_RECORD_OBJECT_STORE_ID`

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/stream.ts
```diff
@@ -482,9 +482,7 @@ app.get("/:parentId/sessions", authorizer({}), async (req, res) => {
   if (record) {
     if (record === "true" || record === "1") {
       query.push(sql`data->>'record' = 'true'`);
-      query.push(sql`data->>'recordObjectStoreId' IS NOT NULL`);
```
leszko, line 487: I'm not sure about the backward compatibility of this query. Did we ever set `recordObjectStoreId` when `record` was `false` 🤔

victorges: I _think_ this was some kind of flag indicating that line 1662 ran on the stream. I'm not 100% sure though tbh

victorges (commented):
We might want to remove the logic on `go-livepeer` that processed those values as well! Or at least make sure that it's not being triggered anymore (which might be the case with these changes).

Agreed that this can only be deployed when we are confident about the new flow. Let's go through this again by then

packages/api/src/controllers/session.ts
```diff
@@ -185,6 +185,6 @@ function removePrivateFields(obj: DBSession) {
 
 const adminOnlyFields: (keyof DBSession)[] = ["deleted", "broadcasterHost"];
 
-const privateFields: (keyof DBSession)[] = ["recordObjectStoreId", "version"];
```
victorges: we should probably still keep this since the legacy objects are still in the DB and we can't return this field (apparently)

packages/api/src/controllers/stream.ts
```diff
@@ -482,9 +482,7 @@ app.get("/:parentId/sessions", authorizer({}), async (req, res) => {
...
     } else if (record === "false" || record === "0") {
-      query.push(sql`data->>'recordObjectStoreId' IS NULL`);
       filterOut = true;
```
victorges, line 489: Might be missing sth like this?
```suggestion
      query.push(sql`(data->>'record' = 'false'` OR data->>'record' IS NULL)`);
      filterOut = true;
```

victorges: not sure if `record` can ever be null, added it just to be safe

packages/api/src/controllers/stream.ts
```diff
@@ -1629,58 +1622,19 @@ app.post("/hook", authorizer({ anyAdmin: true }), async (req, res) => {
...
-          `data integity error: record object store ${stream.recordObjectStoreId} not found or disabled`,
-        ],
-      });
-    }
-    recordObjectStore = ros.url;
-    if (ros.publicUrl) {
-      recordObjectStoreUrl = ros.publicUrl;
-    }
-  }
-
```
victorges, line 1691: omg 😭

packages/api/src/parse-cli.ts
```diff
@@ -305,16 +305,6 @@ export default function parseCli(argv?: string | readonly string[]) {
...
-      "record-object-store-id": {
-        describe:
-          "id of the object store that should be used for `record: true` requests that don't otherwise have an os",
-        type: "string",
-      },
-      "supress-record-in-hook": {
-        describe:
-          "do not return record object store in /stream/hook response, even if it is specified in the stream object",
-        type: "boolean",
-      },
```
victorges, line 341: WDYT of keeping these fields as deprecated (e.g. rewrite description explaining it) just so we have a safer deploy? Otherwise we'll have to deploy, but not before we remove the secret, but not before we update livepeer-infra not to use them (safe path executed in the opposite order lol).

victorges (approved):
1000x yes

packages/api/src/controllers/stream.ts
leszko: 🤣

packages/api/src/controllers/session.ts
leszko: Reverted in 8b722695100a095805efb8ec6f9d3cb3bee1ec23

leszko:
> We might want to remove the logic on `go-livepeer` that processed those values as well! Or at least make sure that it's not being triggered anymore (which might be the case with these changes).
> 
> Agreed that this can only be deployed when we are confident about the new flow. Let's go through this again by then

Yeah, but let's hold off for the moment about removing the recording logic from `go-livepeer`, since it's discussed with the community here: https://github.com/livepeer/Grant-Program/issues/157

packages/api/src/parse-cli.ts
leszko: I think the deployment should work without these fields. Let me try first in staging and if it fails, I'll keep these fields.

victorges: Sounds good :)

PR #1727, opened 2023-05-10
api/health: Add estimate playback latency for streams

victorges:
Adds the missing fields for the stream health exposure regarding an estimation
of the playback latency on HLS and WebRTC playback protocols.

 - Create new fields in the schema
 - Calculate them on the API
 - Fix a log that was missing whitespace

Checked the new field gets returned on the API.

**Checklist**


victorges:
This has been shelved for now while we figure out the right UX for this.

PR #1737, opened 2023-05-16
add missing apis, update description and summary

suhailkakar:

This PR adds the mission APIs to the `api-schema` file, update the summary and description of the APIs and roup the APIs by tag.

- Add the mission APIs to the `api-schema` file 
- Update the summary and description of the APIs 
- Group the APIs by tag

- yarn test 
- openapi validation

**Screenshots (optional)**

**Checklist**


packages/api/src/schema/api-schema.yaml
```diff
@@ -8,8 +8,71 @@ info:
...
 components:
   schemas:
+    request-upload:
+      type: object
+      required:
+        - name
+      properties:
+        name:
+          type: string
+          example: Example name
```
victorges: I believe this one should be replaced with `new-asset-payload` instead (with the only difference that it doesn't have a `url` field)

packages/api/src/schema/api-schema.yaml
```diff
@@ -8,8 +8,71 @@ info:
...
+      properties:
+        url:
+          type: string
+          example: >-
+            https://origin.livepeer.com/api/asset/upload/direct?token=eyJhbGciOiJIUzI1NiJ9.eyJtc2ciOiJoZWxsbyBoYWNrZXIsIHRoZXJlJ3Mgbm90aGluZyBmb3IgeW91IGhlcmUg8J-YhiJ9.1YDjmXsqLcgNyMSzT4kXl_kIni46_EuGX_xfqmC7e0Q
+        tusEndpoint:
+          type: string
+          example: >-
+            https://origin.livepeer.com/api/asset/upload/tus?token=eyJhbGciOiJIUzI1NiJ9.eyJtc2ciOiJoZWxsbyBoYWNrZXIsIHRoZXJlJ3Mgbm90aGluZyBmb3IgeW91IGhlcmUg8J-YhiJ9.1YDjmXsqLcgNyMSzT4kXl_kIni46_EuGX_xfqmC7e0Q
+        asset:
```
victorges: WDYT about referencing the asset schema with `$ref` to avoid the repetition? It will include more fields than will be available on the actual response, but it will also include other fields that are not included here (like `storage`, `playbackPolicy` etc). Those fields are already marked as non-required on the asset schema tho, so I don't think it's a problem to have them show up.

packages/api/src/schema/api-schema.yaml
```diff
@@ -1845,13 +1913,105 @@ components:
           description: The error rate for the asset
         existsBeforeStart:
           type: number
-          description:
+          description: >-
             The percentage of sessions that existed before the asset started
             playing
+    create-asset-url:
```
victorges: Same for `new-asset-payload`. These are **not** the only fields that can be set on the payload.

packages/api/src/schema/api-schema.yaml
```diff
@@ -1845,13 +1913,105 @@ components:
...
+        It is important to note, that the stream object is the core building
+        block of the Livepeer Studio platform. In Livepeer Studio, the stream
+        object has unique configuration data and metadata about all livestream
+        sessions that are associated with the stream.
+
+
+        Historically, a stream object with a parentId represented a single
+        livestream session. This type of "child-session" stream object will be
+        deprecated in an upcoming version of the API in favor of the session
+        object.
```
victorges: This "child session" situation has finally already changed with recording v2! We should remove those remarks.
```suggestion
        POST /stream creates a stream object. The configuration on this stream object
        is inherited by all live sessions using this stream's streamKey.


        It is important to note, that the stream object is the core building
        block of the Livepeer Studio platform. In Livepeer Studio, the stream
        object has unique configuration data and metadata about all livestream
        sessions that are associated with the stream.
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1845,13 +1913,105 @@ components:
...
+
+        The only parameter you are required to set is the name of your stream,
+        but we also highly recommend that you define the profiles parameter with
+        720p, 480p and 360p renditions.
+
+
+        If you do not define transcoding rendition profiles when creating the
+        stream, no transcoding will happen. Your playback video will have the
+        attributes of the source stream only, will not take advantage of
+        adaptive bitrate streaming and will likely buffer.
```
victorges: This was also outdated
```suggestion
        The only parameter you are required to set is the name of your stream,
        but we also highly recommend that you define transcoding profiles
        parameter that suits your specific broadcasting configuration.


        If you do not define transcoding rendition profiles when creating the
        stream, a default set of profiles will be used. These profiles include 240p, 
        360p, 480p and 720p.
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1875,7 +2035,43 @@ paths:
...
+        Example\t                                      \t                      |
+        Description                             |\n|
+        --------------------------------------------- |
+        --------------------------------------------------------------------- |
+        -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|\n|
+        All stream objects with the same userId\t      |
+        /api/stream?streamsonly=1\t     \t     \t     \t     \t     \t     \t     \t  |
+        excludes the historic stream objects with a parentId (stream objects
+        representing a single live stream
+        session)                                                                                                                                                          |\n|
```
victorges: I can't really get what's happening here 😆 
Is this like the markdown from old docs injected here? Does redocly support this? Could you inline the string in another YAML format so it's also readable here? The "create a webhook" one is much nicer

packages/api/src/schema/api-schema.yaml
```diff
@@ -1875,7 +2035,43 @@ paths:
...
+      tags:
+        - Stream
+      summary: Retrieve streams
+      description:
+        "GET /stream to retrieve all stream objects with the same userId. The
+        response will include both parent stream objects and the historic stream
+        objects with a parentId (stream objects representing a single live
+        stream session). This historic type of \"child-session\" stream object
+        will be deprecated in an upcoming version of the API in favor of the
+        session object.\n\nGET /stream?streamsonly=1 to retrieve all stream
```
victorges: We can change the remark here as well
```suggestion
        stream session). The usage of these \"child-session\" stream objects
        is deprecated, use the session endpoints instead. To filter this API not
        to return these child sessions add a `?streamsonly=true` query-string\n\n
        GET /stream?streamsonly=1 to retrieve all stream
```

While at it, mind adding the `streamsonly` query-string to the doc here? Like this https://github.com/livepeer/livepeer-com/blob/bfd3354e5402cb4db57f5f70e893b64d6aff2122/packages/api/src/schema/api-schema.yaml#L2291

packages/api/src/schema/api-schema.yaml
```diff
@@ -1845,13 +1913,105 @@ components:
...
+                    example: 204
+                  height:
+                    type: number
+                    example: 360
+                  bitrate:
+                    type: number
+                    example: 449890
+      required:
+        - type
+        - meta
```
victorges: This was long due, thanks! Would you mind replacing the custom type definition in [playback.ts](https://github.com/livepeer/livepeer-com/blob/b828376275e76abbddf0a524440fd08ee4a21bc8/packages/api/src/controllers/playback.ts#L33) with this one so we don't have to repeat it? (these types are auto-generated on the types.d.ts file, we already have some imports from there in that file)

Adding a suggestion here for a couple things that were missing from the original schema in typescript:
```suggestion
    playback-info:
      type: object
      properties:
        type:
          type: string
          example: vod
          enum: [live, vod, recording]
        meta:
          type: object
          required: [source]
          properties:
            source:
              type: array
              items:
                type: object
                required: [hrn, type, url]
                properties:
                  hrn:
                    type: string
                    example: MP4
                    enum: ["HLS (TS)", "MP4", "WebRTC (H264)"];
                  type:
                    type: string
                    example: html5/video/mp4
                    enum:
                     - html5/application/vnd.apple.mpegurl
                     - html5/video/mp4
                     - html5/video/h264
                  url:
                    type: string
                    example: >-
                      https://asset-cdn.lp-playback.monster/hls/1bde4o2i6xycudoy/static360p0.mp4
                  size:
                    type: number
                    example: 494778
                  width:
                    type: number
                    example: 204
                  height:
                    type: number
                    example: 360
                  bitrate:
                    type: number
                    example: 449890
      required:
        - type
        - meta
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1931,9 +2152,78 @@ paths:
...
+              $ref: "#/components/schemas/update-record-status"
+      responses:
+        "204":
+          description: Success (No content)
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
```
victorges: Actually, we should not document this API (sorry if I probably said otherwise). This has been replaced by the single `PATCH /stream/:id` which you documented above so we don't need this here occupying valuable docs space.

Also remember to delete the `update-record-status` schema above when you delete this

packages/api/src/schema/api-schema.yaml
```diff
@@ -1981,7 +2273,12 @@ paths:
         schema:
           type: string
     get:
-      description: Gets a specific multistream target details
+      tags:
+        - Multistream
+      summary: Retrieve a multistream target
+      description: >-
+        You can retrieve a multistream target with its ID on
+        /multistream/target{id}.
```
victorges: This feels redundant with the OpenAPI spec itself. Would remove the path since it already shows up on the redoc ui
```suggestion
        You can retrieve a multistream target from its ID with this API.
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2138,11 +2544,198 @@ paths:
...
+        To upload an asset, your first need to request for a direct upload URL
+        and only then actually upload the contents of the asset.
+
+
+        This also allows your users to upload files since the upload URL is
+        pre-signed with your credentials and is accessible from the browser. The
+        initial API call to get the upload URL must be done from a backend
+        though, from which an API token is available.
+
+
```
victorges: We should probably remove this paragraph since this API is being used directly from the browser now with CORS keys
```suggestion
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2138,11 +2544,198 @@ paths:
...
+        ## Direct Upload
+
+        For a direct upload, make a PUT request to the URL received in the url
+        field of the response above, with the raw video file as the request
+        body. response above:
+
+
+        ## Resumable Upload
+
+        Livepeer supports resumable uploads via Tus(opens in a new tab). This
```
victorges:
```suggestion
        Livepeer supports resumable uploads via Tus. This
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2138,11 +2544,198 @@ paths:
...
+        a video file.
+
+
+        From the previous section, we generated a URL to upload a video file to
+        Livepeer on POST /api/asset/request-upload. You should use the
+        tusEndpoint field of the response to upload the video file and track the
+        progress:
+
+
+        ```javascript
```
victorges, line 2474: This is good for now, but we definitely need to move this to a guide/tutorial. Now that we have these auto-generated API reference we should make it more objective and technical.

packages/api/src/schema/api-schema.yaml
```diff
@@ -2138,11 +2544,198 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/upload-response"
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
+  /asset/import:
```
victorges: This API is now
```suggestion
  /asset/upload/url:
```
(everything else same, but remember to update the text below that repeats the path)

packages/api/src/schema/api-schema.yaml
```diff
@@ -2249,16 +2852,18 @@ paths:
           required: true
           schema:
             type: string
-      description: create a new room
+      summary: Create a room
```
victorges:
```suggestion
      summary: Create a room user
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2270,9 +2875,9 @@ paths:
           required: true
           schema:
             type: string
-      description: removes a specific user
+      summary: Remove a user
```
victorges:
```suggestion
      summary: Remove a user from the room
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2488,22 +3097,572 @@ paths:
...
+          schema:
+            type: string
+          description: ID of the parent stream
+          required: true
+        - in: query
+          name: record
+          schema:
+            type: integer
+            example: 1
+          description: Flag indicating if the session should be recorded
```
victorges:
```suggestion
          description: Flag indicating if the response should only include recorded sessions
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2488,22 +3097,572 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/task"
+        default:
+          description: Error
+  /transcode:
+    post:
+      tags:
+        - Misc
```
victorges: Can we make this tag `Transcode API`?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2488,22 +3097,572 @@ paths:
...
+      summary: Transcode a video
+      description: >
```
victorges: Omg this is so long. We need a separate guide for this as well instead of including all that in the API ref (not a change for this PR)

packages/api/src/schema/api-schema.yaml
```diff
@@ -2488,22 +3097,572 @@ paths:
...
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
+  "/playback/{id}":
+    get:
+      tags:
+        - Misc
```
victorges: And since this API would end up alone in the `Misc` tag, it could become `Playback`

packages/api/src/schema/api-schema.yaml
```diff
@@ -2488,22 +3097,572 @@ paths:
...
+      summary: Retrieve Playback Info
+      description: >-
+        You can retrieve the playback information by calling the GET playback
+        endpoint with the playbackId.
```
victorges:
```suggestion
        You can retrieve the playback information about any playback identifier by
        calling this endpoint. It accepts playbackIds from any resource in the API
        as well as dStorage IDs for assets.
```

victorges (approved):
LGTM! Made some suggestions but feel free to pick the ones you like and merge without the need of another review.

Thanks for going through all these, a lot of manual work 😩

packages/api/src/schema/api-schema.yaml
suhailkakar: Yes, it is a table :D

I know it is confusing, but yes the openapi support markdown

packages/api/src/schema/api-schema.yaml
victorges: Started using `new-asset-payload` on the API schemas, but kept the `url` field tech debt for now. It will be shown for the `/request-upload` API even though it is not used there. To be fixed later.

packages/api/src/schema/api-schema.yaml
victorges: Same fix.

victorges (approved):
Final LGTM 😄

PR #1746, opened 2023-05-23
billing: get billing usage

gioelecerati:

**Screenshots (optional)**

**Checklist**


packages/api/src/config.js
```diff
@@ -2,49 +2,31 @@ exports.products = {
...
-    ],
+    lookupKeys: ["scale_price"],
+    usage: [],
   },
-  prod_3: {
+  pay_as_you_go_1: {
     order: 3,
     name: "Pay-As-You-Go",
-    lookupKeys: ["price_3"],
+    lookupKeys: ["transcoding_usage", "tstreaming_usage", "tstorage_usage"],
```
adamsoffer, line 107: @gioelecerati Nice. Is this a typo? Should be `streaming_usage` and `storage_usage`?

adamsoffer: @gioelecerati Is the reason you named this product `pay_as_you_go_1` and not `pay_as_you_go` because there might be multiple pay-as-you-go products?

gioelecerati: Yes, it's a typo when I created the prices, but considering it's staging I left it there. Lookup keys needs to be univoque anyway, so when I create the prod products I will make the right things!

About the name, yeah I used the index in case they are going to be future products.
For prod, I was going to name it `pay_as_you_go_prod_1` but I am open to rename everything differently if that's not ok

adamsoffer: Gotcha. Something to keep in mind is that the product team may want to rename a product in the future. In this case, it might make sense to keep the product ids generic (e.g. `prod_0`, `prod_1`, `prod_2`, `prod_3`, `prod_4`, `prod_5`) and then rely on the Stripe's product object `name` field for names (Hacker, Growth, Scale, Pay-as-you-go, Enterprise).

wdyt?

packages/api/src/controllers/user.ts
```diff
@@ -892,11 +904,27 @@ app.post(
       payload.stripeCustomerSubscriptionId
     );
 
+    // Temporarily skip updating the subscription if user is selecting a new plan
+    if (
+      payload.stripeProductId !== "prod_0" &&
+      payload.stripeProductId !== "prod_1" &&
+      payload.stripeProductId !== "prod_2"
```
victorges, line 914: WDYT of creating some helper like `isDeprecatedStripeProduct` or sth like that?

packages/api/src/controllers/user.ts
```diff
@@ -826,6 +826,18 @@ app.post(
...
+    if (
+      stripeProductId !== "prod_0" &&
+      stripeProductId !== "prod_1" &&
+      stripeProductId !== "prod_2"
+    ) {
+      await db.user.update(user.id, {
+        newStripeProductId: stripeProductId,
+        stripeCustomerSubscriptionId: subscription.id,
+      });
+      res.send(subscription);
```
victorges, line 840: Probably missing a `return` here if we want to stop the response here?

packages/api/src/controllers/user.ts
```diff
@@ -892,11 +904,27 @@ app.post(
...
+      payload.stripeProductId !== "prod_0" &&
+      payload.stripeProductId !== "prod_1" &&
+      payload.stripeProductId !== "prod_2"
+    ) {
+      // Update user's product subscription in our db
+      await db.user.update(user.id, {
+        newStripeProductId: payload.stripeProductId,
+        planChangedAt: Date.now(),
+      });
+      res.send(subscription);
```
victorges, line 921: Same possibly missing `return`

victorges (commented):
Reviewed API code, will check frontend usage

victorges (approved):
Sent some comments on Vercel as well, other than that LGTM!

PR #1748, opened 2023-05-25
Source playback

mjh1:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -359,7 +359,7 @@ export async function withPlaybackUrls(
   asset: WithID<Asset>,
   os?: ObjectStore
 ): Promise<WithID<Asset>> {
-  if (asset.status.phase !== "ready") {
+  if (asset.status.phase !== "ready" && !asset.sourcePlaybackReady) {
     return asset;
   }
   try {
@@ -367,6 +367,12 @@ export async function withPlaybackUrls(
   } catch (err) {
     console.error("Error getting asset object store", err);
     return asset;
+  }
+  if (asset.sourcePlaybackReady) {
+    return {
+      ...asset,
+      playbackUrl: getPlaybackUrl(config, ingest, asset, os),
+    };
   }
   return {
     ...asset,
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -359,7 +359,8 @@ export async function withPlaybackUrls(
   asset: WithID<Asset>,
   os?: ObjectStore
 ): Promise<WithID<Asset>> {
-  if (asset.status.phase !== "ready" && !asset.sourcePlaybackReady) {
+  if (asset.files?.length < 1) {
+    // files is only set when playback is available
     return asset;
   }
   try {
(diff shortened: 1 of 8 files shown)


mjh1:
@victorges aside from lack of test coverage, LGTU?

packages/api/src/controllers/asset.ts
```diff
@@ -361,6 +361,12 @@ export async function withPlaybackUrls(
     console.error("Error getting asset object store", err);
     return asset;
   }
+  if (asset.sourcePlaybackReady) {
```
victorges: An alternative could be also to just use the existence of `files` in the asset, specifically for a playback manifest. Those are only set on termination today and are now being set in the middle of processing. Not required change just wondering, WDYT?

packages/api/src/task/scheduler.ts
```diff
@@ -207,6 +213,31 @@ export class TaskScheduler {
...
+    const asset = await db.asset.get(task.outputAssetId);
+    await this.updateAsset(task.outputAssetId, {
+      sourcePlaybackReady: true,
+      files: event.output.upload.assetSpec.files,
+      status: {
+        phase: asset.status?.phase,
+        progress: asset.status?.progress,
+        errorMessage: asset.status?.errorMessage,
+        updatedAt: Date.now(),
+      },
```
victorges: **nit** What you want is probably
```suggestion
      status: {
        ...asset.status,
        updatedAt: Date.now(),
      },
```

packages/www/components/AssetDetails/AssetPlayerBox/index.tsx
```diff
@@ -8,7 +8,7 @@ import FileUploadingProgress from "./FileUploadingProgress";
 import ProcessingProgress from "./ProcessingProgress";
 
 const AssetStatusBox = ({ asset }: { asset?: Asset }) => {
-  if (asset?.status?.phase === "ready" && asset.playbackUrl) {
+  if (asset?.playbackUrl) {
```
victorges, line 11: Hah nice, makes the most sense

victorges (approved):
Yeah LGTM!

packages/api/src/controllers/asset.ts
mjh1: Yeah I was thinking the same really. Does this work: https://github.com/livepeer/studio/pull/1748/commits/a88980c404b69e9ace01d3ecfa8a7fad638e6d5d
or maybe i should also check the return from `getPlaybackUrl` is not undefined

mjh1: Actually no if it's undefined then it'll be hidden from the response object anyway

mjh1:
@victorges Does the test look ok to you?  https://github.com/livepeer/studio/pull/1748/commits/3099be80033adf723a2c2c8988c3a2db96d49ace

victorges:
@mjh1 it does! Thanks for creating that, will make it much easier to create future tests on the scheduler:)

packages/api/src/controllers/asset.ts
victorges: Yeah like that exactly! Does it work fine? Your tests still seemed to check the existence of the boolean

mjh1: Yep worked ok, but yes i'm still using the bool in one other place so will do a follow up PR

PR #1752, opened 2023-05-29
Add Verifiable API endpoints

leszko:

Adds Verifiable Video API as described in [Verified Video Prototype Design Document](https://www.notion.so/livepeer/Prototype-Design-Document-da700144f713493b83c12075a9166c73).

Note that this PR does not include pinning Attestation Metadata to IPFS. This will be added as a separate PR.

- Add DB schema for `attestation`
- Add `/attestation` API into the `/expriement` API
- Add basic operations for attestations

Fix VID-210

**Screenshots (optional)**

**Checklist**

- [ ] ~My change requires a change to the documentation.~ (not yet)


packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -0,0 +1,103 @@
...
+};
+
+app.post("/", validatePost("attestation"), async (req, res) => {
+  const { primaryType, domain, message, signature } = req.body;
+
+  if (primaryType !== "VideoAttestation") {
+    return res.status(400).json({ errors: ["invalid primaryType"] });
+  }
+  if (!_.isEqual(domain, DOMAIN)) {
+    console.log(domain);
```
iameli: Should either remove or clean this up if we're looking to provide debugging context

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -0,0 +1,103 @@
...
+    console.log(DOMAIN);
+    return res.status(400).json({ errors: ["invalid domain"] });
+  }
+  if (!verifySigner(message, signature)) {
+    return res.status(400).json({ errors: ["invalid signature"] });
+  }
+
+  const attestationMetadata = await db.attestation.create({
+    id: uuid(),
```
iameli: Hmm, what would you think about storing attestations by hash? You can get it from ethers.js [using the TypedDataEncoder interface](https://docs.ethers.org/v5/api/utils/hashing/#TypedDataEncoder); it seems kind of arbitrary to me to have an opaque UUID in here when the process _already_ calculates a unique CID.

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -0,0 +1,103 @@
...
+    );
+  }
+  const [output, newCursor] = await db.attestation.find(query, {
+    limit,
+    cursor,
+    order: "data->>'createdAt' DESC",
+    process: ({ data, count: c }) => {
+      if (count) {
+        res.set("X-Total-Count", c);
+      }
```
iameli, line 95: No reason to omit the header if it's zero, right?

```suggestion
      res.set("X-Total-Count", c ?? 0);
```

iameli (changes requested):
WDYT about using the content CIDs from the EIP-712 signature as the primary key instead of a UUID?

iameli:
Also, is there anything in here preventing two people from claiming ownership of the same video?

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -0,0 +1,103 @@
...
+  name: "Verifiable Video",
+  version: "1",
+};
+
+app.post("/", validatePost("attestation"), async (req, res) => {
+  const { primaryType, domain, message, signature } = req.body;
+
+  if (primaryType !== "VideoAttestation") {
+    return res.status(400).json({ errors: ["invalid primaryType"] });
+  }
```
victorges: **nit:** If you prefer, can enforce this on the JSON schema instead with an `enum`

packages/api/src/controllers/experiment/attestation.ts
victorges: I think we wanna omit if it wasn't requested (`count`) query string, not when the count itself (`c`) is zero.

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -0,0 +1,103 @@
...
+    console.log(DOMAIN);
+    return res.status(400).json({ errors: ["invalid domain"] });
+  }
+  if (!verifySigner(message, signature)) {
+    return res.status(400).json({ errors: ["invalid signature"] });
+  }
+
+  const attestationMetadata = await db.attestation.create({
+    id: uuid(),
+    createdAt: Date.now(),
```
victorges, line 45: Is there some kind of timestamp in the message as well? If not, that seems like a good addition. If yes, we could consider using that as the `createdAt` value here (especially validating that it has to be recently signed I guess?)

victorges (approved):
LGTM!

adamsoffer:
Nitpick: can we rename the namespace `/experiment` to `/experimental`? That's the more common namespacing convention for experimental API features.

packages/api/src/controllers/experiment/attestation.ts
leszko: Yeah, sorry. leftover. Updated in e397ed1c1d13bf078b73616f70f7e51c19a0670f

packages/api/src/controllers/experiment/attestation.ts
leszko: Good idea. Updated in 49015f05065772a5b3c0ee6056cd02e2ee4fefe4

packages/api/src/controllers/experiment/attestation.ts
leszko: Yeah, good, idea. We can use the hash of the message. Updated in ce91974a5294a9f5f753a0727a55faa6243276e9. PTAL.

packages/api/src/controllers/experiment/attestation.ts
leszko: I think the `createdAt` should stay server-generated.

But right, the validation for the user-generate `timestamp` makes sense. Added here: 93693ea50585236bc7719a990781b2f243fe4511.

packages/api/src/controllers/experiment/attestation.ts
leszko: Yeah, I think we can omit it. Also I implemented this the same way have it in other parts of the code.

leszko:
> Nitpick: can we rename the namespace `/experiment` to `/experimental`? That's the more common namespacing convention for experimental API features.

We place all experiments under `/experiment/*`. Please take a look at the doc created by @victorges 
https://www.notion.so/livepeer/Feature-Experimentation-Playbook-2f97f92d6b0d4f2aba11167c76ad5775

leszko:
> WDYT about using the content CIDs from the EIP-712 signature as the primary key instead of a UUID?

We generated IPFS CID in the async manner right now (it's actually done by Task Runner). So, I think (at least for now) we need to keep a separate ID.

leszko:
> Also, is there anything in here preventing two people from claiming ownership of the same video?

No, currently no. I described it [here](https://www.notion.so/livepeer/Prototype-Design-Document-da700144f713493b83c12075a9166c73?pvs=4#b6b91cf0f0b04f38a0c61a7da58aa754). PTAL.

iameli:
> > WDYT about using the content CIDs from the EIP-712 signature as the primary key instead of a UUID?
> 
> We generated IPFS CID in the async manner right now (it's actually done by Task Runner). So, I think (at least for now) we need to keep a separate ID.

~~I'm not saying "no ID" or using the IPFS CID — I'm talking about using the EIP-712 hash value as the ID rather than a UUID. task-runner can use that just as easily as it could use a UUID — I don't see the advantage of a UUID at that point.~~

~~EIP-712 payloads are already content-addressable, why not use the CID we already have?~~

D'oh, just realized you already did this and we just miscommunicated about my other comment. LGTM!

packages/api/src/controllers/experiment/attestation.ts
iameli: Ahhh got it, I interpreted `if (count)` as `if (c)`. Cool.

iameli (approved):
LGTM, no problem with shipping this.

I'm not certain about https://github.com/livepeer/studio/pull/1752/commits/93693ea50585236bc7719a990781b2f243fe4511 — I can imagine in the future that we'd want to be able to import signatures done somewhere else in the past. But it'll be easy enough for us to relax that restriction in the future.

leszko:
> LGTM, no problem with shipping this.
> 
> I'm not certain about [93693ea](https://github.com/livepeer/studio/commit/93693ea50585236bc7719a990781b2f243fe4511) — I can imagine in the future that we'd want to be able to import signatures done somewhere else in the past. But it'll be easy enough for us to relax that restriction in the future.

I gave it a second though and let's maybe just restrict to signatures done in the past.

PR #1753, opened 2023-05-30
Add attestation support to playback API

leszko:

Add support for returning playback info form `/api/playback/<attestation-id>`.

VID-211

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/playback.ts
+++ b/packages/api/src/controllers/playback.ts
@@ -115,7 +115,7 @@ export async function getResourceByPlaybackId(
     if (asset.status.phase !== "ready") {
       throw new UnprocessableEntityError("asset is not ready for playback");
     }
-    if (asset.userId !== user?.id && !cutoffDate) {
+    if (asset.userId !== user?.id && cutoffDate) {
       console.log(
         `Returning cross-user asset for playback. ` +
           `userId=${user?.id} userEmail=${user?.email} origin=${origin} ` +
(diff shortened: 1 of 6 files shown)


iameli (approved):
LGTM

packages/api/src/controllers/playback.ts
```diff
@@ -114,7 +115,7 @@ export async function getResourceByPlaybackId(
     if (asset.status.phase !== "ready") {
       throw new UnprocessableEntityError("asset is not ready for playback");
     }
-    if (asset.userId !== user?.id && !isCrossUserQuery) {
+    if (asset.userId !== user?.id && !cutoffDate) {
```
victorges: I think this condition is inverted now

packages/api/src/schema/api-schema.yaml
```diff
@@ -1929,6 +1929,9 @@ components:
                   bitrate:
                     type: number
                     example: 449890
+            attestation:
+              $ref: "#/components/schemas/attestation"
```
victorges, line 1933: Hmm is this like a circular reference?

victorges (approved):
LGTM

packages/api/src/controllers/playback.ts
leszko: Right 🤦 Updated in 15337119dbab5508deb71934225b18c4c5a38b08

packages/api/src/schema/api-schema.yaml
leszko: No, it points correctly here: https://github.com/livepeer/studio/blob/42aa0e87e9869f7b29396efefa10ab22b6e6ab91/packages/api/src/schema/db-schema.yaml#L371

victorges: Ohh I only realized now that this was a field and not a component schema 😅

PR #1755, opened 2023-06-01
Add support for signing with Flow Wallet

leszko:

Add support for signing with the Flow Wallet. Please check the detailed information in the design doc: https://www.notion.so/livepeer/Prototype-Design-Document-da700144f713493b83c12075a9166c73?pvs=4#e47d78831e6146bb9bea145ae6ae8584

Note that I had to change the `attestation.id` back to using `uuid()`. If we want to use the same abstraction for EIP-712 and Flow, then the ID cannot be generated from the typed data (since typed data is only in EIP-712). CC: @iameli 

If you would like to test it, then you can use Flipper Developer wallet (don't use Lilico, because it has some bugs while creating the signature 🤷 )

- Verify signatures according to both: EIP-712 and Flow
- Add `signatureType` (which can be automatically resolved from the payload)

[[VID-216](https://linear.app/livepeer/issue/VID-216/studio-support-flow-wallet-signatures)](https://linear.app/livepeer/issue/VID-216/studio-support-flow-wallet-signatures)

**Screenshots (optional)**

**Checklist**


iameli:
Hmm, understood that Flow doesn't have a typed data construct, unfortunate...

That said, there's still a hash, yeah? And reproducing that hash to verify the signature will still be an important part of verifying code? At that point I'd still be in favor of using that hash for the primary key of the Attestation document... it's still content-addressible even if it's content-addressible by our not-future-proof JSON-hashing thing. If a CID exists, may as well use that CID as the primary key, I think?

Opaque database identifiers scare me in this field. If we use them, people will start building on top of them, and then all of a sudden nothing works without our centralized database.

iameli:
I don't think I have strong feelings about the ad-hoc signing method we're throwing together for Flow signing. That said, we _could_ still handle it by converting the input data to an EIP-712 body and hashing that, and maybe that allows there to be fewer code paths on our end? "Signing hashes of JSON strings" is what Scuttlebutt does and it's one of the biggest criticisms of that project — it basically means you need to use custom JSON stringifiers in every implementing language and there are a lot of corner cases to get right to make the stringification deterministic. So I guess I'm marginally in favor of EIP-712 for signing here but if you don't want to then I don't mind that much.

packages/api/src/controllers/experiment/attestation.test.ts
```diff
@@ -126,6 +126,32 @@ describe("Attestation API", () => {
   });
 
   describe("POST /", () => {
+    it("should support Flow Wallet signatures", async () => {
```
victorges, line 150: Can we add a test for it to reject invalid signatures? Maybe cross-user

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -26,22 +28,28 @@ const TYPES = {
...
+    message,
+    signature,
+    signatureType
+  );
+  if (!verfiedSignatureType) {
+    return res.status(400).json({ errors: ["invalid signature"] });
+  }
 
-  const id = ethers.TypedDataEncoder.hash(DOMAIN, TYPES, message);
+  const id = uuid();
```
victorges: Agreed on trying to use at least a hash here

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -50,15 +58,64 @@ app.post("/", validatePost("attestation"), async (req, res) => {
...
+  try {
+    const verifiedSigner = ethers.verifyTypedData(
+      DOMAIN,
+      TYPES,
+      message,
+      signature
+    );
+
+    return verifiedSigner === message.signer;
+  } catch (e) {
```
victorges, line 97: I wonder if these error messages won't be useful to the user at some point

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -50,15 +58,64 @@ app.post("/", validatePost("attestation"), async (req, res) => {
...
+async function verifyFlowSignature(
+  message: any,
+  signature: string
+): Promise<boolean> {
+  try {
+    // These parameters are required to use the fcl library, even though we don't use on-chain verification
+    await fcl.config({
+      "flow.network": "testnet",
+      "accessNode.api": "https://access-testnet.onflow.org",
+    });
```
victorges: Do we/Can we call this config function multiple times, before every query? Might wanna consider moving it to app initialization e.g. `app-router.ts` or sth like that

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -50,15 +58,64 @@ app.post("/", validatePost("attestation"), async (req, res) => {
...
+      {
+        f_type: "CompositeSignature",
+        f_vsn: "1.0.0",
+        addr: message.signer,
+        keyId: 0,
+        signature: signature,
+      },
+    ];
+
+    return await fcl.AppUtils.verifyUserSignatures(
```
victorges, line 117: Curious do you know why their verification logic is async?

packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -50,15 +58,64 @@ app.post("/", validatePost("attestation"), async (req, res) => {
...
-  const verifiedSigner = ethers.verifyTypedData(
-    DOMAIN,
-    TYPES,
-    message,
-    signature
-  );
+async function verifySignature(
+  message: any,
+  signature: string,
+  signatureType?: string
```
victorges: **nit:** For extra typeness here you can reference the attestation field
```suggestion
  signatureType?: Attestation["signatureType"]
```

victorges (approved):
LGTM, after resolving the ID/hash issue

packages/api/src/controllers/experiment/attestation.ts
leszko: ok! You're both right. Updated to hash: cbdaccdf44ae7f2128424f1962dced96e2cc5886

leszko:
> Hmm, understood that Flow doesn't have a typed data construct, unfortunate...
> 
> That said, there's still a hash, yeah? And reproducing that hash to verify the signature will still be an important part of verifying code? At that point I'd still be in favor of using that hash for the primary key of the Attestation document... it's still content-addressible even if it's content-addressible by our not-future-proof JSON-hashing thing. If a CID exists, may as well use that CID as the primary key, I think?
> 
> Opaque database identifiers scare me in this field. If we use them, people will start building on top of them, and then all of a sudden nothing works without our centralized database.

Yes, you're actually right. Updated in cbdaccdf44ae7f2128424f1962dced96e2cc5886

packages/api/src/controllers/experiment/attestation.ts
leszko: I think they (by default) check the signatures on-chain. That is why all this logic is async. However, we (currently) don't check anything on-chain. Just their library is done that way.

leszko:
> I don't think I have strong feelings about the ad-hoc signing method we're throwing together for Flow signing. That said, we _could_ still handle it by converting the input data to an EIP-712 body and hashing that, and maybe that allows there to be fewer code paths on our end? "Signing hashes of JSON strings" is what Scuttlebutt does and it's one of the biggest criticisms of that project — it basically means you need to use custom JSON stringifiers in every implementing language and there are a lot of corner cases to get right to make the stringification deterministic. So I guess I'm marginally in favor of EIP-712 for signing here but if you don't want to then I don't mind that much.

Let's maybe keep it as it is right now. I totally get your point. But IIUC, "converting the input data to an EIP-712 body" means encoding it before signing, so it won't be human-readable in the Flow wallet anymore. In that case, since this feature is mainly to present to users right now, I'd prefer keeping the stringified JSON.

Saying that, we plan to spec out the typed data structures for the Flow network. We may work on that in June/July.

packages/api/src/controllers/experiment/attestation.ts
leszko: I think that no, the user will just get "invalid signature". Here you can't read much more data from this error message.

packages/api/src/controllers/experiment/attestation.ts
leszko: Moved in 429881b1073e849eccde9c7359b4015167bfbaa2

packages/api/src/controllers/experiment/attestation.test.ts
leszko: Updated in e25a2ac02b4c98893638298020875cf891cb6ce3

PR #1759, opened 2023-06-07
Call livekit API to check egress status

mjh1:

Previously we were recording the ID of the egress in our DB and trying to use that to determine whether egress was active. This doesn't always work as egress gets shut down on livekit's side after a room becomes inactive, while on in studio we would think egress is still active. This PR fixes those issues by always checking through livekit's API for the existence of an egress.

Tested with a local version of studio and modified the existing unit test.

**Checklist**

### DIFF
--- a/packages/api/src/controllers/room.ts
+++ b/packages/api/src/controllers/room.ts
@@ -35,14 +35,14 @@ app.post("/", authorizer({}), async (req, res) => {
     req.config.livekitApiKey,
     req.config.livekitSecret
   );
-  const room = await svc.createRoom({
+  await svc.createRoom({
     name: id,
     // timeout in seconds
     emptyTimeout: 15 * 60,
     maxParticipants: 10,
     metadata: JSON.stringify({ userId: req.user.id }),
   });
-  console.log("livekit room created", room);
+  console.log(`livekit room created. userId=${req.user.id} roomId=${id}`);
 
   await db.room.create({
     id: id,
@@ -113,6 +113,7 @@ app.post(
   authorizer({}),
   validatePost("room-egress-payload"),
   async (req, res) => {
+    // check the room exists and the user is correct
     await getRoom(req);
 
     const egressClient = new EgressClient(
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/room.ts
```diff
@@ -110,10 +113,18 @@ app.post(
   authorizer({}),
   validatePost("room-egress-payload"),
   async (req, res) => {
-    const room = await getRoom(req);
+    await getRoom(req);
```
victorges, line 117: I assume this is only called to validate the room exists? Could you add a short comment to make that clear?

packages/api/src/controllers/room.ts
```diff
@@ -146,33 +152,43 @@ app.post(
...
     req.config.livekitSecret
   );
+  const currentEgress = await egressClient.listEgress(req.params.roomId);
+
+  let found = false;
+  for (const egress of currentEgress) {
+    if (isEgressRunning(egress)) {
+      found = true;
+      const info = await egressClient.stopEgress(egress.egressId);
+      console.log("egress stopped", info);
```
victorges: Can you make these logs more structured e.g. add `roomId=${}`, `egressId=`, etc. Might still want to log the full info in the end anyway (can double-stringify it so it's parsable by Loki, tho the full JSON is prolly more useful for us to copy and inspect it individually anyway)

packages/api/src/controllers/room.ts
```diff
@@ -110,10 +113,18 @@ app.post(
   authorizer({}),
   validatePost("room-egress-payload"),
   async (req, res) => {
-    const room = await getRoom(req);
+    await getRoom(req);
 
-    if (room.egressId !== undefined && room.egressId != "") {
```
victorges, line 115: Another approach here could just be to fetch the egress by ID and check if `isEgressRunning` as well, before assuming it is. I like having the reference to the external objects in our DB, also keeping it consistent with the rooms referencing behavior themselves, and it might be more efficient after many recorded calls are made on the same room. 

Just a suggestion tho, not a must. Any perf issues will take a couple years to happen on any room lol and the rest is non-behavioral.

victorges (approved):
LGTM

packages/api/src/controllers/room.ts
mjh1: Thanks for questioning this, since livekit said to me that egress was a per session thing I didn't think they would build up historically in these API responses but turns out they do. The livekit API has some slightly odd quirks, I've just found that despite the egress info being preserved, if you try to start an egress for a room without an active session it just tells you that the room doesn't exist! That causes another bug for us so I will fix that in this PR!

mjh1: Livekit commented that the API will only return up to 24 hours worth of results so no worries about an infinitely growing list over time 👍

PR #1760, opened 2023-06-08
URL encode the object store credentials for transcode API

mjh1:
We saw an issue reported by a user where special characters were not escaped which caused the download to fail.


victorges (approved):
LGTM

PR #1761, opened 2023-06-08
usage: api reference

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -3211,26 +3211,7 @@ paths:
           content:
             application/json:
               schema:
-                type: object
-                description: |
-                  A simplified metric object about aggregate usage of an
-                  user.
-                properties:
-                  UserID:
-                    $ref: >-
-                      #/components/schemas/usage-metric/properties/UserID
-                  CreatorID:
-                    $ref: >-
-                      #/components/schemas/usage-metric/properties/CreatorID
-                  DeliveryUsageMins:
-                    $ref: >-
-                      #/components/schemas/usage-metric/properties/DeliveryUsageMins
-                  TotalUsageMins:
-                    $ref: >-
-                      #/components/schemas/usage-metric/properties/TotalUsageMins
-                  StorageUsageMins:
-                    $ref: >-
-                      #/components/schemas/usage-metric/properties/StorageUsageMins
+                $ref: "#/components/schemas/usage-metric"
         default:
           description: Error
           content:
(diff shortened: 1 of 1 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -1806,6 +1806,32 @@ components:
...
+    usage-metric:
+      type: object
+      description: |
+        An individual metric about usage of a user.
+      required:
+        - UserID
+        - CreatorID
+        - DeliveryUsageMins
+        - TotalUsageMins
+        - StorageUsageMins
```
victorges, line 1818: I missed on the `livepeer-data` review that these fields did not have their JSON field names specified. We currently have `lowerCamel` on the public Studio API and `snake_case` for Catalyst APIs, so this would create yet another casing scheme which IMO gets pretty messy. It is not an absolute requirement here, especially not for this PR that is only a reference to the other one, but it would be good not to add yet another casing to our APIs.

The fix can be done relatively simply as well, just by adding a `json` tag on the same line where you have the `bigquery` tags for the usage rows. Can probably be a later follow-up anyway tho, maybe on the next changes we make to the usage queries, since AFAIK this API is still in beta etc

packages/api/src/schema/api-schema.yaml
```diff
@@ -3145,6 +3171,72 @@ paths:
...
+                      #/components/schemas/usage-metric/properties/CreatorID
+                  DeliveryUsageMins:
+                    $ref: >-
+                      #/components/schemas/usage-metric/properties/DeliveryUsageMins
+                  TotalUsageMins:
+                    $ref: >-
+                      #/components/schemas/usage-metric/properties/TotalUsageMins
+                  StorageUsageMins:
+                    $ref: >-
+                      #/components/schemas/usage-metric/properties/StorageUsageMins
```
victorges: Do we not return all the fields here? Can just reference the whole schema:
```suggestion
                $ref: #/components/schemas/usage-metric
```

You are missing the `required` field for example. It was only done like this on the `/data/views/query/total` API because there, in fact, we don't return all the fields.

victorges (approved):

packages/api/src/schema/api-schema.yaml
```diff
@@ -1806,6 +1806,32 @@ components:
               leftAt:
                 type: integer
                 description: the time the participant left
+    usage-metric:
+      type: object
+      description: |
+        An individual metric about usage of a user.
+      required:
+        - UserID
```
adamsoffer, line 1814: @gioelecerati Should these be camelcased? ie `userId`?

PR #1762, opened 2023-06-08
api/controllers: Throw errors on bad filters

victorges:
This is to start throwing errors on bad filters specified on the request which improve
API ergonomy. Every time I try to write one myself I end up doing something wrong
and have no feedback that it is wrong.

We've been logging those cases for a while now to make sure no one is using, and
[no one is.](https://eu-metrics-monitoring.livepeer.live/grafana/explore?orgId=1&left=%7B%22datasource%22:%22P8E80F9AEF21F6940%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P8E80F9AEF21F6940%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bapp%3D%5C%22prod-livepeer-api%5C%22%7D%20%7C%3D%20%5C%22parseFilters:%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%22now-14d%22,%22to%22:%22now-7d%22%7D%7D)

 - Wrap `parseFilter` into a function that wraps errors as a 400
 - Change logs into thrown errors
 - Check that `filter.value` is provided as well (not logged before, but feels relatively specific/safe)
 - Check that the specified comparison is valid (not logged before, but feels relatively specific/safe)

call the API with bad filters

Only my own internal ones.

**Checklist**

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -441,8 +441,10 @@ function parseFiltersRaw(fieldsMap: FieldsMap, val: string): SQLStatement[] {
               case "lte":
                 comparison = "<=";
                 break;
-              default:
+              case "eq":
                 comparison = "=";
+              default:
+                throw new Error(`unknown comparison: "${key}"`);
             }
             q.push(
               sql``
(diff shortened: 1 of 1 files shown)


gioelecerati (approved):
LGTM

PR #1769, opened 2023-06-09
access-control: increase webhook-timeout and follow redirects

gioelecerati:

- Increased webhook timeout to 30 seconds - 5 seconds is not enough. The time to do an RPC call to infura to check if you have an NFT, and it's gone. Especially if you have multiple conditions in your webhook logic

Also, address redirects

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/util.ts
+++ b/packages/api/src/util.ts
@@ -84,6 +84,64 @@ export const fetchWithTimeout = (
     );
   });
 
+export const fetchWithTimeoutAndRedirects = async (
+  url: string,
+  options: RequestInitWithTimeout,
+  redirectCount = 5
+): Promise<Response> => {
+  // Throw error if maximum number of redirects has been exceeded
+  if (redirectCount < 0) {
+    throw new Error("Maximum number of redirects exceeded");
+  }
+
+  options = { ...options, redirect: "manual" };
+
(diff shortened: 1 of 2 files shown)


packages/api/src/util.ts
```diff
@@ -84,6 +84,64 @@ export const fetchWithTimeout = (
...
+          return;
+        }
+
+        clearTimeout(timeout);
+
+        // Handle redirects
+        if (response.status >= 300 && response.status < 400) {
+          const newUrl = response.headers.get("location");
+          if (!newUrl) {
+            reject("Redirect with no location");
```
victorges, line 118: Can you add some info on the err msg?
```suggestion
            reject(`Redirect response status=${response.status} with no location`);
```

packages/api/src/util.ts
```diff
@@ -84,6 +84,64 @@ export const fetchWithTimeout = (
...
+            return;
+          }
+
+          const newResponse = await fetchWithTimeoutAndRedirects(
+            newUrl,
+            options,
+            redirectCount - 1
```
victorges, line 125: Neat

victorges (approved):
We ended up having to add support for that hahah
LGTM

PR #1770, opened 2023-06-12
api: Add support for trusted IPFS gateway inputs

victorges:

This is to support users to provide gateway URLs as input and we parse the corresponding
dStorage protocol URL in case the gateway belongs to a trusted list.

Will fix the use case of Lens which is using their own gateway as input.

 - Add a logic to translate gateway URLs into dStorage URLs on upload
 - Add API to migrate existing assets
 - Some tests I can't run locally

`yarn test`

Ad-hoc

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -115,14 +115,18 @@ function parseUrlToDStorageUrl(
   const urlObj = new URL(url);
   const path = urlObj.pathname;
 
-  const pathElements = path.split("/");
-  const isIpfs = pathElements.length < 3 || pathElements[1] !== "ipfs";
+  let pathElements = path.split("/");
+  if (pathElements.length > 0 && pathElements[0] === "") {
+    pathElements = pathElements.slice(1);
+  }
+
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -106,12 +108,49 @@ function cleanAssetTracks(asset: WithID<Asset>) {
       };
 }
 
+function parseUrlToDStorageUrl(
```
0xcadams, line 111: You could also use `parseIpfs` for this 😄 

https://github.com/livepeer/livepeer.js/blob/main/packages/core/src/utils/storage/ipfs.ts

0xcadams (approved):

packages/api/src/controllers/asset.ts
victorges: Yeah i wanted something like that but couldn't find one for arweave. Since we already gate this transformation through a trusted gateway list though, we can be pretty sure that if the CID is invalid the import is going to fail and the user not get a fake association with a CID or tx ID. So that should be enough for now.

PR #1773, opened 2023-06-15
billing: automated billing & pay as you go usage billing

gioelecerati:

- Create new products
- Create pay as you go products
- Report overusage
- Display upcoming invoices
- Automated billing

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+  // Return a response to acknowledge receipt of the event
+  return res.sendStatus(200);
+});
+
+async function sleep(millis) {
+  return new Promise((resolve) => setTimeout(resolve, millis));
+}
+
+// Migrate existing users to stripe
+app.post("/migrate-users-to-stripe", async (req, res) => {
```
victorges: Let's delete this endpoint

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+interface OverUsageItem {
+  units: number;
+  total: number;
+}
+
+interface BillingUsageData {
+  DeliveryUsageMins: number;
+  TotalUsageMins: number;
+  StorageUsageMins: number;
+}
```
victorges: can remove

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+
+interface OverUsageItem {
+  units: number;
+  total: number;
+}
+
+interface BillingUsageData {
+  DeliveryUsageMins: number;
+  TotalUsageMins: number;
+  StorageUsageMins: number;
```
victorges: Reminder to make it lowerCamel

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+  scale_1: "prod_O9XtcfOSMjSD5L",
+};
+
+export const reportUsage = async (req, adminToken) => {
+  const [users] = await db.user.find(
+    [
+      sql`users.data->>'stripeProductId' IN ('growth_1', 'scale_1', 'prod_O9XtHhI6rbTT1B','prod_O9XtcfOSMjSD5L')`,
+    ],
+    {
+      limit: 9999999999,
```
victorges, line 26: This limit will be forced to 1000 internally. We should ideally make this paginated or otherwise create a way to force a query without a limit (but make sure you don't allow _users_ to make that)

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+      useReplica: true,
+    }
+  );
+
+  const [oldProPlanUsers] = await db.user.find(
```
victorges, line 34: Can you add a `// TODO` somewhere to remove this temp/test plan logic?

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+  const payAsYouGoUsers = [...users, ...oldProPlanUsers];
+
+  let updatedUsers = [];
+  for (const user of payAsYouGoUsers) {
+    const userSubscription = await req.stripe.subscriptions.retrieve(
+      user.stripeCustomerSubscriptionId
+    );
+
+    const billingCycleStart = userSubscription.current_period_start * 1000; // 1685311200000 // Test date
+    const billingCycleEnd = userSubscription.current_period_end * 1000; // 1687989600000 // Test date
```
victorges: can remove comments

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+      {} as Record<string, string>
+    );
+
+    try {
+      // Invoice items based on overusage
+      await Promise.all(
+        products[user.stripeProductId].usage.map(async (product) => {
+          if (product.name === "Transcoding") {
+            await req.stripe.subscriptionItems.createUsageRecord(
+              subscriptionItemsByLookupKey["transcoding_usage"],
```
victorges: Ideally we would use the lookup keys in the config instead of hardcoded strings. Probably fine for now.

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+    // create a map of subscription items by their lookup keys
+    const subscriptionItemsByLookupKey = subscriptionItems.data.reduce(
+      (acc, item) => {
+        acc[item.price.lookup_key] = item.id;
+        return acc;
+      },
+      {} as Record<string, string>
+    );
+
+    try {
```
victorges, line 48: Would be nice to wrap the whole iteration with the `try/catch` to make sure we don't fail-fast on errors in the above calls. Ideally could also move the `try` body to a helper func so we can have a clearer `try/call/catch/updateResult` logic here

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+  return {
+    updatedUsers: updatedUsers,
+  };
+};
+
+const calculateOverUsage = async (product, usage) => {
+  const limits = {
+    transcoding: product?.usage[0].limit,
+    streaming: product?.usage[1].limit,
+    storage: product?.usage[2].limit,
```
victorges: Better rely on the config `name` instead of fixed indexes

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+  };
+
+  const overUsage = {
+    TotalUsageMins: Math.max(usage?.TotalUsageMins - limits.transcoding, 0),
```
victorges: **nit:** Might make sense to remove all the `?` in this func since the objs are never null

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+      headers: {
+        Authorization: `Bearer ${adminToken}`,
+      },
+    }
+  ).then((res) => res.json());
+
+  return usage;
+};
+
+// Webhook handler for asynchronous events.
```
victorges: Can you update this comment to explain when/by whom this is called?

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+    // Notify via email
+    const invoice = event.data.object;
+    const [users] = await db.user.find(
+      { stripeCustomerId: invoice.customer },
+      { useReplica: false }
+    );
+
+    // notify help@livepeer.org
+    await sendgridEmailPaymentFailed({
+      email: "help@livepeer.org",
```
victorges, line 271: Might be worth a CLI arg for this.

Also, I think the implementation isn't this an array?
```ts
  const [supportName, supportEmail] = email;
```

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+      userId: users[0].id,
+      invoiceId: invoice.id,
+    });
+  }
+
+  // Return a response to acknowledge receipt of the event
+  return res.sendStatus(200);
+});
+
+async function sleep(millis) {
```
victorges, line 282: We already have this on an `utils` file, can reuse that

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+
+      // sleep for a 200 ms to get around stripe rate limits
+      await sleep(200);
+    }
+  }
+
+  res.json(users);
+});
+
+app.post("/migrate-personal-users", async (req, res) => {
```
victorges: This should be admin-only (`authorizer({anyAdmin: true})` or sth like that)

Edit: maybe its fine cause theres a requried secret below

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+
+    if (currentUsers.length === 0) {
+      keepGoing = false;
+      continue;
+    }
+
+    users = users.concat(currentUsers);
+
+    for (let index = 0; index < currentUsers.length; index++) {
+      let user = currentUsers[index];
```
victorges: **nit:** can be for-of
```suggestion
    for (const user of currentUsers) {
```

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+
+  const batchSize = 100;
+  let cursor = null;
+  let users = [];
+  let keepGoing = true;
+  let firstUserId = null;
+
+  let migratedUsers = [];
+
+  while (keepGoing) {
```
victorges: Can consider removing this outer loop and calling the API multiple times instead. Can make the `batchSize` configurable via query-string to fine-tune it during execution

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+  let cursor = null;
+  let users = [];
+  let keepGoing = true;
+
+  let migratedUsers = [];
+  let firstUserId = null;
+
+  while (keepGoing) {
+    const [currentUsers, newCursor] = await db.user.find(
+      [sql`users.data->>'stripeProductId' = 'prod_1'`],
```
victorges: Might make sense to check the `suspended` field of users as well (soccer)

packages/api/src/controllers/stripe.ts
```diff
@@ -0,0 +1,720 @@
...
+          stripeCustomerPaymentMethodId: null,
+        });
+
+        // sleep for a 200 ms to get around stripe rate limits
+        await sleep(200);
+
+        // Update user's newStripeProductId to null
+        await db.user.update(user.id, {
+          newStripeProductId: null,
+        });
```
victorges: Move to the same update above

packages/api/src/controllers/usage.ts
```diff
@@ -32,6 +34,19 @@ app.post(
   async (req, res) => {
     let { fromTime, toTime } = req.query;
 
+    const authHeader = req.headers.authorization;
+    let token = "";
+    if (authHeader) {
+      token = authHeader.split(" ")[1];
+    }
```
victorges: You can probably just use req.token (and even avoid the temproary var, just sending as arg below)
```suggestion
    const token = req.token;
```

packages/api/src/controllers/usage.ts
```diff
@@ -32,6 +34,19 @@ app.post(
...
+
+    // New automated billing usage report
+    let result = await reportUsage(req, token);
+
+    res.status(200);
+    res.json(result);
+    return;
```
victorges: Delete the code below instead

packages/api/src/controllers/user.ts
```diff
@@ -908,7 +894,7 @@ app.post(
     );
 
     // Temporarily skip updating the subscription if user is selecting a new plan
-    if (
+    /*if (
```
victorges, line 909: Remove commented code

packages/api/src/controllers/user.ts
```diff
@@ -963,25 +949,63 @@ app.post(
 
     */
```
victorges, line 962: Remove commented code above as well

packages/api/src/controllers/user.ts
```diff
@@ -963,25 +949,63 @@ app.post(
...
-            deleted: true,
-            clear_usage: true,
-            price: item.price.id,
-          })),
-          ...items.data.map((item) => ({
-            price: item.id,
-          })),
-        ],
+    let updatedSubscription;
+    if (products[payload.stripeProductId].deprecated) {
```
victorges, line 965: We might not need this if our frontend is the only one calling

packages/api/src/controllers/user.ts
```diff
@@ -963,25 +949,63 @@ app.post(
...
+              clear_usage: true,
+              price: item.price.id,
+            })),
+            ...items.data.map((item) => ({
+              price: item.id,
+            })),
+          ],
+        }
+      );
+    } else {
```
victorges, line 985: TBC here

packages/api/src/controllers/stripe.ts
victorges: Tho keep in mind suspended users may be unsuspended. Maybe we want to migrate them anyway

packages/api/src/controllers/user.ts
```diff
@@ -963,25 +949,63 @@ app.post(
...
+      let payAsYouGoItems = [];
+      if (products[payload.stripeProductId].payAsYouGo) {
```
victorges: might wanna be safer with a `?` on the product, or first check if the product exists before using and returning an explicit error if not.

packages/api/src/controllers/user.ts
```diff
@@ -963,25 +949,63 @@ app.post(
...
+        payload.stripeCustomerSubscriptionId,
+        {
+          billing_cycle_anchor: "now", // reset billing anchor when updating subscription
+          items: [
+            ...subscriptionItems.data.map((item) => {
+              // Check if the item is metered
+              const isMetered = item.price.recurring.usage_type === "metered";
+              return {
+                id: item.id,
+                deleted: true,
```
victorges, line 1008: What does this mean?

packages/api/src/controllers/user.ts
```diff
@@ -1026,6 +1050,31 @@ app.post(
   }
 );
 
+app.post(
+  "/retrieve-upcoming-invoice",
```
victorges, line 1066: Should this be a `GET /user/invoice/:stripeCustomerId` instead? 
Or maybe even just `GET /user/invoice` and the API just uses`req.user.stripeCustomerId` to grab the invoice.

packages/www/components/PlanForm/index.tsx
```diff
@@ -428,12 +428,12 @@ const PlanForm = ({
                 <Box css={{ mt: "$4" }}>
                   <Text variant="neutral">
                     You are currently using the{" "}
-                    {products[user.newStripeProductId]?.name ||
-                      products[stripeProductId]?.name}{" "}
+                    {products[stripeProductId]?.name ||
+                      products[user.newStripeProductId]?.name}{" "}
```
victorges, line 438: Is there any case where `products[stripeProductId]?.name` won't exist? Maybe the old order still makes sense as users get migrated to the new plans and have the `newStripeProductId` cleared out (then we can remove this logic completely).

packages/www/components/Plans/index.tsx
```diff
@@ -251,19 +251,19 @@ const Plans = ({
...
               <Heading as="h3" size="3" css={{ mb: "$3", fontWeight: 600 }}>
-                {products["hacker_1"].name}
+                {products["prod_O9XuIjn7EqYRVW"].name}
               </Heading>
               <Box css={{ mb: "$4", fontSize: "$2" }}>Free</Box>
               <PlanForm
                 text={
                   dashboard
-                    ? newStripeProductId === "hacker_1"
+                    ? stripeProductId === "hacker_1"
```
victorges: Is this ever true? Didn't you say that the actual production plans will be in the `prod_xxx` format? Same in all similar comparisons below.

packages/www/components/UpcomingInvoiceTable/index.tsx
```diff
@@ -1,70 +1,141 @@
...
-          <Td>${transcodingPrice} / min</Td>
-          <Td>
-            {usage &&
-              `$${(
-                (usage.sourceSegmentsDuration / 60) *
-                transcodingPrice
-              ).toFixed(2)}`}
-          </Td>
-        </Tr>
+    <>
```
victorges: I think this surrounding `<> </>` is unnecessary since you only have 1 element inside. It's only necessary to create a fake parent when you want to return a list of elements from a react component. But I'm not an expert here.

packages/www/components/UsageSummary/index.tsx
```diff
@@ -84,13 +101,28 @@ const UsageSummary = () => {
...
+        usage = {
+          TotalUsageMins: 0,
+          DeliveryUsageMins: 0,
+          StorageUsageMins: 0,
+        },
+      ] = await getBillingUsage(fromTime, toTime);
+
       if (res.status == 200) {
         setUsage(usage);
+        doCaculateOverUsage(usage);
```
victorges: Did you miss an `await`?

packages/www/components/UsageSummary/index.tsx
```diff
@@ -101,10 +133,73 @@ const UsageSummary = () => {
...
+        TotalUsageMins: Math.max(usage?.TotalUsageMins - limits.transcoding, 0),
+        DeliveryUsageMins: Math.max(
+          usage?.DeliveryUsageMins - limits.streaming,
+          0
+        ),
+        StorageUsageMins: Math.max(usage?.StorageUsageMins - limits.storage, 0),
+      };
+
+      return overUsage;
+    };
```
victorges: Don't we have this same logic on the API? Feels like it should be exposed and called from here, not re-calculated in the frontend. 

An alternative could be to just fetch (directly or indirectly) the stripe state here? We should avoid making too many calls to BigQuery which is expensive, so it would be good if we could just re-use the state from the cronjob instead of re-querying and calculating on the dashboard on every reload.

You might want to consider leaving this as a future optimization as well, can you make sure to keep this tracked somewhere? For example can be tasks on the Linear projects for all these bits we talked about.

Maybe the same for the function below, though not sure if I saw the same thing on the backend.

packages/www/hooks/use-api/endpoints/user.ts
```diff
@@ -227,9 +235,34 @@ export const getBillingUsage = async (
...
+      } as BillingUsageData | ApiError,
+    ];
+  }
+
   return [res, usage as BillingUsageData | ApiError];
 };
 
+export const getUpcomingInvoice = async (
+  stripeCustomerId: string
+): Promise<[Response, any | ApiError]> => {
```
victorges, line 250: Can you add proper typing (no `any` for our own stuff) if/when you have a chance? Seems like there's already an `UpcomingInvoice` type declared?

packages/www/pages/dashboard/billing/index.tsx
```diff
@@ -63,12 +80,97 @@ const Billing = () => {
...
+      if (overusage) {
+        const oBill = await calculateOverUsageBill(overusage);
+        setOverUsageBill(oBill);
+        let [res, uInvoice] = await getUpcomingInvoice(user.stripeCustomerId);
+        setUpcomingInvoice(uInvoice?.invoices);
+        setUpcomingInvoiceTotal((uInvoice?.invoices.total / 100) | 0);
       }
     };
+
+    const calculateOverUsage = async (usage) => {
```
victorges, line 112: Is this the same code from the other component? If it's not easy to re-use the backend logic, can you move it to a helper and avoid repeating it between the 2 frontend components?

packages/www/pages/dashboard/usage/index.tsx
```diff
@@ -47,7 +51,14 @@ const Usage = () => {
...
   const doSetTimeStep = async (ts: string) => {
     setTimestep(ts);
-    const [res, usage] = await getBillingUsage(from, to, null, ts);
+    const [
+      res,
+      usage = {
+        TotalUsageMins: 0,
+        DeliveryUsageMins: 0,
+        StorageUsageMins: 0,
+      },
```
victorges, line 60: I think this default is redundant with the one you already added to `getBillingUsage` impl?

packages/www/pages/dashboard/usage/index.tsx
```diff
@@ -84,25 +95,37 @@ const Usage = () => {
...
+      toTime: any,
+      status: any
+    ) => {
+      fromTime = fromTime * 1000;
+      toTime = toTime * 1000;
 
-      doSetFrom(fromTime);
-      doSetTo(toTime);*/
+      // if subscription is cancelled, get current month data
+      if (status === "canceled") {
```
victorges, line 107: Hm, seems like this logic could be inside `getBillingUsage` instead?

packages/www/pages/dashboard/usage/index.tsx
```diff
@@ -84,25 +95,37 @@ const Usage = () => {
...
+
+      let [res, usage] = await getBillingUsage(fromTime, toTime);
 
-      const [res, usage] = await getBillingUsage(
-        fromTime * 1000,
-        toTime * 1000
-      );
       if (res.status == 200) {
+        if (!usage) {
+          usage = {
```
victorges, line 117: Same for redundant default.

victorges:
@gioelecerati Tested the frontend but I got 0 usage on all users I tried. Even though on the home page it even shows a billing amount that seems to imply some usage, when I click on "view billing" or "view usage details" inside that there is `0.0` in all items. Not sure what is going on there.

This one is billing-test-7:
Screenshot 2023-07-11 at 20 14 24

But on home it showed `Upcoming invoice: $1775`

victorges (approved):
Didn't get to see it working E2E on the frontend as the usage was all 0, but it's probably some simple bug on the bigqueries or something.

LGTM either way, LFG!

packages/www/public/sitemap-0.xml
victorges, line 1: FYI you can always just revert the changes in this file. It gets updated automatically when you build and i have no idea why, but no need to keep the changes afaik.

gioelecerati:
> @gioelecerati Tested the frontend but I got 0 usage on all users I tried. Even though on the home page it even shows a billing amount that seems to imply some usage, when I click on "view billing" or "view usage details" inside that there is `0.0` in all items. Not sure what is going on there.
> 
> This one is billing-test-7: Screenshot 2023-07-11 at 20 14 24
> 
> But on home it showed `Upcoming invoice: $1775`

It's normal you are getting 0 as usage becase we removed the fixed timestamps with the june billing cycle (which is where the test usage data is residing). So July is empty!

packages/api/src/controllers/user.ts
gioelecerati: If the plan is metered (e.g. growth or scale), we clear the usage on plan update, because different plans has different overusage limits

packages/www/components/Plans/index.tsx
gioelecerati: Yes, they are staging product, just updated all the IDs to prod product for the deploy

packages/api/src/controllers/user.ts
gioelecerati: This api is not getting used any way for now, so I will leave it like this and update it later

PR #1774, opened 2023-06-21
Add endpoints to get and update room participants

mjh1:
Add endpoints to the rooms API to be able to get and update participants.

### DIFF
--- a/packages/api/src/controllers/room.ts
+++ b/packages/api/src/controllers/room.ts
@@ -190,16 +190,7 @@ app.post(
         ttl: 5 * 60,
       }
     );
-    at.addGrant({
-      roomJoin: true,
-      room: req.params.roomId,
-      ...(req.body.roomAdmin !== undefined && {
-        roomAdmin: req.body.roomAdmin,
-      }),
-      ...(req.body.canPublish !== undefined && {
-        canPublish: req.body.canPublish,
-      }),
-    });
+    at.addGrant({ roomJoin: true, room: req.params.roomId });
     const token = at.toJwt();
 
     res.status(201);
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/room.ts
```diff
@@ -190,7 +190,16 @@ app.post(
...
-    at.addGrant({ roomJoin: true, room: req.params.roomId });
+    at.addGrant({
+      roomJoin: true,
+      room: req.params.roomId,
+      ...(req.body.roomAdmin !== undefined && {
+        roomAdmin: req.body.roomAdmin,
+      }),
+      ...(req.body.canPublish !== undefined && {
+        canPublish: req.body.canPublish,
+      }),
```
mjh1: @victorges do you know of a neater way of doing this?

victorges: Not really :/
The only trick I know, which I'm not sure would work here, is that `JSON.serialize` does not serialize `undefined` fields (as there's no value for that in JSON). So if this eventually gets JSON-serialized, you could just set `roomAdmin: req.body.roomAdmin` and it will work out. If we don't really control that code and think it might change in the future, I guess it could be risky.

One nit I'd suggest is destructuring the `req.body` into local variables, so these expressions are a little simpler. Something like this on line 184
```ts
const { name, roomAdmin, canPublish} = req.body;
```
then use the shoter vars around here. But it'd still need the `...(!x ? {} : { x })` logic anyway

victorges: Oh another idea! Just noticed these are simple booleans, which probably have a default on the LiveKit side. Assuming the default for both is `false`, we can do 
```suggestion
      roomAdmin: req.body.roomAdmin ?? false,
      canPublish: req.body.canPublish ?? false,
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1071,6 +1071,10 @@ components:
       properties:
         name:
           type: string
+        roomAdmin:
+          type: boolean
+        canPublish:
+          type: boolean
```
victorges: Can you add some docs (`description`) here to show up in API reference?

victorges (approved):
LGTM

packages/api/src/controllers/room.ts
```diff
@@ -220,6 +220,54 @@ app.delete("/:roomId/user/:participantId", authorizer({}), async (req, res) => {
...
+    const svc = new RoomServiceClient(
+      req.config.livekitHost,
+      req.config.livekitApiKey,
+      req.config.livekitSecret
+    );
+    await svc.updateParticipant(
+      req.params.roomId,
+      req.params.participantId,
+      undefined,
+      req.body.canPublish !== undefined
```
victorges, line 263: What about that `admin` (or sth like that) other config?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2923,6 +2932,56 @@ paths:
...
+            type: string
+        - name: userId
+          in: path
+          required: true
+          schema:
+            type: string
+      summary: Get user details
+      responses:
+        "200":
+          description: Success
```
victorges, line 2953: Can you add the response schema here as well? (`content` field)

You don't even need to define it as a separate component above if you don't already have it (tho fine to create it if you prefer). You can inline it like new assets endpoint: https://github.com/livepeer/studio/pull/1774/files#diff-bc9d02c3e9f64abe5a66b2dff920f42d87e200509c0b087384b1644bfad52c2cL2744

packages/api/src/schema/api-schema.yaml
```diff
@@ -2923,6 +2932,56 @@ paths:
...
+            type: string
+        - name: userId
+          in: path
+          required: true
+          schema:
+            type: string
+      summary: Update a users permissions
+      responses:
+        "204":
+          description: Success
```
victorges, line 2994: Same for specifying response schema

victorges (approved):
LGTM!

packages/api/src/schema/api-schema.yaml
mjh1: This is just a 204 no content

packages/api/src/controllers/room.ts
mjh1: Didn't end up wanting to expose that 👍 we think the ability to change publish permission should be enough

PR #1775, opened 2023-06-22
Store Attestation to IPFS

leszko:

Add storing Attestation in IPFS.

Related task-runner PR: https://github.com/livepeer/task-runner/pull/202

- create a new task type `export-data` which can export any content into IPFS (and potentially other storage providers)
- schedule `export-data` task while creating a new attestation
- update attestation data while receiving the task result
- allow querying attestation by IPFS CID

https://linear.app/livepeer/issue/VID-214/studio-task-runner-add-pinning-video-metadata-to-ipfs

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/experiment/attestation.ts
+++ b/packages/api/src/controllers/experiment/attestation.ts
@@ -85,7 +85,7 @@ app.post("/", validatePost("attestation"), async (req, res) => {
     },
   };
 
-  db.attestation.update([sql`id = ${attestationMetadata.id}`], {
+  db.attestation.update(attestationMetadata.id, {
     storage: attestationMetadata.storage,
   });
 
--- a/packages/api/src/controllers/experiment/attestation.ts
+++ b/packages/api/src/controllers/experiment/attestation.ts
@@ -4,9 +4,11 @@ import { validatePost } from "../../middleware";
 import { ethers } from "ethers";
 import { makeNextHREF, toStringValues } from "../helpers";
 import _ from "lodash";
+import sql from "sql-template-strings";
 import * as fcl from "@onflow/fcl";
 import stringify from "fast-stable-stringify";
 import { Attestation } from "../../schema/types";
+import { taskScheduler } from "../../task/scheduler";
 
 const app = Router();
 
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/experiment/attestation.ts
```diff
@@ -65,10 +67,41 @@ app.post("/", validatePost("attestation"), async (req, res) => {
...
+  attestationMetadata.storage = {
+    status: {
+      phase: "waiting",
+      tasks: {
+        pending: task.id,
+      },
+    },
+  };
+
+  db.attestation.update([sql`id = ${attestationMetadata.id}`], {
```
victorges: **nit:** `update` supports a string input for the id
```suggestion
  db.attestation.update(attestationMetadata.id, {
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1323,8 +1286,25 @@ components:
               $ref: "#/components/schemas/task/properties/params/properties/upload"
             export:
               $ref: "#/components/schemas/export-task-params"
-            transcode:
+            export-data:
```
victorges: Any reason to use kebab-case here? If not, would prefer to keep the camelCase almost-consistency in the API today

packages/api/src/schema/api-schema.yaml
```diff
@@ -2052,7 +2015,97 @@ components:
...
+                $ref: "#/components/schemas/ipfs-file-info/properties"
+                updatedAt:
+                  readOnly: true
+                  type: number
+                  description: |
+                    Timestamp (in milliseconds) at which IPFS export task was updated
+                  example: 1587667174725
+            status:
+              $ref: "#/components/schemas/storageStatus"
+    ipfs:
```
victorges: Can you rename this to `ipfs-export-params`?

packages/api/src/task/scheduler.ts
```diff
@@ -202,6 +202,22 @@ export class TaskScheduler {
           });
         }
         break;
+      case "export-data":
+        if (task.params["export-data"].type === "attestation") {
```
victorges: I believe you need to handle the failure case as well (IIRC a `failTask` function in this scheduler) to update the status to `failed` etc. Otherwise it'd be pending forever

victorges (approved):
LGTM, just some nits/naming and a simple change on scheduler

packages/api/src/schema/api-schema.yaml
```diff
@@ -1483,6 +1463,20 @@ components:
                       description: |
                         URL to access metadata file via HTTP through an IPFS
                         gateway
+            exportData:
```
victorges, line 1466: Even though our API schema itself is mostly in `camelCase`, I think the components in the OpenAPI specs are in `kebab-case`. So this should be `export-data` instead (and `storage-status`, etc). 

This is way less important than the actual user-facing fields tho, so don't bother if it's a pain to rename it now

packages/api/src/schema/api-schema.yaml
leszko: Updated in baf2406b29246e9e5db495bd099522d5e128a7c3

Though, I think we don't have any consistency here. Maybe we should think about cleaning up this whole file/api at some point.

packages/api/src/task/scheduler.ts
leszko: Good point! Updated in 58a5a60c263cff7638266e9da03a382c2510dcce

packages/api/src/schema/api-schema.yaml
leszko: Renamed `storage-status`. But I'll keep `exportData` since it's not a component, just a key/field. And we use camelCase for other cases like it.

packages/api/src/schema/api-schema.yaml
leszko: Updated

packages/api/src/schema/api-schema.yaml
victorges: I see, good to know. Yeah we def need to fix this schema further. It actually gets a bunch of errors if you feed it to some OpenAPI validator and we probably need to setup some linter or sth if we want to guarantee any real consistency.

PR #1783, opened 2023-06-27
Don't leak IDs from API endpoints

thomshutt:
This stops returning information from **write** API endpoints where the return value is unused.

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -140,7 +140,7 @@ async function createSubscription(
     customer: stripeCustomerId,
   });
   if (existing.data.length > 0) {
-    return new InternalServerError("existing subscription already found");
+    return null;
   }
 
   const prices = await stripe.prices.list({
@@ -314,6 +314,10 @@ app.post("/", validatePost("user"), async (req, res) => {
       defaultProductId,
       customer.id
     );
+    if (!subscription) {
+      res.status(400);
+      res.json({ errors: ["error creating subscription"] });
+    }
     stripeFields = {
       stripeCustomerId: customer.id,
       stripeCustomerSubscriptionId: subscription.id,
@@ -829,6 +833,11 @@ app.post(
       stripeProductId,
       stripeCustomerId
     );
+    if (!subscription) {
+      return res
+        .status(400)
+        .send({ errors: ["could not create subscription"] });
+    }
 
     if (
       stripeProductId !== "prod_0" &&
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -578,7 +578,7 @@ app.post("/verify-email", validatePost("verify-email"), async (req, res) => {
       });
     }
   }
-  res.status(200);
+  res.status(200).json({});
 });
 
 app.post(
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -676,7 +676,7 @@ app.post(
     );
     if (newToken) {
       res.status(201);
-      res.json(newToken);
+      res.json({});
     } else {
       res.status(403);
       res.json({ errors: ["error creating password reset token"] });
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -716,8 +716,10 @@ app.post(
         stripeCustomerId: customer.id,
       });
       res.status(201);
+      res.json(customer);
+    } else {
+      res.status(400);
     }
-    res.json(customer);
   }
 );
 
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -723,6 +723,7 @@ app.post(
       res.json(customer);
     } else {
       res.status(400);
+      res.json({ errors: ["error creating customer"] });
     }
   }
 );
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -131,7 +131,7 @@ async function getOrCreateCustomer(stripe: Stripe, email: string) {
   return await stripe.customers.create({ email });
 }
 
-async function getOrCreateSubscription(
+async function createSubscription(
   stripe: Stripe,
   stripeProductId: StripeProductIDs,
   stripeCustomerId: string
@@ -140,7 +140,7 @@ async function getOrCreateSubscription(
     customer: stripeCustomerId,
   });
   if (existing.data.length > 0) {
-    return existing.data[0];
+    return new InternalServerError("existing subscription already found");
   }
 
   const prices = await stripe.prices.list({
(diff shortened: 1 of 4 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -140,7 +140,7 @@ async function getOrCreateSubscription(
     customer: stripeCustomerId,
   });
   if (existing.data.length > 0) {
-    return existing.data[0];
+    return null;
```
thomshutt, line 143: This was the only bit I wasn't sure about - I'm not sure what the correct way to control flow in this way is in Typescript

victorges: I think this function should stay as a "get or create". The idea I think is that a user should always have a subscription on the stripe side, but only one. So this func is idempotent and after being called we have that guarantee and the existing or new subscription is returned.

If you want to hide fields from the subscription in the response instead, WDYT of creating a function like [`cleanUserFields`](https://github.com/livepeer/livepeer-com/blob/8891a083255bd13ffa4f4b10422e71a267da2231/packages/api/src/controllers/user.ts#L48)? Or perhaps not even return anything (204) if we don't need it in the frontend. 

Would only need to change [this bit](https://github.com/livepeer/livepeer-com/blob/8891a083255bd13ffa4f4b10422e71a267da2231/packages/www/hooks/use-api/endpoints/user.ts#L343) I guess? The `subscription` object is returned as the response in some apis here

packages/api/src/controllers/user.ts
```diff
@@ -574,7 +578,7 @@ app.post("/verify-email", validatePost("verify-email"), async (req, res) => {
       });
     }
   }
-  res.status(200).json(cleanUserFields(user));
+  res.status(200).json({});
```
victorges, line 581: I think this one might be needed in some places in the end [1], I think we should update `cleanIUserFields` to remove the PII fields instead. Possibly only for JWT or CORS API key requests, which go to a frontend (backends should be able to fetch it). WDYT?

[1] check this on that [user.ts file](https://github.com/livepeer/livepeer-com/blob/8891a083255bd13ffa4f4b10422e71a267da2231/packages/www/hooks/use-api/endpoints/user.ts) (impl & usage) I mentioned in the other comment

packages/api/src/controllers/user.ts
```diff
@@ -716,8 +720,11 @@ app.post(
         stripeCustomerId: customer.id,
       });
       res.status(201);
+      res.json(customer);
```
victorges, line 723: Interesting that we have a log right above saying this API is deprecated. Perhaps check Loki to see if it's still being used and let's kill it if not?

victorges (commented):
Thanks for fixing this! I think there are some details to be adjusted

packages/api/src/controllers/user.ts
thomshutt: Yeah, I think it's actually redundant in the "create user" flow because it gets created separately as part of that but wasn't sure if it's used elsewhere

packages/api/src/controllers/user.ts
thomshutt: Hiding fields makes me nervous, it feels like there's a lot of scope for things to go wrong as names change, different endpoints need to access different bits etc.

We can discuss in the other thread about what's needed for the "subscription" creation but are you saying something is depending on the email verification trigger endpoint returning a user object? It wasn't clear from the linked file

victorges: Yeah agree, I'm [guilty as hell](https://github.com/livepeer/livepeer-com/blob/6c8f97eb3f70ec632f4b6c0624dcdc06412b0822/packages/api/src/controllers/asset.ts#L552) but also do not enjoy the practice. I think on this API it would be better than my mess though, since we know exactly what we are returning and make the clean-up knowing exactly the schema being cleaned-up (thx JS).

> are you saying something is depending on the email verification trigger endpoint returning a user object?

To be honest I don't know! I sent the file just so you could check what response types were needed or not. I wasn't sure if you had found that file since there were no changes to it, but maybe that is really the case and it's a simpler fix for us 😌

victorges: To be clear, if we are not using those response bodies in our frontend, let's just remove them like you are doing! Less cruft here and attack vectors in the future.

thomshutt: > I wasn't sure if you had found that file since there were no changes to it
Yeah, I have been looking through that too, but definitely appreciate extra eyes on whether I overlooked something in there.

AFAICT for this `verify-email` endpoint, the frontend does nothing with the response. I've also pushed this branch to staging and the page + email sending works correctly as before.

packages/api/src/controllers/user.ts
thomshutt: Right, I don't think we have any flows calling this where we'd expect the user to have an existing subscription though and so I thought it might be cleaner to explicitly mark this as a "create" until we need that.

packages/api/src/controllers/user.ts
thomshutt: Checked again and I was wrong - the bit in the "create user" for (POST to "/user") is wrapped in a `if (req.stripe)` and so I think we are still relying on the frontend calling this endpoint

packages/api/src/controllers/user.ts
victorges: Oh I see. I have never touched this code myself so I don't really know what I'm talking about lol

I thought the flow was sth like:
 - Registration: `POST /user` always creates a subscription with the default product id (free), so every user always has a subscription, it just might be the free one. Here it would be a create.
 - When the user changes their plan, `POST /create-subscription` is called, which has a misleading name and actually _updates_ the existing subscription, doesn't actually create one. Here it would be a get.

I haven't checked the frontend code to know what API it calls when user changes plan tho, I just assumed that partt.

If that's not the case then LGTM!

packages/api/src/controllers/user.ts
victorges: @thomshutt FYI `req.stripe` doesn't come from the request. It's a config [injected on every request object](https://github.com/livepeer/livepeer-com/blob/6c8f97eb3f70ec632f4b6c0624dcdc06412b0822/packages/api/src/app-router.ts#L165) :harold_pain:
Not sure if this changes anything tho or I misunderstood.

victorges (approved):
LGTM! Only replied on existing discussions but you already know this code better than me :P

packages/api/src/controllers/user.ts
thomshutt: Ah okay, maybe we do receive it then - I'd prefer not to remove for this PR since I've tested in its current form, but will do a followup with that removed and check if the signup flow still does what we expect

packages/api/src/controllers/user.ts
thomshutt: Hmm possibly, I can't tell from the code exactly what we do when changing plan so will do some testing

gioelecerati: `create-subscription` is a deprecated api anyway, so the only place where this function is getting called is on user creation. 
When the user changes plan, `update-subscription` is called, which uses the fields into the user table.

packages/api/src/controllers/user.ts
gioelecerati: I see you're mentioning it here as well, I double checked in Loki, `create-subscription` is not getting called. Either way, this would still be a `create` call, so there are no places where it's used to get the subscription

gioelecerati (approved):
LGTM - Will have some conflicts with the automated billing PR that I will resolve, but apart of that this shouldn't break anything

PR #1785, opened 2023-06-28
Add support for setting publish permission when adding participants

mjh1:


packages/api/src/controllers/room.ts
```diff
@@ -190,7 +190,11 @@ app.post(
         ttl: 5 * 60,
       }
     );
-    at.addGrant({ roomJoin: true, room: req.params.roomId });
+    at.addGrant({
+      roomJoin: true,
+      room: req.params.roomId,
+      canPublish: req.body.canPublish ?? true,
```
victorges, line 196: noice

victorges (approved):

PR #1786, opened 2023-06-29
Only display tooltip when playback is ready

mjh1:
Fixes the tooltip logic to only display when a playback url is available.

- upload asset
- wait for upload to complete
- observe asset in list


victorges (approved):
LGTM

PR #1787, opened 2023-06-29
api: prettier, ignore the data directory

iameli-streams:
Otherwise it tries to lint it and causes errors if it doesn't have permissions to do so.


victorges (approved):
Who are you @iameli-streams

iameli:
@victorges My stream-safe account; no private repos or keys or whatnot

PR #1788, opened 2023-06-30
Improve rooms API reference

mjh1:

### DIFF
--- a/packages/api/src/controllers/room.ts
+++ b/packages/api/src/controllers/room.ts
@@ -259,7 +259,7 @@ app.get("/:roomId/user/:participantId", authorizer({}), async (req, res) => {
   res.json({
     id: participant.identity,
     state: participant.state,
-    joinedAt: participant.joinedAt,
+    joinedAt: participant.joinedAt * 1000,
     name: participant.name,
     permission: participant.permission,
     isPublisher: participant.isPublisher,
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -1055,11 +1055,15 @@ components:
           example: d32ae9e6-c459-4931-9898-e86e2f5e7e16
         joinUrl:
           type: string
-          description: Joining URL
+          description:
+            Joining URL - use this for Livepeer's default meeting app (see the
+            multiparticipant streaming guide for more info).
           example: https://meet.livepeer.chat
         token:
           type: string
-          description: Joining JWT
+          description:
+            Joining JWT - this can be used if you have a custom meeting app (see
+            the multiparticipant streaming guide for more info).
           example: token
     get-room-user-response:
       type: object
@@ -1070,8 +1074,8 @@ components:
           example: d32ae9e6-c459-4931-9898-e86e2f5e7e16
         joinedAt:
           type: integer
-          description: The time the user joined
-          example: 1687771565
+          description: Timestamp (in milliseconds) at which the user joined
+          example: 1687517025261
         name:
           type: string
           description: The display name of the user
@@ -1080,6 +1084,7 @@ components:
           type: boolean
           description: Whether a user is allowed to publish audio/video tracks
           example: true
+          default: true
     room-egress-payload:
       type: object
       required:
@@ -1100,6 +1105,7 @@ components:
           type: boolean
           description: Whether a user is allowed to publish audio/video tracks
           example: true
+          default: true
     transcode-payload:
       additionalProperties: false
       required:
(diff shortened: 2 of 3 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -1033,9 +1033,53 @@ components:
...
+          type: string
+          description: The ID of the room
+          example: d32ae9e6-c459-4931-9898-e86e2f5e7e16
+    room-user-response:
+      type: object
+      properties:
+        id:
+          type: string
+          description: The ID of the user
+          example: d32ae9e6-c459-4931-9898-e86e2f5e7e16
```
victorges, line 1055: Are user IDs also UUIDs? Thought they could be sth else

packages/api/src/schema/api-schema.yaml
```diff
@@ -1033,9 +1033,53 @@ components:
...
+        joinUrl:
+          type: string
+          description: Joining URL
+          example: https://meet.livepeer.chat
+        token:
+          type: string
+          description: Joining JWT
+          example: token
```
victorges: Does join URL already contain JWT? Might be worth being more explanatory in the description and explain how they can be used.

packages/api/src/schema/api-schema.yaml
```diff
@@ -1033,9 +1033,53 @@ components:
...
+    get-room-user-response:
+      type: object
+      properties:
+        id:
+          type: string
+          description: The ID of the user
+          example: d32ae9e6-c459-4931-9898-e86e2f5e7e16
+        joinedAt:
+          type: integer
+          description: The time the user joined
```
victorges: Can you specify the unit? Can copy the same format we use for other `...At` timestamps in this file

packages/api/src/schema/api-schema.yaml
```diff
@@ -1033,9 +1033,53 @@ components:
...
+          example: 1687771565
+        name:
+          type: string
+          description: The display name of the user
+          example: name
+        isPublisher:
+          type: boolean
+          description: Whether a user is allowed to publish audio/video tracks
+          example: true
```
victorges, line 1086: Can you add a `default` here too? I assume it is `true`? Default also serves as example

packages/api/src/schema/api-schema.yaml
```diff
@@ -1763,8 +1810,10 @@ components:
             properties:
               eventName:
                 type: string
+                example: event name
```
victorges: This description is literally the same words as the field itself xD

victorges (commented):
draft pre-review

packages/api/src/schema/api-schema.yaml
mjh1: That was me just setting the example so that we didn't get 'null' coming out in the docs

packages/api/src/schema/api-schema.yaml
mjh1: Yep, I just generate them here: https://github.com/livepeer/studio/blob/master/packages/api/src/controllers/room.ts#L201

packages/api/src/controllers/room.ts
```diff
@@ -259,7 +259,7 @@ app.get("/:roomId/user/:participantId", authorizer({}), async (req, res) => {
   res.json({
     id: participant.identity,
     state: participant.state,
-    joinedAt: participant.joinedAt,
+    joinedAt: participant.joinedAt * 1000,
```
victorges, line 262: Is this the only place where we return a timestamp from livekit?

packages/api/src/schema/api-schema.yaml
```diff
@@ -1054,6 +1104,7 @@ components:
         canPublish:
           type: boolean
           description: Whether a user is allowed to publish audio/video tracks
+          example: true
```
victorges, line 1107: Does this one default to true as well? A `default` flag might be more useful than an `example` in these cases where the default behavior is not the "zero" value for the type.
```suggestion
          default: true
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1767,8 +1819,10 @@ components:
             properties:
               eventName:
                 type: string
+                example: event name
               timestamp:
```
victorges: Could be worth a description with the unit as well

victorges (approved):
LGTM

packages/api/src/schema/api-schema.yaml
mjh1: Makes sense, the auto generated docs seem to need the example field though, so should I set both? I'll do that for now, but let me know.

packages/api/src/controllers/room.ts
mjh1: There's another one too, are you thinking it needs a helper function?

packages/api/src/schema/api-schema.yaml
victorges: Ah, yeah, do that. My comment was based on existing OpenAPI tools that we've used in the past like redoc, but I'm actually not sure what our current implementation is doing.

PR #1795, opened 2023-07-10
Revert removing recording fields from webhook

leszko:

During the Recording V2 work we removed two fields from the webhook payload:
- `recordingUrl`
- `mp4Url`

This PR adds them back for the sake of backwards-compatibility.

**Screenshots (optional)**

**Checklist**

- [ ] ~~My change requires a change to the documentation.~~
- [ ] ~~I have updated the documentation accordingly.~~
- [ ] ~~I have added tests to cover my changes.~~

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -264,6 +264,12 @@ export class TaskScheduler {
       const asset = await db.asset.get(task.outputAssetId);
       if (asset && asset.source.type == "recording" && asset.source.sessionId) {
         const session = await db.session.get(asset.source.sessionId);
+        const externalSession = await toExternalSession(
+          this.config,
+          session,
+          null,
+          true
+        );
         if (session) {
           await this.queue.publishWebhook("events.recording.ready", {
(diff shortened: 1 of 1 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -264,6 +264,12 @@ export class TaskScheduler {
       const asset = await db.asset.get(task.outputAssetId);
       if (asset && asset.source.type == "recording" && asset.source.sessionId) {
         const session = await db.session.get(asset.source.sessionId);
+        const externalSession = await toExternalSession(
+          this.config,
+          session,
+          null,
+          true
+        );
         if (session) {
```
victorges: Should this `toExternalSession` go to inside this `if`? Since the `session` object can apparently be null (don't know how, but the check may be there for a reason)

victorges (approved):
LGTM

packages/api/src/task/scheduler.ts
leszko: Yes, right. Updated.

PR #1797, opened 2023-07-12
Expose participant metadata for rooms

mjh1:

This was a customer requested feature. We are enabling the option to store user defined metadata for each participant.

Local testing plus the same testing on staging.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/room.ts
+++ b/packages/api/src/controllers/room.ts
@@ -128,6 +128,16 @@ app.post(
       }
     }
 
+    const svc = new RoomServiceClient(
+      req.config.livekitHost,
+      req.config.livekitApiKey,
+      req.config.livekitSecret
+    );
+    const rooms = await svc.listRooms([req.params.roomId]);
+    if (rooms.length < 1) {
+      throw new BadRequestError(`room not currently running`);
+    }
+
     const stream = await db.stream.get(req.body.streamId);
     if (
       !stream ||
@@ -206,6 +216,7 @@ app.post(
         name: req.body.name,
         identity: id,
         ttl: 5 * 60,
+        metadata: req.body.metadata,
       }
     );
     at.addGrant({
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/room.ts
```diff
@@ -128,6 +128,16 @@ app.post(
       }
     }
 
+    const svc = new RoomServiceClient(
+      req.config.livekitHost,
+      req.config.livekitApiKey,
+      req.config.livekitSecret
+    );
+    const rooms = await svc.listRooms([req.params.roomId]);
+    if (rooms.length < 1) {
```
thomshutt, line 137: Do we need to error if it's > 1 or safe to just ignore that case?

thomshutt (approved):

packages/api/src/controllers/room.ts
mjh1: Good point yeah it shouldn't be possible unless there was an issue on livekit's side, but makes sense to add a check to be made aware of it if it ever did happen.

victorges (commented):
posthumous LGTM

packages/api/src/controllers/room.ts
```diff
@@ -281,7 +297,7 @@ app.put(
     await svc.updateParticipant(
       req.params.roomId,
       req.params.participantId,
-      undefined,
+      req.body.metadata,
       req.body.canPublish !== undefined
```
victorges, line 301: nit/style suggestion for the future: as we add more fields to the request payload, I think it's useful sometimes to make it clear in the beginning of the API handler what are all the user input that will be considered in the func. TypeScript has some nice syntaxes for doing it in a clean way and getting all types to the fields as well, which improves code safety.

It could look like this:
```ts
const { roomId, participantId } = req.params as Record<string, string>; // might want to use `toStringValues` helper instead
const { name, metadata, canPublish } = req.body as RoomUserUpdatePayload;
```
(or you can destructure the whole thing in a single statement, but I find that one a bit weirder sometimes)

Then use all the well-typed fields in the function, which can often also cut some extra chars (lol) when constructiong objects (e.g. instead of `{ metadata: metadata }` one could just do `{ metadata }`)

Everything only style anyway, just tripping ;)

PR #1800, opened 2023-07-14
api: we don't use this table anymore, right?

iameli-streams:
I don't think we do. And it's the only thing remaining that stops us from using CockroachDB, which I'd kind of like to do.


victorges:
@iameli I'm actually not sure, but would love to get rid of it too. Would you mind querying the production DB and checking the timestamp of the latest records? Just to make sure we're not adding info to it anymore (I'm pretty sure we're not querying it, maybe only in some lost metabase dashboards).

iameli-streams:
this got merged

PR #1801, opened 2023-07-15
put www in api

iameli-streams:
A few changes related to putting @livepeer/www in the box.

* The API server now embeds the Next.js production server and hosts the Studio frontend on all routes that aren't `/api`. This shouldn't be a breaking change for anything, as our production routes only send `/api` traffic to the API server anyway.
* The `pkg` static binary for `@livepeer/www` has been removed. It was only ever used in livepeer-in-a-box, and livepeer-in-a-box wants the API too.
* I got frustrated trying to adapt the API Docker build to the new system -- it got confusing quickly, as the API server now has a dependency on a built `www` existing -- so I threw my hands up and just put the `pkg`ed version into the container. In theory, this should have identical behavior while being as small as a regular Docker build. I'll make sure to test it pretty extensively though.
* Per discussion, I have added a CI "canary" of sorts to verify that www can be built into static HTML/JS/CSS files, which is something that might be useful in the future if we were embedding the frontend in a Go binary without a V8 runtime. We're not doing that yet, though, so all we do is verify that the static build works at all.
* I wanted to make us CockroachDB compatible. The only things we had that weren't supported by Cockroach were clauses that looked like `user_email character varying(512) COLLATE pg_catalog."default" NOT NULL` in the CDN usage tables that we don't use for anything anymore. So I just took them out. Probably we could adapt them if they're worth preserving, but I doubt that they are.


0xcadams (commented):
Wait, so we wouldn’t use Vercel anymore for deployment? I don’t agree with that change - I think that would be a net negative and I’m not sure what we gain, other than syncing deployments (which our backend is in node, if we want to make these sync, this can be deployed to Vercel)

iameli:
@0xcadams I think we covered this on Discord, but just to clarify publicly: nah, we'll still use Vercel, this is just for a different use case

.github/workflows/docker.yaml
```diff
@@ -51,6 +48,9 @@ jobs:
 
       - name: Login to DockerHub
         uses: docker/login-action@v2
+        if:
```
hjpotter92, line 51: i think this `if` clause isn't really necessary.

.github/workflows/static.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - "v*"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  static:
+    name: Do a static build
+    runs-on: ubuntu-20.04
```
hjpotter92: not `ubuntu-22`? (or `ubuntu-latest`)

packages/api/Dockerfile
```diff
@@ -1,30 +1,13 @@
-FROM	node:16	as	builder
-WORKDIR	/app
+FROM node:18 as builder
```
hjpotter92, line 1: i was contemplating if we switch it to `node:lts` but our builds from caches might not pick up when lts switches from 18 to 20 (and also dependabot won't know what to do)

packages/api/Dockerfile
```diff
@@ -1,30 +1,13 @@
...
-
-COPY	packages/api/tsconfig.json	.
-COPY	packages/api/src	src
-RUN	yarn run prepare
-
-FROM	node:16
 WORKDIR /app
+COPY . .
+RUN	yarn install --frozen-lockfile --ignore-scripts
+RUN yarn run prepare:lerna
```
hjpotter92, line 6: do we not need `NODE_ENV` here for builds?

packages/api/Dockerfile
```diff
@@ -1,30 +1,13 @@
...
-ENV	NODE_ENV production
-
-COPY	packages/api/package.json yarn.lock ./
-RUN	yarn install --frozen-lockfile
-COPY --from=builder	/app/dist	dist
-
-ARG	VERSION
-ENV	VERSION ${VERSION}
-ARG	GITHUB_SHA
-ENV	GITHUB_SHA ${GITHUB_SHA}
```
hjpotter92, line 25: version/github-sha are used to report running version to metrics.

.github/workflows/docker.yaml
iameli-streams: It failed when I ran it from a `iameli-streams` fork.

packages/api/Dockerfile
iameli-streams: I hit some problem when jumping to 20, comfortable with 18 for now

packages/api/Dockerfile
iameli-streams: NODE_ENV=production was used to only install production dependencies for prod containers. The pkg version automatically only packages what's needed for production so it doesn't matter... unless we have any logic that depends on that, I suppose...

iameli-streams: Oh lol I guess we do ```
▶ git grep NODE_ENV | wc -l
23
```

File: packages/api/Dockerfile (hunk 5 above)
Comment (iameli) [reply packages/api/Dockerfile:25]: added

File: packages/api/Dockerfile [hunk 6]
```diff
@@ -1,30 +1,18 @@
-FROM	node:16	as	builder
-WORKDIR	/app
+FROM node:18 as builder
```
victorges, line 1: Can you correspondingly upgrade the node version in our CI as well? e.g. https://github.com/livepeer/studio/pull/1801/files?diff=split&w=1#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R31

packages/api/package.json
```diff
@@ -109,7 +109,8 @@
     "whatwg-fetch": "^3.4.0",
     "winston": "^3.2.1",
     "wolfy87-eventemitter": "^5.2.9",
-    "yargs": "^17.3.1"
+    "yargs": "^17.3.1",
+    "@livepeer.studio/www": "0.15.0"
```
victorges, line 113: Isn't this a cyclic dependency? Or is that fine?

packages/api/src/app-router.ts
```diff
@@ -260,7 +262,8 @@ export default async function makeApp(params: CliArgs) {
   if (fallbackProxy) {
     app.use(proxy({ target: fallbackProxy, changeOrigin: true }));
   }
-  app.use(errorHandler());
```
victorges, line 263: I think we still need the errorHandler, it is not a "not found handler"!
https://expressjs.com/en/guide/error-handling.html#writing-error-handlers

packages/api/src/app-router.ts
```diff
@@ -260,7 +262,8 @@ export default async function makeApp(params: CliArgs) {
...
+  const wwwHandler = await require("@livepeer.studio/www");
```
victorges: Dos this import need to be dynamic? Could have it as 
`import wwwPromise from '...'`
on top then here just 
`.use(await wwwPromise)`

It's generally better for TypeScript inference (and maybe some build tooling) so I generally try to avoid raw `require` function.

packages/api/src/index.ts
```diff
@@ -18,12 +18,15 @@ const prefix = "livepeer_api_";
...
-  return (
-    process.env.VERSION ||
-    process.env.GITHUB_SHA ||
-    require("child_process").execSync("git rev-parse HEAD").toString().trim() ||
-    "undefined"
-  );
+  let version = process.env.VERSION || process.env.GITHUB_SHA;
+  if (!version) {
+    try {
+      require("child_process").execSync("git rev-parse HEAD").toString().trim();
```
victorges: don't you need to assign `version` or `return` here?

packages/www/lib/utils/index.tsx
```diff
@@ -109,6 +109,23 @@ export const getStripe = () => {
   return stripePromise;
 };
 
+/**
+ * should stripe be enabled in this context?
+ */
+export const shouldStripe = () => {
```
victorges, line 115: would be really funny if someone typo'd the `e` out 🌚

packages/www/next.config.js
```diff
@@ -337,40 +334,56 @@ const config = {
   },
 };
 
-module.exports = withSentryConfig(
-  withPlugins(
+if (!isStaticBuild) {
+  config = {
+    ...config,
+    i18n: {
+      locales: ["en", "es"],
```
victorges, line 341: do we have our dashboard in spanish? 0o

packages/www/pages/_app.tsx
```diff
@@ -65,12 +66,8 @@ const livepeerClient = createReactClient({
...
     // eventually should move to using JWT from user's login
     apiKey: "",
-    baseUrl: isStaging()
-      ? "https://livepeer.monster/api"
-      : "https://livepeer.studio/api",
-    webrtcIngestBaseUrl: isStaging()
-      ? "https://webrtc.livepeer.monster/webrtc"
-      : "https://webrtc.livepeer.studio/webrtc",
+    baseUrl: `${getEndpoint()}/api`,
+    webrtcIngestBaseUrl: `${getEndpoint()}/webrtc`,
```
victorges, line 70: Neat! Keep in mind that the webrtc endpoints had a `webrtc.` subdomain though. maybe specify it as an optional arg to `getEndpoint()`? Or does it work without it?

packages/www/server.js
```diff
@@ -1,47 +1,24 @@
...
-  };
-  console.log(JSON.stringify(description));
-  process.exit(0);
-}
-
-app.prepare().then(() => {
-  createServer(handle).listen(cli.port, (err) => {
-    if (err) throw err;
-    console.log(`> Ready on http://localhost:${cli.port}`);
+const getApp = async () => {
```
victorges, line 10: I have no idea what's going on around here, first time I see this file lol

victorges (commented):
LGTM! Will leave for @0xcadams for the official approval

packages/api/package.json
iameli: I thought that would be a problem too, but it's apparently fine in the context of a Yarn workspace. It helps that there's no actual circular import occurring — www imports api for certain schema definitions, and builds them into a bundle using the `next build` process; api then imports `server.js` from `www` which only serves out that compiled bundle.

packages/api/src/index.ts
iameli: Oh. Yeah.

packages/www/lib/utils/index.tsx
iameli: lmao

packages/www/next.config.js
iameli: I guess not but we're prepared for it

packages/www/server.js
iameli: It's this! Except now we're exporting a handler function instead of an actual HTTP server. https://nextjs.org/docs/pages/building-your-application/configuring/custom-server

packages/api/Dockerfile
iameli-streams: Done

packages/api/src/app-router.ts
iameli-streams: Good catch, I was about to break a lot of error handling. I went ahead and added it to `prefixRouter` instead of `app` so its behavior kicks in for all `/api` endpoints but Next.js can still handle other errors however it wants.

packages/www/pages/_app.tsx
iameli-streams: I'm still confused what webrtc. is for and was planning to let this get to staging and verify that hitting the same URLs via `playback.livepeer.monster` resulted in the same result :shrug:

0xcadams: Yeah, it's the same, and we didn't have a giant reason for doing playback different from ingest. The only reason was for possible separation of concerns, if we ever want to route ingest differently somehow. But this looks good

0xcadams (approved):
LGTM!

packages/www/pages/_app.tsx
iameli-streams: Right on. I like using the same one as much as possible, for exactly the use case I'm working on of making it easy to set up a single-node server that does everything.

PR #1806, opened 2023-07-20
Live Multistream Target Switching

leszko:

[Design Doc](https://www.notion.so/livepeer/Live-Multistream-Target-Switching-Simplified-Design-Doc-575e0150a0084fceb32d7d630dcb69db#ddb9daca3e93485fafde9625811efd95)

**Notes:**
- I thought for some time about how to unit test this change, but ended up with something very complex, so decided to not include the unit tests. Saying that if you see any easy way to add tests, let me know.\
- I added a query for streams related to the given multistream target; we may consider adding a GIN index; on the other hand I added `userId` to the query, so that may be good enough

- UI: Allow changing Mutlistram targets while the stream is live
- Trigger Catalyst event when the multistream target is updated

fix https://linear.app/livepeer/issue/VID-296/live-multistream-update-studio-api-and-ui

**Screenshots (optional)**

**Checklist**

- [ ] ~I have added tests to cover my changes.~


packages/api/src/controllers/helpers.ts
```diff
@@ -523,6 +523,27 @@ export function isValidBase64(str: string) {
...
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      resource: "stream",
+      playback_id: playbackId,
+    }),
+  };
+
+  fetch(url, options);
```
victorges: I think we need to `await` or `return` the `fetch` promise, unless this is meant to happen in background. If that's the case, we might want to do a `.catch()` on the unawaited promises, to log the error and avoid possibly crashing the node process (not 100% sure this still happens). We also probably don't want to make this function `async`, with it handling the internal errors instead

victorges (approved):
LGTM!

packages/api/src/controllers/helpers.ts
leszko: Updated to be a sync execution (added `await`) in 82c5773adf73bf4ea0c195d02e64ddbb4a321d87

PR #1807, opened 2023-07-20
fix schema references

suhailkakar:

The current version of `api-schema.yaml` is having errors on OpenAPI validations and cannot be used by any parser and etc. This PR fixes those errors.

 - moved `object-store-patch-payload `, `stream-set-active-payload`, and `signing-key-response-payload` to `db-schema.yaml`
 - updated `playbackid` to `playbackId` (same as param) 
 - for 4 components removed and `$ref` and instead just added the params

- `yarn compile-schemas` and `yarn test`

**Screenshots (optional)**

**Checklist**

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -159,53 +159,6 @@ components:
                 Timestamp (in milliseconds) at which the webhook last was
                 triggered
               example: 1587667174725
-    webhook-response:
-      type: object
-      required:
-        - webhookId
-        - statusCode
-      additionalProperties: false
-      properties:
-        id:
-          type: string
-          readOnly: true
-          example: de7818e7-610a-4057-8f6f-b785dc1e6f88
-        webhookId:
-          readOnly: true
-          type: string
-        eventId:
-          readOnly: true
-          type: string
-        createdAt:
-          readOnly: true
-          type: number
-          description: |
-            Timestamp (in milliseconds) at which webhook response object was
-            created
-          example: 1587667174725
-        duration:
-          type: number
-        statusCode:
-          type: number
-          default: 0
-        response:
-          type: object
-          additionalProperties: false
-          properties:
-            body:
-              type: string
-            headers:
-              type: object
-              additionalProperties:
-                type: array
-                items:
-                  type: string
-            redirected:
-              type: boolean
-            status:
-              type: number
-            statusText:
-              type: string
     stream:
       type: object
       required:
@@ -386,27 +339,6 @@ components:
           $ref: "#/components/schemas/stream/properties/record"
         multistream:
           $ref: "#/components/schemas/stream/properties/multistream"
-    deactivate-many-payload:
-      type: object
-      additionalProperties: false
-      properties:
-        ids:
-          type: array
-          minItems: 1
-          items:
-            type: string
-    asset-patch-payload:
-      type: object
-      additionalProperties: false
-      properties:
-        name:
-          $ref: "#/components/schemas/asset/properties/name"
-        creatorId:
-          $ref: "#/components/schemas/input-creator-id"
-        playbackPolicy:
-          $ref: "#/components/schemas/playback-policy"
-        storage:
-          $ref: "#/components/schemas/new-asset-payload/properties/storage"
     stream-patch-payload:
       type: object
       additionalProperties: false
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -1000,11 +986,10 @@ components:
                   properties:
                     spec:
                       oneOf:
-                        - type: "null"
+                        - type: object
                         - $ref: >-
                             #/components/schemas/asset/properties/storage/properties/ipfs/properties/spec
                 - type: boolean
-                - type: "null"
```
suhailkakar: @victorges can you please confirm if this is correct?

packages/api/src/schema/api-schema.yaml
```diff
@@ -1663,15 +1648,6 @@ components:
         disabled:
           type: boolean
           description: Disable the signing key to allow rotation safely
-    signing-key-response-payload:
```
victorges, line 1666: This is actually an API object. It should stay here and we should actually change https://github.com/livepeer/livepeer-com/blob/e98108915d4ead04eaec4ba75240084355422d00/packages/api/src/schema/api-schema.yaml#L3466 to reference it instead of the other one which is wrong.

packages/api/src/schema/api-schema.yaml
```diff
@@ -2102,7 +2078,18 @@ components:
...
+                  description: CID of the file on IPFS
+                url:
+                  readOnly: true
+                  type: string
+                  description: URL with IPFS scheme for the file
+                gatewayUrl:
+                  readOnly: true
+                  type: string
+                  description:
+                    URL to access file via HTTP through an IPFS gateway
```
victorges: Actually! I remembered that better than repeating each and every field, we can actually $ref the individual fields and only repeat the list here :)

Can you change the places where you fixed the "inheritance $ref" to be like this instead? suggestion+example:
```suggestion
                cid:
                  $ref: "#/components/schemas/ipfs-file-info/properties/cid"
                url:
                  $ref: "#/components/schemas/ipfs-file-info/properties/url"
                gatewayUrl:
                  $ref: "#/components/schemas/ipfs-file-info/properties/gatewayUrl"
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2115,8 +2102,23 @@ components:
       type: object
       additionalProperties: false
       properties:
-        $ref: >-
-          #/components/schemas/asset/properties/storage/properties/ipfs/properties/spec/properties
+        nftMetadataTemplate:
```
victorges, line 1906: Same for refing the fields instead of repeating their specs

packages/api/src/schema/db-schema.yaml
```diff
@@ -1101,3 +1101,37 @@ components:
...
+          type: number
+          description: Timestamp (in milliseconds) at which the stream started.
+          example: 1587667174725
+    signing-key-response-payload:
+      type: object
+      required:
+        - publicKey
+        - privateKey
+      properties:
+        $ref: "#/components/schemas/signing-key/properties"
```
victorges: This will need to change to individual field `$ref` as well

victorges (approved):
Actually LGTM as the change to $ref individual fields is pretty simple, but let me know if you want another review after you make that change

packages/api/src/schema/api-schema.yaml
suhailkakar: Yes, it is way better! updated it!

packages/api/src/schema/api-schema.yaml
```diff
@@ -1000,11 +1000,10 @@ components:
                   properties:
                     spec:
                       oneOf:
-                        - type: "null"
+                        - type: object
```
victorges: Why this change? I think this is changing the schema. Continuing on the other comment.

packages/api/src/schema/api-schema.yaml
victorges: This is official JSON schema syntax: https://json-schema.org/understanding-json-schema/reference/null.html

Do you know why does these OpenAPI tools not support it? Either way, we probably need to do more here to actually represent a field that can only be `null`.

victorges (approved):
Apart from the null thing, LGTM

packages/api/src/schema/api-schema.yaml
suhailkakar: I believe with `nullable: true` it should be fine?

packages/api/src/schema/api-schema.yaml
```diff
@@ -995,12 +1005,12 @@ components:
                 pinned files, specify an object with a spec field. False or null
                 means to unpin from IPFS, but it's unsupported right now.
               oneOf:
-                - type: object
+                - type: "null"
```
victorges: I think you over-reverted here haha

packages/api/src/schema/api-schema.yaml
victorges: It isn't, because it will start accepting any kind of objects as well not only `null`. We would have to add `nullable` to an existing `oneOf` option here, but the problem is that the only one currently is a `$ref`.

I think there might be a way if we redefined the `spec` object here to tag it as `nullable` (maybe `$ref`'ing the `properties` field instead to avoid too much repetition)

If we decide to ignore these `null` for now, make sure to revert this to `type: "null"`

packages/api/src/schema/api-schema.yaml
suhailkakar: oh, sorry! updated it!

PR #1808, opened 2023-07-20
www/use-analyzer: allow analyzer to run in more contexts

iameli-streams:
@victorges, is there any problem these days with disregarding the `region` parameter and just going right to `https://livepeer.studio/data`?

### DIFF
--- a/packages/www/hooks/use-analyzer.tsx
+++ b/packages/www/hooks/use-analyzer.tsx
@@ -2,6 +2,7 @@ import EventSource from "eventsource";
 import { useApi } from "hooks";
 import { useContext, useMemo, createContext, ReactNode } from "react";
 import { isStaging, isDevelopment, HttpError } from "../lib/utils";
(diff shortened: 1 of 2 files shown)


packages/www/hooks/use-analyzer.tsx
```diff
@@ -140,11 +141,7 @@ export namespace events {
 }
 
 const makeUrl = (region: string, path: string) => {
-  if (isDevelopment() && useLocalAnalyzer) {
-    return `http://localhost:8080/data${path}`;
-  }
-  const tld = isStaging() || isDevelopment() ? "monster" : "com";
-  return `https://${region || defaultRegion}.livepeer.${tld}/data${path}`;
+  return `${getEndpoint()}/data${path}`;
```
victorges, line 144: While at it, can you add the /api prefix to this?

victorges (approved):
No need for the region stuff anymore @iameli-streams !

PR #1817, opened 2023-07-26
fix: fix warnings on stream

0xcadams:


adamsoffer (commented):
@0xcadams Mind adding a screenshot of the changes? I'm getting an error while trying to go live from the browser on staging.
CleanShot 2023-07-26 at 13 26 23@2x

PR #1830, opened 2023-08-02
Live Multistream Target Switching (#1806) (#1828)

leszko:
It's the same PR that was reverted some time ago (#1806)


victorges (approved):
LGTM

PR #1832, opened 2023-08-03
Prevent toggle recording in Studio API when stream is live

leszko:
fix VID-326


packages/api/src/controllers/stream.ts
```diff
@@ -1375,10 +1382,17 @@ app.patch("/:id/record", authorizer({}), async (req, res) => {
     res.status(400);
     return res.json({ errors: ["can't set for session"] });
   }
+  const record = req.body.record;
   if (req.body.record === undefined) {
```
victorges: WDYT of adding a check if `record` field is a boolean here as well? Otherwsie we can end up using different values in different places (there's a hacky conversion to boolean on line 1398)
```suggestion
  if (typeof record !== 'boolean') {
```

victorges (approved):
LGTM

packages/api/src/controllers/stream.ts
leszko: Yeah, good idea.

PR #1833, opened 2023-08-03
billing: fix email webhook & free tier pay as you go & cleanup

gioelecerati:

Key Changes:

- The function sendgridEmailPaymentFailed has been modified to handle different scenarios for sending emails. It checks if the last payment failed within the last 3 days before sending an email, both for payment failed and missing payment method for hackers.

- Introduced logic to handle users on a "pay as you go" plan. It checks if a user's last streamed date is within the last month and then processes their usage.
- There's a check to exclude users with emails containing "@livepeer.org" from the usage report.
- Logic has been added to handle payment failures for users on specific plans.
- The webhook endpoint has been updated to handle the "invoice.payment_failed" event. It sends a notification email when a payment fails.
- A new endpoint subscribe-hackers-to-pay-as-you-go has been introduced. This endpoint handle the migration for hacker users to a "pay as you go" plan.

- Some parts of the code related to updating user subscriptions have been removed, like unused fields for migration and comments

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -15,6 +15,7 @@ import { CreatorId, InputCreatorId, ObjectStore } from "../schema/types";
 import { BadRequestError } from "../store/errors";
 
 const ITERATIONS = 10000;
+const PAYMENT_FAILED_TIMEFRAME = 3 * 24 * 60 * 60 * 1000;
 
 const crypto = new Crypto();
 
@@ -304,10 +305,10 @@ export async function sendgridEmailPaymentFailed({
   invoiceUrl,
   templateId,
 }) {
-  // If user.lastPaymentFailed is newer than 3 days, don't send email
+  // If user.lastEmailAboutPaymentFailure is newer than 3 days, don't send email
   if (
-    user.lastPaymentFailed &&
-    user.lastPaymentFailed > Date.now() - 3 * 24 * 60 * 60 * 1000
+    user.lastEmailAboutPaymentFailure &&
+    user.lastEmailAboutPaymentFailure > Date.now() - PAYMENT_FAILED_TIMEFRAME
   ) {
     return false;
   }
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -299,12 +299,47 @@ export async function sendgridEmail({
 export async function sendgridEmailPaymentFailed({
   email,
   sendgridApiKey,
-  userId,
+  user,
   invoiceId,
+  invoiceUrl,
+  templateId,
 }) {
+  // If user.lastPaymentFailed is newer than 3 days, don't send email
+  if (
+    user.lastPaymentFailed &&
+    user.lastPaymentFailed > Date.now() - 3 * 24 * 60 * 60 * 1000
+  ) {
+    return false;
+  }
+
   const [supportName, supportEmail] = email;
+
+  let subject: string;
+  let text: string;
+
+  if (!invoiceId) {
+    subject = `Free tier user reached usage limit`;
+    text = `User ${user.email} reached usage limit`;
+  } else {
+    subject = `Payment failed for invoice ${invoiceId} for user ${email}`;
+    text = `User ${user.email} failed to pay invoice ${invoiceId}`;
+  }
+
   const msg = {
-    text: `User ${userId} failed to pay invoice ${invoiceId}`,
+    personalizations: [
+      {
+        to: [{ email }],
+        bcc: undefined,
+        dynamicTemplateData: {
+          subject,
+          preheader: `Stripe payment failed`,
+          text,
+          buttonText: `View invoice`,
+          buttonUrl: invoiceUrl,
+          unsubscribe: invoiceUrl,
+        },
+      },
+    ],
     from: {
       email: supportEmail,
       name: supportName,
@@ -313,10 +348,13 @@ export async function sendgridEmailPaymentFailed({
       email: supportEmail,
       name: supportName,
     },
+    // email template id: https://mc.sendgrid.com/dynamic-templates
+    templateId: templateId,
   };
 
   SendgridMail.setApiKey(sendgridApiKey);
   await SendgridMail.send(msg);
+  return true;
 }
 
 export function sendgridValidateEmail(email: string, validationApiKey: string) {
--- a/packages/api/src/controllers/stripe.ts
+++ b/packages/api/src/controllers/stripe.ts
@@ -1,4 +1,4 @@
-import { Router } from "express";
+import { Router, Request } from "express";
 import { db } from "../store";
 import Stripe from "stripe";
 import { products } from "../config";
@@ -6,6 +6,8 @@ import sql from "sql-template-strings";
 import fetch from "node-fetch";
 import qs from "qs";
 import { sendgridEmailPaymentFailed } from "./helpers";
+import { WithID } from "../store/types";
+import { User } from "../schema/types";
 
 const app = Router();
 
@@ -17,7 +19,7 @@ const productMapping = {
   scale_1: "prod_O9XtcfOSMjSD5L",
 };
 
-export const reportUsage = async (req, adminToken) => {
+export const reportUsage = async (req: Request, adminToken: string) => {
   const [users] = await db.user.find(
     [
       sql`users.data->>'stripeProductId' IN ('growth_1', 'scale_1', 'prod_O9XtHhI6rbTT1B','prod_O9XtcfOSMjSD5L')`,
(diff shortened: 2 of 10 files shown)


packages/api/src/controllers/helpers.ts
```diff
@@ -299,12 +299,47 @@ export async function sendgridEmail({
...
+  invoiceUrl,
+  templateId,
 }) {
+  // If user.lastPaymentFailed is newer than 3 days, don't send email
+  if (
+    user.lastPaymentFailed &&
+    user.lastPaymentFailed > Date.now() - 3 * 24 * 60 * 60 * 1000
+  ) {
+    return false;
+  }
```
victorges: **nit:** Can you move this check to whoever calls this function? This function has a single purpose of sending an email, also reflected on its name, and I think it's misleading to start adding other logic to it like awareness about the whole building strategy.

Just for code clarity though, not a required change.

packages/api/src/controllers/helpers.ts
```diff
@@ -299,12 +299,47 @@ export async function sendgridEmail({
...
-  userId,
+  user,
   invoiceId,
+  invoiceUrl,
+  templateId,
 }) {
+  // If user.lastPaymentFailed is newer than 3 days, don't send email
+  if (
+    user.lastPaymentFailed &&
+    user.lastPaymentFailed > Date.now() - 3 * 24 * 60 * 60 * 1000
```
victorges: Can you make this a configurable CLI argument or at least a constant in the top of this file?

packages/api/src/controllers/helpers.ts
```diff
@@ -299,12 +299,47 @@ export async function sendgridEmail({
...
+    subject = `Payment failed for invoice ${invoiceId} for user ${email}`;
+    text = `User ${user.email} failed to pay invoice ${invoiceId}`;
+  }
+
   const msg = {
-    text: `User ${userId} failed to pay invoice ${invoiceId}`,
+    personalizations: [
+      {
+        to: [{ email }],
+        bcc: undefined,
```
victorges, line 333: Should these emails go to an internal address we can track?

packages/api/src/controllers/stripe.ts
```diff
@@ -41,7 +41,18 @@ export const reportUsage = async (req, adminToken) => {
     }
   );
 
-  const payAsYouGoUsers = [...users, ...oldProPlanUsers];
+  // Current date in milliseconds
+  let currentDateMillis = new Date().getTime();
+  let oneMonthMillis = 31 * 24 * 60 * 60 * 1000;
```
victorges: This should also be a constant.

packages/api/src/controllers/stripe.ts
```diff
@@ -41,7 +41,18 @@ export const reportUsage = async (req, adminToken) => {
...
+
+  // One month ago unix millis timestamp
+  let cutOffDate = currentDateMillis - oneMonthMillis;
```
victorges: **nit:** Use `const` instead of `let` whenever you won't need to update the variables. Makes the code generally safer and easier to read on colorful IDEs (consts are usually displayed differently than mutable values)

packages/api/src/controllers/stripe.ts
```diff
@@ -67,6 +78,16 @@ export const reportUsage = async (req, adminToken) => {
 };
 
 async function reportUsageForUser(req, user, adminToken) {
+  if (user.email.includes("@livepeer.org")) {
```
victorges: This is exploitable by creating an email on any other domain that starts with livepeer.org e.g. `dontchargeme@livepeer.org.justcast.it`

Perhaps considering checking for a suffix instead. There's probably a function for it but I don't remember.

Additionally, you might want to ignore any user with `isTestUser` or `isAdmin` (or sth like that).

packages/api/src/controllers/stripe.ts
```diff
@@ -67,6 +78,16 @@ export const reportUsage = async (req, adminToken) => {
 };
 
 async function reportUsageForUser(req, user, adminToken) {
```
victorges: Can you add types here? All these should already exist so no reason to use `any` here.

packages/api/src/controllers/stripe.ts
```diff
@@ -92,6 +113,36 @@ async function reportUsageForUser(req, user, adminToken) {
     subscription: user.stripeCustomerSubscriptionId,
   });
 
+  if (
+    user.stripeProductId.includes("hacker_1") ||
+    user.stripeProductId.includes("prod_O9XuIjn7EqYRVW")
+  ) {
+    // If they don't have a card and in overusage, send a notification email
+    // If they have a card and in overusage, report the usage
```
victorges: Is this comment applicable? It seems like it doesn't send the usage for happy users that have a card, it just sends the payment failed email for all users instead.

packages/api/src/controllers/stripe.ts
```diff
@@ -92,6 +113,36 @@ async function reportUsageForUser(req, user, adminToken) {
...
+        email: "help@livepeer.org",
+        sendgridApiKey: req.config.sendgridApiKey,
+        user,
+        invoiceId: null,
+        invoiceUrl: null,
+        templateId: req.config.sendgridTemplateId,
+      });
+
+      if (emailSent) {
+        db.user.update(user.id, {
```
victorges: Missing `await`.

packages/api/src/controllers/stripe.ts
```diff
@@ -261,29 +327,42 @@ app.post("/webhook", async (req, res) => {
   } else if (event.type === "invoice.payment_failed") {
     // Notify via email
     const invoice = event.data.object;
+    let invoiceUrl = invoice.invoice_pdf;
```
victorges, line 336: Is this right? A "url" is assigned from a "pdf", might be a mistake here.

packages/api/src/controllers/stripe.ts
```diff
@@ -379,6 +453,30 @@ app.post("/migrate-personal-user", async (req, res) => {
           `User ${user.email} has ${subscriptionItems.data.length} subscription items`
         );
 
+        let payAsYouGoItems = [];
+
+        const payAsYouGoPrices = await req.stripe.prices.list({
+          lookup_keys: products["pay_as_you_go_1"].lookupKeys,
+        });
+        payAsYouGoItems = payAsYouGoPrices.data.map((item) => ({
```
victorges: I think this works? no reason to declare the var early right
```suggestion
        const payAsYouGoPrices = await req.stripe.prices.list({
          lookup_keys: products["pay_as_you_go_1"].lookupKeys,
        });
        const payAsYouGoItems = payAsYouGoPrices.data.map((item) => ({
```

packages/api/src/controllers/stripe.ts
```diff
@@ -379,6 +453,30 @@ app.post("/migrate-personal-user", async (req, res) => {
...
+          price: item.id,
+        }));
+
+        if (!req.body.actually_migrate) {
+          res.status(500).json({
```
victorges: Did you really mean a 500 here? Can be confusing when testing dry run
```suggestion
          res.json({
```

packages/api/src/schema/db-schema.yaml
```diff
@@ -995,18 +995,12 @@ components:
...
-            - pay_as_you_go_1
-            - prod_4
-            - prod_O9XuIjn7EqYRVW
-            - prod_O9XtHhI6rbTT1B
-            - prod_O9XtcfOSMjSD5L
-            - prod_O9XuWMU1Up6QKf
+        toMigrate:
+          type: boolean
+          default: true
+        lastPaymentFailed:
```
victorges: Is this `lastPaymentFailed` or `lastEmailAboutPaymentFailure`? Might want to add at least a description explaining what really is this field, in case it is not what the name implies (the other one is too verbose I don't like it either haha)

victorges (approved):
Would've been useful breaking up the changes in separate PRs, kinda massive to understand what is related to what, but the code looks generally ok. LGTM.

packages/api/src/schema/db-schema.yaml
gioelecerati: Good point! Yeah changing to `lastEmailAboutPaymentFailure`

packages/api/src/controllers/stripe.ts
gioelecerati: Yeah, it's a direct link to the invoice so when shih-yu receive the email, he can check it. It's coming with the stripe webhook, and this is the only useful url, unless I build one myself that points to the dashboard

packages/api/src/controllers/stripe.ts
gioelecerati: Right! Should have been endsWith, changing it! nice catch

packages/api/src/controllers/stripe.ts
gioelecerati: Yeah it's referring to the check below the comment, I moved it inside the scope of the condition

packages/api/src/controllers/helpers.ts
gioelecerati: the `email` receiver it's already an internal address, do you think we should send it to a secondary email?

victorges: That's fine then, I thought the email was going to the user

PR #1836, opened 2023-08-07
Fix Studio CI Tests

hjpotter92:
hopefully it fixes the tests


packages/api/src/app-router.ts
```diff
@@ -268,10 +266,19 @@ export default async function makeApp(params: CliArgs) {
...
   if (frontend) {
-    const wwwHandler = await createFrontend();
-    app.use(wwwHandler);
+    wwwHandler = await createFrontend();
   }
+  const errHandler = errorHandler();
+
+  app.use((err, _req, res, _next) => {
+    if (frontend && !_req.path.startsWith(httpPrefix)) {
+      wwwHandler(_req, res, _next);
```
leszko: @iameli I'm not sure this works, because I don't know in which scenarios it is used 🙃 Could you double-check?

hjpotter92: it's being used to setup a proxy from api server to frontend for `livepeer-in-a-box` project.

leszko: Will this work after my change? 🙃

hjpotter92: should be easy to verify by updating the `manifest.yaml` file for that project to your commit in this project. once the build pipeline finishes: https://github.com/livepeer/studio/actions/runs/5938452782/job/16102930628?pr=1836

hjpotter92: *branch + commit sha

victorges: You probably want to check if `err` is null, otherwise this could eat api handler exceptions afaik 

It would be weird for this to be called without an error tho as per the [express error handler docs](https://expressjs.com/en/guide/error-handling.html#writing-error-handlers)

packages/api/src/test-server.ts
```diff
@@ -52,19 +52,28 @@ params.ingest = [
...
 ];
-params.amqpUrl = `amqp://localhost:5672/${testId}`;
+params.amqpUrl = `amqp://127.0.0.1:5672/${testId}`;
 if (!params.insecureTestToken) {
   params.insecureTestToken = uuid();
 }
 params.listen = true;
 params.requireEmailVerification = true;
 params.livekitHost = "livekit";
+params.frontend = false;
```
victorges, line 62: Do we need the app-router changes even with this params change?

victorges (approved):
LGTM either way, especially after confirming if this works when running with the frontend, incl  api error responses etc

packages/api/src/test-server.ts
leszko: I think we still need those changes, because in the actual environment (staging, prod), we still want to return API errors for `/api/*`, not return HTML.

packages/api/src/app-router.ts
leszko: Ok, update to handle `err` in the frontend handler.

leszko (approved):

PR #1838, opened 2023-08-08
api: add esbuild version

iameli:
So the `pkg` deployment is useful but not very good for development. It takes like three full minutes to build it.

This PR adds a different approach; instead of bundling a single static binary including the V8 runtime, it uses esbuild to bundle the entire API server into a single JavaScript file, including all the node_modules dependencies and whatnot. This builds in about ~3 seconds and is executable in any context with a working Node.js runtime, which Catalyst is presumed to have for the forseeable future.


packages/api/package.json
```diff
@@ -37,15 +37,15 @@
     "test:build": "parcel build --no-autoinstall --no-minify --bundle-node-modules -t browser --out-dir ../dist-worker ../src/worker.js",
     "coverage": "yarn run test --coverage",
     "pkg": "pkg --out-path=bin --no-bytecode --public-packages \"*\" --public . --compress GZip ",
-    "pkg:local": "yarn run prepare && yarn run pkg --targets=$(node -p '`node18-${os.platform()}-${os.arch()}`')"
+    "pkg:local": "yarn run prepare && yarn run pkg --targets=$(node -p '`node18-${os.platform()}-${os.arch()}`')",
```
victorges, line 40: Do we still need to keep these?

packages/api/src/app-router.ts
```diff
@@ -197,19 +196,6 @@ export default async function makeApp(params: CliArgs) {
     app.use(`/${insecureTestToken}`, insecureTest());
   }
 
-  // Populate Kubernetes getOrchestrators and getBroadcasters is provided
-  if (kubeNamespace) {
-    app.use(
```
victorges, line 202: This unused?

packages/api/src/middleware/validators.js
```diff
@@ -1,5 +1,5 @@
 // import { schemaWalk } from '@cloudflare/json-schema-walker'
-import validators from "../schema/validators";
+import validators from "../../dist/schema/validators";
```
victorges: Won't this break the builds before a build is ever made? Chicken&egg

packages/api/src/parse-cli.ts
```diff
@@ -78,14 +80,19 @@ export type CliArgs = ReturnType<typeof parseCli>;
...
 // elegant way, I'd love to hear it!
-let args;
+let allOptions;
 const originalOpts = yargs.options;
-yargs.options = function (arg) {
-  args = arg;
-  return originalOpts.call(this, arg);
+yargs.options = function (...args) {
+  if (!allOptions) {
+    allOptions = args[0];
```
victorges, line 87: Hm, I think we sent all process args to that `yargsToMist` function. Won't this break that, not being able to display the help from `-j` arg?

If `args` is actually a list of lists, mind adding a type to the `allOptions` above making it clear what it is?

packages/api/src/parse-cli.ts
```diff
@@ -78,14 +80,19 @@ export type CliArgs = ReturnType<typeof parseCli>;
...
+  }
+  return originalOpts.call(this, ...args);
 };
 
 export default function parseCli(argv?: string | readonly string[]) {
+  if (!argv) {
+    argv = process.argv.slice(2);
```
victorges, line 94: I thought `yargs` did this automatically

victorges (approved):

packages/api/package.json
iameli: Yeah, we're still planning to use the pkg version as the primary thing in the box (if only because I can't get the embedded frontend working well with esbuild) so it's still useful to be able to build that when necessary. But it takes like ~3min instead of ~3sec for the esbuild, so we'll use esbuild in the box dev environment.

packages/api/src/app-router.ts
iameli: Think so. https://livepeer.com/api/region and https://livepeer.com/api/broadcaster both do their own thing.

packages/api/src/parse-cli.ts
iameli: Apparently not the way we're using it here. I'm not super clear why I had to make these changes but esbuild really didn't want to work with the way we were invoking yargs.

packages/api/src/parse-cli.ts
iameli: Yeah, I need to add a type definition to this stuff to unbreak CI. Literally all I'm trying to do here is get the first `.options` call that we're invoking here; all the subsequent calls are apparently internal https://github.com/livepeer/studio/blob/357a24e9a39f5c441cc89270a4415cc857138c93/packages/api/src/parse-cli.ts#L97

packages/api/src/middleware/validators.js
iameli: Nah, the babel compilation doesn't care that the module doesn't exist, it literally just transpiles the source. The schema generation is [its own thing](https://github.com/livepeer/studio/blob/357a24e9a39f5c441cc89270a4415cc857138c93/packages/api/src/compile-schemas.js) that only ever puts files in the `dist` directory. Previously we'd only ever execute from `dist` but now the `esbuild` compiles directly out of `src`, so we need to point to their actual location.

iameli: Wait no you're totally right it writes it to the `src` directory, fixing

PR #1844, opened 2023-08-15
billing: migration: don't update anchor on hacker subscribe

gioelecerati:

Removing the anchor `now` on hacker subscription to pay as you go plans
This is needed because on hacker migration the billing cycle would get reset to today instead of keeping the original subscription date.

**Screenshots (optional)**

**Checklist**


victorges (approved):
Neat

PR #1847, opened 2023-08-16
Populate recordingURL when source playback is ready

mjh1:

Populate recordingURL when source playback is ready

- Populate recordingURL as soon as source playback is available
- Refactor the function to simplify and set the recording v2 status based on the recording vod asset
Tested in staging.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -444,8 +444,12 @@ export async function getRecordingFields(
   );
   return {
     recordingStatus: recordingStatus,
-    recordingUrl: pathJoin(base, "index.m3u8"),
-    mp4Url: pathJoin(base, "source.mp4"),
+    ...(!isReady && !forceUrl
+      ? null
+      : {
+          recordingUrl: pathJoin(base, "index.m3u8"),
+          mp4Url: pathJoin(base, "source.mp4"),
+        }),
   };
 }
 
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -217,11 +217,14 @@ async function getRecordingUrls(
   // Recording V2
   if (session.version === "v2") {
     const asset = await db.asset.getBySessionId(session.id);
-    if (!asset || ["waiting", "processing"].includes(asset.status?.phase)) {
-      // Recording processing in progress
+    if (!asset) {
       return;
     }
     const assetWithPlayback = await withPlaybackUrls(config, ingest, asset);
+    if (!assetWithPlayback.playbackUrl) {
+      // Recording processing in progress
+      return;
+    }
     return {
       recordingUrl: assetWithPlayback.playbackUrl,
       mp4Url: assetWithPlayback.downloadUrl,
@@ -447,9 +450,6 @@ export async function getRecordingFields(
   const isUnused = !session.lastSeen && session.createdAt < readyThreshold;
 
   const recordingStatus = isReady ? "ready" : isUnused ? "none" : "waiting";
-  if (!isReady && !forceUrl) {
-    return { recordingStatus };
-  }
   const recordingUrls = await getRecordingUrls(config, ingest, session);
   if (!recordingUrls) {
     return { recordingStatus: "waiting" };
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -447,9 +450,6 @@ export async function getRecordingFields(
   const isUnused = !session.lastSeen && session.createdAt < readyThreshold;
 
   const recordingStatus = isReady ? "ready" : isUnused ? "none" : "waiting";
-  if (!isReady && !forceUrl) {
```
mjh1, line 450: @victorges do you know what the `forceUrl` param was used for? It's not used anywhere other than these lines I removed. I removed them because I want to just check for the presence of a playbackURL rather than checking the recording status field (to allow us to get source playback through the recordingURL field).

victorges: This is for the API to return a recordingURL even if it might not have been complete yet. I think this was more useful on the recording v1 to be honest, where recordings were playable even while the stream was live and we only had an artificial wait on the API side to make sure the stream was finished.

I think it would be a noop in this flow here, since the `getRecordingUrls` would return `undefined` anyway and we wouldn't be able to force URLs. Does that seem correct? If so, then it should be completely fine removing this.

packages/api/src/controllers/stream.ts
```diff
@@ -447,9 +450,6 @@ export async function getRecordingFields(
   const isUnused = !session.lastSeen && session.createdAt < readyThreshold;
 
   const recordingStatus = isReady ? "ready" : isUnused ? "none" : "waiting";
```
victorges, line 439: I think we should also change this to be more compatible with recording v2. This logic works fine for v1, where the recording readiness was loosely based on when the stream ended (`lastSeen`), but I think we have more accurate information here for recording v2.

So WDYT of fixing this status as well? You can use the same `session.version === "v2"` to tell if we should use the asset state for this, and/or maybe even move this logic to `getRecordingUrls`, or implement it based on the response from `getRecordingUrls`. 

Can also consider having a new `recordingStatus` value here for source playback like `"source"`. But otherwise it should be fine to keep it as `waiting` anyway, just like the asset that gets a `playbackUrl` but stays in `processing`.

victorges (approved):
LGTM with a suggestion for an additional change for the `recordingStatus`

packages/api/src/controllers/stream.ts
mjh1: Ah nice I was wondering what that time based logic was about :)

mjh1: I've done a little refactor which I think makes things much simpler, see what you think

packages/api/src/controllers/stream.ts
```diff
@@ -442,22 +411,42 @@ export async function getRecordingFields(
...
-  return { recordingStatus, ...recordingUrls };
+  const base = pathJoin(
+    ingest,
+    `recordings`,
+    session.lastSessionId ?? session.id
+  );
+  return {
+    recordingStatus: recordingStatus,
+    recordingUrl: pathJoin(base, "index.m3u8"),
+    mp4Url: pathJoin(base, "source.mp4"),
```
victorges: I think you're only missing a check here to only include the URLs when `isReady || forceUrl`. Otherwise we'd be returning URLs for sessions that don't have a recording for v1 (OTOH this will never be the case anymore as we don't create v1 sessions anymore).

Just in case anyway, you can do it with some ts literal trick as well
```suggestion
    ...(!isReady && !forceUrl ? null : {
      recordingUrl: pathJoin(base, "index.m3u8"),
      mp4Url: pathJoin(base, "source.mp4"),
    })
```

victorges (approved):
LGTM

packages/api/src/controllers/stream.ts
victorges: Neat refact, I think it's simpler to follow the logic now with v1/v2 totally separate

packages/api/src/controllers/stream.ts
mjh1: ah well spotted thanks

PR #1854, opened 2023-08-18
billing: enterprise products

gioelecerati:

- Adding enterprise products
- Fixed a typo for storage limits on the hacker plan
- Created an admin API to subscribe users to enteprise plan

**Screenshots (optional)**

**Checklist**


packages/api/src/config.js
```diff
@@ -298,4 +298,50 @@ export const products = {
...
+      },
+      {
+        name: "Storage",
+        description: "Storage (minutes)",
+        price: 0,
+      },
+    ],
+    monthlyPrice: 0,
+  },
+  prod_OTTbwpzxNLMNSh: {
```
victorges, line 324: Curious what's the difference between the 2?

OK, I saw below that one of them is the staging? Can you add any reference here to specify that while we don't store this in the DB properly? `Enterprise (Staging)` could already do it

packages/api/src/controllers/stripe.ts
```diff
@@ -699,4 +700,122 @@ app.post("/migrate-pro-user", async (req, res) => {
   });
 });
 
+app.post(
+  "/subscribe-to-enterprise",
+  authorizer({ anyAdmin: true }),
```
victorges, line 705: Should this be here or under the `/user` APIs? Where is the API to change plans today?

packages/www/pages/dashboard/billing/index.tsx
```diff
@@ -375,10 +377,11 @@ const Billing = () => {
               </Flex>
             </Heading>
           </Flex>
-          {!products[user.stripeProductId]?.order ? (
+          {!standardProducts.includes(products[user.stripeProductId]?.name) ? (
             <Text variant="neutral">
-              The Hacker plan is free of charge up to 1000 minutes per month and
-              limited to 10 concurrent viewers per account.
```
victorges, line 381: Don't we want this text anymore?

victorges (approved):
LGTM

packages/www/pages/dashboard/billing/index.tsx
gioelecerati: Not for hackers! Because they are going to see their upcoming invoice, only custom plan users will see a text

packages/api/src/controllers/stripe.ts
gioelecerati: The update-subscription api is just for users, this one is for admins to move customers to the enterprise plans. After all the migrations are done, we would need to refactor the whole stripe.ts, remove all the migrations api and make only a couple admin apis to handle sporadic migrations

packages/api/src/config.js
gioelecerati: Right, one is staging, I will change the name to specify it

PR #1859, opened 2023-08-22
vod: lower amount of max queued vod jobs per user

gioelecerati:

Just a placeholder if needed due to recent incident 

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/parse-cli.ts
+++ b/packages/api/src/parse-cli.ts
@@ -320,7 +320,7 @@ export default function parseCli(argv?: string | readonly string[]) {
       vodMaxScheduledTasksPerUser: {
         describe:
           "maximum number of tasks that can be in the VOD execution queue for a given user",
-        default: 100,
+        default: 25,
         type: "number",
       },
       "ingest-region": {


packages/api/src/parse-cli.ts
```diff
@@ -320,7 +320,7 @@ export default function parseCli(argv?: string | readonly string[]) {
       vodMaxScheduledTasksPerUser: {
         describe:
           "maximum number of tasks that can be in the VOD execution queue for a given user",
-        default: 100,
+        default: 25,
```
victorges, line 323: This should be changed in the deployment config, no need to change the code. It was made an arg/env exactly so we don't need to make code changes and version deploys to change its value.

victorges (changes requested):

packages/api/src/parse-cli.ts
gioelecerati: Yeah sorry, this was just a placeholder for yesterday night, we actually changed the deployment config! Closing this :)

PR #1862, opened 2023-08-24
Populate recordingURL when source playback is ready (#1847)

mjh1:
This is to re-roll #1847 which was reverted previously in 6336f953b66035f7c280dd428c13544056b87a3e


thomshutt (approved):

thomshutt:
LGTM but looks like tests are failing

PR #1863, opened 2023-08-24
api: allow non-subdomain based bucket endpoints

iameli-streams:
Turns out all we needed to do was to upgrade to Tus 1.0 beta.


packages/api/src/controllers/helpers.ts
```diff
@@ -222,8 +223,9 @@ export async function getObjectStoreS3Config(
     region,
     bucket,
     signingRegion: region,
-    endpoint: `${protocol}//${url.host}`,
+    endpoint: `${protocol}//${url.host}/${bucket}`,
     forcePathStyle: true,
+    s3BucketEndpoint: true,
```
victorges: Where does this field come from and what's the difference from `forcePathStyle`? If I try this locally I get that the field does not exist in this type, maybe that's why CI is failing as well.

packages/api/src/controllers/helpers.test.ts
```diff
@@ -182,6 +183,40 @@ describe("controllers/helpers", () => {
...
+    it("should reject urls with subpaths", async () => {
+      let threw = false;
+      try {
+        await getObjectStoreS3Config({
+          url: "s3+https://user:password@example.com/testbucket/subpath/foo",
+        });
+      } catch (e) {
+        threw = true;
+      }
+      expect(threw).toBe(true);
```
victorges: 😎 
```suggestion
      await expect(getObjectStoreS3Config({
          url: "s3+https://user:password@example.com/testbucket/subpath/foo",
        }).rejects.toThrow("Invalid OS URL path: /testbucket/subpath/foo");
```

packages/api/src/controllers/helpers.test.ts
```diff
@@ -182,6 +183,40 @@ describe("controllers/helpers", () => {
       );
     });
   });
+
+  describe("getObjectStoreS3Config", () => {
+    it("should handle a few URL formats", async () => {
+      const result = await getObjectStoreS3Config({
```
victorges: No idea why that function was made async lol. Can probably make it sync if you can

packages/api/src/controllers/helpers.ts
iameli-streams: I found the documentation here <https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html> and just verified that flipping it off reintroduces the problem. Do you need to add it to the type just above that function in your local setup?

CI's failing because presigned URLs are now getting signed with `/bucket-name/bucket-name`

packages/api/src/controllers/helpers.test.ts
iameli-streams: Ahhh I tried this except without the `rejects` part

packages/api/src/controllers/helpers.test.ts
iameli-streams: Haha yeah I reached the same conclusion

packages/api/src/controllers/helpers.ts
victorges: It doesn't need to be added to the `OSS3Config` type cause we already "contains" `S3ClientConfig`. Maybe the version needs to be updated?

> CI's failing because presigned URLs are now getting signed with /bucket-name/bucket-name

I think this would happen with only your change to the endpoint above. I'm suspecting this new flag is being a noop

iameli-streams: Just tested it again, `s3BucketEndpoint: false` definitely reintroduces the problem. Maybe the types are out-of-date?

victorges: @iameli-streams maybe tus is using a different S3 library than ourselves 👀

iameli-streams:
@victorges Completely reworked to just upgrade Tus and change the types and stuff. Works beautifully.

packages/api/package.json
```diff
@@ -53,6 +53,9 @@
     "@peculiar/webcrypto": "^1.2.3",
     "@sendgrid/client": "^7.6.0",
     "@sendgrid/mail": "^7.6.0",
+    "@tus/file-store": "^1.0.0-beta.2",
+    "@tus/s3-store": "^1.0.0-beta.7",
+    "@tus/server": "^1.0.0-beta.7",
```
victorges, line 58: just to make sure this requires absolutely no change to the clients right?

packages/api/src/controllers/asset.ts
```diff
@@ -1,10 +1,15 @@
...
+import { Request, RequestHandler, Router, Response } from "express";
 import jwt, { JwtPayload } from "jsonwebtoken";
 import { v4 as uuid } from "uuid";
-import mung from "express-mung";
-import tus from "tus-node-server";
+import {
+  Server as TusServer,
+  EVENTS as TUS_EVENTS,
+  Upload as TusUpload,
+} from "@tus/server";
```
victorges, line 10: Alternative:
```suggestion
import * as tus from "@tus/server";
```

then just use `tus.Server`, `tus.EVENTS`, etc

packages/api/src/controllers/asset.ts
```diff
@@ -968,16 +974,16 @@ export const setupTus = async (objectStoreId: string): Promise<void> => {
 async function createTusServer(objectStoreId: string) {
   const os = await getActiveObjectStore(objectStoreId);
   const s3config = await getObjectStoreS3Config(os);
-  const tusServer = new tus.Server({
+  const datastore = new TusS3Store({
+    partSize: 8 * 1024 * 1024,
+    s3ClientConfig: s3config,
```
victorges, line 979: What about `tmpDirPrefix`? Or no need anymore (I think that one's just a temporary local dir tho, should be fine to change it anywya)

packages/api/src/controllers/asset.ts
```diff
@@ -968,16 +974,16 @@ export const setupTus = async (objectStoreId: string): Promise<void> => {
...
     path: "/upload/tus",
     namingFunction,
   });
-  tusServer.datastore = new tus.S3Store({
-    ...s3config,
-    partSize: 8 * 1024 * 1024,
-    tmpDirPrefix: "tus-tmp-files",
-  });
-  tusServer.on(tus.EVENTS.EVENT_UPLOAD_COMPLETE, onTusUploadComplete(false));
+  tusServer.on(TUS_EVENTS.POST_FINISH, onTusUploadComplete(false));
```
victorges, line 986: Is this event the same as the other one? Or better?

At some point we had a couple "lost" uploads, that uploaded the whole file but never called the complete callback. Not sure what could happen here

packages/api/src/controllers/helpers.ts
```diff
@@ -189,11 +188,9 @@ export function deleteCredentials(objectStoreUrl: string): string {
     .replace(secretAccessKey, "***");
 }
 
-export type OSS3Config = S3ClientConfig &
-  Pick<TusS3Opts, "region" | "bucket"> & {
-    accessKeyId: string;
-    secretAccessKey: string;
-  };
+export type OSS3Config = S3ClientConfig & {
+  bucket: string;
```
victorges, line 192: Do we need this for our own stuff? Maybe could even get rid as well 😍

victorges (approved):
Much nicer!

packages/api/src/controllers/asset.ts
iameli: I couldn't find it documented anywhere in the new thing. Dunno.

packages/api/src/controllers/asset.ts
iameli: I think it's the replacement. Different function signature, same information.

packages/api/src/controllers/helpers.ts
iameli: Nah I copied this type definition directly from the Tus internals; their config just adds a `bucket` field to exactly what they push to the AWS client

PR #1865, opened 2023-08-31
fix code coverage

iameli:
Verbose


packages/api/package.json
```diff
@@ -35,7 +35,7 @@
     "test": "jest -i --silent \"${PWD}/src\"",
     "test:dev": "jest \"${PWD}/src\" -i --silent --watch",
     "test:build": "parcel build --no-autoinstall --no-minify --bundle-node-modules -t browser --out-dir ../dist-worker ../src/worker.js",
-    "coverage": "yarn run test --coverage",
+    "coverage": "yarn run test --coverage --verbose",
```
victorges, line 38: Is this required?

victorges (approved):

packages/api/package.json
iameli: I don't think I ever actually fixed anything

victorges: lol ok hahah

PR #1867, opened 2023-08-31
billing: hackers with pay as you go

gioelecerati:

A migration API to subscribe all hackers to pay as you go plans

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stripe.ts
+++ b/packages/api/src/controllers/stripe.ts
@@ -340,6 +340,7 @@ app.post("/hacker/migration/pay-as-you-go", async (req, res) => {
     return res.json({ errors: ["unauthorized"] });
   }
   let migration = [];
+  let invoiceUsage = req.body.invoiceUsage;
 
   let hackerPlan = "prod_O9XuIjn7EqYRVW";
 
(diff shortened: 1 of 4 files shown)


packages/api/src/controllers/stripe.ts
```diff
@@ -82,7 +82,8 @@ export const reportUsage = async (req: Request, adminToken: string) => {
 async function reportUsageForUser(
   req: Request,
   user: WithID<User>,
-  adminToken: string
+  adminToken: string,
+  actuallyReport: boolean = true
```
victorges: ts tip: `: boolean` is auto inferred when you add the default value
```suggestion
  actuallyReport = true
```

packages/api/src/controllers/stripe.ts
```diff
@@ -120,7 +121,7 @@ async function reportUsageForUser(
   });
 
   if (
-    user.stripeProductId.includes("hacker_1") ||
+    (actuallyReport && user.stripeProductId.includes("hacker_1")) ||
```
victorges, line 133: Shouldn't `actuallyReport` be respected when the product ID is `prod_O9XuIjn7EqYRVW` as well? Maybe a parentheses issue going on here or maybe `actuallyReport` has a different meaning than it implies?

packages/api/src/controllers/stripe.ts
```diff
@@ -543,17 +565,21 @@ app.post("/subscribe-hackers-to-pay-as-you-go", async (req, res) => {
       ],
     });
     await db.user.update(user.id, {
-      isActiveSubscription: false,
+      migrated: true,
```
victorges: Glad we're merging all these different flags in one (and having the default being `false` is always safer for any code that forgets to check if it exists!)

victorges (approved):
LGTM

packages/api/src/controllers/stripe.ts
```diff
@@ -375,34 +379,47 @@ app.post("/webhook", async (req, res) => {
...
-    }
-  );
+  let hackerPlan = "prod_O9XuIjn7EqYRVW";
 
-  if (users.length == 0) {
-    res.json({
-      errors: ["no users found"],
-    });
-    return;
+  if (req.body.staging === true) {
```
victorges, line 347: You should probably check something in the config instead of allowing prod to execute staging logic and staging to execute prod logic depending on a request parameter. e.g.
```suggestion
  if (req.config.frontendDomain === 'livepeer.monster') {
```
https://github.com/livepeer/livepeer-infra/blob/66a29742e156bc65887b5ce70941bb1d785a5eb0/values/api.values.yaml#L40

Ideally all these product IDs would come from config and/or DB, as we've already talked though :)

PR #1868, opened 2023-09-01
fix box www fallback

iameli:
4b2fb4cd1 broke the fallback because `wwwHandler` was only registered as a four-argument error handler instead of a regular three-argument handler. I think this fixes while keeping the API handler behavior

### DIFF
--- a/packages/api/src/app-router.ts
+++ b/packages/api/src/app-router.ts
@@ -254,12 +254,16 @@ export default async function makeApp(params: CliArgs) {
   }
   const apiErrorHandler = errorHandler();
 
-  app.use((err, _req, res, _next) => {
-    if (frontend && !_req.path.startsWith(httpPrefix)) {
-      wwwHandler(err, _req, res, _next);
-    }
-    apiErrorHandler(err, _req, res, _next);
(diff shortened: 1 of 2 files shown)


hjpotter92 (approved):

leszko (changes requested):
The change looks fine, but I see the test suite is failing. Could you check?

packages/api/src/app-router.ts
```diff
@@ -254,12 +254,16 @@ export default async function makeApp(params: CliArgs) {
...
   const apiErrorHandler = errorHandler();
 
-  app.use((err, _req, res, _next) => {
-    if (frontend && !_req.path.startsWith(httpPrefix)) {
-      wwwHandler(err, _req, res, _next);
-    }
-    apiErrorHandler(err, _req, res, _next);
-  });
+  if (frontend) {
+    app.use((req, res, next) => {
```
victorges, line 258: Hmmmm, maybe in the end what we really wanted here was doing an `app.all('*', ...)` which registers a catch-all fallback route instead of `app.use` which registers a middleware?

PR #1875, opened 2023-09-01
api: attestation test was missing an await for clearDatabase

iameli:
This was timing out and crashing tests I think


victorges (approved):
nice catch

PR #1877, opened 2023-09-05
Add support for a failover recording bucket

mjh1:

I am enabling support for a failover/secondary recordings bucket to allow us to switch to a new storage location. We can set up the new bucket as the secondary, switch the recording output over to that bucket, then in this code whenever we can't find recordings at the old location we can fetch from the secondary (new) location instead.

Tested with local dev environment

**Screenshots (optional)**

**Checklist**


mjh1:
@victorges let me know what you think, thanks! (I've tested it locally using the 'box' but hopefully will be easy to add a unit test too)

packages/api/src/parse-cli.ts
```diff
@@ -296,6 +296,10 @@ export default function parseCli(argv?: string | readonly string[]) {
         describe: "object store ID used by Catalyst to store recordings",
         type: "string",
       },
+      secondaryRecordObjectStoreId: {
+        describe: "object store ID used by Catalyst to store recordings",
```
victorges: nit
```suggestion
        describe: "secondary object store ID used by Catalyst to store recordings",
```

packages/api/src/webhooks/cannon.ts
```diff
@@ -558,16 +561,44 @@ export default class WebhookCannon {
...
+
+      const secondaryOs = await db.objectStore.get(
+        this.secondaryRecordObjectStoreId
+      );
+      if (secondaryOs) {
+        let params = {
+          method: "GET",
+          timeout: 5 * 1000,
+        };
+        const resp = await fetchWithTimeout(url, params);
```
victorges, line 579: I _think_ `fetchWithTimeout` might throw on timeout, so we prolly want a try/catch here (might be worth a helper func like `fileExists(url)`)

packages/api/src/webhooks/cannon.ts
```diff
@@ -558,16 +561,44 @@ export default class WebhookCannon {
...
+        session.id,
+        "output.m3u8"
+      );
+
+      const secondaryOs = await db.objectStore.get(
+        this.secondaryRecordObjectStoreId
+      );
+      if (secondaryOs) {
+        let params = {
+          method: "GET",
```
victorges: I wonder if a `HEAD` would be possible here instead

victorges (commented):
@mjh1 the logic makes sense to me!

packages/api/src/webhooks/cannon.ts
mjh1: Oh yep good idea

packages/api/src/webhooks/cannon.ts
```diff
@@ -585,6 +605,24 @@ export default class WebhookCannon {
...
+      timeout: 5 * 1000,
+    };
+    const resp = await fetchWithTimeout(url, params);
+
+    if (resp.status == 200) {
+      return true;
+    }
+  } catch (e) {
+    console.log("error fetching " + url, e);
+    return false;
```
victorges: _Maybe_ only return `false` if this is a timeout error, otherwise we might wrongly fallback to the secondary when the file did exist but there was a temporary error or something. 

Actually, a temporary network issue could result in a timeout as well, so maybe we don't want to `catch` anything here after all and just let exceptions be themselves (lol sorry for the back and forth, I know I just suggested it). On an exception, [this code](https://github.com/livepeer/livepeer-com/blob/b4b41a178b814008fe79b2298ac506d4115fae63/packages/api/src/webhooks/cannon.ts#L103) will actually `nack` the AMQP message to be retried later.

victorges (approved):
LGTM + a revert suggestion from sth I had suggested myself 🙈

packages/api/src/webhooks/cannon.ts
mjh1: ah of course yep makes sense

victorges (approved):
LGTM

PR #1891, opened 2023-09-20
Remove transcode and import tasks

mjh1:
This is no longer used and has been removed from task-runner

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -304,7 +304,6 @@ export function getStaticPlaybackInfo(
 }
 
 export function getPlaybackUrl(
-  { vodCatalystObjectStoreId }: CliArgs,
   ingest: string,
   asset: WithID<Asset>,
   os: ObjectStore
@@ -321,17 +320,7 @@ export function getPlaybackUrl(
     (f) => f.type === "catalyst_hls_manifest"
   );
   if (catalystManifest) {
-    if (os.id !== vodCatalystObjectStoreId) {
-      return pathJoin(os.publicUrl, asset.playbackId, catalystManifest.path);
-    }
-    // TODO: Set up Catalyst playback or a CDN in front of the default object store.
     return pathJoin(os.publicUrl, asset.playbackId, catalystManifest.path);
-    // return pathJoin(
-    //   "https://playback.livepeer.monster:10443/", // TODO: Make this a cli arg
-    //   "hls",
-    //   `asset+${asset.playbackId}`,
-    //   "index.m3u8"
-    // );
   }
   return undefined;
 }
@@ -380,7 +369,7 @@ export async function withPlaybackUrls(
     ...(asset.status.phase === "ready" && {
       downloadUrl: getDownloadUrl(config, ingest, asset, os),
     }),
-    playbackUrl: getPlaybackUrl(config, ingest, asset, os),
+    playbackUrl: getPlaybackUrl(ingest, asset, os),
   };
 }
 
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -321,17 +320,7 @@ export function getPlaybackUrl(
...
-      return pathJoin(os.publicUrl, asset.playbackId, catalystManifest.path);
-    }
-    // TODO: Set up Catalyst playback or a CDN in front of the default object store.
     return pathJoin(os.publicUrl, asset.playbackId, catalystManifest.path);
-    // return pathJoin(
-    //   "https://playback.livepeer.monster:10443/", // TODO: Make this a cli arg
-    //   "hls",
-    //   `asset+${asset.playbackId}`,
-    //   "index.m3u8"
-    // );
```
victorges, line 350: We cannot change any logic for using assets from the old pipeline until we migrate all of them to Catalyst. We can only stop _generating_ them for now.

packages/api/src/task/scheduler.ts
```diff
@@ -162,27 +162,6 @@ export class TaskScheduler {
         });
         await this.triggerRecordingReadyWebhook(task);
         break;
-      case "transcode":
```
victorges, line 165: Can you also remove any references on the API schema (which generates all the model typings here as well) to the removed tasks? e.g. there's a `task.type` enum field which is allowed to be `transcode` or `import` today. We should remove that (even tho the old DB objects might have them, that's not a problem)

victorges, line 165: Shouldn't we remove `case "import"` above as well?

victorges (changes requested):
We cannot remove the backward compatibility with the old assets yet, we should only stop generating them.

packages/api/src/controllers/asset.ts
mjh1: ahh.. of course

mjh1: But the result is the same from this logic, either way we just return `return pathJoin(os.publicUrl, asset.playbackId, catalystManifest.path);` right? That's why I removed the if statement

victorges: lol I'm sorry, I had just landed a 36h 🛬 trip yesterday morning, maybe I was a zombie 🧟

packages/api/src/schema/api-schema.yaml
```diff
@@ -3747,304 +3714,6 @@ paths:
                 $ref: "#/components/schemas/task"
         default:
           description: Error
-  /transcode:
```
victorges: ~~wow didnt even know we had this documented~~

Oh, actually this one is for the `transcode-file` task, so we should keep it. The `transcode` task was in `/asset/:id/transcode` if I'm not mistaken, which isn't documented anywhere.

victorges (approved):
LGTM! Just gotta bring back the `/transcode` docs

packages/api/src/schema/api-schema.yaml
mjh1: whoops yep you're right

packages/www/components/AssetDetails/AssetEventLogTab.tsx
```diff
@@ -30,15 +30,7 @@ const AssetEventLogTab = ({ asset }: { asset: Asset }) => {
               task.outputAssetId === asset.id || task.inputAssetId === asset.id
           )
           ?.map((task) => ({
-            type: task?.params?.import?.url
```
mjh1, line 33: @0xcadams @victorges do you know if this part of the code is still relevant? I've just removed the lines that rely on the now deleted import task

victorges: Hmm, maybe it would still be relevant to show the info from the old import tasks, tho those are so far in the past now I bet no one will be looking at them anymore. I say we just clean this code as well.

victorges: WDYT of transforming this `import` based logic into `upload` tho? Should be relatively easy (tho would be a new feature in a sense, not sth we have today. totally optional)

victorges: Yeah this is what the event tab is looking like today for a freshly uploaded asset, looks pretty buggy to me tbh (with `import` task it shows an event of 'asset uploaded' etc). Would be a nice to have to fix this, tho as I said it would not be a regression to just leave as is if you prefer, and we can later prioritize fixing this events tab.

Screenshot 2023-09-25 at 10 51 47

mjh1: Ah gotcha, i'll take a look

victorges (approved):
Still LGTM anyway as that UI bit is pretty lopri

packages/www/components/AssetDetails/AssetEventLogTab.tsx
```diff
@@ -30,7 +30,16 @@ const AssetEventLogTab = ({ asset }: { asset: Asset }) => {
...
           )
           ?.map((task) => ({
-            type: task?.params?.export?.["ipfs"] ? "ipfs-upload" : "other",
+            type:
+              asset.source?.type == "url"
+                ? "url-import"
+                : asset.source?.type == "directUpload"
+                ? "file-import"
+                : asset.source?.type == "recording"
+                ? "recorded-session"
```
victorges: The primary key on this UI is actually the `task` ("events"), not the asset, so I think this actually won't be equivalent. When we're showing data about an export task, we would actually show it as another entry for the asset upload instead. 

To fix you could surround these `asset.source?.type` checks under a `["upload", "import"].includes(task?.type)` condition instead, and only then you check the `asset`. WDYT?

Otherwise, can probably just replace `import` with `upload` on that previous implementation and let the `import` ones be ~brokenish.

mjh1: Ah.. yep. I thought there must've been something wrong with me doing that ha, yeah that fix by checking the task type makes sense 👍

victorges (approved):
👌

PR #1892, opened 2023-09-21
clip: define clip api

gioelecerati:

- Created a clip endpoints to create clips
- Fixed the query to get last session and added a new one to get it from the session table
- Return recording url on the playback request
- Created a new task `clip`

This is only part of the work, still missing the webhooks and the job handling

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  let asset = await validateAssetPayload(
+    id,
+    uPlaybackId,
+    content.userId,
+    Date.now(),
+    defaultObjectStoreId(req),
+    req.config,
+    {
+      name: req.body.name || `clip_${uPlaybackId}`,
```
victorges: supernit: for consistency with [the assets from recording](https://github.com/livepeer/livepeer-com/blob/4a3856a0e7efdf4fa6ecf9505c43062dc5301ecd/packages/api/src/webhooks/cannon.ts#L558), lets use dash here
```suggestion
      name: req.body.name || `clip-${uPlaybackId}`,
```

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+    },
+    // TODO: actual type `clip` - currently something's wrong with the types when setting up the schema
+    { type: "directUpload", sourceId: content.id }
```
victorges: This should be **really** easy to implement so please don't skip it on this first PR! Just add the enum on the schema and set it here.

Additionally, we should differentiate between streams/sessions and asset IDs, because our IDs have become non-globally-unique since Recordings V2 (sessions and assets have the same ID).

So I'd suggest sth like this
```suggestion
    { type: "clip", ...(isStream, ? { sessionId: session.id } : { assetId: content.id }) }
```
WDYT? (the syntax doesn't have to be like that, just an example)

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+const app = Router();
+
+app.post("/", validatePost("clip-payload"), async (req, res) => {
+  const playbackId = req.body.playbackId;
+  const userId = req.user.id;
+
+  const id = uuid();
+  let uPlaybackId = await generateUniquePlaybackId(id);
+
+  const content = await db.stream.getByPlaybackId(playbackId); //||
```
victorges, line 58: Shouldn't this be reading from the session table instead? Conceptually, the clips are made on the session, not the channel (parent stream, being fetched here). 
e.g. A time range like "01:23 to 03:46" doesn't really make sense when it references a channel instead of a specific session (unless we're using absolute timestamps, are we?)

edit: Ok, I see that you fetch the session below, but the problem is that the clip asset is referencing the stream but not the session. Might be the case to just move the `validateAssetPayload` to after fetching the session ~L84, so we can create the asset payload with the proper reference to the session instead (from the session we can always find the channel, but not the other way around)

packages/api/src/store/stream-table.ts
```diff
@@ -296,7 +296,17 @@ export default class StreamTable extends Table<DBStream> {
...
-      sql`SELECT data FROM stream  WHERE data->>'parentId'=${id} ORDER BY data->'createdAt' DESC LIMIT 1`.setName(
+      sql`SELECT data FROM stream  WHERE data->>'parentId'=${id} ORDER BY data->>'createdAt' DESC LIMIT 1`.setName(
+        `${this.name}_by_parentid_last_session`
+      ),
+      opts
+    );
+    return res.rowCount < 1 ? null : (res.rows[0].data as DBStream);
+  }
+
+  async getLastSessionFromSessionsTable(id: string, opts?: QueryOptions) {
```
victorges: nit: This function should be in a `sessions-table.ts` file/class instead, while this class accesses the `stream` table. The whole parent/child stream/session is already pretty confusing, we should not make them look even more interchangeable.

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+      .json({ errors: ["Clipping for assets is not implemented yet"] });
+    return;
+    //const os = await db.objectStore.get(req.config.vodCatalystObjectStoreId);
+    //url = pathJoin(os.publicUrl, content.playbackId, content.id, "index.m3u8");
+  }
+
+  asset = await createAsset(asset, req.queue);
+
+  const task = await req.taskScheduler.createAndScheduleTask(
+    "clip",
```
victorges, line 121: Since you're creating a new task type, you have to modify `task/scheduler.ts` to know how to handle it _on every place that we handle `upload` task today_ (otherwise, for example, the asset will never get populated AFAIK)

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (isStream) {
+    if (!content.record) {
+      res.status(400).json({
+        errors: ["Recording must be enabled on a live stream to create clips"],
+      });
+    }
+    let recordingResponse = await getRecordingUrl(content, req);
+    url = recordingResponse.url;
+    session = recordingResponse.session;
+  } else {
```
victorges: Related to the stream vs session discussion, it feels like this logic should be in some kind of `fetchClipContent` function, and this handler should never even care/access the `Stream` object directly. The stream object is just an implementation detail to get to the session, but clips should conceptually only relate with the sessions themselves.

i.e. on line 28 we can call some function that returns `DBSession | DBAsset` and that's all this API handler cares about

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+    {
+      clip: {
+        clipStrategy: {
+          playbackId,
+          startTime: req.body.startTime,
+          endTime: req.body.endTime,
+        },
+        catalystPipelineStrategy: catalystPipelineStrategy(req),
+        url,
+        sessionId: session.id,
```
victorges, line 131: This will explode if session is not defined
```suggestion
        sessionId: session?.id,
```

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+        inputId: content.id,
```
victorges, line 132: Do we need this?

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+    },
+    null,
+    asset,
+    userId
+  );
+
+  res.json({
+    task,
+    asset,
+  });
```
victorges: These objects need to be cleaned up corresponding to the `asset.ts` logic. Otherwise we can leak secrets on the public API. You can likely just copy this middleware: https://github.com/livepeer/livepeer-com/blob/0b9b4903becfe7b5413f290f285bdf35a699f9bc/packages/api/src/controllers/asset.ts#L577

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+        inputId: content.id,
+      },
+    },
+    null,
+    asset,
+    userId
+  );
+
+  res.json({
+    task,
```
victorges: We should return only the task ID here consistently with the asset API.
```suggestion
    task: { id: task.id },
```

packages/api/src/controllers/clip.ts
```diff
@@ -0,0 +1,154 @@
...
+      "output.m3u8"
+    );
+  }
+
+  /*resp = await fetchWithTimeout(url, params);
+
+  if (resp.status != 200) {
+    throw new Error("Recording not found");
+  }*/
+
```
victorges: should remove commented code
```suggestion
```

packages/api/src/controllers/playback.ts
```diff
@@ -71,6 +73,14 @@ function newPlaybackInfo(
       url: webRtcUrl,
     });
   }
+  if (recordingUrl) {
+    playbackInfo.meta.recordings = [];
```
victorges: WDYT of calling this something else? We already have an existing "recordings" definitions in our API, which is not what this is. This looks like it returns the URLs of all recordings from this stream, which is not what it does. 

Maybe `historicPlaybackUrl` or `dvrPlayback` or anything else. Doesn't matter if it's maybe a little too verbose IMO since it will mostly be used internally, but it is better than using a misleading name (remember that this will show up in our API reference as well, this is a public-facing API)

packages/api/src/controllers/playback.ts
```diff
@@ -183,7 +193,9 @@ async function getPlaybackInfo(
   ingest: string,
   id: string,
   isCrossUserQuery: boolean,
-  origin: string
+  origin: string,
+  withRecordings?: boolean,
+  recordingObjectStoreId?: string
```
victorges: For clarity:

```suggestion
  recordCatalystObjectStoreId?: string
```

packages/api/src/controllers/stream.ts
```diff
@@ -401,6 +401,31 @@ app.get("/", authorizer({}), async (req, res) => {
...
+    }
+
+    const os = await db.objectStore.get(objectStoreId);
+    url = pathJoin(os.publicUrl, session.playbackId, session.id, "output.m3u8");
+  } catch (e) {
+    console.log(`
+      Error getting recording playback url: ${e}
+    `);
+    return null;
+  }
```
victorges, line 424: 1. Shouldn't this have that same logic to try fetching the manifest then falling back to the secondary OS if not found? (we might wanna have a helper for that after all, it'd be repeated in 3 places now)
2. I don't think we should ignore errors. This should explode if we have problems accessing the DB. It's better to have the client retry than returning a false result here (as if the recording URL didnt' exist)

packages/api/src/controllers/stream.ts
```diff
@@ -401,6 +401,31 @@ app.get("/", authorizer({}), async (req, res) => {
...
+export async function getRecordingPlaybackUrl(
+  stream: DBStream,
+  objectStoreId: string
+) {
+  let url: string;
+
+  try {
+    const session = await db.stream.getLastSessionFromSessionsTable(stream.id);
+
+    if (!session) {
```
victorges, line 413: Should we check `session.record`? (or `stream.record`)

packages/api/src/schema/db-schema.yaml
```diff
@@ -949,6 +968,40 @@ components:
...
+                    startTime:
+                      type: number
+                      description: Start time of the clip in milliseconds
+                    endTime:
+                      type: number
+                      description: End time of the clip in milliseconds
+                    playbackId:
+                      type: string
+                      description: Playback ID of the stream or asset to clip
+                catalystPipelineStrategy:
```
victorges, line 991: Wonder if we can/should just reference some common field to avoid repeating this

victorges (approved):
LGTM with a couple mostly functional suggestions.

FTR in some places I used the "channel" term, meaning a parent stream object. Correspondingly, child streams would be the "transcoding sessions" and the sessions the "ingest sessions"

packages/api/src/controllers/playback.ts
victorges: > (remember that this will show up in our API reference as well, this is a public-facing API)

alternatively, we can move the new field to the `db-schema.yaml` file so it doesn't show up in API reference. That shoul be fine as well

packages/api/src/controllers/clip.ts
gioelecerati: While processing, the clip is going to reside in the recordings storage under playbackId/sessionId/clip_assetPlaybackId/clip.m3u8 so the original playbackId of the content is needed to build this URL down the pipeline - although there is room for improvement in task runner and there are some hacks in catalyst-api that are going to get resolved and have a cleaner task here in studio.

Also, now the asset is going to use the normal vod storage for the output location generation, so at the end the output asset won't have the sessionId in the output URL - will need to put it in the source object to reference clips to sessions

packages/api/src/controllers/clip.ts
gioelecerati: I added "clip" to the cases in scheduler.ts - not sure yet if they whole upload code is going to work the same when the output is coming back, but we address this in a subsequent pr

packages/api/src/controllers/clip.ts
gioelecerati: Added a check if the recording session is missing

packages/api/src/controllers/clip.ts
gioelecerati: Yes - for now we need this down the pipeline to build URLs, will probably get better later

packages/api/src/controllers/stream.ts
gioelecerati: We're checking it up in the code, on the api call

packages/api/src/controllers/clip.ts
gioelecerati: This makes a lot of sense, changed it!

packages/api/src/controllers/clip.ts
gioelecerati: Moved into a separate function after talking with max

packages/api/src/controllers/stream.ts
gioelecerati: 1. implemented - as suggested by max!
2. I removed this try catch with that

packages/api/src/controllers/clip.ts
victorges: I think this is independent of where the files are stored. On the API level, the clips should reference the user-accessible content where the clip was based from. In case of a livestream, the clip is not made on top of a "channel", but on top of a specific session, so this is what has to be referenced on the asset side. 

> While processing, the clip is going to reside in the recordings storage under playbackId/sessionId/clip_assetPlaybackId/clip.m3u8

Not sure what you mean by "recording storage", but isn't the clip just gonna be a regular asset stored under `/playbackId/index.m3u8` (the asset `playbackId`, not the original one)? Feels like you're referring to an intermediate object created during catalyst processing, which doesn't really matter here.

packages/api/src/controllers/clip.ts
victorges: Is there any way that there wouldn't be a session here? Maybe it's confusing because there's all the commented out / dummy branches for assets, which are disabled now. Might be simpler to not have those asset flows for now and just implement them later when we do want asset clipping.

packages/api/src/controllers/clip.ts
victorges: The input URL is already built here and sent on the `url` field above. The output URL should be based only on the asset `playbackId`, so it'd not be needed later. What am I missing?

packages/api/src/controllers/clip.ts
```diff
@@ -68,18 +56,36 @@ app.post("/", validatePost("clip-payload"), async (req, res) => {
         errors: ["Recording must be enabled on a live stream to create clips"],
       });
     }
-    session = await db.stream.getLastSessionFromSessionsTable(content.id);
-    const os = await db.objectStore.get(req.config.recordCatalystObjectStoreId);
-    url = pathJoin(os.publicUrl, session.playbackId, session.id, "output.m3u8");
+    let runningRecording = await getRunningRecording(content, req);
+    url = runningRecording.url;
+    session = runningRecording.session;
```
victorges: Neat! :)

btw a ts tip is using deconstruction syntax (not requesting a change, just sharing if you like it)
```suggestion
    ({ url, session }) = await getRunningRecording(content, req);
```

packages/api/src/controllers/clip.ts
```diff
@@ -103,9 +109,57 @@ app.post("/", validatePost("clip-payload"), async (req, res) => {
...
-    task,
+    task: { id: task.id },
     asset,
   });
 });
 
+async function getRunningRecording(content: DBStream, req: Request) {
+  let objectStoreId: string;
+
+  const session = await db.stream.getLastSessionFromSessionsTable(content.id);
```
victorges: I've sent a PR to fix this query, please merge it into this PR (base branch is already set, just review/make any changes/merge it)

https://github.com/livepeer/studio/pull/1896

packages/api/src/controllers/clip.ts
```diff
@@ -103,9 +109,57 @@ app.post("/", validatePost("clip-payload"), async (req, res) => {
...
+    }*/
+    objectStoreId = req.config.secondaryRecordObjectStoreId;
+  } else {
+    objectStoreId = req.config.recordCatalystObjectStoreId;
+  }
+
+  return {
+    url,
+    session,
+    objectStoreId,
```
victorges, line 192: This is unused isnt it?

packages/api/src/task/scheduler.ts
```diff
@@ -130,7 +130,7 @@ export class TaskScheduler {
     let assetSpec: Asset;
     switch (task.type) {
       case "import":
-      case "upload":
+      case "upload" || "clip":
```
victorges: This will not work, `"upload" || "clip"` evaluates to `"upload"`. You have to do the same thing as the line above.

victorges (approved):
LGTM!

packages/api/src/controllers/clip.ts
gioelecerati: Right! I was using it before, when assets where ending up in the same store, removing it

packages/api/src/task/scheduler.ts
gioelecerati: Oh right! thank you

PR #1893, opened 2023-09-21
api: user: added email verification messages

gioelecerati:

- Added text when signing in as unverified, asking to check for a verification email
- Added a text when signing up, informing that we sent a verification email
- Modified text for `email already registered` asking to sign in or check the verification email

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -522,7 +522,7 @@ app.post("/token", validatePost("user"), async (req, res) => {
     res.status(403);
     return res.json({
       errors: [
-        "user has not been verified - please check your inbox for verification email",
+        "This account has not been verified - please check your inbox for a verification email.",
       ],
     });
   }
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -414,11 +414,14 @@ app.post("/", validatePost("user"), async (req, res) => {
     return res.json({ errors: ["user not created"] });
   }
 
+  let isTest =
+    process.env.NODE_ENV === "test" || process.env.NODE_ENV === "development";
+
   if (
     req.config.requireEmailVerification &&
     !user.emailValid &&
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -409,6 +413,16 @@ app.post("/", validatePost("user"), async (req, res) => {
...
 
+  if (
+    req.config.requireEmailVerification &&
+    !user.emailValid &&
+    !user.admin &&
+    !req.config.isTest
+  ) {
+    res.status(403);
+    return res.json({ errors: ["we just sent you a verification email"] });
+  }
```
victorges: I don't think this check makes sense. Wasn't this user just created with `emailValid = validUser` var? Either verification will be disabled and this always passes, or verification is enabled and the registration request always fails. (this is the registration API right?)

packages/api/src/test-server.ts
```diff
@@ -58,6 +58,7 @@ if (!params.insecureTestToken) {
 }
 params.listen = true;
 params.requireEmailVerification = true;
+params.isTest = true;
```
victorges: This is redundant, we should just use the `NODE_ENV` environment variable when we need to know if it is a test. e.g. https://github.com/livepeer/livepeer-com/blob/1fd399ff3387dd915a8b7a3ac36e123ce17af64b/packages/api/src/app-router.ts#L125-L126

victorges (approved):

packages/api/src/controllers/user.ts
gioelecerati: yeah - it's failing a returning the text to be displayed under the form - similar to "email already registered" appearing just over the sign up button

I agree this should'nt be a 403 and a better way to display messages should exist

victorges: Definitely should not be a 403.

PR #1894, opened 2023-09-22
billing: usage notifications

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/notification.ts
+++ b/packages/api/src/controllers/notification.ts
@@ -6,8 +6,14 @@ import { WithID } from "../store/types";
 import { Request } from "express";
 import { User } from "../schema/types";
 
+export const HACKER_DISABLE_CUTOFF_DATE = 1697752800000;
+
 export const getUsageNotifications = async (
-  usagePercentages,
+  usagePercentages: {
+    TotalUsageMins: number;
+    DeliveryUsageMins: number;
+    StorageUsageMins: number;
+  },
   user: WithID<User>
 ) => {
   let notifications = [];
@@ -27,13 +33,21 @@ export const getUsageNotifications = async (
     usagePercentages.DeliveryUsageMins > 75 ||
     usagePercentages.StorageUsageMins > 75
   ) {
-    await db.user.update(user.id, {
-      notifications: {
-        usage: {
-          notification75: true,
-        },
-      },
+    notifications.push({
+      type: "notification75",
+      title: "Usage Warning",
+      message: `Your usage is over 75% of your limit.`,
     });
+  } else if (usagePercentages.TotalUsageMins >= 100) {
+    notifications.push({
+      type: "notification100",
+      title: "Usage Warning",
+      message: `You have exceeded your usage limit.`,
+    });
+
+    if (user.createdAt > HACKER_DISABLE_CUTOFF_DATE) {
+      // disable user
+    }
   } else {
     if (
       user.notifications?.usage?.notification75 ||
@@ -44,6 +58,7 @@ export const getUsageNotifications = async (
           usage: {
             notification75: false,
             notification90: false,
+            notification100: false,
           },
         },
       });
--- a/packages/api/src/controllers/notification.ts
+++ b/packages/api/src/controllers/notification.ts
@@ -1,0 +1,107 @@
+import { sendgridEmail } from "./helpers";
+import { frontendUrl, unsubscribeUrl } from "./user";
+import { db } from "../store";
+import { sendgridEmailPaymentFailed } from "./helpers";
+import { WithID } from "../store/types";
+import { Request } from "express";
+import { User } from "../schema/types";
+
+export const getUsageNotifications = async (
+  usagePercentages,
+  user: WithID<User>
+) => {
+  let notifications = [];
+
+  if (
+    usagePercentages.TotalUsageMins > 90 ||
+    usagePercentages.DeliveryUsageMins > 90 ||
+    usagePercentages.StorageUsageMins > 90
+  ) {
+    notifications.push({
+      type: "notification90",
+      title: "Usage Warning",
+      message: `Your usage is over 90% of your limit.`,
+    });
+  } else if (
+    usagePercentages.TotalUsageMins > 75 ||
+    usagePercentages.DeliveryUsageMins > 75 ||
+    usagePercentages.StorageUsageMins > 75
+  ) {
+    await db.user.update(user.id, {
+      notifications: {
+        usage: {
+          notification75: true,
+        },
+      },
+    });
+  } else {
+    if (
+      user.notifications?.usage?.notification75 ||
+      user.notifications?.usage?.notification90
+    ) {
+      await db.user.update(user.id, {
+        notifications: {
+          usage: {
+            notification75: false,
+            notification90: false,
+          },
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/notification.ts
victorges, line 1: WDYT of moving this file to a folder like `./controllers/utils` or move the helpers to `helpers.ts`? The `controllers` folder is meant to hold only files matching APIs, with the only exception today being that existing `helpers.ts` file. 

In the future would make sense to move the generic `helpers` to to separate folder as well, tho I think it's enough for now just moving this to that separate folder.

packages/api/src/schema/db-schema.yaml
```diff
@@ -1060,9 +1060,22 @@ components:
         migrationInvoice:
           type: boolean
           default: false
-        lastEmailAboutPaymentFailure:
-          type: number
-          example: 1587667174725
+        notifications:
```
victorges, line 1075: neat

packages/api/src/controllers/usage.ts
```diff
@@ -2,10 +2,12 @@ import Router from "express/lib/router";
 import { db } from "../store";
 import { authorizer, validatePost } from "../middleware";
 import { products } from "../config";
-import { reportUsage } from "./stripe";
 import fetch from "node-fetch";
 import qs from "qs";
 import { NotFoundError } from "../store/errors";
+import { WithID } from "../store/types";
+import { User } from "../schema/types";
```
victorges, line 9:
```suggestion
import { User, WithID } from "../store/types";
```

packages/api/src/controllers/usage.ts
```diff
@@ -63,6 +65,68 @@ export const calculateOverUsage = async (product, usage) => {
...
+    TotalUsageMins: Math.max(
+      (usage?.TotalUsageMins / (limits.transcoding || 0)) * 100,
+      0
+    ),
+    DeliveryUsageMins: Math.max(
+      (usage?.DeliveryUsageMins / (limits.streaming || 0)) * 100,
+      0
+    ),
+    StorageUsageMins: Math.max(
+      (usage?.StorageUsageMins / (limits.storage || 0)) * 100,
```
victorges: Risk of dividing by `0` in all these. perhaps `1`? Or maybe you want a top-level ternary on `limits.transcoding ? ...` instead

packages/api/src/controllers/usage.ts
```diff
@@ -63,6 +65,68 @@ export const calculateOverUsage = async (product, usage) => {
   return overUsage;
 };
 
+export const getUsagePercentageOfLimit = async (product, usage) => {
+  let limits: any = {};
```
victorges: ts to the rescue (from `any`) 😆 
```suggestion
  let limits: Record<string, number> = {};
```

packages/api/src/controllers/notification.ts
```diff
@@ -0,0 +1,107 @@
+import { sendgridEmail } from "./helpers";
+import { frontendUrl, unsubscribeUrl } from "./user";
+import { db } from "../store";
+import { sendgridEmailPaymentFailed } from "./helpers";
+import { WithID } from "../store/types";
+import { Request } from "express";
+import { User } from "../schema/types";
+
+export const getUsageNotifications = async (
+  usagePercentages,
```
victorges: Can you add a type to this? Can even inline it e.g.
```suggestion
  usagePercentages: {TotalUsageMins: number, DeliveryUsageMins: number, StorageUsageMins: number},
```

packages/api/src/controllers/notification.ts
```diff
@@ -0,0 +1,107 @@
...
+  user: WithID<User>
+) => {
+  let notifications = [];
+
+  if (
+    usagePercentages.TotalUsageMins > 90 ||
+    usagePercentages.DeliveryUsageMins > 90 ||
+    usagePercentages.StorageUsageMins > 90
```
victorges: I think these fields have confusing names. They imply the unit is minutes, when I believe it is actually a percentage? WDYT of renaming to e.g. `TotalUsageLimitPercent`, `just `...Percent` or sth like that?

packages/api/src/controllers/notification.ts
```diff
@@ -0,0 +1,107 @@
...
+  ) {
+    notifications.push({
+      type: "notification90",
+      title: "Usage Warning",
+      message: `Your usage is over 90% of your limit.`,
+    });
+  } else if (
+    usagePercentages.TotalUsageMins > 75 ||
+    usagePercentages.DeliveryUsageMins > 75 ||
+    usagePercentages.StorageUsageMins > 75
```
victorges: An alternative is grabbing the `Math.max` first then checking once each time

packages/api/src/controllers/notification.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (
+    usagePercentages.TotalUsageMins > 90 ||
+    usagePercentages.DeliveryUsageMins > 90 ||
+    usagePercentages.StorageUsageMins > 90
+  ) {
+    notifications.push({
+      type: "notification90",
+      title: "Usage Warning",
+      message: `Your usage is over 90% of your limit.`,
+    });
```
victorges: shouldnt you update the user object flag here?

packages/api/src/controllers/notification.ts
```diff
@@ -0,0 +1,107 @@
...
+  } else if (
+    usagePercentages.TotalUsageMins > 75 ||
+    usagePercentages.DeliveryUsageMins > 75 ||
+    usagePercentages.StorageUsageMins > 75
+  ) {
+    await db.user.update(user.id, {
```
victorges: shouldnt you push a notification here? 

also, when are the `notifications.usage.notification75/90` checked?

packages/api/src/controllers/notification.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (
+      user.notifications?.usage?.notification75 ||
+      user.notifications?.usage?.notification90
+    ) {
+      await db.user.update(user.id, {
+        notifications: {
+          usage: {
+            notification75: false,
+            notification90: false,
+          },
```
victorges: This seems unnecessary, could add a comment explaining why do you need this?

victorges (approved):
LGTM!

packages/api/src/controllers/usage.ts
gioelecerati: ouch! fixed

packages/api/src/controllers/notification.ts
gioelecerati: Oh right, it was mixed up, all db.set and all checks are done in the function `notifyUser`

```
export const notifyUser = async (notifications, user, req) => {
```

packages/api/src/controllers/notification.ts
gioelecerati: it's happening in `notifyUser`

packages/api/src/controllers/notification.ts
gioelecerati: This is resetting the notifications sent to false for each type when all three types of usage are under 75% (so at the reset of the billing cycle)

PR #1897, opened 2023-09-27
playback: improve dvr playback error handling

gioelecerati:
…ning recording

- Updating the playback endpoint to return the dvrPlayback url only if available
- We're returning two separate error messages if recording is disabled for streams or if the recording is not available so that the player UI or the clipping UI can display a proper message
- We're omitting the `url` field in this case so that the player won't show the clipping icon
- Moving the getRunningSession method to the session controller

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/playback.ts
+++ b/packages/api/src/controllers/playback.ts
@@ -90,7 +90,7 @@ function newPlaybackInfo(
       });
     } else {
       playbackInfo.meta.dvrPlayback.push({
-        error: "no running recordings available.",
+        error: "no running recordings available for this stream.",
       });
     }
   }
@@ -232,8 +232,7 @@ async function getPlaybackInfo(
 
   if (stream) {
     let url: string;
-    let fetchRunningRecording = withRecordings && stream.record;
-    if (fetchRunningRecording) {
+    if (withRecordings) {
       ({ url } = await getRunningRecording(stream, req));
     }
     return newPlaybackInfo(
@@ -244,7 +243,7 @@ async function getPlaybackInfo(
       null,
       stream.isActive ? 1 : 0,
       url,
-      fetchRunningRecording,
+      withRecordings,
       stream.record
     );
   }
--- a/packages/api/src/controllers/playback.ts
+++ b/packages/api/src/controllers/playback.ts
@@ -80,7 +80,7 @@ function newPlaybackInfo(
     playbackInfo.meta.dvrPlayback = [];
     if (!streamRecord) {
       playbackInfo.meta.dvrPlayback.push({
-        message: "Error: Recording is not enabled for this stream.",
+        error: "recording is not enabled for this stream.",
       });
     } else if (recordingUrl) {
       playbackInfo.meta.dvrPlayback.push({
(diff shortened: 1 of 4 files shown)


0xcadams (approved):
LGTM

packages/api/src/controllers/session.ts
```diff
@@ -187,4 +189,56 @@ const adminOnlyFields: (keyof DBSession)[] = ["deleted", "broadcasterHost"];
 
 const privateFields: (keyof DBSession)[] = ["recordObjectStoreId", "version"];
 
+export async function getRunningRecording(content: DBStream, req: Request) {
```
victorges, line 192: Should we use this on `webhook/cannon.ts` as well? I think that's where the implementation came from right, not sure if it became different tho

packages/api/src/controllers/playback.ts
```diff
@@ -73,14 +76,25 @@ function newPlaybackInfo(
...
+  if (withRecordings) {
     playbackInfo.meta.dvrPlayback = [];
-    playbackInfo.meta.dvrPlayback.push({
-      hrn: "HLS (TS)",
-      type: "html5/application/vnd.apple.mpegurl",
-      url: recordingUrl,
-    });
+    if (!streamRecord) {
+      playbackInfo.meta.dvrPlayback.push({
+        message: "Error: Recording is not enabled for this stream.",
```
victorges: Instead of `message: "Error: *"`WDYT of just calling the field `error:` with a clean error message? I don't think we should ever use this `message` for anything else.

packages/api/src/controllers/playback.ts
```diff
@@ -195,7 +209,7 @@ async function getPlaybackInfo(
   isCrossUserQuery: boolean,
   origin: string,
   withRecordings?: boolean,
-  recordCatalystObjectStoreId?: string
+  req?: Request
```
victorges: This is duplicated, the request is already the first argument to this function.

Either extract more fields from that argument, or replace the deconstruction syntax with grabbing the whole `req` value and use it in the implementation.

packages/api/src/controllers/playback.ts
```diff
@@ -274,7 +288,7 @@ app.get("/:id", async (req, res) => {
     isEmbeddablePlayer,
     origin,
     withRecordings,
-    req.config.recordCatalystObjectStoreId
+    req
```
victorges: Exact same argument as line 285 :)

victorges (approved):

packages/api/src/controllers/playback.ts
gioelecerati: Yeah ok makes sense! I was using `error` before, just felt uncomfortable with having a 2XX response and an error message  inside, but at the end it's just specific to the dvrPlayback, so, changing it to error

victorges: Yeah, it would be just hiding the fact that it is only used for errors though. In general, backends should not be creating messages to the end-users. This would be just a flag for the frontend or sth (and the 403 on user registration). This will be tech debt if we eventually internationalize our dashboard.

PR #1900, opened 2023-09-28
api: allow overriding default stream profiles

iameli-streams:
Two changes here:

1. Add `--default-stream-profiles` allowing the default profiles for a stream object to be provided at runtime (the profiles that are chosen if there aren't any provided for a stream)
2. Changes those default profiles to use `profile: "H264ConstrainedHigh"` by default. We want this in almost all cases for livestreaming because it's necessary to support WebRTC.

### DIFF
--- a/packages/api/src/parse-cli.ts
+++ b/packages/api/src/parse-cli.ts
@@ -76,6 +76,34 @@ function coerceJsonValue<T>(flagName: string) {
   };
 }
 
+function coerceJsonProfileArr(flagName: string) {
+  return (str: string): FfmpegProfile[] => {
+    let profiles;
+    const validator = profileValidator as ValidateFunction;
+    try {
+      profiles = JSON.parse(str);
+    } catch (e) {
+      throw new Error(`--${flagName} JSON parsing error: ${e.message}`);
+    }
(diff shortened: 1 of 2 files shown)


packages/api/src/parse-cli.ts
```diff
@@ -453,6 +456,73 @@ export default function parseCli(argv?: string | readonly string[]) {
...
+          {
+            name: "720p0",
+            fps: 0,
+            bitrate: 3000000,
+            width: 1280,
+            height: 720,
+            profile: "H264ConstrainedHigh",
+          },
+        ] as FfmpegProfile[]),
+        coerce: (str: string): FfmpegProfile[] => {
```
victorges: **nit:** Mind moving this to the top of the file with a descriptive name, like the other `coerce...` funcs around there? Or perhaps even just adjust `coerceJsonArray` to receive the validator as an optional argument 👀 but thats probably a stretch

victorges (approved):
LGTM

packages/api/src/parse-cli.ts
iameli: Done!

PR #1904, opened 2023-09-29
api: user: added disabled field

gioelecerati:

- Added disabled field to user to be used when usage limit is reached for the free tier

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -1114,6 +1114,15 @@ app.post(
       );
     }
 
+    if (
+      products[payload.stripeProductId].name == "Growth" ||
+      products[payload.stripeProductId].name == "Scale"
+    ) {
+      await db.user.update(user.id, {
+        disabled: false,
+      });
+    }
+
     // Update user's product subscription in our db
     await db.user.update(user.id, {
       stripeProductId: payload.stripeProductId,
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -18,6 +18,7 @@ import {
   UpdateSubscription,
   User,
   SuspendUserPayload,
+  DisableUserPayload,
 } from "../schema/types";
 import { db } from "../store";
 import { InternalServerError, NotFoundError } from "../store/errors";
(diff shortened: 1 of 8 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -523,6 +524,66 @@ app.patch(
   }
 );
 
+app.patch(
+  "/:id/disabled",
```
victorges, line 533: Could also be just `PATCH /user/:id` as it's patching the user object

packages/api/src/middleware/auth.ts
```diff
@@ -118,6 +118,12 @@ function authenticator(): RequestHandler {
       throw new ForbiddenError(`user is suspended`);
     }
 
+    if (user.disabled) {
+      throw new ForbiddenError(
+        `user is disabled - you maxed out your usage for the month, please input a payment method`
+      );
+    }
```
victorges: Disabled users cannot be handled the same way as suspended users. They need specific handling on the operations we want to block. This will make the account unusable, just like for suspended, in which case there's no point in creating the new property.

victorges: You can make checks about the user being disabled here instead:
- for live: https://github.com/livepeer/livepeer-com/blob/8e19a6f062630f8013f1fc769c0c9e04dd23b08b/packages/api/src/controllers/stream.ts#L1649
- for vod: https://github.com/livepeer/livepeer-com/blob/9411c8279173c7b68c31c92a7700399ec7ba40ac/packages/api/src/task/scheduler.ts#L346

packages/api/src/controllers/user.ts
victorges: Actually, is this even an API? I thought it would be automatically enabled when hitting billing limits, and automatically cleared once payment is made.

gioelecerati: It's an api similar to :id/suspended to be used by admins

packages/api/src/middleware/auth.ts
gioelecerati: Moved this check there!

PR #1906, opened 2023-09-29
api: get clips

gioelecerati:

- Create a get clips api

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/clip.ts
+++ b/packages/api/src/controllers/clip.ts
@@ -13,8 +13,7 @@ import {
 import { generateUniquePlaybackId } from "./generate-keys";
 import { v4 as uuid } from "uuid";
 import { DBSession } from "../store/session-table";
-import { fetchWithTimeout } from "../util";
-import { DBStream } from "../store/stream-table";
+import { authorizer } from "../middleware";
 import { toExternalAsset } from "./asset";
 import mung from "express-mung";
 import { Asset } from "../schema/types";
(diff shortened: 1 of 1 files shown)


emranemran (approved):

packages/api/src/controllers/clip.ts
```diff
@@ -160,4 +165,100 @@ app.post("/", validatePost("clip-payload"), async (req, res) => {
...
+  createdAt: { val: `asset.data->'createdAt'`, type: "int" },
+  updatedAt: { val: `asset.data->'status'->'updatedAt'`, type: "int" },
+  userId: `asset.data->>'userId'`,
+  playbackId: `asset.data->>'playbackId'`,
+  sourceType: `asset.data->'source'->>'type'`,
+  sourceSessionId: `asset.data->'source'->>'sessionId'`,
+  sourceAssetId: `asset.data->'source'->>'assetId'`,
+} as const;
+
+app.get("/:id", authorizer({}), async (req, res) => {
```
victorges, line 179: As discussed offline, let's move this to the `/asset` root since this is a GET an API on the asset resource and there's no separate "clip" object. (the POST is fine for now, since the input can be other things than an asset so it could be confusing inside /asset as well)

This is ok to merge as is though, but should not be documented to end-users and we should move it before doing so

packages/api/src/controllers/clip.ts
```diff
@@ -160,4 +165,100 @@ app.post("/", validatePost("clip-payload"), async (req, res) => {
...
+    case "session":
+      query.push(sql`asset.data->'source'->>'sessionId' = ${content.id}`);
+      break;
+    case "stream":
+      const sessionQuery = [];
+      sessionQuery.push(sql`data->>'parentId' = ${content.id}`);
+      const [sessions] = await db.session.find(sessionQuery, {
+        order: `data->>'lastSeen' DESC NULLS LAST`,
+        cursor,
+      });
```
victorges, line 234: This should be a temp hack, there might be a _ton_ (thousands) of streams here for any long living stream (say, lot radio 🛸 ). proper fix should involve a join (on the regular asset list query there's a join with the `user` table, check it out)

victorges (approved):
LGTM

PR #1908, opened 2023-10-03
Increase final retry delay

mjh1:
This is to help out with the situation where a catalyst deploy is happening, use a longer delay for task retries to have a better chance of the app being ready. This has been tested locally e2e.

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -428,7 +428,10 @@ export class TaskScheduler {
     };
 
     task = await this.updateTask(task, { status });
-    await sleep(retries * TASK_RETRY_BASE_DELAY);
+    // increase sleep for final retry
+    const sleepDur =
+      (retries == MAX_RETRIES ? retries + 1 : retries) * TASK_RETRY_BASE_DELAY;
+    await sleep(sleepDur);
     await this.scheduleTask(task, retries);
   }
 
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -37,6 +37,7 @@ function sqlQueryGroup(values: string[]) {
 
 const MAX_RETRIES = 2;
 const TASK_RETRY_BASE_DELAY = 30 * 1000;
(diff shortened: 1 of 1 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -428,7 +428,10 @@ export class TaskScheduler {
     };
 
     task = await this.updateTask(task, { status });
-    await sleep(retries * TASK_RETRY_BASE_DELAY);
+    // increase sleep for final retry
+    const sleepDur =
+      (retries == MAX_RETRIES ? retries + 1 : retries) * TASK_RETRY_BASE_DELAY;
```
mjh1: @victorges I'm increasing the final retry delay from 60s to 90s, sound ok? should we go any higher?

thomshutt: If there aren't downsides to going higher then I'd prefer that - 90s might still be tight

victorges: how about 3 mins?

victorges (approved):

packages/api/src/task/scheduler.ts
```diff
@@ -428,7 +429,12 @@ export class TaskScheduler {
     };
 
     task = await this.updateTask(task, { status });
-    await sleep(retries * TASK_RETRY_BASE_DELAY);
+    // increase sleep for final retry
+    const sleepDur =
+      retries == MAX_RETRIES
+        ? TASK_FINAL_RETRY_DELAY
+        : retries * TASK_RETRY_BASE_DELAY;
```
victorges, line 436: The precedence here is certainly what we expect, right? As in it won't be `TASK_FINAL_RETRY_DELAY * TASK_RETRY_BASE_DELAY` on the last retry 😂

victorges (approved):
Neat!

packages/api/src/task/scheduler.ts
mjh1: I had the same thought haha! I added in some brackets but then the prettier command removes them. But yes I've tested and it's all good

PR #1916, opened 2023-10-11
stream: fix idle state race condition

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -897,18 +897,17 @@ app.post(
     // compatibility with /setactive recordging/webhooks handling)
     const issues =
       payload.is_active && !payload.is_healthy
-        ? payload.human_issues ||
-          (payload.issues ? [payload.issues] : undefined)
-        : undefined;
+        ? payload.human_issues || (payload.issues ? [payload.issues] : null)
+        : null;
     const patch: Partial<DBSession & DBStream> = {
-      isHealthy: payload.is_active ? payload.is_healthy : undefined,
+      isHealthy: payload.is_active ? payload.is_healthy : null,
       issues,
       // do not clear the `lastSeen` field when the stream is not active
       ...(payload.is_active ? { lastSeen: Date.now() } : null),
     };
 
     // Since we might need to delete some fields, use replace instead of update
-    await db.stream.replace({ ...stream, ...patch });
+    await db.stream.update(stream.id, patch);
 
     if (payload.session_id) {
       const session = await db.session.get(payload.session_id, {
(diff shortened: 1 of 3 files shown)


leszko (approved):

packages/api/src/controllers/stream.ts
```diff
@@ -897,18 +897,17 @@ app.post(
     // compatibility with /setactive recordging/webhooks handling)
     const issues =
       payload.is_active && !payload.is_healthy
-        ? payload.human_issues ||
-          (payload.issues ? [payload.issues] : undefined)
-        : undefined;
+        ? payload.human_issues || (payload.issues ? [payload.issues] : null)
+        : null;
```
victorges, line 901: In case of an array, we have a reasonable default value that does indicate "no issues". This would be the empty array `[]`. WDYT of using that instead of `null` here?

packages/api/src/controllers/stream.ts
```diff
@@ -897,18 +897,17 @@ app.post(
...
-      isHealthy: payload.is_active ? payload.is_healthy : undefined,
+      isHealthy: payload.is_active ? payload.is_healthy : null,
       issues,
       // do not clear the `lastSeen` field when the stream is not active
       ...(payload.is_active ? { lastSeen: Date.now() } : null),
     };
 
     // Since we might need to delete some fields, use replace instead of update
-    await db.stream.replace({ ...stream, ...patch });
+    await db.stream.update(stream.id, patch);
```
victorges, line 910: neat

packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
```diff
@@ -233,7 +233,7 @@ const SaveTargetDialog = ({
   return (
     <AlertDialog open={isOpen} onOpenChange={onOpenChange}>
       <AlertDialogContent
-        css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4" }}
+        css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4", zIndex: 1 }}
```
victorges, line 236: this fixes the player on top thing? 👀

victorges (approved):
lgtm

packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
gioelecerati: yeah!

packages/api/src/controllers/stream.ts
gioelecerati: Mmm yeah makes sense! The tests where checking for `toBeUndefined` tho, will change them to look for an empty array, sounds good?

victorges: Yeah that made sense! Keeping as null sounds ok anyway :)

PR #1920, opened 2023-10-13
clip: clip by session id

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/session.ts
+++ b/packages/api/src/controllers/session.ts
@@ -157,8 +157,8 @@ app.get("/:id", authorizer({}), async (req, res) => {
 
   let originalRecordingUrl: string | null = null;
 
-  if (req.query.sourceRecording == "true") {
-    const { url } = await getSourceRecordingBySessionId(req.params.id, req);
+  if (req.query.sourceRecording === "true") {
+    const { url } = await buildRecordingUrl(session, req);
     originalRecordingUrl = url;
     session.sourceRecordingUrl = originalRecordingUrl;
   }
@@ -216,26 +216,6 @@ const adminOnlyFields: (keyof DBSession)[] = ["deleted", "broadcasterHost"];
 
 const privateFields: (keyof DBSession)[] = ["recordObjectStoreId", "version"];
 
-export async function getSourceRecordingBySessionId(
-  sessionId: string,
-  req: Request
-) {
-  const session = await db.session.get(sessionId);
-
-  if (!session) {
-    console.log(`
-      session: recording by session id: no recording found for session ${sessionId}
-    `);
-    return {
-      url: null,
-      session: null,
-      objectStoreId: null,
-    };
-  }
-
-  return await buildRecordingUrl(session, req);
-}
-
 export async function getRunningRecording(content: DBStream, req: Request) {
   let objectStoreId: string;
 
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/session.ts
```diff
@@ -145,13 +145,24 @@ app.get("/:id", authorizer({}), async (req, res) => {
   let session = await db.session.get(req.params.id);
   if (
     !session ||
-    ((session.userId !== req.user.id || session.deleted) && !req.user.admin)
+    ((session.userId !== req.user.id || session.deleted) &&
+      !req.user.admin &&
+      !LVPR_SDK_EMAILS.includes(req.user.email))
```
victorges, line 150: 😭

packages/api/src/controllers/session.ts
```diff
@@ -145,13 +145,24 @@ app.get("/:id", authorizer({}), async (req, res) => {
...
   ) {
     // do not reveal that session exists
     res.status(404);
     return res.json({ errors: ["not found"] });
   }
   res.status(200);
+
+  let originalRecordingUrl: string | null = null;
+
+  if (req.query.sourceRecording == "true") {
```
victorges: use triple equals in javascript.
```suggestion
  if (req.query.sourceRecording === "true") {
```

packages/api/src/controllers/session.ts
```diff
@@ -145,13 +145,24 @@ app.get("/:id", authorizer({}), async (req, res) => {
...
+    const { url } = await getSourceRecordingBySessionId(req.params.id, req);
+    originalRecordingUrl = url;
+    session.sourceRecordingUrl = originalRecordingUrl;
```
victorges, line 163: Keep in mind that these will not be available forever right. Do we have any check to return a clear error to the user when the recording is already deleted?

packages/api/src/controllers/session.ts
```diff
@@ -204,6 +216,26 @@ const adminOnlyFields: (keyof DBSession)[] = ["deleted", "broadcasterHost"];
...
+    return {
+      url: null,
+      session: null,
+      objectStoreId: null,
+    };
+  }
+
+  return await buildRecordingUrl(session, req);
+}
+
```
victorges: This function seems unnecessary. Every caller of it already has the `session` object available. Then we are doing another round-trip to the DB to fetch it again. Callers can just use `buildRecordingUrl` directly.

victorges (approved):

packages/api/src/controllers/session.ts
gioelecerati: Yeah, made this in case we later want to include session source recording URL in the playback API endpoint as well with /api/playback/SESSION_ID, but I can remove it, don't think it's going to happen

packages/api/src/controllers/session.ts
gioelecerati: I know

packages/api/src/controllers/session.ts
gioelecerati: @emranemran How long are these source recording saved?

emranemran: 6 days

PR #1921, opened 2023-10-14
vod: specify content-length when creating presigned urls

gioelecerati:

This is coming out of nothing, just a thing that came to my mind.

With this, we allow developers to specify a `contentLength` parameter in the `request-upload` api that generates presigned URLs.
This way, that URL is going to be valid exclusively for files with that exact size in bytes.

This enables tier enforcing for platforms, for example, a platform providing free tier functionalities with a max asset size of 50MB for users would generate the presigned URL in the backend collecting the content-length of the user file and providing it to our function.

If the user tries to upload a different file, a 403 status code would get returned.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -508,12 +508,17 @@ async function genUploadUrl(
   playbackId: string,
   objectStoreId: string,
   jwtSecret: string,
-  aud: string
+  aud: string,
+  contentLength?: number
 ) {
   const uploadedObjectKey = `directUpload/${playbackId}`;
   const os = await getActiveObjectStore(objectStoreId);
 
-  const presignedUrl = await getS3PresignedUrl(os, uploadedObjectKey);
+  const presignedUrl = await getS3PresignedUrl(
+    os,
+    uploadedObjectKey,
+    contentLength
+  );
   const uploadToken = jwt.sign({ playbackId, presignedUrl, aud }, jwtSecret, {
     algorithm: "HS256",
   });
@@ -877,7 +882,8 @@ app.post(
       playbackId,
       vodObjectStoreId,
       jwtSecret,
-      jwtAudience
+      jwtAudience,
+      req.body.contentLength
     );
 
     const ingests = await req.getIngest();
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -1064,7 +1069,17 @@ app.put("/upload/direct", async (req, res) => {
...
+  proxy.on("error", function (err, req, proxyRes) {
+    console.error("Proxy error:", err);
+    if (!res.headersSent) {
+      res.status(403);
+      res.end();
+      proxyRes.end();
+    } else {
+      res.end();
+      proxyRes.end();
+    }
```
victorges, line 1081:
```suggestion
    if (!res.headersSent) {
      res.status(403);
    }
    res.end();
    proxyRes.end();
```

packages/api/src/controllers/asset.ts
```diff
@@ -1064,7 +1069,17 @@ app.put("/upload/direct", async (req, res) => {
       );
     }
   });
-
+  proxy.on("error", function (err, req, proxyRes) {
+    console.error("Proxy error:", err);
+    if (!res.headersSent) {
+      res.status(403);
```
victorges, line 1075: would this work?
```suggestion
      res.status(proxyRes.status);
```

packages/api/src/controllers/helpers.ts
```diff
@@ -223,13 +223,22 @@ export async function getObjectStoreS3Config(
...
+  contentLength?: number
+) {
   const config = await getObjectStoreS3Config(os);
   const s3 = new S3Client(config);
-  const putCommand = new PutObjectCommand({
+  let putCommandConfig = {
     Bucket: config.bucket,
     Key: objectKey,
-  });
+    ContentLength: null,
```
victorges, line 236: Could be just `ContentLength: contentLength || null`

packages/api/src/schema/db-schema.yaml
```diff
@@ -920,6 +920,8 @@ components:
           description: Object store ID where the asset is stored
           writeOnly: true
           example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+        contentLength:
+          type: number
```
victorges, line 924: This should be in `api-schema` since its public

victorges (approved):
LGTM

PR #1922, opened 2023-10-14
user: patch api & change email

gioelecerati:

- Create user patch endpoint
- Allow to change emails, sending a verification email for the new one
- create a field `newEmail` in the db to handle verification
- Changing emails for admins is always disallowed
- Admins can update info of users

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -480,6 +480,109 @@ app.post("/", validatePost("user"), async (req, res) => {
   res.json(user);
 });
 
+app.patch("/", authorizer({}), async (req, res) => {
+  const { email } = req.body;
+  const userId = req.user.id;
+
+  // Validate the new email
+  const emailValid = validator.validate(email);
+  if (!emailValid) {
+    return res.status(422).json({ errors: ["Invalid email"] });
+  }
+
+  const isEmailRegisteredAlready = await isEmailRegistered(email);
+  if (isEmailRegisteredAlready) {
+    return res.status(409).json({
+      errors: ["This email is already registered. Please choose another one."],
+    });
+  }
+
+  const emailValidToken = uuid();
+
+  if (req.user.admin) {
+    if (!req.body.userId) {
+      console.log(`
+        Admin user ${req.user.id} attempted to change email without providing userId
+      `);
+      return res
+        .status(400)
+        .json({ errors: ["userId is required for admins"] });
+    }
+
+    const user = await db.user.get(req.body.userId);
+
+    if (user.admin) {
+      return res
+        .status(400)
+        .json({ errors: ["Cannot change email of admins"] });
+    }
+
+    if (!user) {
+      return res.status(404).json({ errors: ["User not found"] });
+    }
+
+    await db.user.update(req.body.userId, {
+      email: email,
+    });
+
+    res.status(200);
+    return res.json({ message: "Email updated successfully." });
+  }
+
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -449,6 +449,109 @@ app.post("/", validatePost("user"), async (req, res) => {
   res.json(user);
 });
 
+app.patch("/", authorizer({}), async (req, res) => {
```
victorges: This should have the ID in the path
`PATCH /user/:id`

packages/api/src/controllers/user.ts
```diff
@@ -449,6 +449,109 @@ app.post("/", validatePost("user"), async (req, res) => {
...
+
+    const user = await db.user.get(req.body.userId);
+
+    if (user.admin) {
+      return res
+        .status(400)
+        .json({ errors: ["Cannot change email of admins"] });
+    }
+
+    if (!user) {
```
victorges, line 528: Using `user` before checking if its not null

packages/api/src/controllers/user.ts
```diff
@@ -449,6 +449,109 @@ app.post("/", validatePost("user"), async (req, res) => {
...
+      email: email,
+      supportAddr: req.config.supportAddr,
+      sendgridTemplateId: req.config.sendgridTemplateId,
+      sendgridApiKey: req.config.sendgridApiKey,
+      subject: "Verify Your New Email Address for Livepeer Studio",
+      preheader: "Email Verification Needed!",
+      buttonText: "Verify Email",
+      buttonUrl: frontendUrl(
+        req,
+        `/verify-new-email?${qs.stringify({
```
victorges, line 557: Check what page this will open

packages/api/src/controllers/user.ts
```diff
@@ -449,6 +449,109 @@ app.post("/", validatePost("user"), async (req, res) => {
...
+      text: ["Please verify your new email address."].join("\n\n"),
+    });
+
+    res.status(200).json({ message: "Verification email sent." });
+  } catch (err) {
+    res.status(400).json({ errors: [`Error sending email: ${err}`] });
+  }
+});
+
+app.get("/verify/new-email", async (req, res) => {
```
victorges, line 572: Might be able to use the exact same flow (front + api) from current initial email verification. Maybe just add some ifs/params there to hadnle the "update email" differently from "registration'

packages/api/src/controllers/user.ts
```diff
@@ -449,6 +449,109 @@ app.post("/", validatePost("user"), async (req, res) => {
...
 
+app.patch("/", authorizer({}), async (req, res) => {
+  const { email } = req.body;
+  const userId = req.user.id;
+
+  // Validate the new email
+  const emailValid = validator.validate(email);
+  if (!emailValid) {
+    return res.status(422).json({ errors: ["Invalid email"] });
+  }
```
victorges: Make sure to lowercase the email as well
`const lowercaseEmail = email.toLowerCase();`

packages/api/src/schema/api-schema.yaml
```diff
@@ -1771,6 +1771,9 @@ components:
           type: string
           description: user email address
           example: useremail@gmail.com
+        newEmail:
+          type: string
+          description: temporary field for email change
```
victorges: This should be in `db-schema` since it is an internal field.

victorges (approved):
LGTM

PR #1923, opened 2023-10-16
api: Nuke streams asynchronously w/ events

victorges:
Avoids a direct API->Mist connection and should be much more reliable with mapics' reconciliation logic.

- Remove previous direct-mist nuking code
- Add logic to send event when nuking streams instead
- Pairs with https://github.com/livepeer/catalyst-api/pull/922

Probably better to test it in the box... Will look for help

Nuking

**Checklist**


leszko (approved):
Really nice how the logic simplified 👏

packages/api/src/controllers/user.ts
```diff
@@ -31,8 +31,9 @@ import {
   sendgridValidateEmail,
   toStringValues,
   FieldsMap,
+  triggerCatalystStreamUpdated,
```
iameli-streams: Unused?

PR #1927, opened 2023-10-17
api: abstract requesterId

gioelecerati:

Make `generateRequesterId` an helper function and put it into the task root, so that it's usable for rate-limiting outside of clips

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/clip.ts
+++ b/packages/api/src/controllers/clip.ts
@@ -24,9 +24,11 @@ import {
   parseOrder,
   makeNextHREF,
   generateRequesterId,
+  sqlQueryGroup,
 } from "./helpers";
 import sql from "sql-template-strings";
 import { DBStream } from "../store/stream-table";
+import { getProcessingTasksByRequesterId } from "./task";
 
 const app = Router();
 export const LVPR_SDK_EMAILS = ["livepeerjs@livepeer.org"];
@@ -56,31 +58,6 @@ app.use(
     return data;
   })
 );
-
-async function getProcessingClipsByRequesterId(
-  requesterId: string,
-  ownerId: string
-): Promise<WithID<Task>[]> {
-  const scheduledAt = Date.now() - 5 * 60 * 1000;
-  const [tasks] = await db.task.find([
-    sql`data->>'userId' = ${ownerId}`,
-    sql`coalesce((data->>'scheduledAt')::bigint, 0) > ${scheduledAt}`,
-  ]);
-  let runningClips: WithID<Task>[] = [];
-  const tasksByRequester = tasks.filter((t) => t.requesterId === requesterId);
-  tasksByRequester.forEach((task) => {
-    if (task.type === "clip") {
-      if (
-        task.status.phase === "running" ||
-        task.status.phase === "pending" ||
-        task.status.phase === "waiting"
-      )
-        runningClips.push(task);
-    }
-  });
-
-  return runningClips;
-}
 
 app.post("/", validatePost("clip-payload"), async (req, res) => {
   const playbackId = req.body.playbackId;
@@ -127,12 +104,14 @@ app.post("/", validatePost("clip-payload"), async (req, res) => {
     throw new NotFoundError("Content not found");
   }
 
-  const processingClips = await getProcessingClipsByRequesterId(
+  const processingClips = await getProcessingTasksByRequesterId(
     requesterId,
-    req.user.id
+    req.user.id,
+    "clip",
+    5
   );
 
-  if (processingClips.length >= MAX_PROCESSING_CLIPS) {
+  if (processingClips.length >= MAX_PROCESSING_CLIPS && !clippingUser.admin) {
     throw new ForbiddenError("Too many clips are being processed.");
   }
 
@@ -336,15 +315,5 @@ export const getClips = async (
 
   return res.json(output);
 };
-
-function sqlQueryGroup(values: string[]) {
-  const query = sql`(`;
-  values.forEach((value, i) => {
-    if (i) query.append(`, `);
-    query.append(sql`${value}`);
-  });
-  query.append(`)`);
-  return query;
-}
 
 export default app;
--- a/packages/api/src/controllers/clip.ts
+++ b/packages/api/src/controllers/clip.ts
@@ -61,16 +61,26 @@ app.use(
 );
 
 async function getProcessingClipsByRequesterId(
-  requesterId: string
+  requesterId: string,
+  ownerId: string
 ): Promise<WithID<Asset>[]> {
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/clip.ts
```diff
@@ -61,49 +58,35 @@ app.use(
...
-    sql`data->'source'->>'requesterId' = ${requesterId}`,
-    sql`coalesce((data->>'createdAt')::bigint, 0) > ${createdAfter}`,
+  requesterId: string,
+  ownerId: string
+): Promise<WithID<Task>[]> {
+  const scheduledAt = Date.now() - 5 * 60 * 1000;
+  const [tasks] = await db.task.find([
+    sql`data->>'userId' = ${ownerId}`,
+    sql`coalesce((data->>'scheduledAt')::bigint, 0) > ${scheduledAt}`,
   ]);
```
victorges: Can we have that as a dedicated query function on `task-table`? Like the other rate limiting queries, with a clear name so we can easily reason about in all the rate-limiting logics we create all around. e.g. this is probably just `getPendingClipTasks` or something

packages/api/src/controllers/clip.ts
```diff
@@ -61,49 +58,35 @@ app.use(
...
+  let runningClips: WithID<Task>[] = [];
+  const tasksByRequester = tasks.filter((t) => t.requesterId === requesterId);
+  tasksByRequester.forEach((task) => {
+    if (task.type === "clip") {
+      if (
+        task.status.phase === "running" ||
+        task.status.phase === "pending" ||
+        task.status.phase === "waiting"
```
victorges: This should be filtered directly on the query. You might also be able to re-use [this index](https://github.com/livepeer/livepeer-com/blob/23b1aa453bbef83981ad2547169c7149f0436e35/packages/api/src/store/task-table.ts#L77) if you query by `updatedAt` instead of `scheduledAt` (and pending tasks are gonna be updated at least every minute)

packages/api/src/controllers/clip.ts
```diff
@@ -61,49 +58,35 @@ app.use(
...
+  ownerId: string
+): Promise<WithID<Task>[]> {
+  const scheduledAt = Date.now() - 5 * 60 * 1000;
+  const [tasks] = await db.task.find([
+    sql`data->>'userId' = ${ownerId}`,
+    sql`coalesce((data->>'scheduledAt')::bigint, 0) > ${scheduledAt}`,
   ]);
+  let runningClips: WithID<Task>[] = [];
+  const tasksByRequester = tasks.filter((t) => t.requesterId === requesterId);
+  tasksByRequester.forEach((task) => {
```
victorges: Btw we should either write this iteratively (`forEach`, or even better a literate `for (.. of ..)`), or functionally `filter`. Ideally do not mix the 2 patterns otherwise it's just a back-and-forth to read through it.

Using the functional approach (generally my preference unless it's too complex of a pipeline), you would just chain 2 or more `.filter()` calls with conditions for which to filter the tasks.

packages/api/src/controllers/helpers.ts
```diff
@@ -12,6 +12,9 @@ import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 import base64url from "base64url";
 import { CreatorId, InputCreatorId, ObjectStore } from "../schema/types";
 import { BadRequestError } from "../store/errors";
+import * as nativeCrypto from "crypto";
+
+const SALT = nativeCrypto.randomBytes(32).toString("hex");
```
victorges: Reminder to move this to a CLI arg.

packages/api/src/controllers/helpers.ts
```diff
@@ -234,6 +237,35 @@ export async function getS3PresignedUrl(os: ObjectStore, objectKey: string) {
   return getSignedUrl(s3, putCommand, { expiresIn });
 }
 
+export async function generateRequesterId(req: Request, playbackId: string) {
+  const origin =
```
victorges: I would use a different name than `origin`, since this is already it is already [well-defined in HTTP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) and is not what it means here. Why not just call this `ip`?

packages/api/src/controllers/helpers.ts
```diff
@@ -234,6 +237,35 @@ export async function getS3PresignedUrl(os: ObjectStore, objectKey: string) {
...
+    req.headers["cf-connecting-ip"] ||
+    req.headers["true-client-ip"] ||
+    req.headers["x-forwarded-for"];
+
+  let requesterId: string;
+  if (!origin) {
+    console.log(`
+        unable to determine origin of requester for user=${req.user.id} when clipping playbackId=${playbackId}
```
victorges:
```suggestion
        unable to determine ip of requester for user=${req.user.id} when clipping playbackId=${playbackId}
```

packages/api/src/controllers/helpers.ts
```diff
@@ -234,6 +237,35 @@ export async function getS3PresignedUrl(os: ObjectStore, objectKey: string) {
...
+    `);
+    requesterId = `UNKNOWN-${playbackId}`;
+  } else {
+    console.log(`
+         user=${req.user.id} requesterId generated for playbackId=${playbackId} from origin=${origin}
+    `);
```
victorges: Since this is just the success case, probably too verbose to send to prod? Also leaks the IP->requestId association (unless we do want that to be queryable internally, wdyt)

packages/api/src/controllers/helpers.ts
```diff
@@ -234,6 +237,35 @@ export async function getS3PresignedUrl(os: ObjectStore, objectKey: string) {
...
+    let originString = Array.isArray(origin) ? origin.join(",") : origin;
```
victorges: another simpler alternative is to just make `ip.toString()` whenever you use it. if it's already a string, the same string is returned

victorges (approved):
lgtm with some comments

packages/api/src/controllers/clip.ts
gioelecerati: Right! I moved it in the task table and made the `clipType` as optional parameter, so it can used not just for clips (`getProcessingTaskByRequesterId`)

packages/api/src/controllers/task.ts
```diff
@@ -94,6 +95,29 @@ export function taskParamsWithoutCredentials(
   return result;
 }
 
+export async function getProcessingTasksByRequesterId(
```
victorges, line 98: I meant `task-table.ts`, not the task controller! The tables do the queries, the controllers handle APIs. No need to open a PR to fix this though, commenting for the future.

PR #1928, opened 2023-10-17
api/helpers: throw an error upon unsuccessful catalyst fetch

iameli-streams:
We presently just return the response which won't actually let us know if there's a problem.


victorges (approved):
noice

PR #1929, opened 2023-10-17
stream nuking: bit of cleanup

iameli-streams:
Only nuke parents and don't display an error on the frontend.

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -31,10 +31,10 @@ import {
   sendgridValidateEmail,
   toStringValues,
   FieldsMap,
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -110,7 +110,10 @@ export async function terminateUserStreams(
   req: Request,
   userId: string
 ): Promise<void> {
-  const [streams] = await db.stream.find({ userId });
+  const [streams] = await db.stream.find([
+    sql`stream.data->>'parentId' IS NULL`,
```
victorges, line 114: Good catch 😬

victorges (approved):

PR #1931, opened 2023-10-19
api: fix resend validation email

gioelecerati:

- Check if the email is valid with the validator in the `verify-email` api
- Resend the validation email on token generation fail 

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -560,6 +560,8 @@ app.post("/token", validatePost("user"), async (req, res) => {
     !isTest &&
     user.createdAt > EMAIL_VERIFICATION_CUTOFF_DATE
   ) {
+    // Resend the verification email
+    await sendVerificationEmail(req, user, user.stripeProductId);
     res.status(403);
     return res.json({
       errors: [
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -560,6 +560,8 @@ app.post("/token", validatePost("user"), async (req, res) => {
     !isTest &&
     user.createdAt > EMAIL_VERIFICATION_CUTOFF_DATE
   ) {
+    // Resend the verification email
+    await sendVerificationEmail(req, user, user.stripeProductId);
```
victorges, line 564: So we're gonna resend the verification email automatically after a signin attempt, is that it? Perhaps you want to change line 561 to check the last verification email timestamp instead of the user creation ts (not sure we have that tho)

packages/api/src/controllers/user.ts
```diff
@@ -628,7 +638,10 @@ app.post("/verify-email", validatePost("verify-email"), async (req, res) => {
     } = req.config;
 
     try {
+      // This is a test of the Sendgrid email validation API. Remove this
```
victorges, line 641: nice doc

victorges (approved):

PR #1947, opened 2023-11-01
Return VOD thumbs VTT URL when available

mjh1:


packages/api/src/controllers/asset.ts
```diff
@@ -345,6 +346,28 @@ export function getPlaybackUrl(
...
+    asset.playbackId,
+    "thumbnails",
+    "thumbnails.vtt"
+  );
+
+  let params = {
+    method: "HEAD",
+    timeout: 5 * 1000,
+  };
+  let resp = await fetchWithTimeout(vttUrl, params);
```
leszko: Why do you need to make this HTTP call here? Can't you just return `vttUrl`?

leszko (commented):
Added a comment. Also, the CI is failing, could you check?

packages/api/src/controllers/asset.ts
mjh1: Because the vtt will not always be present, it's not switched on for all users yet.

mjh1: Tidied this up now to check the returned files on the asset

leszko (approved):
Looks good. One thing we could add is a unit test, somewhere here: https://github.com/livepeer/studio/blob/b5db9dc13b06d0633308344463aca62ae2349651/packages/api/src/controllers/playback.test.ts#L198

PR #1956, opened 2023-11-03
user: disallow temp email & better gate logs

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -10,6 +10,7 @@ import { products } from "../config";
 import hash from "../hash";
 import logger from "../logger";
 import { authorizer, validatePost } from "../middleware";
+import { isFakeEmail } from "fakefilter";
 import {
   CreateCustomer,
   CreateSubscription,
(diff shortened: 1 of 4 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -317,6 +318,12 @@ app.post("/", validatePost("user"), async (req, res) => {
     return;
   }
 
+  if (isFakeEmail(email)) {
+    res.status(422);
+    res.json({ errors: ["invalid email"] });
+    return;
+  }
```
victorges, line 325: Actuallyyyy we already have an even fancier email validation done with the Sendgrid API here: https://github.com/livepeer/livepeer-com/blob/142c9e588e964f014c80248ba004db153387d6da/packages/api/src/controllers/user.ts#L402

WDYT of making that non-background instead (or as well)? Meaning we return this error when Sendgrid says the email is fake as well. You can check the logs from that validation function to see some data on previous validations. The times I checked it seemed pretty good but we didn't go forward with enabling it.

victorges (approved):

PR #1960, opened 2023-11-07
Support patching profiles in a stream

leszko:


packages/api/src/controllers/stream.ts
```diff
@@ -1349,16 +1349,34 @@ app.patch(
...
     if (record != undefined && stream.isActive && stream.record != record) {
       res.status(400);
       return res.json({
         errors: ["cannot change 'record' field while stream is active"],
       });
     }
+    if (
+      profiles != undefined &&
+      stream.isActive &&
+      stream.profiles != profiles
```
victorges: **nit:** I think this last check is always true since the `profiles` array is a newly allocated instance parsed from the request. Even 2 empty arrays are not really equal 
```bash
$ node
Welcome to Node.js v16.20.1.
Type ".help" for more information.
> [] == []
false
```

We can probably remove this check or can do [a deep equality check using `lodash`](https://www.geeksforgeeks.org/lodash-_-isequal-method/) which is already available in the api project (just removing sounds fine tho)

victorges (approved):
LGTM

packages/api/src/controllers/stream.ts
leszko: Thanks! Removed in b98b95a67ee5c1ce0d51de17b0a6f76ca6f4b4e6

PR #1962, opened 2023-11-07
usage: enable hourly reporting

gioelecerati:

This enables hourly reporting for usage

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/usage.ts
+++ b/packages/api/src/controllers/usage.ts
@@ -338,17 +338,7 @@ app.post(
   authorizer({ anyAdmin: true }),
   validatePost("usage"),
   async (req, res) => {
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/usage.ts
```diff
@@ -385,6 +375,10 @@ app.post(
       }
     }
 
+    // New automated billing usage report
```
victorges, line 378: Do we still need all the old reporting logic? Could be good to clean that up as well.

victorges (approved):
LGTM

packages/api/src/controllers/usage.ts
gioelecerati: We need it! the explorer is still relying on that

PR #1964, opened 2023-11-08
Fix Stream Idle by using the isActive from webhook, not stream

leszko:
One more fix for another scenario in which stream can stay active while it's not active anymore.

Here's what was the issue:
1. We get more two concurrent requests `/health/hook` and `/setactive` and they both change the `stream` object.
2. In `health/hook` we first get the stream data, then patch it, and then replace the whole `stream` object in db
3. So, here is what may happen:
 - `/health/hook` function is called, it gets the current stream with `isActive: true`
 - `/setactive` function is called to change the stream active to `isActive: false`
 - `/setactive` updates DB and the stream in DB is `isActive: false`
 - /health/hook` replaces the stream object with `isActive: true`

In result, we have the stream with `isActive: true`, which is incorrect.

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -902,6 +902,7 @@ app.post(
     const patch: Partial<DBSession & DBStream> = {
       isHealthy: payload.is_active ? payload.is_healthy : undefined,
       issues,
+      isActive: payload.is_active,
       // do not clear the `lastSeen` field when the stream is not active
       ...(payload.is_active ? { lastSeen: Date.now() } : null),
     };


gioelecerati (approved):

packages/api/src/controllers/stream.ts
```diff
@@ -902,6 +902,7 @@ app.post(
     const patch: Partial<DBSession & DBStream> = {
       isHealthy: payload.is_active ? payload.is_healthy : undefined,
       issues,
+      isActive: payload.is_active,
```
victorges, line 905: As per the comment on L895 above, this can introduce other race conditions with recordings instead, which is the case if the health hook arrives before than the `/setactive` hook, with the `is_active` being `false`. This will cause the recordings not to be generated for that stream because the `/setactive` will detect the change is a noop here https://github.com/livepeer/livepeer-com/blob/2d1e24eaf8ef9eaa089936bea47373d696f85a93/packages/api/src/controllers/stream.ts#L1120

@gioelecerati had previously fixed this on https://github.com/livepeer/studio/pull/1916, by replacing the "read and replace" with a single "patch" of the stream object. It seems to have been reverted later for an unclear reason ("until we're sure it's not causing issues") on https://github.com/livepeer/studio/pull/1954, but I believe it is the better fix for this not to introduce other potential race conditions, which I believe this version does.

@gioelecerati was there any conclusion about the other fix introducing any issues?

PR #1968, opened 2023-11-08
api: bulk creation of multistream targets

gioelecerati:

Allow to bulk create multi stream targets

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -688,10 +688,10 @@ components:
               - user not found
     multistream-target-input:
       oneOf:
-        - $ref: "#/multistream-target"
+        - $ref: "#/components/schemas/multistream-target"
         - type: array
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -686,6 +686,12 @@ components:
             example:
               - id not provided
               - user not found
+    multistream-target-input:
+      oneOf:
+        - $ref: "#/multistream-target"
```
victorges: I don't think this ref will work. I simplified it on the message I wrote.

victorges (approved):

packages/api/src/schema/api-schema.yaml
gioelecerati: lol, I just copy pasted it without actually look the string content! fixing it

PR #1969, opened 2023-11-09
stream: fix idle state race condition (#1916)

leszko:
I think we should resurrect this PR from @gioelecerati . I'm testing it in staging now and it seems to work fine. I didn't experience any issues with webhooks or anything. That makes me think that maybe the webhook issues experienced by the customer were unrelated.

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -896,18 +896,17 @@ app.post(
     // compatibility with /setactive recordging/webhooks handling)
     const issues =
       payload.is_active && !payload.is_healthy
-        ? payload.human_issues ||
-          (payload.issues ? [payload.issues] : undefined)
-        : undefined;
+        ? payload.human_issues || (payload.issues ? [payload.issues] : [])
+        : null;
     const patch: Partial<DBSession & DBStream> = {
-      isHealthy: payload.is_active ? payload.is_healthy : undefined,
+      isHealthy: payload.is_active ? payload.is_healthy : null,
       issues,
       // do not clear the `lastSeen` field when the stream is not active
       ...(payload.is_active ? { lastSeen: Date.now() } : null),
     };
 
     // Since we might need to delete some fields, use replace instead of update
-    await db.stream.replace({ ...stream, ...patch });
+    await db.stream.update(stream.id, patch);
 
     if (payload.session_id) {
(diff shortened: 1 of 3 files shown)


gioelecerati (approved):
Now with tests! LGTM

packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
```diff
@@ -233,7 +233,7 @@ const SaveTargetDialog = ({
   return (
     <AlertDialog open={isOpen} onOpenChange={onOpenChange}>
       <AlertDialogContent
-        css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4" }}
+        css={{ maxWidth: 450, px: "$5", pt: "$4", pb: "$4", zIndex: 1 }}
```
gioelecerati: This is not needed, it was an unrelated change that got merged in another PR from @0xcadams

packages/api/src/controllers/stream.ts
```diff
@@ -896,18 +896,17 @@ app.post(
...
+        : null;
     const patch: Partial<DBSession & DBStream> = {
-      isHealthy: payload.is_active ? payload.is_healthy : undefined,
+      isHealthy: payload.is_active ? payload.is_healthy : null,
       issues,
       // do not clear the `lastSeen` field when the stream is not active
       ...(payload.is_active ? { lastSeen: Date.now() } : null),
     };
 
     // Since we might need to delete some fields, use replace instead of update
```
victorges: I think this comment can be removed

victorges (approved):
LGTM!

adamsoffer (approved):
LGTM

packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
leszko: Reverted in 614e408f7c7a468b4d321b6076a3d9ad1edfde45

packages/api/src/controllers/stream.ts
leszko: Removed in bd9959fa85ca4c1f86d1a638960c4e7e226cc161

PR #1978, opened 2023-11-20
api: asset: rerun tasks

gioelecerati:

Allow to rerun tasks by task id or asset id

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -1208,9 +1208,14 @@ app.patch(
   }
 );
 
-app.post("/:id/retry", authorizer({ anyAdmin: true }), async (req, res) => {
+app.post("/:id/retry", authorizer({}), async (req, res) => {
   const { id } = req.params;
   const asset = await db.asset.get(id);
+
+  if (!req.user.admin && asset.userId !== req.user.id) {
+    throw new ForbiddenError(`users may only retry their own assets`);
+  }
+
   if (!asset) {
     throw new NotFoundError(`Asset not found`);
   }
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -1208,6 +1208,39 @@ app.patch(
   }
 );
 
+app.post("/:id/retry", authorizer({ anyAdmin: true }), async (req, res) => {
```
victorges: Either way, do these APIs need to be admin-only? Feels like it could be useful for users to retry their tasks if they want (dashboard integration would be a plus). Remember to add a check for the user Id if you remove the admin restriction.

packages/api/src/controllers/task.ts
```diff
@@ -279,6 +279,34 @@ app.post(
...
+    return res.status(500).json({ errors: ["user not found"] });
+  }
+
+  if (!req.user.admin && req.user.id !== task.userId) {
+    return res.status(403).json({
+      errors: ["users may only retry their own tasks"],
+    });
+  }
+
+  await req.taskScheduler.retryTask(task, "retry requested by user");
```
victorges: This will not work. The task is already at its last retry and this function will also have a sleep of up to 3 minutes which is unacceptable in an API call. You need to go inside this function you are calling and make sure it does what you want.

victorges (changes requested):

packages/api/src/controllers/task.ts
victorges: Also, this error message doesn't make sense The task stores the error message from the last retry that failed, while here you are replacing it with something that is not an error message. You should either keep the same message or clear it and restart the retry count to 0 (I think the ideal flow would be keeping the error message & retry count, just retrying until a higher retry count like from 3 to 6 instead of 0 to 3?)

victorges (approved):
LGTM

PR #1979, opened 2023-11-20
asset: never reprocess assets when duplicated

gioelecerati:

Just a proposal to never reprocess assets when the import url is duplicated per userId

**Screenshots (optional)**

**Checklist**


victorges:
As per our discussion on Discord, I am unsure about this since it might not make sense all users to share the same asset objects between their end-users because the underlying videos is the same. Different users wouldn't be able to set custom access control conditions on their version for example. IMO we can optimize this on the infrastructure layer, like sharing the storage for the assets with the same content, but it is still useful to have the separate asset "metadata objects" imo.

PR #1980, opened 2023-11-20
api/schema: Add docs on stream terminate API

victorges:


hthillman:
Whose review is needed here?

PR #1982, opened 2023-11-22
access-control: move rate limit to catalyst

gioelecerati:

Moving the rate limiting in memory to catalyst, to get around caching
Goes along with https://github.com/livepeer/catalyst-api/pull/986

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -18,14 +18,12 @@ import { signatureHeaders, storeTriggerStatus } from "../webhooks/cannon";
 import { Response } from "node-fetch";
 import { fetchWithTimeoutAndRedirects } from "../util";
 import fetch from "node-fetch";
-import { WithID } from "../store/types";
-import { DBStream } from "../store/stream-table";
 import { HACKER_DISABLE_CUTOFF_DATE } from "./utils/notification";
 import { isFreeTierUser } from "./helpers";
 import { cache } from "../store/cache";
 
 const WEBHOOK_TIMEOUT = 30 * 1000;
-const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 5;
+const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 30;
 const app = Router();
 
 interface HitRecord {
@@ -135,10 +133,13 @@ app.post(
 
     const playbackPolicyType = content.playbackPolicy?.type ?? "public";
 
-    if (user.createdAt < HACKER_DISABLE_CUTOFF_DATE) {
-      let limitReached = freeTierLimitReached(content, user, req);
-      if (limitReached) {
-        throw new ForbiddenError("Free tier user reached viewership limit");
+    let response = {};
+
+    if (user.createdAt > HACKER_DISABLE_CUTOFF_DATE) {
+      if (isFreeTierUser(user)) {
+        response = {
+          rateLimit: MAX_ALLOWED_VIEWERS_FOR_FREE_TIER,
+        };
       }
     }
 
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -128,17 +125,18 @@ app.post(
...
+      response = {
+        rateLimit: MAX_ALLOWED_VIEWERS_FOR_FREE_TIER,
+      };
     }
 
     switch (playbackPolicyType) {
       case "public":
-        res.status(204);
-        return res.end();
+        res.status(200);
```
victorges, line 158: Make sure catalyst-api is not checking for a 204 explicitly (e.g. just checking for a 2xx). If it is expecting a 204, we need to first deploy a version of catalyst that allows different status codes here.

packages/api/src/controllers/access-control.ts
```diff
@@ -270,37 +268,4 @@ app.get("/public-key", async (req, res) => {
     });
 });
 
-async function freeTierLimitReached(
```
victorges: So happy to get rid of this! There's even a bonus: catalyst is also much more capable of actually limiting the number of _concurrent viewers_ instead of "number of viewers that start per minute"! Might be a lower hanging fruit for the future

packages/api/src/controllers/access-control.ts
```diff
@@ -128,17 +125,18 @@ app.post(
...
-    if (user.createdAt < HACKER_DISABLE_CUTOFF_DATE) {
-      let limitReached = await freeTierLimitReached(content, user, req);
-      if (limitReached) {
-        throw new ForbiddenError("Free tier user reached viewership limit");
-      }
+    let response = {};
+
+    if (isFreeTierUser(user) && user.createdAt > HACKER_DISABLE_CUTOFF_DATE) {
+      response = {
+        rateLimit: MAX_ALLOWED_VIEWERS_FOR_FREE_TIER,
```
victorges: A couple things regarding this schema:
- Since this schema is defined by Catalyst and implemented by the API, we should probably stick to Catalyst's design choices. Specifically, this should be using `snake_case` instead of` camelCase`
- "Rate limit" is a completely generic term and I think we should be more specific here, especially to allow other limits to be added seamlessly in the future. Specifically, I was thinking of having a `rate_limit` _object_ instead, and inside the object we define a specific limit we want _by name_.

e.g. this whole payload could be
```suggestion
        rate_limit: {
          max_view_starts_per_minute: MAX_ALLOWED_VIEWERS_FOR_FREE_TIER, 
        }  
```
(or sth like that, if the limit is actually something else)

victorges (approved):
LGTM with a couple suggestions

packages/api/src/controllers/access-control.ts
Shih-Yu: @gioelecerati do you think it is possible to limit the number of concurrent viewers for this work as Victor mentioned that it is a low hanging fruit and not worry about it in the future?

gioelecerati: Yes, just rebased it and tested, waiting for green and merge

thomshutt (approved):

packages/api/src/controllers/access-control.ts
```diff
@@ -135,10 +133,13 @@ app.post(
 
     const playbackPolicyType = content.playbackPolicy?.type ?? "public";
 
-    if (user.createdAt < HACKER_DISABLE_CUTOFF_DATE) {
```
victorges, line 138: Wow I was re-reviewing this as it's going with my deploy now and just realized that this condition was inverted? Were we not even trying to limit any new users (i.e. pirates)?? 😆

PR #1983, opened 2023-11-27
Handle update task failure

mjh1:
To address this: https://linear.app/livepeer/issue/VID-428/asset-still-uploading-for-almost-a-week

We should mark the task as failed rather than just leaving it in the previous state. Tested this by placing an error throw line into the code, created an upload task and saw it set to failed, with the old code it was left as `pending`.

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -406,15 +406,25 @@ export class TaskScheduler {
 
   async scheduleTask(task: WithID<Task>, retries = 0) {
     const timestamp = Date.now();
-    await this.updateTask(task, {
-      // only update scheduledAt on the first schedule (retries == 0)
-      scheduledAt: retries ? undefined : timestamp,
-      status: {
-        phase: "waiting",
-        updatedAt: timestamp,
-        retries,
-      },
-    });
+    try {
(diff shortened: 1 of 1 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -406,15 +406,25 @@ export class TaskScheduler {
...
-    await this.updateTask(task, {
-      // only update scheduledAt on the first schedule (retries == 0)
-      scheduledAt: retries ? undefined : timestamp,
-      status: {
-        phase: "waiting",
-        updatedAt: timestamp,
-        retries,
-      },
-    });
+    try {
```
victorges, line 409: wonder if a single try block wouldnt be simpler. then we can log the error and can figure out where it came from, and fail the task with a generic message like "Failed to enqueue task"

victorges (approved):

packages/api/src/task/scheduler.ts
mjh1: Yeah that totally makes sense! So just move the original `try` line up above the `updateTask` call right?

victorges: exactly!

victorges (approved):

PR #1984, opened 2023-11-27
feat: added login and update livepeer.js

0xcadams:


adamsoffer:
@hthillman do we really need all these extra fields? I'm in favor of least amount of friction as possible for signups. Nearly all developer platforms these days ask for an email address or allow you to sign up with github and nothing else. I feel even first and last name are superfluous.

CleanShot 2023-11-27 at 17 42 24@2x
CleanShot 2023-11-27 at 17 42 42@2x
CleanShot 2023-11-27 at 17 43 43@2x
CleanShot 2023-11-27 at 17 45 06@2x

hthillman:
@adamsoffer @0xcadams 

My only real requirement here is Company Name, if someone is using Studio for work. This is so that we can correctly associate Studio accounts with our sales pipeline.

I am okay with removing everything else other than email.

hthillman:
@adamsoffer all the rest is legacy, in general we should revisit the FTUX

hthillman:
I would also love to allow sign-in with github +google but assuming that's a bigger lift that should be tackled when we do seat management

0xcadams:
Okay great - so I can remove name fields and keep company name.

FWIW these signups usually have more screens to them once you enter email/password. The low friction is usually on entering email and then they add more signup friction as a second step.

hthillman:
@0xcadams yeah this is a good callout. IMO we need a wholesale rethink of the account creation flow. We can get a lot of valuable data during this process to help us fine-tune the onboarding and give users a clear and workflow-specific path to success

0xcadams:
Okay, I removed first name and last name fields and simplified the UI. Let me know if this looks good to merge.

Screenshot 2023-11-28 at 2 55 48 PM

Screenshot 2023-11-28 at 2 55 44 PM

hthillman:
@0xcadams 

Nit: Studio -> Livepeer Studio

Also, wdyt about changing  "Are you using Studio for work" to "I am using Livepeer Studio for work"? That feels more apt since there is a checkbox

Lastly, does this information get logged to Hubspot as well as the DB? or just the DB

0xcadams:
@victorges @gioelecerati could you guys take a quick look at this and give any feedback on why it's failing to create a user? The API is responding with `user not created` and not sure why. Maybe firstName/lastName are required fields in the DB?

hthillman:
bump @victorges @gioelecerati! When you have a moment, could you give Chase a review?

victorges:
@0xcadams just pushed a commit to fix the issue I believe you were having: replication delay in our DB.

I reviewed the user registration flow and there's no validation on the first/last name fields. I did find that there is a logic that creates the user and immediately reads it back though. The problem was that it was reading from a read replica so sometimes it would not find the user depending on the replication lag (I could not reproduce the issue, cause it's not a consistent thing). I believe the issue is unrelated to the change here, but I've fixed it anyway. Should be good to go.

Do notice that you won't be able to reproduce the fix until we actually merge this and deploy to staging/production. You can try connecting to a VPN in Europe tho and I believe you will never get the error (the primary DB is used for reading as well).

packages/www/components/Register/index.tsx
```diff
@@ -88,66 +96,73 @@ const Login = ({ id, buttonText, onSubmit, loading, errors }) => {
...
+            <Checkbox
+              onCheckedChange={(checked) =>
+                setIsCompanyAffiliated(checked === true)
+              }
+              checked={isCompanyAffiliated}
+            />
+            <Box css={{ ml: "$1" }}>I am using Livepeer Studio for work</Box>
+          </Flex>
+
+          {isCompanyAffiliated ? (
```
victorges: Can we put this new field after the email/password? Feels kinda weird to be the first thing when it's some very optional information. Especially since we are defaulting this to `true` so it's always displayed there.

Would feel like a more natural flow to at least start with the standard registration fields like user/password and only then deal with auxiliary info/metadata like this.

packages/www/components/Register/index.tsx
```diff
@@ -9,14 +16,15 @@ import BroadcastingProvider from "../Login/BroadcastingProvider";
...
-const Login = ({ id, buttonText, onSubmit, loading, errors }) => {
+const Register = ({ id, buttonText, onSubmit, loading, errors }) => {
   const router = useRouter();
   const [firstName, setFirstName] = useState("");
   const [lastName, setLastName] = useState("");
   const [organization, setOrganization] = useState("");
   const [phone, setPhone] = useState("");
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
+  const [isCompanyAffiliated, setIsCompanyAffiliated] = useState(true);
```
victorges: Especially if we move the fields to after user/password fields, would be much cleaner to default this to `false` IMO. Users might click it on automatic as if it were a terms of service thing and then they see the new field.

packages/www/components/Site/Pricing/faq.md
```diff
@@ -44,9 +44,8 @@ a Hacker account, your account will be disabled until you select a paid plan.
 ### **What currencies does Livepeer Studio accept?**
 
 USD is the only currency accepted right now via the Livepeer Studio dashboard.
-Crypto-native payments are coming later this year. If you prefer to pay with
-cryptocurrency now, you may reach out
-at [help@livepeer.org](mailto:help@livepeer.org)
+Crypto-native payments are coming soon. If you prefer to pay with cryptocurrency
```
victorges, line 47: Can we fix this as well while at it https://discord.com/channels/423160867534929930/611650333243867170/1187168731504320562

victorges (approved):
LGTM

packages/www/components/Register/index.tsx
0xcadams: Yeah it changes the flow of the form, though, so I'm not sure about that. There are different fields depending on how you respond to this. cc @hthillman

PR #1986, opened 2023-11-30
refactor: fixed openapi errors (api-schema)

suhailkakar:
- fixed the openapi validation errors  
- added security scheme
- removed the description from a few apis which are no longer needed (the api-reference is self explanatory)

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -3570,12 +3570,16 @@ paths:
   /access-control/signing-key:
     post:
       summary: Create a signing key
-      description: |
+      description: >
 
-        The publicKey is a representation of the public key, encoded as base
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -3773,22 +3571,12 @@ paths:
...
-        - The publicKey is a representation of the public key, encoded as base
-        64 and is passed as a string,
-        - The privateKey is displayed only on creation. This is the only moment
+        The publicKey is a representation of the public key, encoded as base
+        64 and is passed as a string, and  the privateKey is displayed only on creation. This is the only moment
         where the client can save the private key, otherwise it will be lost.
         Remember to decode your string when signing JWTs.
-        \
-        \
-        Up to 10 signing keys can be generated, after that you must delete at
```
victorges, line 3790: This might be relevant to keep

victorges (approved):
neat

PR #1988, opened 2023-12-06
multistream: atomic add and remove targets from a given stream

gioelecerati:

Allow atomic creation of targets into a given stream and atomic remove

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -1327,6 +1327,83 @@ app.patch(
   }
 );
 
+app.post(
+  "/:id/multistream",
+  authorizer({}),
+  validatePost("target-add-payload"),
+  async (req, res) => {
+    const payload = req.body;
+
+    const stream = await db.stream.get(req.params.id);
+
+    if (!stream || stream.deleted) {
+      res.status(404);
+      return res.json({ errors: ["stream not found"] });
+    }
+
+    if (stream.userId !== req.user.id) {
+      res.status(403);
+      return res.json({ errors: ["stream not found"] });
+    }
+    let streamTargets = stream.multistream;
+
+    let multistream: DBStream["multistream"];
+
+    multistream = streamTargets;
+    multistream.targets.push(payload);
+    multistream = await validateMultistreamOpts(
+      req.user.id,
+      stream.profiles,
+      multistream
+    );
+
+    let patch: StreamPatchPayload & Partial<DBStream> = {
+      multistream,
+    };
+
+    await db.stream.update(stream.id, patch);
+
+    res.status(204);
+    res.end();
+  }
+);
+
+app.delete("/:id/multistream/:targetId", authorizer({}), async (req, res) => {
+  const { id, targetId } = req.params;
+
+  const stream = await db.stream.get(id);
+
+  if (!stream || stream.deleted) {
+    res.status(404);
+    return res.json({ errors: ["stream not found"] });
+  }
+
+  if (stream.userId !== req.user.id) {
+    res.status(403);
+    return res.json({ errors: ["stream not found"] });
+  }
+  let streamTargets = stream.multistream;
+
+  let multistream: DBStream["multistream"];
+
+  multistream = streamTargets;
+  multistream.targets = multistream.targets.filter((t) => t.id !== targetId);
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -1327,6 +1327,83 @@ app.patch(
...
+    if (stream.userId !== req.user.id) {
+      res.status(403);
+      return res.json({ errors: ["stream not found"] });
+    }
+    let streamTargets = stream.multistream;
+
+    let multistream: DBStream["multistream"];
+
+    multistream = streamTargets;
+    multistream.targets.push(payload);
```
victorges: this should work without mutating objects or intermediate duplicate vars
```suggestion
    let multistream: DBStream["multistream"] = { 
      targets: [...(stream.multistream?.targets ?? []), ...payload]
    }
```

victorges: This can get a null reference if the stream has no multistream config (fixed on suggestion above)

packages/api/src/controllers/stream.ts
```diff
@@ -1327,6 +1327,83 @@ app.patch(
...
+    };
+
+    await db.stream.update(stream.id, patch);
+
+    res.status(204);
+    res.end();
+  }
+);
+
+app.delete("/:id/multistream/:targetId", authorizer({}), async (req, res) => {
```
victorges, line 1372: Shouldn't this be similar to the "multi-add" endpoint, receiving a list of IDs in the payload? And if you do so, it should have a similar REST  naming pattern with `POST /:id/name-of-the-action` (in this case `remove-multistream-target` or sth like that)

packages/api/src/controllers/stream.ts
```diff
@@ -1327,6 +1327,83 @@ app.patch(
   }
 );
 
+app.post(
+  "/:id/multistream",
```
victorges: The path should have a clear name of what **action** is being taken on this API, while "multistream" does not mean much, it's only the name of the field. I'd rename this to either `add-multistream` or even `add-multistream-target`

packages/api/src/controllers/stream.ts
```diff
@@ -1327,6 +1327,83 @@ app.patch(
...
+  if (stream.userId !== req.user.id) {
+    res.status(403);
+    return res.json({ errors: ["stream not found"] });
+  }
+  let streamTargets = stream.multistream;
+
+  let multistream: DBStream["multistream"];
+
+  multistream = streamTargets;
+  multistream.targets = multistream.targets.filter((t) => t.id !== targetId);
```
victorges, line 1391: Same null reference.

packages/api/src/controllers/stream.ts
```diff
@@ -1327,6 +1327,83 @@ app.patch(
...
+  const stream = await db.stream.get(id);
+
+  if (!stream || stream.deleted) {
+    res.status(404);
+    return res.json({ errors: ["stream not found"] });
+  }
+
+  if (stream.userId !== req.user.id) {
+    res.status(403);
+    return res.json({ errors: ["stream not found"] });
```
victorges: Either the error should be a 404 or the error message should be a "forbidden" thing

packages/api/src/schema/api-schema.yaml
```diff
@@ -273,6 +273,44 @@ components:
         sessionId:
           type: string
           description: Session ID of the stream to clip
+    target:
```
victorges, line 276: Is this different from the existing `multistream-target` that is referenced inside here? If so, in what way? Can you give this a clear name explaining why it exists and we're not just using the other definition? 

e.g. `multistream-target-input` would be already much clearer if it is a `multistream-target` to be received on request payload.

packages/api/src/schema/api-schema.yaml
```diff
@@ -395,43 +433,7 @@ components:
                 streamed to
               items:
                 type: object
```
victorges: should remove all the fields when referencing
```suggestion
```

victorges (approved):
There are some minor issues in the code which could have been easily caught on tests. Can you please include some unit tests here?

Other comments are mostly style/optional.

victorges (approved):
Neat

Issue #1985, opened 2023-11-28
Fix frontend API embed


iameli:
Broken right now due to a Next.js upgrade. I may give up entirely on booting the production app and just embed the static version instead.

Muritavo:
Hey there @iameli. A question popped in my mind now.
 
From what I see, you work with Livepeer team correct? If not, feel free to disconsider this question.

Would it be feasible to split the frontend startup from the API? 
From what I could understand from exploring the code, it looks like the API only has the responsibility of starting the NextJS server when the args is required, so this decision could be moved up in the hierarchy and the decision could be made at the code that decides when to start the API.

This way we could have separated MAKE commands for frontend and API.

PR #1989, opened 2023-12-07
Fixing frontend failing to start

Muritavo:
- Restored NextJS version to previous version to fix next.config.js not found
- Using require to import @vercel/analytics/react because of pkg bug https://github.com/vercel/pkg/issues/1291
- Including CJS dependencies into pkg filesystem

The docker images are unable to start the NextJS application. I "fixed" (quotes because I couldn't find the core problem on the related libraries) 2 problems:
- 1º: next.config.js could not be found, even though the file was there and the suggested import to fix it was the same file 🤔.
Captura de Tela 2023-12-06 às 20 48 49

- 2º: After fixing the first problem, a second one appeared. It couldn't find @vercel/analytics.

Both problems could only be reproduced when packaging the using the "pkg" command. I think both are related to [this unclosed bug](https://github.com/vercel/pkg/issues/1291) on pkg repository. Maybe it's something related to their virtual file system. More research needed.

- 1º: Downgrading the NextJS back to the previous lock file version seems to resolve the issue.
- 2º: I changed the "import" to "@vercel/analytics/react" to a "require". This makes it load the .cjs extension module. Because of how "pkg" decides which dependencies files to include, I had to include on package.json an explicit indication for it to copy the files from @vercel/analytics node_modules folder. 

Since I don't have the necessary knowledge to test docker images, I tested it manually only.
- Firstly I assured that the error was occurring locally with the latest version
- Found the culprit commit
- Downgraded NextJS version (fixing the 1º problem)
- Tested it locally after each command and found that the pkg command was related to the @vercel import error
- Changed the import to require and tested it again, opening the NextJS application (fixing the 2º problem)
- Only tested the NextJS application  and login. The changes shouldn't require testing other features.

Probably these:
#1985 

**Checklist**

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -182,7 +182,8 @@
       "../../node_modules/dom-helpers/**/*",
       "../../node_modules/react-icons/**/*",
       "../../node_modules/next-mdx-remote/**/*",
-      "../../node_modules/livekit-server-sdk/**/*"
+      "../../node_modules/livekit-server-sdk/**/*",
+      "../../node_modules/@vercel/analytics/**/*"
     ]
   },
   "jest": {
--- a/packages/www/pages/_app.tsx
+++ b/packages/www/pages/_app.tsx
@@ -25,7 +25,9 @@ import {
 import "../css/hubspot.scss";
 import { isStaging } from "lib/utils";
 import { getEndpoint } from "../hooks/use-api";
-import { Analytics } from "@vercel/analytics/react";
+
+const { Analytics } =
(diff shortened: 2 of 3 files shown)


Muritavo:
This pull request have been ignored and I'm closing it since the culprit commit has been reverted. :(

adamsoffer:
@Muritavo Thanks so much for this PR. I'm really sorry the Studio team never got eyes on it. We want to encourage OSS contributors to get involved in the codebase and will make sure future PRs get addressed within a week tops.

PR #1991, opened 2023-12-08
access-control: admin public key access

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -137,7 +137,9 @@ app.post(
 
     if (
       playbackPolicyType !== "public" &&
-      req.body.pub === req.config.accessControlAdminPubkey
+      req.body.pub === req.config.accessControlAdminPubkey &&
+      req.body.pub !== "" &&
+      req.body.pub
     ) {
       res.status(204);
       return res.end();
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -135,6 +135,13 @@ app.post(
       }
     }
 
+    if (req.body.pub && playbackPolicyType !== "public") {
+      if (req.body.pub === req.config.accessControlAdminPublicKey) {
+        res.status(204);
+        return res.end();
+      }
+    }
+
     switch (playbackPolicyType) {
       case "public":
         res.status(204);
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -135,11 +135,12 @@ app.post(
       }
     }
 
-    if (req.body.pub && playbackPolicyType !== "public") {
-      if (req.body.pub === req.config.accessControlAdminPublicKey) {
-        res.status(204);
-        return res.end();
-      }
+    if (
+      playbackPolicyType !== "public" &&
+      req.body.pub === req.config.accessControlAdminPublicKey
+    ) {
+      res.status(204);
+      return res.end();
     }
 
     switch (playbackPolicyType) {
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -137,7 +137,7 @@ app.post(
 
     if (
       playbackPolicyType !== "public" &&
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -135,6 +135,13 @@ app.post(
       }
     }
 
+    if (req.body.pub && playbackPolicyType !== "public") {
+      if (req.body.pub === req.config.accessControlAdminPublicKey) {
+        res.status(204);
+        return res.end();
+      }
+    }
```
victorges: could have it in a single check
```suggestion
    if (playbackPolicyType !== "public" && req.body.pub === req.config.accessControlAdminPublicKey) {
        res.status(204);
        return res.end();
    }
```

packages/api/src/controllers/access-control.ts
```diff
@@ -135,6 +135,13 @@ app.post(
       }
     }
 
+    if (req.body.pub && playbackPolicyType !== "public") {
+      if (req.body.pub === req.config.accessControlAdminPublicKey) {
```
victorges: Curious on what's the frontend process to get catalyst to send a random string on the `pub` field? I thought this field was a public key used to sign the JWT, which it validates on its side. Is that what's happening?

victorges (approved):
Neat

packages/api/src/controllers/access-control.ts
gioelecerati: Not sure what you mean, but Catalyst is getting the signed JWT from the frontend, decoding it, extracting the public key (after checking the claims) and sending it to studio, which verifies is the incoming recovered pubkey is owned by the actual user of the streams.

So, as far as I receive the admin pubkey here, we know that it's coming from a signed JWT

thomshutt (approved):

thomshutt (approved):

thomshutt:
LGTM but tests are failing

packages/api/src/controllers/signing-key.ts
```diff
@@ -241,4 +242,48 @@ signingKeyApp.patch(
   }
 );
 
+signingKeyApp.get(
+  "/jwt/:playbackId",
+  authorizer({ anyAdmin: true, noApiToken: true }),
```
victorges: You should just use `admin:` (not `anyAdmin:`) instead which is the general standard on these dashboard only apis (like /user APIs I think). Don't remember what's the difference with `noApiToken` tho 🤔 
```suggestion
  authorizer({ admin: true }),
```

packages/api/src/controllers/signing-key.ts
```diff
@@ -241,4 +242,48 @@ signingKeyApp.patch(
...
+      iat: Math.floor(issuedAtSec),
+    };
+
+    const options: SignOptions = {
+      algorithm: "ES256",
+    };
+
+    const token = sign(payload, adminKey, options);
+
+    return res.send(token);
```
victorges: might wanna wrap in an object just in case this grows
```suggestion
    return res.send({ token });
```

packages/www/pages/app/stream/[id].tsx
```diff
@@ -257,6 +259,10 @@ const ID = () => {
         throw new Error(info.errors.toString());
       }
       setStream(info.stream);
+      if (stream.playbackPolicy?.type != "public" && info.stream.playbackId) {
```
Comment (victorges) [inline packages/www/pages/app/stream/[id].tsx]: Can we tell if the user is an admin here? Would be nice to avoid this logic if it's not an admin

victorges (approved):

packages/www/pages/app/stream/[id].tsx
Comment (gioelecerati) [reply packages/www/pages/app/stream/[id].tsx]: Isn't this the admin UI anyway? Non admin users can never access this

Comment (victorges) [reply packages/www/pages/app/stream/[id].tsx]: I don't think so. It is just the old dashboard, and it shows additional information if the user is admin (so maybe we can tell that here somehow 🤔 )

Comment (gioelecerati) [reply packages/www/pages/app/stream/[id].tsx]: Old dashboard? :open_mouth: This is new to me lol - added the admin check!

PR #1992, opened 2023-12-08
access-control: better logs & dashboard update

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -138,9 +138,14 @@ app.post(
     if (
       playbackPolicyType !== "public" &&
       req.body.pub === req.config.accessControlAdminPubkey &&
+      req.config.accessControlAdminPubkey !== "" &&
+      req.config.accessControlAdminPubkey &&
       req.body.pub !== "" &&
       req.body.pub
     ) {
+      console.log(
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -138,9 +138,14 @@ app.post(
     if (
       playbackPolicyType !== "public" &&
       req.body.pub === req.config.accessControlAdminPubkey &&
+      req.config.accessControlAdminPubkey !== "" &&
+      req.config.accessControlAdminPubkey &&
```
victorges, line 142: This check already contains the above since `""` is falsy
```
$ node
Welcome to Node.js v21.4.0.
> if ("") {
... console.log("true")
... } else {
... console.log("false")
... }
false
```

packages/api/src/controllers/access-control.ts
```diff
@@ -138,9 +138,14 @@ app.post(
...
       req.body.pub !== "" &&
       req.body.pub
     ) {
+      console.log(
+        `access-control: gate: allowing admin access for contentId=${content.id} playbackId=${playbackId}, user=${user.id} is admin, pubKey=${req.config.accessControlAdminPubkey} is adminPubKey`
```
victorges, line 147: Hmm, this part is weird: `user=${user.id} is admin`

Shouldn't you actually log `user.admin` here? I don't think there's a guarantee on `user` being an admin here (in fact its probably not, since it's the owner of the video, not the caller)

victorges (approved):

PR #1993, opened 2023-12-12
New API to switch stream to CDN

mjh1:


packages/api/src/controllers/stream.ts
```diff
@@ -1015,6 +1015,22 @@ app.post("/:id/heartbeat", authorizer({ anyAdmin: true }), async (req, res) => {
   res.status(204).end();
 });
 
+app.put(
```
iameli, line 1018: I like when we're a bit more generic about this stuff — would it be possible to add it as a field to the PATCH handler so we don't need a new endpoint entirely? We can still check if the user is an admin if that field is present in the payload.

packages/api/src/schema/db-schema.yaml
```diff
@@ -715,6 +715,9 @@ components:
         deleted:
           type: boolean
           description: Set to true when stream deleted
+        routeToCDN:
+          type: boolean
+          description: If stream should be routed through CDN
```
pwilczynskiclearcode, line 720: Currently it's 0.0-100.0 float number in catalyst-api https://github.com/livepeer/catalyst-api/blob/main/main.go#L103 .
We might want it in here instead of boolean

mjh1: Yeah makes sense 👍

packages/api/src/controllers/stream.ts
mjh1: Yeah that's fair enough, I was considering that way so I'll go for it 👍

mjh1:
Not useful enough to continue with for now

PR #1994, opened 2023-12-14
Add quality param for livestreams

leszko:


emranemran (approved):

packages/api/src/schema/api-schema.yaml
```diff
@@ -53,6 +53,10 @@ components:
         fpsDen:
           type: integer
           minimum: 1
+        quality:
```
victorges, line 56: I wonder if we can just merge the `ffmpeg-profile` and `profile` now?

victorges (approved):
LGTM

PR #1997, opened 2023-12-19
db: cache objects with ttl

gioelecerati:

Create a cache to store DB objects and save us from spikes of queries

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -23,6 +23,7 @@ import { DBStream } from "../store/stream-table";
 import { getViewers } from "./usage";
 import { HACKER_DISABLE_CUTOFF_DATE } from "./utils/notification";
 import { isFreeTierUser } from "./helpers";
+import { cacheGetOrSet } from "../store/cache";
 
 const WEBHOOK_TIMEOUT = 30 * 1000;
 const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 5;
@@ -102,9 +103,13 @@ app.post(
   validatePost("access-control-gate-payload"),
   async (req, res) => {
     const playbackId = req.body.stream.replace(/^\w+\+/, "");
-    const content =
-      (await db.stream.getByPlaybackId(playbackId)) ||
-      (await db.asset.getByPlaybackId(playbackId));
+
+    let content = await cacheGetOrSet(`acl-content-${playbackId}`, async () => {
+      return (
+        (await db.stream.getByPlaybackId(playbackId)) ||
+        (await db.asset.getByPlaybackId(playbackId))
+      );
+    });
 
     res.set("Cache-Control", "max-age=120,stale-while-revalidate=600");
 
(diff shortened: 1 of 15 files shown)


packages/api/src/store/cache.ts
```diff
@@ -0,0 +1,23 @@
+import NodeCache from "node-cache";
+
+export const cache = new NodeCache({ stdTTL: 60, maxKeys: 1000 });
```
pwilczynskiclearcode: Should we document it somewhere that changes to user access may take 1 minute to have effect?

packages/api/src/middleware/auth.ts
```diff
@@ -118,10 +119,13 @@ function authenticator(): RequestHandler {
       throw new ForbiddenError(`user is suspended`);
     }
 
-    req.user = user;
+    // make copies in case someone mutates these
+    req.token = { ...tokenObject };
```
gioelecerati: This is causing several tests to fail with `access disallowed to api keys`

packages/api/src/middleware/auth.ts
```diff
@@ -108,7 +108,8 @@ function authenticator(): RequestHandler {
       );
     }
 
-    user = await db.user.get(userId);
+    user = await db.user.get(userId, { cache: true });
```
gioelecerati: We got a problem here - tests are failing when checking if playback is disallowed to suspended users because the cached user is not suspended

gioelecerati: Also, some other tests are failing due to the User being unverified. We are relying too much to users not being cached

packages/api/src/middleware/auth.ts
victorges: Hmmm

packages/api/src/middleware/auth.ts
victorges: Yeah we need to update the tests either not to cache at all or to wait until it's expired before checking

victorges: Oh a better option should be a test function to clear the cache

packages/api/src/middleware/auth.ts
victorges: Understood it! problem was when we didnt have an API key here, we were still setting the `token` var to an empty object. Fixed, good catch.

packages/api/src/middleware/auth.ts
victorges: done

gioelecerati:
LGTM but I cannot approve it myself because I opened this PR :D

packages/api/src/controllers/access-control.ts
```diff
@@ -136,7 +136,7 @@ app.post(
     const playbackPolicyType = content.playbackPolicy?.type ?? "public";
 
     if (user.createdAt < HACKER_DISABLE_CUTOFF_DATE) {
-      let limitReached = await freeTierLimitReached(content, user, req);
+      let limitReached = freeTierLimitReached(content, user, req);
```
gioelecerati, line 139: Oh! right, wasn't needed!

gioelecerati:
Looking good :heart:

packages/api/src/store/cache.ts
victorges: I think a little read after write inconsistency is somewhat expected in APIs like this (and we also already had _some_ of it due to database read replicas). Since we are trying to only cache objects that are mostly immutable right now (users, object stores, api&access keys, etc) I'd rather keep that behavior implicit for now. 

Wouldn't oppose to document it if anyone has a strong opinion about it though.

victorges (approved):
LGTM 🥸

PR #1998, opened 2023-12-19
access-control: allow to suspend keys

gioelecerati:

Allow to suspend keys

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -1,5 +1,5 @@
 import signingKeyApp from "./signing-key";
-import { validatePost } from "../middleware";
+import { authorizer, validatePost } from "../middleware";
 import { Request, Router } from "express";
 import _ from "lodash";
 import { db } from "../store";
@@ -23,6 +23,7 @@ import { DBStream } from "../store/stream-table";
 import { getViewers } from "./usage";
 import { HACKER_DISABLE_CUTOFF_DATE } from "./utils/notification";
 import { isFreeTierUser } from "./helpers";
+import { v4 as uuid } from "uuid";
 
 const WEBHOOK_TIMEOUT = 30 * 1000;
 const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 5;
@@ -150,6 +151,16 @@ app.post(
         res.status(204);
         return res.end();
       case "jwt":
+        if (req.body.jwt) {
+          const [res] = await db.accessControlKey.find({
+            key: req.body.jwt,
+          });
+          if (res[0]) {
+            if (res[0].suspended) {
+              throw new ForbiddenError("Jwt is suspended");
+            }
+          }
+        }
         if (!req.body.pub) {
           console.log(`
             access-control: gate: no pub provided for playbackId=${playbackId}, disallowing playback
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -250,6 +271,30 @@ app.post(
...
+  });
+
+  if (!keys[0]) {
+    let doc = await db.accessControlKey.create({
+      id: uuid(),
+      key,
+      suspended: true,
+    });
+
+    res.status(200).json({ id: doc.id, suspended: true });
```
mjh1, line 289: do we need a return here or does `res.status` handle that

packages/api/src/controllers/access-control.ts
```diff
@@ -250,6 +271,30 @@ app.post(
   }
 );
 
+app.post("/suspend/:key", authorizer({ anyAdmin: true }), async (req, res) => {
```
mjh1, line 274: Can we add a get handler to check whether something is suspended and/or a handler for listing all entries

gioelecerati: Added!

packages/api/src/controllers/access-control.ts
gioelecerati: It's already handled by .status().json()

packages/api/src/controllers/access-control.ts
```diff
@@ -150,6 +151,16 @@ app.post(
         res.status(204);
         return res.end();
       case "jwt":
+        if (req.body.jwt) {
```
victorges, line 154: If the type is jwt, isn't this payload field obligatory? Looking like a redundant check, or maybe we want an inverted check right in the beginning to return an error/403 right away if a JWT is not provided

packages/api/src/controllers/access-control.ts
```diff
@@ -210,6 +221,16 @@ app.post(
             "Content is gated and requires an access key"
           );
         }
+
+        const [keys] = await db.accessControlKey.find({
```
victorges, line 225: I don't like how we are adding 2 more IOs in the API that is already slow due to db access. Do we need to fetch the access key here? Didn't we already fetch it somewhere in this handler to validate the jwt?

packages/api/src/controllers/access-control.ts
```diff
@@ -150,6 +151,16 @@ app.post(
         res.status(204);
         return res.end();
       case "jwt":
+        if (req.body.jwt) {
+          const [res] = await db.accessControlKey.find({
```
victorges, line 155: Another thing... This logic here basically means that we can't cache this API call in catalyst API. It would need to be a unique cache key for each and every user that plays the stream, which sounds like a lot.

Could we implement this blacklist on catalyst API side instead? It could just get a list of jwts as a config and we don't add any more IO in this flow. It's probably fine for fishtank if we update that config ourselves for now.

packages/api/src/controllers/access-control.ts
```diff
@@ -250,6 +271,45 @@ app.post(
...
+
+  const [keys] = await db.accessControlKey.find({
+    key,
+  });
+
+  if (!keys[0]) {
+    let id = uuid();
+    let doc = await db.accessControlKey.create({
+      id,
+      key,
```
victorges, line 285: Is this actually a suspended jwt?

packages/api/src/schema/db-schema.yaml
```diff
@@ -1150,6 +1153,20 @@ components:
         deleted:
           type: boolean
           default: false
+    access-control-key:
```
victorges, line 1156: Nit: I think this name is misleading/unclear. WDYT of naming it explicitly like `access-control-blocked-jwt`?

victorges (commented):
Not approving because I think this might increase the load on this API which is already causing problems. Let's discuss on the threads

gioelecerati:
closing in favor of https://github.com/livepeer/catalyst-api/pull/1048

PR #2004, opened 2023-12-29
www: Implement JWT refreshing logic

victorges:

This is to stop generating JWTs to access the dashboard that have no expiration time. This
causes a security risk if these tokens ever leak, since there's no way to revoke them and they'd
be valid forever.

The new design involves:
- Adding an `exp` to all the JWTs ("access tokens") that we sign, making sure they can only be used temporarily (24h)
- Creating a separate "refresh token" abstraction that is saved on the DB. The client only deals with a vanity ID, just like for API keys. These last for 30 days and can be used to generate new access tokens.
- The refresh tokens themselves are also refreshed when they are close to their expiration (currently 30 days). If a user doesn't accept the dashboard during the refresh token refresh period (20% of ttl, or 6 days), they will need to login again.
- A single refresh token can only be used to generate 1 access token every "half life" (so one every 12h). This is to prevent reuse of this token, which will be immediately revoked if reused.
  - This is a pretty simple security measure. We could create many other restrictions on top of these now that we have them on the DB, like enforcing only the same IP or device to use them.
- There is also some API & logic to migrate from the old never-expiring JWTs to the new scheme.
- Never-expiring JWTs won't be accepted anymore after a cut-off date, date when we should kill the dead migration code.

- Create the new refresh and migration APIs
- Create client code to call these APIs to refresh them tokens
- Write some API tests

- `yarn test`
- todo: e2e, maybe in the box, maybe in staging

The never-expiring JWTs.

**Checklist**

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -32,7 +32,7 @@
     "go-livepeer": "run-p \"go-livepeer:**\" --",
     "go-livepeer:broadcaster": "bin/livepeer -broadcaster -datadir ./bin/broadcaster -orchAddr 127.0.0.1:3086 -rtmpAddr 0.0.0.0:3035 -httpAddr :3085 -cliAddr :3075 -v 6 -authWebhookUrl http://127.0.0.1:3004/api/stream/hook -orchWebhookUrl http://127.0.0.1:3004/api/orchestrator",
     "go-livepeer:orchestrator": "bin/livepeer -orchestrator -datadir ./bin/orchestrator -transcoder -serviceAddr 127.0.0.1:3086 -cliAddr :3076 -v 6",
-    "test": "POSTGRES_CONNECT_TIMEOUT=120000 jest -i --silent \"${PWD}/src.+user.test.ts\"",
+    "test": "POSTGRES_CONNECT_TIMEOUT=120000 jest -i --silent \"${PWD}/src\"",
     "test:dev": "jest \"${PWD}/src\" -i --silent --watch",
     "test:build": "parcel build --no-autoinstall --no-minify --bundle-node-modules -t browser --out-dir ../dist-worker ../src/worker.js",
     "coverage": "yarn run test --coverage",
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -32,7 +32,7 @@
     "go-livepeer": "run-p \"go-livepeer:**\" --",
     "go-livepeer:broadcaster": "bin/livepeer -broadcaster -datadir ./bin/broadcaster -orchAddr 127.0.0.1:3086 -rtmpAddr 0.0.0.0:3035 -httpAddr :3085 -cliAddr :3075 -v 6 -authWebhookUrl http://127.0.0.1:3004/api/stream/hook -orchWebhookUrl http://127.0.0.1:3004/api/orchestrator",
     "go-livepeer:orchestrator": "bin/livepeer -orchestrator -datadir ./bin/orchestrator -transcoder -serviceAddr 127.0.0.1:3086 -cliAddr :3076 -v 6",
-    "test": "POSTGRES_CONNECT_TIMEOUT=120000 jest -i --silent \"${PWD}/src\"",
+    "test": "POSTGRES_CONNECT_TIMEOUT=120000 jest -i --silent \"${PWD}/src.+user.test.ts\"",
     "test:dev": "jest \"${PWD}/src\" -i --silent --watch",
     "test:build": "parcel build --no-autoinstall --no-minify --bundle-node-modules -t browser --out-dir ../dist-worker ../src/worker.js",
     "coverage": "yarn run test --coverage",
--- a/packages/api/src/controllers/user.test.ts
+++ b/packages/api/src/controllers/user.test.ts
@@ -7,6 +7,10 @@ import { TestClient, clearDatabase, setupUsers } from "../test-helpers";
 import serverPromise, { TestServer } from "../test-server";
 import sql from "sql-template-strings";
 import { NEVER_EXPIRING_JWT_CUTOFF_DATE } from "../middleware";
+import {
+  REFRESH_TOKEN_MIN_REUSE_DELAY_RATIO,
+  REFRESH_TOKEN_REFRESH_THRESHOLD,
+} from "./user";
 
 let server: TestServer;
 let mockUser: User;
@@ -405,6 +409,7 @@ describe("controllers/user", () => {
 
     beforeEach(async () => {
       client = new TestClient({ server });
+      // JWT TTL args are in seconds but we use millis here, so multiply by 1000
       ACCESS_TOKEN_TTL = server.jwtAccessTokenTtl * 1000;
       REFRESH_TOKEN_TTL = server.jwtRefreshTokenTtl * 1000;
 
@@ -489,6 +494,26 @@ describe("controllers/user", () => {
         const refreshTokenObj = await db.jwtRefreshToken.get(refreshToken);
         expect(refreshTokenObj.lastSeen).toBeLessThanOrEqual(Date.now());
         expect(refreshTokenObj.lastSeen).toBeGreaterThanOrEqual(beforeRefresh);
+        expect(refreshTokenObj.revoked).toBeFalsy();
+      });
+
+      it("should NOT allow using refresh token too often", async () => {
+        let res = await client.post("/user/token/refresh", { refreshToken });
+        expect(res.status).toBe(201);
+
+        mockTimeDelay(
+          ACCESS_TOKEN_TTL * REFRESH_TOKEN_MIN_REUSE_DELAY_RATIO - 60 * 1000
+        );
+
+        res = await client.post("/user/token/refresh", { refreshToken });
+        expect(res.status).toBe(401);
+        const { errors } = await res.json();
+        expect(errors[0]).toBe(
+          "refresh token has already been used too recently"
+        );
+
+        const refreshTokenObj = await db.jwtRefreshToken.get(refreshToken);
+        expect(refreshTokenObj.revoked).toBe(true);
       });
 
       it("should not accept access token after expiration", async () => {
@@ -515,7 +540,9 @@ describe("controllers/user", () => {
       });
 
       it("should generate a new refresh token when close to expiration", async () => {
-        mockTimeDelay(0.9 * REFRESH_TOKEN_TTL);
+        mockTimeDelay(
+          REFRESH_TOKEN_TTL * (1 - REFRESH_TOKEN_REFRESH_THRESHOLD) + 60 * 1000
+        );
 
         const res = await client.post("/user/token/refresh", {
           refreshToken,
@@ -530,14 +557,6 @@ describe("controllers/user", () => {
         const oldRefreshTokenObj = await db.jwtRefreshToken.get(refreshToken);
         expect(oldRefreshTokenObj.revoked).toBe(true);
       });
-    });
-
-    // This is an evil reminder to delete all the code for migrating from the
-    // never-expiring JWTs. Once this test starts failing, we should delete the
-    // section below from the tests and all the code paths that mention
-    // "never-expiring JWTs".
-    it("should delete JWT migration dead code after it is done", async () => {
-      expect(Date.now()).toBeLessThanOrEqual(NEVER_EXPIRING_JWT_CUTOFF_DATE);
     });
 
     describe("never-expiring JWTs migration", () => {
--- a/packages/api/src/controllers/user.test.ts
+++ b/packages/api/src/controllers/user.test.ts
@@ -1,10 +1,12 @@
 import { v4 as uuid } from "uuid";
+import jwt from "jsonwebtoken";
 
 import { User } from "../schema/types";
 import db from "../store/db";
 import { TestClient, clearDatabase, setupUsers } from "../test-helpers";
 import serverPromise, { TestServer } from "../test-server";
 import sql from "sql-template-strings";
+import { NEVER_EXPIRING_JWT_CUTOFF_DATE } from "../middleware";
 
 let server: TestServer;
 let mockUser: User;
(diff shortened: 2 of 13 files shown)


packages/api/package.json
```diff
@@ -32,7 +32,7 @@
     "go-livepeer": "run-p \"go-livepeer:**\" --",
     "go-livepeer:broadcaster": "bin/livepeer -broadcaster -datadir ./bin/broadcaster -orchAddr 127.0.0.1:3086 -rtmpAddr 0.0.0.0:3035 -httpAddr :3085 -cliAddr :3075 -v 6 -authWebhookUrl http://127.0.0.1:3004/api/stream/hook -orchWebhookUrl http://127.0.0.1:3004/api/orchestrator",
     "go-livepeer:orchestrator": "bin/livepeer -orchestrator -datadir ./bin/orchestrator -transcoder -serviceAddr 127.0.0.1:3086 -cliAddr :3076 -v 6",
-    "test": "POSTGRES_CONNECT_TIMEOUT=120000 jest -i --silent \"${PWD}/src\"",
+    "test": "POSTGRES_CONNECT_TIMEOUT=120000 jest -i --silent \"${PWD}/src.+user.test.ts\"",
```
leszko: Why do you need to add `user.test.ts` explicitly here (and other tests don't need it? 🤔

packages/api/src/controllers/user.test.ts
```diff
@@ -393,6 +396,222 @@ describe("controllers/user", () => {
...
+  describe.only("JWT token refresh", () => {
+    let client: TestClient;
+    let ACCESS_TOKEN_TTL: number;
+    let REFRESH_TOKEN_TTL: number;
+
+    let mockUserObj: User;
+
+    beforeEach(async () => {
+      client = new TestClient({ server });
+      ACCESS_TOKEN_TTL = server.jwtAccessTokenTtl * 1000;
```
leszko, line 413: What is the value of `server.jwtAccessTokenTtl` and why do we multiply by `1000`? Could you add some comments?

packages/api/src/controllers/user.test.ts
```diff
@@ -393,6 +396,222 @@ describe("controllers/user", () => {
...
+        const body = await res.json();
+        expect(body.refreshToken).toBeUndefined();
+
+        await expectValidAccessToken(body.token);
+      });
+
+      it("should generate a new refresh token when close to expiration", async () => {
+        mockTimeDelay(0.9 * REFRESH_TOKEN_TTL);
+
+        const res = await client.post("/user/token/refresh", {
```
leszko, line 547: So, the same endpoint `/user/token/refresh` returns `refreshToken` ONLY AFTER it's almost expired. Is that correct? Why don't we allow always refreshing the `refreshToken`?

packages/api/src/controllers/user.test.ts
```diff
@@ -393,6 +396,222 @@ describe("controllers/user", () => {
...
+        const oldRefreshTokenObj = await db.jwtRefreshToken.get(refreshToken);
+        expect(oldRefreshTokenObj.revoked).toBe(true);
+      });
+    });
+
+    // This is an evil reminder to delete all the code for migrating from the
+    // never-expiring JWTs. Once this test starts failing, we should delete the
+    // section below from the tests and all the code paths that mention
+    // "never-expiring JWTs".
+    it("should delete JWT migration dead code after it is done", async () => {
```
leszko: I suggest removing it and creating a GH Issue (or Linear) with some date. Or some calendar reminder. I don't like it, because if we forget about this code, then suddenly the test will start to fail and maybe it won't be a good time, e.g. , imagine we have a quick fix to deploy to prod and someone starts to see this error.

packages/api/src/controllers/user.test.ts
```diff
@@ -393,6 +396,222 @@ describe("controllers/user", () => {
...
+        );
+
+        mockTime(NEVER_EXPIRING_JWT_CUTOFF_DATE - 24 * 60 * 60 * 1000);
+      });
+
+      it("should still allow accessing API", async () => {
+        await expectValidAccessToken(neverExpiringJwt);
+      });
+
+      it("should NOT allow accessing API after cut-off date", async () => {
```
leszko, line 597: Do we plan to communicate anything to the users that their JWT may stop working? I wonder if we have some users using never-expiring tokens and if it can cause any issues to them.

packages/api/src/controllers/user.ts
```diff
@@ -36,13 +37,16 @@ import {
...
 import sql from "sql-template-strings";
+import { CliArgs } from "../parse-cli";
 
 const adminOnlyFields = ["verifiedAt", "planChangedAt"];
 
 const salesEmail = "sales@livepeer.org";
 const infraEmail = "infraservice@livepeer.org";
 const freePlan = "prod_O9XuIjn7EqYRVW";
 
+const REFRESH_TOKEN_REFRESH_THRESHOLD = 0.2;
```
leszko: So, we allow refreshing `refreshToken` after 80% of it's TTL passed. Is that correct?

packages/api/src/controllers/user.ts
```diff
@@ -636,9 +653,105 @@ app.post("/token", validatePost("user"), async (req, res) => {
...
+      refreshTokenObj.expiresAt < now ||
+      refreshTokenObj.revoked
+    ) {
+      console.log(
+        `Refresh attempt with invalid refresh token=${refreshToken} expiresAt=${refreshTokenObj?.expiresAt} revoked=${refreshTokenObj?.revoked}`
+      );
+      return res.status(401).json({ errors: ["invalid refresh token"] });
+    }
+
+    if (now - refreshTokenObj.lastSeen < req.config.jwtAccessTokenTtl / 2) {
```
leszko: What's the use of this line? Why do we revoke token? Do we maybe have a test scenario for this?

packages/api/src/controllers/user.ts
```diff
@@ -636,9 +653,105 @@ app.post("/token", validatePost("user"), async (req, res) => {
...
+      refreshTokenObj.expiresAt - now <
+      fullTokenTtl * REFRESH_TOKEN_REFRESH_THRESHOLD
+    ) {
+      // issue a new token and revoke the old one if it's close to expiring
+      await db.jwtRefreshToken.update(refreshToken, {
+        lastSeen: now,
+        revoked: true,
+      });
+
+      newRefreshToken = uuid();
```
leszko, line 721: Shouldn't we first create a new refresh token and then revoke the old one? 

Otherwise, what happens, if we revoked the old one, but failed to create a new one?

packages/api/src/schema/db-schema.yaml
```diff
@@ -464,6 +464,45 @@ components:
...
+      additionalProperties: false
+      properties:
+        id:
+          type: string
+          example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+        userId:
+          type: string
+          index: true
+          example: 66E2161C-7670-4D05-B71D-DA2D6979556F
+        lastSeen:
```
leszko, line 479: What is the use of `lastSeen` field?

packages/www/hooks/use-api/index.tsx
```diff
@@ -61,12 +61,26 @@ const makeContext = (
...
       // todo: not every endpoint will return JSON
       const body = await res.json();
       // todo: this can go away once we standardize on body.errors
       if (!Array.isArray(body.errors) && typeof body.error === "string") {
         body.errors = [body.error];
       }
+
+      const tokenExpired =
+        res.status === 401 && body.errors?.[0] === "access token expired";
+      if (tokenExpired && state.refreshToken && !isRetry) {
```
leszko: Why do we need to check `isRetry` here?

packages/api/package.json
victorges: Oh no, forgot to revert this part 😅

packages/api/src/controllers/user.test.ts
victorges: Added a comment explaining the 1000. Their default value (in `parse-cli.ts`) doesn't matter for these tests though, as we just grab these values not to have to repeat any constants here.

packages/api/src/controllers/user.test.ts
victorges: Exactly! The idea is that the access tokens are refreshed more often and exist entirely ~offchain~ outside the DB, relying on JWT signatures instead. Generating a new refresh token every time the access token was refreshed would mean a new DB entry for every user every day [1] they accessed the dashboard, which seems too much. There are some platforms that do do that, refreshing the refresh token on every access token refresh, but I preferred to keep it a bit cheaper for now and maybe we build up on the security checks later.

[1] And that's just the current default TTL. Imagine if we go lower like access keys lasting for only 1h or sth like that

packages/api/src/controllers/user.test.ts
victorges: Oh no I thought no one would read the tests! hahaha
Removed 🌚

packages/api/src/controllers/user.test.ts
victorges: No, JWTs are an internal thing only used by our dashboard. We've only used them internally to fake API calls from the dashboard (🙋🏻‍♂️), but users should not be relying on this in any way, which is also completely undocumented (the fact that the dashboard authorizes differently). In these cases I'd rather break anyone that reverse engineered and was exploiting this.

packages/api/src/controllers/user.ts
victorges: Right, which for the default 30 days means 6 days til expiration

packages/api/src/controllers/user.ts
victorges: Yeah I moved these around a bit... I thought it would be worse if we created a new refresh token and then failed to revoke the old one, then there's 2 refresh tokens out there for the same session. I'd rather have the user logout and try again in this rare scenario (and If we think about it, if we have any failure here the user will get a status error and won't get even an access token, let alone a refresh token)

> Otherwise, what happens, if we revoked the old one, but failed to create a new one?

In practice, any error here will cause a status code error and the user won't get anything (not even a new access key). Then their dashboard is likely going to log out, or maybe the next time they try to make an API call, it will:
- 401 the same way, since their access token is expired
- try to refresh the token, which will 401 since their refresh token is now revoked
- then the dashboard logs out and they can log in again

Even though the log out is a bit cumbersome, this is a pretty rare scenario (1 transaction succeeding and next one failing), so I'd rather optimize for not having more than 1 refresh token active for the same session.

packages/api/src/schema/db-schema.yaml
victorges: It is currently used to make sure we don't use the same refresh token to generate multiple access tokens. A refresh token is supposed to stay within the browser, and used only when the corresponding access key expires. By updating the `lastSeen` we make sure that we only generate enough access keys for that single session. If the refresh token leaks and a malicious user tries to generate access tokens for the victim account, they'd eventually get revoked automatically (we can be smarter about this later)

packages/api/src/controllers/user.ts
victorges: This is the only use for the `lastSeen` field which I explained here: https://github.com/livepeer/studio/pull/2004#discussion_r1440700535

The logic here is that a single refresh token should not be used to generate multiple access tokens. We could limit this even further to only allow 1 to be generated every `jwtAccessTokenTtl`, but I added a `/2` in there just to give it a bit of leeway and avoid random clock skew errors (so one could generate 1 access tokens every 12h, lasting 24h each).

I'll add a comment here and maybe restructure the expression to be clearer. And good catch, just added a test as well! (there was even a bug, `jwtAccessTokenTtl` needs a `* 1000` here too)

packages/www/hooks/use-api/index.tsx
victorges: Because we retry the API call that failed on L80, and we don't want to have an infinite loop here trying to refresh the token in case something went wrong

packages/api/src/controllers/user.test.ts
leszko: Ok, makes sense. Thanks for the explanation.

packages/api/src/controllers/user.test.ts
leszko: Ok, makes sense. Thanks for the explanation 🙏

packages/api/src/controllers/user.ts
leszko: Maybe stupid question, I'm not a frontend person, but when you have multiple tabs opened, then don't each need a separate JWT token? What happens if one tab refreshes the token, but we don't allow multiple refreshes?

packages/api/src/schema/db-schema.yaml
leszko: Maybe stupid question, I'm not a frontend person, but when you have multiple tabs opened, then don't each need a separate JWT token? What happens if one tab refreshes the token, but we don't allow multiple refreshes?

leszko (approved):
Looks good 👍 

Added one more inline comment. Plus, did you test it with Livepeer in a box? I tried and got some frontend errors, so maybe better to double-check.

victorges:
> did you test it with Livepeer in a box? I tried and got some frontend errors, so maybe better to double-check.

Wasn't able to as I shifted to some other tasks later this week. I'll only be able to get back to this after my PTO next week, so on 15/01 (or 16 since that's apparently a holiday). Maybe I'll be able to test it tomorrow, but I wouldn't dare to deploy this big change and then leave on PTO xD

But in general agree that this absolutely needs some E2E testing, this is a pretty big change!

packages/api/src/schema/db-schema.yaml
victorges: Both cookies and local storage (where our frontend stores these tokens) are shared between all tabs! If it weren't like that I guess you'd need to login again on every tab you have on a website.

> The main features of localStorage are: Shared between all tabs and windows from the same origin. The data does not expire. It remains after the browser restart and even OS reboot.

victorges: So what happens is that all tabs will share the same access (jwt) and refresh (id) tokens, and when one of them gets a 401 from the API it will use the refresh token and update the access token on the local storage. Then all other tabs will get this new access token already and won't get 401s themselves.

victorges:
Rebase done.

PR #2006, opened 2024-01-03
Return targetId while creating a multistream target

leszko:
Linear: https://linear.app/livepeer/issue/PS-196/adding-target-id-to-response-of-atomic-creation

### DIFF
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -523,7 +523,7 @@ describe("controllers/stream", () => {
           spec: { name: "target-name", url: "rtmp://test/test" },
         }
       );
-      expect(res2.status).toBe(204);
+      expect(res2.status).toBe(200);
     });
 
     describe("set active and heartbeat", () => {
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -524,6 +524,8 @@ describe("controllers/stream", () => {
         }
       );
       expect(res2.status).toBe(200);
+      const body = await res2.json();
+      expect(body.id).toBeDefined();
     });
 
     describe("set active and heartbeat", () => {
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -115,11 +115,7 @@ async function validateMultistreamTarget(
   return { ...specless, id: created.id };
 }
 
-async function validateMultistreamOpts(
-  userId: string,
-  profiles: Profile[],
-  multistream: MultistreamOptions
-): Promise<MultistreamOptions> {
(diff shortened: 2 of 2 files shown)


packages/api/src/controllers/stream.test.ts
```diff
@@ -523,7 +523,7 @@ describe("controllers/stream", () => {
           spec: { name: "target-name", url: "rtmp://test/test" },
         }
       );
-      expect(res2.status).toBe(204);
+      expect(res2.status).toBe(200);
```
victorges, line 526: Should we check the body here too?

packages/api/src/controllers/stream.ts
```diff
@@ -132,7 +128,15 @@ async function validateMultistreamOpts(
...
   profileNames.add("source");
+  return profileNames;
+}
 
+async function validateMultistreamOpts(
+  userId: string,
+  profiles: Profile[],
+  multistream: MultistreamOptions
+): Promise<MultistreamOptions> {
+  const profileNames = toProfileNames(profiles);
```
victorges, line 139: neat

packages/api/src/controllers/stream.ts
```diff
@@ -1346,8 +1350,14 @@ app.post(
       return res.json({ errors: ["stream not found"] });
     }
 
+    const newTarget = await validateMultistreamTarget(
```
victorges, line 1353: nice catch

packages/api/src/controllers/stream.ts
```diff
@@ -1364,8 +1374,8 @@ app.post(
 
     await triggerCatalystStreamUpdated(req, stream.playbackId);
 
-    res.status(204);
-    res.end();
+    res.status(200);
+    res.json(newTarget);
```
victorges: You should call `db.multistreamTarget.cleanWriteOnlyResponse` [(e.g.)](https://github.com/livepeer/livepeer-com/blob/37ab514c365bf0ccae3f787dc6afb8a13785fe2b/packages/api/src/controllers/multistream.ts#L85) on this object before returning it on the response payload. It will clean up the private URL field (and maybe some other), which is not supposed to go in API responses.

It's theoretically not a huge issue here, since the user just sent the URL on the request, but I think we should keep it consistent (and eventually there could be other secret fields).

victorges (approved):
LGTM, only 1 required change to clean up the response object

packages/api/src/controllers/stream.test.ts
leszko: Good idea, updated!

packages/api/src/controllers/stream.ts
leszko: Updated. Thanks for spotting that!

PR #2013, opened 2024-01-17
api: Create logic for automatically cleaning up dead streams

victorges:

This creates an API and a corresponding cronjob to clean-up stream that are marked active
but haven't been updated in a while. This is an improvement from the lazy approach that
only tried to clean-up the streams once they were accessed via the API, which also had the
undesirable effect of triggering expensive mutations from read-only requests.

- Create the new jobs API
- Create the cronjbo action
- Create a separate DB connection pool for background operations like this (and webhooks, and tasks)
- Remove old lazy logic ✨ 

- [x] Ran the queries on the DB
- [x] `yarn test`
- [ ] Try it on staging

Implements ENG-869

**Checklist**


victorges:
Ready for review! Could you take a look @leszko ?

.github/workflows/cron-streams-active-cleanup.yaml
```diff
@@ -0,0 +1,15 @@
+name: "cron: Active streams cleanup"
```
leszko: I wonder if we shouldn't run it as a kubernetes Cron Jobs. I know that currently we don't have a clear agreement on whether cron jobs should be GH Actions or Kubernetes CronJobs, but the advantages I see for integrating it in our Infra:
- Alerts integration (Grafana Pager Duty)
- (super minor) GH Actions load on the GH workers
- (minor) security and passing admin LP token

Anyway, I leave it up to you.

packages/api/src/controllers/stream.ts
```diff
@@ -229,6 +234,7 @@ function activeCleanupOne(
   }
 
   setImmediate(async () => {
+    await cleanupSemaphore.wait();
```
leszko: The "right" solution would be to have a separate connection pool for the cleanup work. This is a workaround. I wonder how difficult would it be to implement a separate connection pool for the cleanup.

.github/workflows/cron-streams-active-cleanup.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - cron: "*/10 * * * *" # run every 10 minutes
+
+jobs:
+  cron:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clean-up active streams
+        run: |
+          curl --request POST \
+          --url 'https://livepeer.studio/api/stream/job/active-cleanup' \
```
leszko: Will this GH Action fail, if the cleanup fails? I think we would like to have it working this way:
1. The call `https://livepeer.studio/api/stream/job/active-cleanup` would be sync, so the GH Action effectively waits for the streams to get cleaned up
2. If it fails, the whole job should fail, so that we know that something is wrong

.github/workflows/cron-streams-active-cleanup.yaml
```diff
@@ -0,0 +1,15 @@
+name: "cron: Active streams cleanup"
+
+on:
+  schedule:
+    - cron: "*/10 * * * *" # run every 10 minutes
```
leszko: How about running it every 5 min and getting rid of all the opportunistic cleanup? I think this job should be the only place where we do the stream cleanup.

Otherwise, we risk not exposing a lot of bugs + it's super wrong that making a GET call to studio makes a change in the Studio DB.

packages/api/src/util.ts
```diff
@@ -28,21 +28,48 @@ export const sleep = (duration) => {
...
-      if (!timeoutMs) return promise;
-      return Promise.race([promise, sleep(timeoutMs)]);
-    },
-  };
+  const sem = new Semaphore(1);
+  sem.wait(); // semaphore from this function starts acquired
+  return sem;
 };
 
+export class Semaphore {
```
leszko: Did you consider using some library for this instead of implementing the Semaphore?

If we end up having our own Semaphore implementation, then how about writing some unit tests?

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -286,7 +286,8 @@ describe("webhook cannon", () => {
       expect(callCount).toBe(1);
       expect(receivedEvent).toBe("stream.started");
 
-      sem = semaphore();
+      // at this point the semaphore is re-acquired cause we just waited (acquired) above
```
leszko: Could you also add unit test for `/job/active-cleanup`

leszko (commented):
Good work @victorges !

Added some comments, but in general I really like the approach and love that we'll start having this functionality!

.github/workflows/cron-streams-active-cleanup.yaml
victorges: Even though I agree it would be much better to run cronjobs, I'd leave that as a separate change for the future, given we have other (more critical) workflows already running on GitHub actions, mainly usage billing. We should probably bring this up as a tech debt to be prioritized eventually though.

packages/api/src/controllers/stream.ts
victorges: I've implemented the separate connection pool here!
https://github.com/livepeer/studio/pull/2013/commits/ef6df9300d01fc866ef380e41360065e29673f5d

I'd recommend reviewing that commit separately, since I touched several different files (also moved webhooks, tasks and usage logic to the jobs conn pool).

packages/api/src/util.ts
victorges: Not necessary anymore given I'm using a separate conn pool instead :)

But FTR, I remember I _did_ search for libraries back then and surprisingly there weren't any suitable! I guess people don't really care about concurrency in node since it's single threaded 🤷‍♂️

packages/api/src/controllers/webhook.test.ts
```diff
@@ -288,8 +290,22 @@ describe("controllers/webhook", () => {
...
+        for (const id of [webhookResJson.id, webhookResJson2.id]) {
+          const res = await client.get(`/webhook/${id}`);
+          expect(res.status).toBe(200);
+          const { status } = (await res.json()) as Webhook;
+          if (status.lastTriggeredAt >= now) {
+            hooksCalled++;
+          }
+        }
+      }
+      expect(hooksCalled).toBe(2);
```
victorges, line 308: FTR: this was the very cryptic failure I was getting from the tests. What has happening was that, after creating the separate jobs DB pool, the test below (`records webhook logs`) started failing because the webhook created _there_ was begin called twice, one additional time with the event triggered in _this_ test.

The problem was that this test never really waited (or even checked) that the webhooks created here were actually triggered. So the test finished while leaving the background runner calling the webhooks for it. This probably caused some flakiness every now and then as well.

What has happening was that, after switching to the separate DB pool, the very first event handled by the `WebhookCannon` took a little longer since it had to setup a new connection with Postgres (that hadn't been already initialized from the usual test setup). This small delay was already enough for the next test to start and create another webhook, which then got called for the same event. I could reproduce this by making a dummy query on the `jobsDb` in this test, which then caused the failure to stop cause the pool was then already initialized.

Fixed this by correcting this test logic tho, which was not even checking that the webhooks were being called. Explaining here only cause it seems so unrelated to the changes here, but they actually were.

.github/workflows/cron-streams-active-cleanup.yaml
victorges: Good call. I've ran [some tests with the GH action](https://github.com/livepeer/studio/actions/workflows/cron-streams-active-cleanup.yaml) and we can get the action to fail simply by adding. `--fail` to the curl command, so it exits with a bad code if the server responds a bad status as well. 

The clean-up implementation itself is already synchronous as well, so that part should be working once the API actually gets deployed.

.github/workflows/cron-streams-active-cleanup.yaml
victorges: Super scary, but I also would love to remove the opportunistic logic. Implemented here! https://github.com/livepeer/studio/pull/2013/commits/f70dfe3b4ff708d21eb97a596db35e45ba03013a

Notice that with a 5 minute frequency, the worst case delay for a stream to be activecleanedup will be 10 minutes. I think making the github action more frequent would not be that good though given limited runners etc, so we can consider that once we move to an actual cron in our infra. WDYT?

packages/api/src/controllers/stream.ts
```diff
@@ -293,65 +294,100 @@ export function getFLVPlaybackUrl(ingest: string, stream: DBStream) {
...
     !isNaN(stream.lastSeen) &&
     Date.now() - stream.lastSeen > ACTIVE_TIMEOUT
   );
 }
 
-function activeCleanupOne(
+/**
+ * Creates an updated stream object with a fixed `isActive` field in case it is lost and needs clean up.
+ */
+function withFixedIsActive(stream: DBStream) {
```
leszko: Does it mean that we can have a stream object without `isActive` field? 

I wonder if we should fix such streams in the clean-up logic. I think the sematic of not having `isActive` field would be that the stream is not active. So, probably no need to clean it up.

.github/workflows/cron-streams-active-cleanup.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
           - name: "staging"
             hostname: "livepeer.monster"
             api-token-secret: LP_STAGING_API_ADMIN_TOKEN
+          - name: production
```
leszko, line 15: Don't we want to run stream clean-up as Kubernetes CronJobs rather than GH Actions? I think it should be simpler to monitor with our existing infra.

leszko (approved):
Added 2 comments. Other than that, LGMT 👍

.github/workflows/cron-streams-active-cleanup.yaml
victorges: Yeah agreed! I don't think it's in scope of this PR though as explained here: https://github.com/livepeer/studio/pull/2013#discussion_r1579080037

packages/api/src/controllers/stream.ts
victorges: So this was just to keep the same logic of returning `isActive: false` after the active timeout for streams that got lost in action. Similar to what had before, but instead of triggering the clean-up job at that point we just monkey patch the API object before returning it, and wait for the clean-up job to pick it up.

I made it like this to make sure we keep retrocompatibility, tho thinking further about it I thought that we could also just not have the monkey patch at all? So we keep returning `isActive: true` for the user for up to ~7 minutes (worst case) until a clean-up job comes and fixes it (and triggers recording).

WDYT? Should we keep the retrocompatibility in the sense "if the stream is not seen in 90s we already start returning `isActive: false`", or should we just remove this monkey patch once and for all?

PR #2014, opened 2024-01-18
don't produce db query related prometheus metrics

pwilczynskiclearcode:


packages/api/src/store/db.ts
```diff
@@ -42,18 +41,6 @@ export interface PostgresParams {
 
 type Table<T> = BaseTable<WithID<T>>;
 
-type QueryHistogramLabels = {
-  query: string;
-  result: string;
-};
-
-const metricHistogram: Histogram<keyof QueryHistogramLabels> = new Histogram({
```
pwilczynskiclearcode, line 50: we already have query insights enabled on gcp: https://console.cloud.google.com/sql/instances/livepeer-com/insights;duration=PT1H;sort_by=TOTAL_EXEC_TIME/executed?organizationId=251137649854&project=livepeer-firestore-production
so we can drop own prometheus metrics related to db query time. 
We send `query` label which can contain extremely long full-text SQL query which is dropped by Victoria Metrics anyway and [pollutes its logs](https://eu-metrics-monitoring.livepeer.live/grafana/goto/2n-YdJcIg?orgId=1)

victorges (approved):

PR #2015, opened 2024-01-18
Start on webhook log API

mjh1:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -279,7 +279,7 @@ app.delete("/", authorizer({}), async (req, res) => {
   res.end();
 });
 
-const logsFieldsMap: FieldsMap = {
+const requestsFieldsMap: FieldsMap = {
   id: `webhook_response.ID`,
   createdAt: { val: `webhook_response.data->'createdAt'`, type: "int" },
   userId: `webhook_response.data->>'userId'`,
@@ -287,7 +287,7 @@ const logsFieldsMap: FieldsMap = {
   statusCode: `webhook_response.data->'response'->>'status'`,
 };
 
-app.get("/:id/logs", authorizer({}), async (req, res) => {
+app.get("/:id/requests", authorizer({}), async (req, res) => {
   let { limit, cursor, all, event, allUsers, order, filters, count } =
     req.query;
   if (isNaN(parseInt(limit))) {
@@ -298,7 +298,7 @@ app.get("/:id/logs", authorizer({}), async (req, res) => {
   }
 
   if (req.user.admin && allUsers && allUsers !== "false") {
-    const query = parseFilters(logsFieldsMap, filters);
+    const query = parseFilters(requestsFieldsMap, filters);
     query.push(sql`webhook_response.data->>'webhookId' = ${req.params.id}`);
     if (!all || all === "false") {
       query.push(sql`webhook_response.data->>'deleted' IS NULL`);
@@ -315,7 +315,7 @@ app.get("/:id/logs", authorizer({}), async (req, res) => {
       cursor,
       fields,
       from,
-      order: parseOrder(logsFieldsMap, order),
+      order: parseOrder(requestsFieldsMap, order),
       process: ({ data, usersdata, count: c }) => {
         if (count) {
           res.set("X-Total-Count", c);
@@ -332,7 +332,7 @@ app.get("/:id/logs", authorizer({}), async (req, res) => {
     return res.json(output);
   }
 
-  const query = parseFilters(logsFieldsMap, filters);
+  const query = parseFilters(requestsFieldsMap, filters);
   query.push(sql`webhook_response.data->>'userId' = ${req.user.id}`);
   query.push(sql`webhook_response.data->>'webhookId' = ${req.params.id}`);
 
@@ -350,7 +350,7 @@ app.get("/:id/logs", authorizer({}), async (req, res) => {
     cursor,
     fields,
     from,
-    order: parseOrder(logsFieldsMap, order),
+    order: parseOrder(requestsFieldsMap, order),
     process: ({ data, count: c }) => {
       if (count) {
         res.set("X-Total-Count", c);
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -94,7 +94,7 @@ app.get("/", authorizer({}), async (req, res) => {
   const query = parseFilters(fieldsMap, filters);
   query.push(sql`webhook.data->>'userId' = ${req.user.id}`);
 
-  if (!all || all === "false") {
+  if (!all || all === "false" || !req.user.admin) {
     query.push(sql`webhook.data->>'deleted' IS NULL`);
   }
 
@@ -336,7 +336,7 @@ app.get("/:id/requests", authorizer({}), async (req, res) => {
   query.push(sql`webhook_response.data->>'userId' = ${req.user.id}`);
   query.push(sql`webhook_response.data->>'webhookId' = ${req.params.id}`);
 
-  if (!all || all === "false") {
+  if (!all || all === "false" || !req.user.admin) {
     query.push(sql`webhook_response.data->>'deleted' IS NULL`);
   }
 
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -8,8 +8,6 @@ import { makeNextHREF, parseFilters, parseOrder, FieldsMap } from "./helpers";
 import { db } from "../store";
 import sql from "sql-template-strings";
 import { UnprocessableEntityError, NotFoundError } from "../store/errors";
-import { WebhookStatusPayload } from "../schema/types";
-import { DBWebhook } from "../store/webhook-table";
 
 function validateWebhookPayload(id, userId, createdAt, payload) {
   try {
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/webhook.ts
```diff
@@ -44,7 +44,7 @@ const fieldsMap: FieldsMap = {
   url: `webhook.data->>'url'`,
   blocking: `webhook.data->'blocking'`,
   deleted: `webhook.data->'deleted'`,
-  createdAt: `webhook.data->'createdAt'`,
+  createdAt: { val: `webhook.data->'createdAt'`, type: "int" },
```
mjh1, line 47: @victorges for some reason the general /webhooks list API stopped working, there was an error about sorting on this field but this seems to fix it. Does that seem ok?
The error was:
```
2024-01-19 12:35:28 [error]: 	runQuery phase=error elapsed=38ms error=unimplemented: can't order by column type jsonb query={"strings":["SELECT  webhook.id as id, webhook.data as data, count(*) OVER() AS count FROM webhook WHERE webhook.data->>'userId' = ","  AND webhook.data->>'deleted' IS NULL  ORDER BY webhook.data->'createdAt' DESC NULLS LAST LIMIT ",""],"values":["9c2936b5-143f-4b10-b302-6a21b5f29c3d",20]}
error: unimplemented: can't order by column type jsonb
```

victorges: Yeah this should be fine. FTR another approach would be changing it to 
```
createdAt: `webhook.data->>'createdAt'`,
```
(`->>` instead of `->`)

Which would read the field as a string instead of jsonb. This would render inaccurate ordering of an int tho as it'd be comparing them as strings, but at least it would use the index that we have on the field (related to https://github.com/livepeer/livepeer-com/blob/16fd70915821dfbbc21bfd6bd8d3b65c90a1ade9/packages/api/src/store/stream-table.ts#L116).

I checked the other APIs we have though, and we are always doing this `type: "int"`, so let's leave it like this here for consistency. Once we fix these indexes we can fix all the places.

packages/api/src/controllers/webhook.ts
```diff
@@ -345,4 +279,93 @@ app.delete("/", authorizer({}), async (req, res) => {
   res.end();
 });
 
+const requestsFieldsMap: FieldsMap = {
```
victorges, line 282: Do you think this `/requests` sub-path is going to grow a lot? If so, it might be worth considering exporting a separate `Router()` on another file, then using it here only with `app.use('/:id/requests', requests)`

Not a required change though, mostly a nit

packages/api/src/controllers/webhook.ts
```diff
@@ -345,4 +279,93 @@ app.delete("/", authorizer({}), async (req, res) => {
...
+    return res.json(output);
+  }
+
+  const query = parseFilters(requestsFieldsMap, filters);
+  query.push(sql`webhook_response.data->>'userId' = ${req.user.id}`);
+  query.push(sql`webhook_response.data->>'webhookId' = ${req.params.id}`);
+
+  if (!all || all === "false") {
+    query.push(sql`webhook_response.data->>'deleted' IS NULL`);
+  }
```
victorges: I believe this option shouldn't be available in non-admin queries. Lmk if you saw this in another handler 👀

packages/api/src/schema/api-schema.yaml
```diff
@@ -241,21 +240,14 @@ components:
           example: 1587667174725
         duration:
           type: number
-        statusCode:
-          type: number
-          default: 0
+        request:
+          type: object
```
victorges, line 247: Could you add the schema for this object as well?

packages/api/src/schema/db-schema.yaml
```diff
@@ -593,10 +593,27 @@ components:
...
+          type: string
+          index: true
+        createdAt:
           index: true
+        response:
+          type: object
+          properties:
+            status:
+              type: number
+              index: true
```
victorges: I couldn't tell if all these fields are in the `api-schema` as well, but I think at least `event` is not. Just to explain the idea between these 2 files:
 - `api-schema` gets generated into our public API docs, so it should contain the exact schema we want documented for our users
 - Then either for internal fields or custom properties we set on our fields (like `index:`), we set in this `db-schema` file.
   - Internal fields should have a `writeOnly` prop here (or custom code) so they get filtered on API responses
   - This gets deep merged on top of the API schema before we generate API code (since it gets deep merged, also no need to include duplicate props here like `type`)
 
So here, any field that is returned to the user (e.g. `event`) should also be in the `api-schema` so it shows up in our docs etc. Ideally with a `description` as well which will go in the doc

Does it make sense? For more info there's a little readme from when I did the schema split: https://github.com/livepeer/livepeer-com/blob/1c0773d9d6f5dfcffe0f149b19abb8dc4076bcd0/packages/api/src/schema/README.md#L1

packages/api/src/webhooks/cannon.ts
```diff
@@ -474,26 +487,31 @@ export default class WebhookCannon {
     event: messages.WebhookEvent,
     resp: Response,
     startTime: [number, number],
-    responseBody: string
+    responseBody: string,
+    stream: DBStream,
+    params
```
victorges, line 492: I believe this type works here: `RequestInitWithTimeout`

packages/api/src/webhooks/cannon.ts
```diff
@@ -474,26 +487,31 @@ export default class WebhookCannon {
...
+      const encodedResponseBody = Buffer.from(
+        responseBody.substring(0, 1024)
+      ).toString("base64");
 
       await this.db.webhookResponse.create({
         id: uuid(),
         webhookId: webhook.id,
         eventId: event.id,
+        event: event.event,
+        userId: webhook.userId,
```
victorges, line 505: Btw do you plan on clearing the existing webhook_response objects or are we gonna be retrocompatible in some way? I'd be fine with just deleting them (or archiving if we wanna be safe), but I wonder if there might be problems if we keep the old objects in the DB and risk getting compatibility errors

packages/api/src/webhooks/cannon.ts
```diff
@@ -474,26 +487,31 @@ export default class WebhookCannon {
...
         duration: hrDuration[0] + hrDuration[1] / 1e9,
-        statusCode: resp.status,
         response: {
           body: encodedResponseBody,
-          headers: resp.headers.raw(),
           redirected: resp.redirected,
           status: resp.status,
           statusText: resp.statusText,
         },
+        request: params,
```
victorges: Can we have an explicit schema here? This params is a raw `fetch` parameters object and I think it could change inadvertently (or have weird stuff like functions in it). Even if we end up just copying specific fields here like url, header, etc, I think it'd still be preferrable. WDYT?

packages/api/src/webhooks/cannon.ts
mjh1: That's a good point yep it would look wrong as we'd have results with different fields appearing etc. I think yeah archiving them would be best. I'll create a ticket to make sure that gets done.

packages/api/src/controllers/webhook.ts
mjh1: Huh, yeah good point! I think I must have copied this from here: https://github.com/livepeer/studio/blob/f71d79de9fccb99889f697d11b7a94af290e349d/packages/api/src/controllers/webhook.ts#L99 I'll fix it there too

packages/api/src/controllers/webhook.ts
mjh1: yep that sounds good, i've branched off this branch for the resend api so i'll implement this there to avoid merge conflicts

packages/api/src/webhooks/cannon.ts
mjh1: I did that initially but then thought we needed everything for the resending. being explicit is safer though, i'll switch it back

packages/api/src/schema/db-schema.yaml
mjh1: Ah yep thanks, will remember for next time! I've sorted that all out now I think

packages/api/src/schema/api-schema.yaml
```diff
@@ -228,32 +228,46 @@ components:
...
           readOnly: true
           type: number
           description: |
-            Timestamp (in milliseconds) at which webhook response object was
+            Timestamp (in milliseconds) at which webhook request object was
             created
           example: 1587667174725
         duration:
           type: number
+          description: The time taken to make the webhook request
```
victorges: Can you add the unit here? Likely millis?

packages/api/src/schema/api-schema.yaml
```diff
@@ -228,32 +228,46 @@ components:
...
         request:
           type: object
+          properties:
```
victorges, line 248: Should we include the URL here too, in case the webhook config changes (I think we talked about this on Discord, don't remember the decision)

victorges (approved):
LGTM just some schema nits

packages/api/src/schema/api-schema.yaml
mjh1: Oh yep good point, also the secret?

victorges: I believe that would be included in the headers and/or payload right?

mjh1: ah yep forgot i already looked at that, we have the sig in the headers

PR #2017, opened 2024-01-22
api: added user custom tags to stream object

gioelecerati:

Added an object `tags` field to the stream object with a max size of 2048 characters for user custom data

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -337,6 +337,11 @@ components:
           example: test_stream
         creatorId:
           $ref: "#/components/schemas/creator-id"
+        tags:
+          type: object
+          description: User input tags associated with the stream
+          additionalProperties:
+            type: string
         lastSeen:
           type: number
           example: 1587667174725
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -337,6 +337,11 @@ components:
           example: test_stream
         creatorId:
           $ref: "#/components/schemas/creator-id"
+        tags:
```
victorges: Thom suggested renaming this to `userTags` to be extra clear. WDYT?
ref: https://www.notion.so/SHARED-Trovo-Technical-Design-Pull-Based-Ingest-8603fca9e09c4feca57da38e89c9777e?d=5966bcbf2d4641b08701d378caf7f201&pvs=4#a42563388f844e11b48c32422d14fb82

victorges (approved):
LGTM, but let's refrain from merging this until we confirm if Trovo is actually going to need it. If all they want is a creator ID tag, they should use the specific field we have for that

thomshutt:
@victorges - @hthillman seemed to think this would be a generally useful feature for more sophisticated customers to be able to integrate. If the work is done then I'd vote for merging regardless of whether Trovo end up using it

victorges:
@thomshutt I see. I come from the fact that we've already had this feature for at least a year on VOD and no one used it. Then we started having confusing experiences on actual features because we had the unused stuff in our schema (which is why we eventually removed it). If we know we're gonna attract different kinds of customers now that are gonna make use of these, I'm fine with merging it.

PR #2021, opened 2024-01-24
Webhook get and resend APIs

mjh1:
Implements get and resend APIs for webhook logs. So now you can retrieve info for a single webhook log and trigger a resend.

- I had to move the `storeResponse` function outside of the cannon object so that it could be called by the newly written `resendWebhook` function. 
- I moved the webhook log API handlers to a separate file for better organisation.

Tested locally and wrote a unit test.

**Checklist**

### DIFF
--- a/packages/api/src/controllers/webhook.ts
+++ b/packages/api/src/controllers/webhook.ts
@@ -7,15 +7,8 @@ import { v4 as uuid } from "uuid";
 import { makeNextHREF, parseFilters, parseOrder, FieldsMap } from "./helpers";
 import { db } from "../store";
 import sql from "sql-template-strings";
-import {
-  UnprocessableEntityError,
-  NotFoundError,
-  ForbiddenError,
-  BadRequestError,
-} from "../store/errors";
-import { resendWebhook } from "../webhooks/cannon";
-import { DBWebhook } from "../store/webhook-table";
-import { WebhookResponse } from "../schema/types";
+import { UnprocessableEntityError, NotFoundError } from "../store/errors";
+import webhookLog from "./webhook-log";
 
 function validateWebhookPayload(id, userId, createdAt, payload) {
   try {
@@ -45,6 +38,8 @@ function validateWebhookPayload(id, userId, createdAt, payload) {
 }
 
 const app = Router();
+
+app.use("/:id/log", webhookLog);
 
 const fieldsMap: FieldsMap = {
   id: `webhook.ID`,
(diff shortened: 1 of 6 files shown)


mjh1:
Annoyingly this PR looks bigger than it really is, I had to move some code around so lots of deletes and additions of the same lines.

mjh1:
Just realised I need to add the new API paths to api-schema.yaml

victorges:
> Annoyingly this PR looks bigger than it really is, I had to move some code around so lots of deletes and additions of the same lines.

If it'd be easy to separate the moving around on a separate commit, it'd be easier to review! no problem if it's not trivial tho

mjh1:
> > Annoyingly this PR looks bigger than it really is, I had to move some code around so lots of deletes and additions of the same lines.
> 
> If it'd be easy to separate the moving around on a separate commit, it'd be easier to review! no problem if it's not trivial tho

I'll try! 👀

mjh1:
> If it'd be easy to separate the moving around on a separate commit, it'd be easier to review! no problem if it's not trivial tho

Done 👍 the first commit is the one with the real changes, the next two are just moving code around

mjh1:
@victorges ready for review now

packages/api/src/controllers/webhook.ts
```diff
@@ -287,9 +295,45 @@ const requestsFieldsMap: FieldsMap = {
   statusCode: `webhook_response.data->'response'->>'status'`,
 };
 
+app.put("/:id/log/:requestId/resend", authorizer({}), async (req, res) => {
```
victorges: Should this be a `POST` instead? Unless we somehow keep the same `requestId` when we resend, but I guess we don't 🤔

packages/api/src/controllers/webhook.ts
```diff
@@ -287,9 +295,45 @@ const requestsFieldsMap: FieldsMap = {
...
+  const webhookResponse = await db.webhookResponse.get(req.params.requestId);
+  await checkRequest(req, webhook, webhookResponse);
+
+  await resendWebhook(webhook, webhookResponse);
+  res.status(202).end();
+});
+
+app.get("/:id/log/:requestId", authorizer({}), async (req, res) => {
+  const webhook = await db.webhook.get(req.params.id);
+  const webhookResponse = await db.webhookResponse.get(req.params.requestId);
```
victorges: Something funny here is that the `id` and `requestId` params are redundant right? I wonder if we'll ever need an API which is just `/webhook/log/:requestId` instead, but let's keep it like this for now I think it's consistent with standards

packages/api/src/controllers/webhook.ts
```diff
@@ -287,9 +295,45 @@ const requestsFieldsMap: FieldsMap = {
...
+  await checkRequest(req, webhook, webhookResponse);
+
+  res.status(200);
+  return res.json(webhookResponse);
```
victorges: Should we call `db.webhookResponse.cleanWriteOnly` on these APIs? Or could include one of those `mung.json` middleware on the `webhook-log` router

packages/api/src/controllers/webhook.ts
```diff
@@ -287,9 +295,45 @@ const requestsFieldsMap: FieldsMap = {
...
+});
+
+async function checkRequest(
+  req,
```
victorges: **ts-nit:** You can change the `express` import to 
```ts
import { Request, Router } from "express";
```
and use the `Request` type here (it will also fix the typing of all the route handlers 🙏)

packages/api/src/controllers/webhook.ts
```diff
@@ -287,9 +295,45 @@ const requestsFieldsMap: FieldsMap = {
...
+  webhook: DBWebhook,
+  webhookResponse: WebhookResponse
+) {
+  if (
+    !req.user.admin &&
+    (req.user.id !== webhook.userId || req.user.id !== webhookResponse.userId)
```
victorges: This will get an NRE if `webhook` or `webhookResponse` are null (checked below)

packages/api/src/webhooks/cannon.ts
```diff
@@ -497,11 +494,11 @@ export default class WebhookCannon {
         responseBody.substring(0, 1024)
       ).toString("base64");
 
-      await this.db.webhookResponse.create({
```
victorges, line 500: Should we remove this field from the class and fix all usages?

packages/api/src/webhooks/cannon.ts
```diff
@@ -626,6 +623,48 @@ export default class WebhookCannon {
   }
 }
 
+export async function resendWebhook(
```
victorges, line 631: Hmmmm, so the resending is synchronous? I thought we would just send an event over to RabbitMQ and let the cannon do a regular delivery, with retries etc. 

If it is synchronous (and blocking), shouldn't the API return a 200 with the new `WebhookResponse` in the payload instead?

packages/api/src/webhooks/cannon.ts
```diff
@@ -626,6 +623,48 @@ export default class WebhookCannon {
...
+  const triggerTime = Date.now();
+  const startTime = process.hrtime();
+  let resp: Response;
+  let responseBody: string;
+  let statusCode: number;
+  let errorMessage: string;
+  try {
+    resp = await fetchWithTimeout(webhookResponse.request.url, {
+      method: webhookResponse.request.method,
+      headers: webhookResponse.request.headers,
```
victorges, line 660: I'm not sure if we should repeat the webhook signature here. We add a timestamp to the signature payload to avoid the possibility of replay attacks, so the webhook handler can check that and make sure it is never reused. But then we are doing a replay "attack" ourselves through the `resend` flow!

OTOH this is a "debug-only" API right? We don't expect users to be calling this on a healthy workflow, and maybe they _do_ want the absolute exact payload.

WDYT? Do we have any reference from other platforms? Might be worth expliciting this behavior on the docs, regardless of which one we choose.

packages/api/src/schema/api-schema.yaml
```diff
@@ -2729,6 +2729,80 @@ paths:
...
+                $ref: "#/components/schemas/webhook-response"
+        default:
+          description: Error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/error"
+  "/webhook/{id}/log/{logId}/resend":
+    put:
+      summary: Resend a webhook
```
victorges, line 2786: It might be worth some further `description:` about what this API does

victorges (approved):
LGTM either way but I have some functional suggestions/discussions (and some less important non-functional/nits)

Reviewed commit by commit (thanks!), so some comments may be in the initial places where the code was written

packages/api/src/controllers/webhook.ts
mjh1: So you're suggesting `POST /:id/log/resend` with the requestID in the request body?

packages/api/src/controllers/webhook.ts
mjh1: Yeah I wasn't too sure which way was preferable

packages/api/src/webhooks/cannon.ts
mjh1: I think I'll save that for a separate PR 👍

packages/api/src/controllers/webhook.ts
mjh1: I found that I needed to use `import Router from "express/lib/router";` otherwise I get errors on the queryparams like limit and cursor:
```TS2322: Type 'string | ParsedQs | string[] | ParsedQs[]' is not assignable to type 'string | number'.   Type 'ParsedQs' is not assignable to type 'string | number'.```

File: packages/api/src/controllers/webhook.ts (hunk 1 above)
Comment (victorges) [reply packages/api/src/controllers/webhook.ts]: No, I think `POST /:id/log/:requestId/resend` is fine! It's just that a `PUT .../:requestId` would seem like it's updating the same request/log, instead of creating a new one. Also I think the general pattern for "actions" (in this case "resend") is to be a `POST` as well

File: packages/api/src/controllers/webhook.ts (hunk 4 above)
Comment (victorges) [reply packages/api/src/controllers/webhook.ts]: Yeah those are actual errors we would get in production! I think if a query-string is specified multiple times it will come as an array to the API handlers. To fix these, I generally convert the query-string to all strings, with `toStringValues`: e.g. https://github.com/livepeer/livepeer-com/blob/80f3aa14daed6a03eba294b490544e0725796472/packages/api/src/controllers/stream.ts#L1010

File: packages/api/src/webhooks/cannon.ts (hunk 7 above)
Comment (mjh1) [reply packages/api/src/webhooks/cannon.ts:631]: As discussed on a call let's keep it blocking but I'll return the response directly 👍

File: packages/api/src/webhooks/cannon.ts (hunk 8 above)
Comment (mjh1) [reply packages/api/src/webhooks/cannon.ts:660]: As discussed, regenerating the sig makes sense 👍

**Comment by mjh1** [comment]
@victorges would you mind just sanity checking the last two commits? thanks!

File: packages/api/src/webhooks/cannon.ts [hunk 10]
```diff
@@ -615,7 +617,10 @@ async function storeResponse(
         method: params.method,
         headers: params.headers,
       },
-    });
+      sharedSecret: sharedSecret,
+    };
+    await db.webhookResponse.create(webhookResponse);
+    return webhookResponse;
   } catch (e) {
```
victorges: Perhaps we should rethrow this error now, otherwise the API will respond with a `200` and an "undefined" body

packages/api/src/webhooks/cannon.ts
```diff
@@ -626,14 +631,24 @@ async function storeResponse(
...
   try {
+    const sigHeaders = signatureHeaders(
+      webhookResponse.request.body,
+      webhookResponse.sharedSecret,
+      Date.now()
+    );
+    webhookResponse.request.headers = {
+      ...webhookResponse.request.headers,
+      ...sigHeaders,
+    };
```
victorges: This timestamp actually goes in the payload as well, so we need to update that too
```suggestion
    const timestamp = Date.now()
    const payload = { ...webhookResponse.request.body, timestamp }
    const sigHeaders = signatureHeaders(
      payload,
      webhookResponse.sharedSecret,
      timestamp
    );
    webhookResponse.request.headers = {
      ...webhookResponse.request.headers,
      ...sigHeaders,
    };
```
(then use `payload` instead of `webhookResponse.request.body` below)

victorges (approved):
great

packages/api/src/webhooks/cannon.ts
mjh1: ah thanks 👍

packages/api/src/webhooks/cannon.ts
mjh1: I moved the catch into the main firehook function where it's still needed. hopefully that looks good to you https://github.com/livepeer/studio/pull/2021/commits/9490476346cd7d70ae71e0167b66582e88c02616?w=1

PR #2024, opened 2024-01-31
api: Create PUT /stream/pull API for idempotent pull stream

victorges:

This creates a separate `PUT /stream/pull` API to replace the `autoStartPull` query-string
that had been created on the regular creation API initially.

This is because the semantics expected from this pull API are now becoming very different
from the regular POST-based Create Stream API that we have today. Specifically, the 2 new
requirements are:
- It should update the stream object if it already exists
- It should block the request until the stream actually starts on the media server

Since blocking a connection is a non-standard and non-recommended behavior, that part
was still kept under a query-string. So this API has 2 possible idempotence results:
- If `waitActive` is NOT specified: when the API succeeds, the stream is guaranteed to be created and the pull to have been triggered.
- If `waitActive` IS specified: when the API succeeds, the stream is guaranteed to be created, the pull triggered, and the media server already started the stream.

In both cases, if a stream with the same `pull.source` already existed, it will have been updated
and then triggered to start.

One caveat is that we make no checks about the stream _already_ being active when this API is called.
In that case, the media server won't actually update the configuration of the ongoing stream, so it is an
exception to the idempotence aimed by this API. We can think of alternatives, like returning an error when
the stream is already active, similar to the PATCH API, but it depends on what the design partner for this
feature expects from this API.

Also notice that I ended up preferring not to use this: https://github.com/livepeer/livepeer-data/pull/180
Doing a simpler poll on the DB instead for the `stream.isActive` field.

- Abstract the "create stream" logic from the existing API into a helper, to be reusable in the new API
- Create separate `PUT /stream/pull` API
- Implement logic to update stream object based on PUT payload
- Implement logic to wait for stream to become active

The pull logic on Catalyst is not implement yet so it's not really possible to test it. Mostly tested calling the API itself.

Implements latest requirements here https://discord.com/channels/423160867534929930/1201860223652859986/1201860228413399040

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -609,6 +609,7 @@ const fieldsMap = {
   createdAt: { val: `asset.data->'createdAt'`, type: "int" },
   updatedAt: { val: `asset.data->'status'->'updatedAt'`, type: "int" },
   userId: `asset.data->>'userId'`,
+  creatorId: `stream.data->'creatorId'->>'value'`,
   playbackId: `asset.data->>'playbackId'`,
   playbackRecordingId: `asset.data->>'playbackRecordingId'`,
   phase: `asset.data->'status'->>'phase'`,
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -542,9 +542,7 @@ describe("controllers/stream", () => {
         expect(updatedStream.profiles).toEqual([]);
 
         const document = await db.stream.get(stream.id);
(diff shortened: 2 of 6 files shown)


mjh1 (approved):
Are you thinking you'll remove the stream pull logic from the create stream handler in a later PR?

victorges:
> Are you thinking you'll remove the stream pull logic from the create stream handler in a later PR?

Yeah exactly, we need the confirmation that they migrated to the new one so we can delete the old logic. Will probably do it when removing the experiment itself.

PR #2028, opened 2024-01-31
fix: fix broadcasting

0xcadams:
This fixes broadcasting by changing the endpoint for WebRTC to be a full URL, as opposed to a relative path.

This commit broke Studio's broadcasting config, and it now expects a full URL and is failing on that: https://github.com/livepeer/livepeer-react/blame/04c040398b917e8d528b00c98c7b350aa0ab6ace/packages/core-web/src/media/browser/webrtc/shared.ts#L267


victorges (approved):
LGTM

PR #2030, opened 2024-01-31
api: fix userTags type

gioelecerati:

Fixing tags type because I noticed they were supporting only `string`

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -653,7 +653,6 @@ components:
           index: true
         userTags:
           type: object
-          additionalProperties: true
         isActive:
           index: true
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/db-schema.yaml
```diff
@@ -653,6 +653,7 @@ components:
           index: true
         userTags:
           type: object
+          additionalProperties: true
```
victorges: This doesn't need to be repeated here! The `db-schema` is deep merged with `api-schema`, so any field that is not here will be just kept!

victorges (approved):

victorges (approved):
cool

PR #2032, opened 2024-02-01
Rename webhook response table

mjh1:


victorges (approved):
Nice!

Keep in mind that all webhook logs will be "deleted" once we deploy this. Might want to run a query on the DB to copy the logs with the new schema to the new table, after we deploy this.

PR #2033, opened 2024-02-01
Add stream config endpoint

mjh1:


packages/api/src/controllers/stream.ts
```diff
@@ -1848,6 +1848,25 @@ app.get("/:id/info", authorizer({}), async (req, res) => {
   res.json(resp);
 });
 
+app.get("/:id/config", authorizer({ anyAdmin: true }), async (req, res) => {
+  let { id } = req.params;
+  let stream = await db.stream.getByPlaybackId(id);
+  if (!stream) {
+    stream = await db.stream.get(id);
+  }
```
victorges: Perhaps by streamKey instead? Catalyst never gets the stream ID AFAIK.

But also, maybe we want to be explicit here by using query-strings instead like `/api/stream/config?playbackId=<pid>` or `/api/stream/config?streamKey=<sk>`

Otherwise, I'm concerned we could for example eventually start allowing playback by stream key or stream id, which could leak to accidental leaks of those for example 🤔  (stream ID is also secret, it kinda gives ownership over the resource if the app uses a frontend API key)

ORRR, to keep it simple, we can allow only playback ID for now and think about this later when this evolves for example to replace the API that mapic calls.

packages/api/src/controllers/stream.ts
```diff
@@ -1848,6 +1848,25 @@ app.get("/:id/info", authorizer({}), async (req, res) => {
...
+  if (!stream) {
```
victorges: Need to check `stream.deleted` and maybe `stream.suspended` as well

packages/api/src/controllers/stream.ts
```diff
@@ -1848,6 +1848,25 @@ app.get("/:id/info", authorizer({}), async (req, res) => {
...
+    res.status(404);
+    return res.json({
+      errors: ["not found"],
+    });
+  }
+
+  res.status(200);
+  res.json({
+    pull: stream.pull,
```
victorges: Do we want to forward the whole pull? We could omit the `location` field if catalyst will make no use of it.
```suggestion
    pull: !stream.pull : null : {
      ...stream.pull,
      location: undefined
    }
```

victorges (approved):
LGTM either way simple asks

PR #2034, opened 2024-02-01
api: Add createdByTokenName field to assets

victorges:
This is needed to do some analytics on how many assets are being created with API keys or from the dashboard.

Additional changes:
- Did some cleanup on `validateAssetPayload` to avoid it growing indefinitely
- Removed the `createdByTokenId` from streams which just increases our leakage risk (API keys IDs are the keys themselves)


packages/api/src/controllers/asset.ts
```diff
@@ -182,15 +182,15 @@ function parseUrlToDStorageUrl(
 }
 
 export async function validateAssetPayload(
+  req: Pick<Request, "body" | "user" | "token" | "config">,
```
gioelecerati, line 185: Lovely

gioelecerati (approved):
LGTM

PR #2040, opened 2024-02-06
Add flv playback URL to playback info endpoint

mjh1:


packages/api/src/controllers/playback.test.ts
```diff
@@ -143,7 +147,44 @@ describe("controllers/playback", () => {
...
+              },
+              {
+                hrn: "WebRTC (H264)",
+                type: "html5/video/h264",
+                url: `${ingest}/webrtc/${stream.playbackId}`,
+              },
+              {
+                hrn: "FLV (H264)",
+                type: "video/x-flv",
+                url: `${ingest}/flv/${stream.playbackId}`,
```
leszko, line 180: Is this the correct URL? Shouldn't it have some suffix like `.flv`?

btw. I tried it in the box and it's not playable, but maybe it's just the box...

mjh1: Yeah it's no suffix like webrtc e.g. this works: `https://livepeercdn.studio/flv/db15j36hse0i9lay`. The box seems to have some different nginx setup going on which stops that working. So we can fix that separately.

leszko: Great. Thanks for explanation.

leszko (approved):

PR #2042, opened 2024-02-07
api: update terminate endpoing to allow non-admin users

emranemran:
Allow any user that owns the stream to call /terminate to stop a running stream.


mjh1 (changes requested):
I think this just has the effect of stopping admins from being able to nuke other users streams so we don't want that. It looks actually like this API is already open to non-admins for their own streams.

emranemran:
> I think this just has the effect of stopping admins from being able to nuke other users streams so we don't want that. It looks actually like this API is already open to non-admins for their own streams.

I see what you're saying - in that case, it seems like we don't need do anything at all and it should just work.

PR #2043, opened 2024-02-08
api: playback: require auth for playback recordings

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/playback.ts
+++ b/packages/api/src/controllers/playback.ts
@@ -17,7 +17,11 @@ import { DBSession } from "../store/session-table";
 import { Asset, PlaybackInfo, Stream, User } from "../schema/types";
 import { DBStream } from "../store/stream-table";
 import { WithID } from "../store/types";
-import { NotFoundError, UnprocessableEntityError } from "../store/errors";
+import {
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/playback.ts
```diff
@@ -316,6 +320,29 @@ app.get("/:id", async (req, res) => {
   const origin = req.headers["origin"] ?? "";
   const isEmbeddablePlayer = embeddablePlayerOrigin.test(origin);
 
+  if (withRecordings) {
```
victorges: Could this be easily moved to inside the `getPlaybackInfo` function? There's probably a place in there where we handle streams explicitly, and I think we should avoid fetching the same data multiple times on a request

victorges (approved):
LGTM anyway but we should avoid fetching the data twice. This API is on the critical path for playback and any millisecond counts!

packages/api/src/controllers/playback.ts
gioelecerati: Makes TONs of sense! Moving it and removing the useless stream fetch

PR #2053, opened 2024-02-13
api: webhook: allow additional headers & payload to be sent in webhooks

gioelecerati:

Allow catalyst to specify two new objects when calling Studio gate
- `webhookPayload`
- `webhookHeaders` 
They will get appended to the JSON payload and the headers respectively when doing a webhook request for access control to an external party

Also added `Trovo-Auth-Version: 1.1` as header to any webhook fired for a stream created via the `pull` api

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -13,7 +13,12 @@ import {
 } from "../store/errors";
 import tracking from "../middleware/tracking";
 import { DBWebhook } from "../store/webhook-table";
-import { Asset, PlaybackPolicy, User } from "../schema/types";
+import {
+  AccessControlGatePayload,
+  Asset,
+  PlaybackPolicy,
+  User,
+} from "../schema/types";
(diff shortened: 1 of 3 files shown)


mjh1 (approved):

packages/api/src/controllers/access-control.ts
```diff
@@ -34,23 +39,38 @@ type GateConfig = {
...
-    }),
+    body: JSON.stringify(jsonPayload),
   };
 
+  if (payload.webhookHeaders) {
+    params.headers = { ...params.headers, ...payload.webhookHeaders };
+  }
+
+  if (isTrovoAuth) {
+    params.headers["Trovo-Auth-Version"] = "1.1";
```
victorges, line 71: This hurts 😆 
Let's make sure to migrate these to per-webhook headers configs later, which Trovo can play around however they want. We should avoid hardcoded logic for specific clients.

PR #2057, opened 2024-02-15
Projects integration

emranemran:
WIP integration branch


0xcadams:
Can we close this now that we landed projects?

PR #2058, opened 2024-02-15
api: acl: bring limit for hackers down to 5 per node

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -23,7 +23,7 @@ import { isFreeTierUser } from "./helpers";
 import { cache } from "../store/cache";
 
 const WEBHOOK_TIMEOUT = 30 * 1000;
-const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 5;
+const MAX_ALLOWED_VIEWERS_FOR_HACKER_TIER_PER_NODE = 5;
 const app = Router();
 
 type GateConfig = {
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -23,7 +23,7 @@ import { isFreeTierUser } from "./helpers";
 import { cache } from "../store/cache";
 
 const WEBHOOK_TIMEOUT = 30 * 1000;
-const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 30;
+const MAX_ALLOWED_VIEWERS_FOR_FREE_TIER = 5;
```
pwilczynskiclearcode: I'd change the variable name to `MAX_ALLOWED_VIEWERS_FOR_HACKER_TIER_PER_NODE` or comment that it's per-node value. It's not obvious in this context.

pwilczynskiclearcode:
With this implementation lower limit is the only option. The problem is that with scaling up the number of nodes (like for Fishtank) we raise the global viewers limit for the hacker plan. Maybe it's acceptable for now.

Also with current 29 nodes. 29 * 5 = 145 which is quite high and still above our alert's thresholds

gioelecerati:
> With this implementation lower limit is the only option. The problem is that with scaling up the number of nodes (like for Fishtank) we raise the global viewers limit for the hacker plan. Maybe it's acceptable for now.
> 
> Also with current 29 nodes. 29 * 5 = 145 which is quite high and still above our alert's thresholds

Yes, until we have a catalyst shared state with the amount of viewers globally, this is the only space we have to move around.
I agree that 145 is high, but consider that playback would need to happen in all regions to reach that number, right?

thomshutt (commented):
I think that bringing this down makes sense - hopefully 5 per node is low enough that the system's not usable for people trying to abuse the free tier

thomshutt (approved):

pwilczynskiclearcode:
@gioelecerati 
>  I agree that 145 is high, but consider that playback would need to happen in all regions to reach that number, right?

Right. I guess that rate-limiting is done after geolocation... so if we reach the limit on a node we won't redirect viewer to another node.

pwilczynskiclearcode (approved):

PR #2061, opened 2024-02-16
Add a webhook log success field to allow filtering in the UI

mjh1:

Adding a success field to webhook logs so that we can easily filter based on this in the UI.

- Added a success field to webhook logs
- Fixed creating webhook logs when there is no http response, in case of timeouts or connection refused etc.

- Tested locally on box environment

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/webhook-log.ts
+++ b/packages/api/src/controllers/webhook-log.ts
@@ -26,6 +26,7 @@ const requestsFieldsMap: FieldsMap = {
   userId: `webhook_log.data->>'userId'`,
   event: `webhook_log.data->>'event'`,
   statusCode: `webhook_log.data->'response'->>'status'`,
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/webhook-log.ts
```diff
@@ -26,6 +26,7 @@ const requestsFieldsMap: FieldsMap = {
   userId: `webhook_log.data->>'userId'`,
   event: `webhook_log.data->>'event'`,
   statusCode: `webhook_log.data->'response'->>'status'`,
+  success: `webhook_log.data->>'success'`,
```
leszko, line 29: I'm not sure how simple it would be, but you can consider adding unit test for this case.

leszko (approved):

PR #2064, opened 2024-02-19
api: flag streams as pull created for auth header in webhook

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -472,12 +472,12 @@ components:
           description: Used to form playback URL
         playbackPolicy:
           $ref: "#/components/schemas/playback-policy"
-        isTrovoAuth:
+        isPullStream:
           type: boolean
           readOnly: true
           description: >-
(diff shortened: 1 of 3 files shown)


thomshutt (approved):

packages/api/src/schema/api-schema.yaml
```diff
@@ -472,12 +472,12 @@ components:
...
           $ref: "#/components/schemas/playback-policy"
-        isTrovoAuth:
+        isPullStream:
           type: boolean
           readOnly: true
           description: >-
-            Wether the stream is authenticated by Trovo. If true, any webhook
-            policy will include the Trovo Auth header
+            Wether the stream is a pull stream. If true, any webhook policy will
+            include the Trovo Auth header
```
victorges, line 480: We can already tell this by the existence of the `pull` field, so this is redundant. It should also be omitted from API schema not to pollute our public documentation. So you can move this to `db-schema` or just omit from the schema altogether (tho I guess this would break typescript stuff, unless you start using the `stream.pull` for it).

PR #2072, opened 2024-02-22
Add extra debug values to session re-use log

mjh1:


mjh1:
@victorges lgtu?

victorges (approved):
Lgtm

PR #2075, opened 2024-02-26
stream/helpers: reduce some delays in await functions

emranemran:


mjh1 (approved):

packages/api/src/controllers/helpers.ts
```diff
@@ -674,7 +674,7 @@ export const triggerCatalystPullStart =
               res.status
             } error=${JSON.stringify(body)}`
           );
-          await sleep(1000);
+          await sleep(250);
```
victorges, line 677: I wonder if this could have bigger risks of concurrency issues, given each of these requests might hit a different catalyst instance. Then we could have multiple instances starting a pull at the same time (not sure what errors could happen after that tho). Let's keep this in mind if we notice any issues (this has been deployed to prod earlier today)

PR #2078, opened 2024-02-27
Projects in Studio

emranemran:
Initial set of changes from myself + @mjh1 to enable "Project" environments in Studio:
* Added boiler plate schema changes for project tables
* Added project.ts to handle creation/retrieval of projects
* Updated asset.ts to use projectId logic where applicable

**A few notes:**
* I'm would like to merge changes as we go so that it's not one giant PR at the end that's difficult to review -- so we're making sure that projectId is not enforced anywhere right now so that we don't break backwards compatibility. Any records with an empty or missing projectId will get categorized as default project in the dashboard. 
* I'm opening up this PR early to collect feedback in case we're going down the wrong path. Please review the changes and suggest anything that might have been missed.
* TODO: fix/add tests...

PS: this is my first large typrscript PR so be gentle :)

### DIFF
--- a/packages/api/src/controllers/api-token.js
+++ b/packages/api/src/controllers/api-token.js
@@ -48,6 +48,7 @@ const fieldsMap = {
   name: { val: `api_token.data->>'name'`, type: "full-text" },
   lastSeen: `api_token.data->'lastSeen'`,
   userId: `api_token.data->>'userId'`,
+  projectId: `api_token.data->>'projectId'`,
   "user.email": { val: `users.data->>'email'`, type: "full-text" },
 };
 
@@ -66,6 +67,9 @@ app.get("/", async (req, res) => {
 
   if (!userId) {
     const query = parseFilters(fieldsMap, filters);
+    query.push(
+      sql`coalesce(api_token.data->>'projectId', '') = ${req.project?.id || ""}`
+    );
 
     let fields =
       " api_token.id as id, api_token.data as data, users.id as usersId, users.data as usersdata";
@@ -102,9 +106,13 @@ app.get("/", async (req, res) => {
       errors: ["user can only request information on their own tokens"],
     });
   }
-
   const query = parseFilters(fieldsMap, filters);
   query.push(sql`api_token.data->>'userId' = ${userId}`);
+  query.push(
+    req.project?.id
+      ? sql`api_token.data->>'projectId' = ${req.project.id}`
+      : sql`api_token.data->>'projectId' IS NULL OR api_token.data->>'projectId' = ''`
+  );
 
   let fields = " api_token.id as id, api_token.data as data";
   if (count) {
@@ -155,6 +163,7 @@ app.post("/", validatePost("api-token"), async (req, res) => {
   await req.store.create({
     id: id,
     userId: userId,
+    projectId: req.query.projectId?.toString(),
     kind: "api-token",
     name: req.body.name,
     access: req.body.access,
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -44,6 +44,7 @@ import {
   NewAssetPayload,
   ObjectStore,
   PlaybackPolicy,
+  Project,
   Task,
 } from "../schema/types";
 import { WithID } from "../store/types";
@@ -182,7 +183,7 @@ function parseUrlToDStorageUrl(
 }
 
 export async function validateAssetPayload(
-  req: Pick<Request, "body" | "user" | "token" | "config">,
+  req: Pick<Request, "body" | "user" | "token" | "config" | "project">,
   id: string,
   playbackId: string,
   createdAt: number,
@@ -234,6 +235,7 @@ export async function validateAssetPayload(
     name: payload.name,
     source,
     staticMp4: payload.staticMp4,
+    projectId: req.project?.id ?? "",
     creatorId: mapInputCreatorId(payload.creatorId),
     playbackPolicy,
     objectStoreId: payload.objectStoreId || (await defaultObjectStoreId(req)),
@@ -613,6 +615,7 @@ const fieldsMap = {
   creatorId: `stream.data->'creatorId'->>'value'`,
   playbackId: `asset.data->>'playbackId'`,
   playbackRecordingId: `asset.data->>'playbackRecordingId'`,
+  projectId: `asset.data->>'projectId'`,
   phase: `asset.data->'status'->>'phase'`,
   "user.email": { val: `users.data->>'email'`, type: "full-text" },
   cid: `asset.data->'storage'->'ipfs'->>'cid'`,
@@ -653,6 +656,12 @@ app.get("/", authorizer({}), async (req, res) => {
   if (!req.user.admin || !all || all === "false") {
     query.push(sql`asset.data->>'deleted' IS NULL`);
   }
+
+  query.push(
+    req.project?.id
+      ? sql`asset.data->>'projectId' = ${req.project.id}`
+      : sql`asset.data->>'projectId' IS NULL OR asset.data->>'projectId' = ''`
+  );
 
   let output: WithID<Asset>[];
   let newCursor: string;
@@ -798,7 +807,11 @@ const uploadWithUrlHandler: RequestHandler = async (req, res) => {
     url,
     encryption: assetEncryptionWithoutKey(encryption),
   });
-  const dupAsset = await db.asset.findDuplicateUrlUpload(url, req.user.id);
+  const dupAsset = await db.asset.findDuplicateUrlUpload(
+    url,
+    req.user.id,
+    req.project?.id
+  );
   if (dupAsset) {
     const [task] = await db.task.find({ outputAssetId: dupAsset.id });
     if (!task.length) {
--- a/packages/api/src/controllers/project.ts
+++ b/packages/api/src/controllers/project.ts
@@ -1,0 +1,145 @@
+import { Request, Router } from "express";
+import { authorizer, validatePost } from "../middleware";
+import { db } from "../store";
+import { v4 as uuid } from "uuid";
+import {
+  makeNextHREF,
+  parseFilters,
+  parseOrder,
+  toStringValues,
+} from "./helpers";
+import {
+  NotFoundError,
+  ForbiddenError,
+  BadRequestError,
+  BadGatewayError,
+  InternalServerError,
+} from "../store/errors";
+import sql from "sql-template-strings";
+import { WithID } from "../store/types";
+import { Project } from "../schema/types";
+
+const app = Router();
+
+const fieldsMap = {
+  id: `project.ID`,
+  name: { val: `project.data->>'name'`, type: "full-text" },
+  createdAt: { val: `project.data->'createdAt'`, type: "int" },
+  userId: `project.data->>'userId'`,
+} as const;
+
+app.get("/", authorizer({}), async (req, res) => {
+  let { limit, cursor, order, all, filters, count } = toStringValues(req.query);
+
+  if (isNaN(parseInt(limit))) {
+    limit = undefined;
+  }
+  if (!order) {
+    order = "updatedAt-true,createdAt-true";
+  }
+
+  const query = [...parseFilters(fieldsMap, filters)];
+
+  if (!req.user.admin || !all || all === "false") {
+    query.push(sql`project.data->>'deleted' IS NULL`);
+  }
+
+  let output: WithID<Project>[];
+  let newCursor: string;
+  if (req.user.admin) {
+    let fields =
+      " project.id as id, project.data as data, users.id as usersId, users.data as usersdata";
+    if (count) {
+      fields = fields + ", count(*) OVER() AS count";
+    }
+    const from = `project left join users on project.data->>'userId' = users.id`;
+    [output, newCursor] = await db.project.find(query, {
+      limit,
+      cursor,
+      fields,
+      from,
+      order: parseOrder(fieldsMap, order),
+      process: ({ data, usersdata, count: c }) => {
+        if (count) {
+          res.set("X-Total-Count", c);
+        }
+        return {
+          ...data,
+          user: db.user.cleanWriteOnlyResponse(usersdata),
+        };
+      },
+    });
+  } else {
+    query.push(sql`project.data->>'userId' = ${req.user.id}`);
+
+    let fields = " project.id as id, project.data as data";
+    if (count) {
+      fields = fields + ", count(*) OVER() AS count";
+    }
+    [output, newCursor] = await db.project.find(query, {
+      limit,
+      cursor,
+      fields,
+      order: parseOrder(fieldsMap, order),
+      process: ({ data, count: c }) => {
+        if (count) {
+          res.set("X-Total-Count", c);
+        }
+        return data;
+      },
+    });
+  }
+
+  res.status(200);
+  if (output.length > 0 && newCursor) {
+    res.links({ next: makeNextHREF(req, newCursor) });
+  }
+
+  return res.json(output);
+});
+
+app.get("/:id", authorizer({}), async (req, res) => {
+  const project = await db.project.get(req.params.id, {
+    useReplica: false,
+  });
+  console.log("YYY :", project, req.params.id);
+
+  if (!project || project.deleted) {
+    res.status(403);
+    return res.json({ errors: ["project not found"] });
+  }
+  console.log("YYY here:", project, req.params.id);
+
+  if (req.user.admin !== true && req.user.id !== project.userId) {
+    throw new ForbiddenError(
+      "user can only request information on their own projects"
+    );
+  }
+  console.log("YYY here here:", project, req.params.id);
+
+  res.json(project);
+});
+
+app.post("/", authorizer({}), async (req, res) => {
+  const { name } = req.body;
+
+  const id = uuid();
+  await db.project.create({
+    id: id,
+    name: name,
+    userId: req.user.id,
+    createdAt: Date.now(),
+  });
+
+  const project = await db.project.get(id, { useReplica: false });
+
+  if (!project) {
+    res.status(403);
+    return res.json({ errors: ["project not created"] });
+  }
+
+  res.status(201);
+  res.json(id);
+});
+
+export default app;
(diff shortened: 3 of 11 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -200,6 +202,13 @@ export async function validateAssetPayload(
     }
   }
 
+  if (payload.projectId) {
+    const project = await getProject(req);
```
leszko: I wonder if we shouldn't add `project` to `req` in https://github.com/livepeer/studio/blob/95e12f9b0136a60ee0ff5c55608c1f5bd7be3278/packages/api/src/middleware/auth.ts#L79

My thinking is that now, `authenticator()` assigns `user` to each request. But from now on, the request (api token) will be associated always with one project, so maybe we should inject it in `authenticator()` and then there would be no need to call `getProject()` everywhere?

packages/api/src/controllers/asset.ts
```diff
@@ -654,6 +674,14 @@ app.get("/", authorizer({}), async (req, res) => {
     query.push(sql`asset.data->>'deleted' IS NULL`);
   }
 
+  if (projectId) {
```
leszko: If we filter by `projectId`, do we also need to filter by `userId` in the line 683?

leszko (commented):
@emranemran @mjh1 kudos for sending the PR early. 

I didn't manage to review it all, but I added some first comments. PTAL.

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  let output: WithID<Project>[];
+  let newCursor: string;
+  if (req.user.admin) {
+    let fields =
+      " project.id as id, project.data as data, users.id as usersId, users.data as usersdata";
+    if (count) {
+      fields = fields + ", count(*) OVER() AS count";
+    }
+    const from = `project left join users on project.data->>'userId' = users.id`;
```
leszko, line 55: Do you think it would be possible to put the SQL queries inside the table? I know it's the same in other controllers, but I wonder if it would make sense to abstract it and put inside the `store/*-table.ts`?

packages/api/src/controllers/asset.ts
emranemran: @victorges and I discussed something like that but Victor suggested we need to maintain compatibility with the SDK I believe. I'm still a bit on the fence on this b/c from the dashboard perspective, project is basically a "filter" to categorize the assets, streams, etc. But maybe philosophically what you're saying makes sense. wdyt @victorges?

packages/api/src/controllers/project.ts
emranemran: We could do that - I just followed the existing patterns in other files.

packages/api/src/controllers/asset.ts
emranemran: Yes, don't we? We're trying to get all assets beloniging to a specific userId broken down by a specific projectId.

leszko: Doesn't one project belong to one user only? Then, filtering by projectId should be enough.

packages/api/src/controllers/project.ts
leszko: Ok, that's probably fine, we can try to clean it up for all controllers in some other PR.

leszko (commented):
Looks good from my end. You can decide with @victorges on the open discussions. 

Plus, please add unit tests.

packages/api/src/store/asset-table.ts
```diff
@@ -121,7 +121,8 @@ export default class AssetTable extends Table<WithID<Asset>> {
 
   async findDuplicateUrlUpload(
     url: string,
-    userId: string
+    userId: string,
+    projectId: string
```
mjh1, line 125: good spot adding this 👍

packages/api/src/controllers/asset.ts
mjh1: I don't think projectId should be its own queryparam like this, we can add it to the `fieldsMap` instead. I also think we should treat it similar to userId in that only admins are allowed to use it. For non-admins we should just take the projectId associated with the api key being used.

packages/api/src/controllers/asset.ts
mjh1: I think we can still maintain compatibility but having the projectId added to the request as Rafal suggests will just neaten the code up a bit. User and project would both exist on the request object then, and we can maintain compatibility by checking both in the controllers.

mjh1: The other thing is we don't want projectId to be part of the payload, we should be taking it from the api key being used. We could potentially leave that for a later PR though and leave this one mainly as just adding the projects controller. 
Edit: I'm working on this bit locally atm, including the `authenticator()` rafal mentioned (branched off this), let me know if you want to pair etc WIP: https://github.com/livepeer/studio/pull/2081

emranemran: @mjh1: I saw your edit comment too late and started making simliar changes to my branch - but I scrapped those and just cherry-picked yours here instead. Could we use this branch to pair for integration? I added another commit to make some edits but currently a bit broken when trying to create new api-keys -- see my comment here: https://github.com/livepeer/studio/pull/2078#issuecomment-1970374411

emranemran:
@mjh1 @leszko: a few questions:
* Let's assume a project exists in the dashboard. Now the user creates an api-key which does a POST. Now looking at `authorizer`, Bearer method is not supported for creation of new api-keys. Tried sending a jwt request in box (using params in `config/full-stack.json`) but it fails when trying to retrieve a user object. Any ideas why? 
* For the migration case when a current user logs in, the `authorizer` will get an empty or missing projectId field via `getRequest` for every GET/POST/etc. Should a default project be created on first login after we migrate?

packages/api/src/controllers/project.ts
```diff
@@ -29,9 +29,16 @@ const app = Router();
...
-  const project = await db.project.get(projectId);
-  if (!project || project.deleted) {
-    throw new NotFoundError(`project not found`);
+  let project;
+
+  if (projectId) {
+    project = await db.project.get(projectId);
+    if (!project || project.deleted) {
+      return {};
+      //throw new NotFoundError(`project not found`);
```
emranemran: Not sure yet how to handle this case yet....just returning empty object for now.

packages/api/src/controllers/asset.ts
leszko: Guys, I think you're on the right track!

Wrt what @mjh1 wrote, I think that we should not send `projectId` in the payload, because `projectId` is derived from API TOKEN. The same way, we don't send `userId` in every request.

leszko:
> * Let's assume a project exists in the dashboard. Now the user creates an api-key which does a POST. Now looking at `authorizer`, Bearer method is not supported for creation of new api-keys. Tried sending a jwt request in box (using params in `config/full-stack.json`) but it fails when trying to retrieve a user object. Any ideas why?

Yeah, I think it make sense, that Bearer is not supported for creating api tokens, because you always need to first create them from the livepeer dashboard (read: jwt). Why doesn't it work in the box? No idea, unfortunately 😞 

> * For the migration case when a current user logs in, the `authorizer` will get an empty or missing projectId field via `getRequest` for every GET/POST/etc. Should a default project be created on first login after we migrate?

This is something we started to discuss at some point. And I think we still have these 3 options:
- Option 1: Add `projectId` for all users with the DB migration (then you don't have the case you described)
- Option 2: Tweak it in the `authorizer()`, so in the authorizer, if a user does not have a project, then return some default project
- Option 3: What you described, so create a project on-a-fly

I think that we discussed that the Option 3 probably makes the least sense. But you know better the code now, so hard to suggest what's the best.

packages/api/src/controllers/asset.ts
mjh1: Ok cool yep let's just use this branch

packages/api/src/middleware/auth.ts
```diff
@@ -26,7 +27,7 @@ export const EMAIL_VERIFICATION_CUTOFF_DATE = 1695765600000; // 2023-09-26T22:00
  * are set to expire after a short time and the client manages using a refresh
  * token for keeping the user logged in.
  */
-export const NEVER_EXPIRING_JWT_CUTOFF_DATE = 1709251200000; // 2024-03-01T00:00:00.000Z
+export const NEVER_EXPIRING_JWT_CUTOFF_DATE = 1740820816000; // 2024-03-01T00:00:00.000Z
```
emranemran: note: just a HACK for now since I wasn't sure why we have this cutoff date.

emranemran: cc @victorges @leszko

leszko: AFAIK you don't need to change it. It was only a cutoff for the "old" JWT mechanism, and we're using now the "new" JWT mechanism. But Victor knows best since he updated "old" => "new" :)

victorges: Yeah this doesn't need to change! Here's the doc on how the migration works: https://www.notion.so/livepeer/Studio-dashboard-authentication-310a74802de54f538e0ecb633f94ee8e?pvs=4

packages/api/src/controllers/api-token.js
```diff
@@ -66,6 +67,11 @@ app.get("/", async (req, res) => {
 
   if (!userId) {
     const query = parseFilters(fieldsMap, filters);
+    query.push(
+      req.project?.id
+        ? sql`api_token.data->>'projectId' = ${req.project.id}`
+        : sql`api_token.data->>'projectId' IS NULL OR api_token.data->>'projectId' = ''`
```
victorges: if you want parity with the ts checks I suggeted, I think these queries could also be:
```suggestion
      sql`coalesce(api_token.data->>'projectId', '') = ${req.project?.id || ''}`
```

packages/api/src/controllers/asset.ts
```diff
@@ -182,7 +184,7 @@ function parseUrlToDStorageUrl(
 }
 
 export async function validateAssetPayload(
-  req: Pick<Request, "body" | "user" | "token" | "config">,
+  req: Pick<Request, "body" | "user" | "token" | "config" | "project">,
```
victorges, line 186: Shouldn't this be `projectId`?

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,163 @@
...
+  NotFoundError,
+  BadRequestError,
+  InternalServerError,
+  UnauthorizedError,
+  NotImplementedError,
+} from "../store/errors";
+
+const app = Router();
+
+export async function getProject(req, projectId) {
```
victorges: Remember to add types. TypeScript will also help you get around!
```suggestion
export async function getProject(req: Request, projectId: string) {
```

packages/api/src/controllers/project.ts
victorges: Why doesn't the throwing of `NotFoundError` work?

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,163 @@
...
+  let project;
+
+  if (!req.user.admin && req.user.id !== project.userId) {
```
victorges: Is this function used? Looks like it will always get a null reference exception here since `project` is unassigned. Overall, looks like its implementation can be reduced to
```ts
const project = projectId ? await db.project.get(projectId) : { name: 'default', userId: req.user.id }
if (!req.user.admin && req.user.id !== project.userId) {
   throw new ForbiddenError(`invalid user`);
}
```
WDYT?

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,163 @@
...
+  res.status(200);
+  if (output.length > 0 && newCursor) {
+    res.links({ next: makeNextHREF(req, newCursor) });
+  }
+
+  return res.json(output);
+});
+
+app.get("/:projectId", authorizer({}), async (req, res) => {
+  const project = await req.project;
```
victorges: Is the project on the request a promise? I think it's not awaited most of the time

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,163 @@
...
+    name: name,
+    userId: req.user.id,
+    createdAt: Date.now(),
+  });
+  res.status(201);
+
+  const project = await db.project.get(id, { useReplica: false });
+
+  if (!project) {
+    res.status(403);
```
victorges, line 137: Probably a 500 makes more sense (I don't see this ever happening anyway, tho I guess I've seen this pattern elsewhere)

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,163 @@
...
+  const { name } = req.body;
+
+  const id = uuid();
+  await db.project.create({
+    id: id,
+    name: name,
+    userId: req.user.id,
+    createdAt: Date.now(),
+  });
+  res.status(201);
```
victorges: This line can be removed

packages/api/src/middleware/auth.ts
```diff
@@ -102,14 +105,17 @@ function authenticator(): RequestHandler {
       }
 
       userId = tokenObject.userId;
+      projectId = tokenObject.projectId;
       // track last seen
       tracking.recordToken(tokenObject);
     } else if (authScheme === "jwt") {
+      // TODO parse projectID for jwt auth
```
victorges: already done!

packages/api/src/middleware/auth.ts
```diff
@@ -149,6 +155,9 @@ function authenticator(): RequestHandler {
     req.token = tokenObject;
     req.user = user;
 
+    project = await getProject(req, projectId);
```
victorges: This probably deserves some caching like user and api key above

packages/api/src/middleware/auth.ts
```diff
@@ -149,6 +155,9 @@ function authenticator(): RequestHandler {
...
+    req.project = project;
```
victorges: Can you add this `project` field to our `Request` type extensions [here](https://github.com/livepeer/livepeer-com/blob/b19fa102e2cae6389ed755bfdd478b4fe4d8fe21/packages/api/src/types/common.d.ts#L44)?

packages/api/src/schema/api-schema.yaml
```diff
@@ -1153,6 +1172,10 @@ components:
             Name of the asset. This is not necessarily the filename, can be a
             custom name or title
           example: filename.mp4
+        projectId:
```
victorges, line 1179: Seems like we still miss a _bunch_ of these. They are probably `readOnly` too 🤔

victorges (commented):
preliminary LGTM! Still missing `projectId` on a ton of resources and checks on the APIs

packages/api/src/controllers/asset.ts
mjh1: I was thinking `project` here because we pull the project from the db and add the whole thing to the request like we do for the user object

packages/api/src/middleware/auth.ts
emranemran: Already have it here actually: 
https://github.com/livepeer/studio/pull/2078/files#diff-2ce1dfa11aee18127c26c8ff761f0ca88d79f008f9ec5a4c10c3bc39d8cd489bR45

Did you mean this or something else?

packages/api/src/controllers/project.ts
emranemran: good catch, removed.

packages/api/src/controllers/project.ts
emranemran: Good point - the function is used and it seemed to work even though project was left unassigned. But I see your point though - updated to use your suggestion as it's simpler and effectively does the same.

packages/api/src/controllers/project.ts
emranemran: Done

packages/api/src/controllers/asset.ts
emranemran: `project` here makes sense I think?

packages/api/src/controllers/api-token.js
emranemran: OK updated to your suggestion and will test to verify it's the same behavior.

emranemran:
> preliminary LGTM! Still missing projectId on a ton of resources and checks on the APIs

So the plan with this PR is to get it ready and merged w/o breaking existing dashboard usage so that @suhailkakar can get started on frontend work. Then follow-up PRs will be used for streams/webhooks/etc and update those code paths for projectId handling.

packages/api/src/middleware/auth.ts
victorges: Oh yeah I meant that. Not sure how i missed it

packages/api/src/controllers/asset.ts
victorges: Ohh I thought this was getting an asset, but it's the request object. `project` is the right thing yeah

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,145 @@
...
+  const project = await db.project.get(req.params.id, {
+    useReplica: false,
+  });
+  console.log("YYY :", project, req.params.id);
+
+  if (!project || project.deleted) {
+    res.status(403);
+    return res.json({ errors: ["project not found"] });
+  }
+  console.log("YYY here:", project, req.params.id);
```
leszko, line 111: leftover?

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  return res.json(output);
+});
+
+app.get("/:id", authorizer({}), async (req, res) => {
+  const project = await db.project.get(req.params.id, {
+    useReplica: false,
+  });
+  console.log("YYY :", project, req.params.id);
```
leszko, line 105: leftover?

packages/api/src/controllers/project.ts
```diff
@@ -0,0 +1,145 @@
...
+    return res.json({ errors: ["project not found"] });
+  }
+  console.log("YYY here:", project, req.params.id);
+
+  if (req.user.admin !== true && req.user.id !== project.userId) {
+    throw new ForbiddenError(
+      "user can only request information on their own projects"
+    );
+  }
+  console.log("YYY here here:", project, req.params.id);
```
leszko, line 118: leftover?

leszko (approved):
LGTM

PR #2080, opened 2024-02-28
fix: moved upload via URL to separate API schema and added docs

0xcadams:


packages/api/src/schema/api-schema.yaml
```diff
@@ -286,20 +286,36 @@ components:
...
         name:
           type: string
-          description: Name of the clip
+          description: >-
+            The optional friendly name of the clip to create.
         sessionId:
           type: string
-          description: Session ID of the stream to clip
+          description: >-
+            The optional session ID of the stream to clip. This can be used to
```
victorges, line 313: Maybe worth explaining the relationship between this and playbackId. I guess that if you dont provide this it clips the stream otherwise it clips a session? And it could alternatively be referenced by the session recording asset playbackId as well? (I don't know, these are actually questions about how it works too haha)

packages/api/src/schema/api-schema.yaml
```diff
@@ -1154,8 +1185,70 @@ components:
...
+          example: filename.mp4
+        staticMp4:
+          type: boolean
+          description: Whether to generate MP4s for the asset.
+          example: true
+        playbackPolicy:
+          $ref: "#/components/schemas/playback-policy"
+        creatorId:
+          $ref: "#/components/schemas/input-creator-id"
+        storage:
```
victorges: Is this the same field as `new-asset-from-url-payload`? If so, we should avoid repeating it and use a `$ref` instead (maybe `new-asset-from-url-payload` references this one, since it's the more specific one)

packages/api/src/schema/api-schema.yaml
```diff
@@ -1154,8 +1185,70 @@ components:
...
+                  additionalProperties: false
+                  properties:
+                    spec:
+                      oneOf:
+                        - type: "null"
+                        - $ref: >-
+                            #/components/schemas/asset/properties/storage/properties/ipfs/properties/spec
+                - type: boolean
+                - type: "null"
+        encryption:
```
victorges, line 1216: same re ref

packages/api/src/schema/api-schema.yaml
```diff
@@ -1154,8 +1185,70 @@ components:
...
+          required:
+            - encryptedKey
+          properties:
+            encryptedKey:
+              type: string
+              writeOnly: true
+              description: >-
+                Encryption key used to encrypt the asset. Only writable in the
+                upload asset endpoints and cannot be retrieved back.
+        c2pa:
```
victorges: same I guess tho this is much smaller. Good to avoid repeating docs at lesat

victorges (approved):
Neat! Thanks for this

packages/api/src/schema/api-schema.yaml
0xcadams: Haha yeah let me get feedback from @emranemran in here as well.

packages/api/src/schema/api-schema.yaml
0xcadams: Yeah, agreed, I will ref these out

packages/api/src/schema/api-schema.yaml
```diff
@@ -286,20 +286,36 @@ components:
...
+            the ClipTrigger in the UI Kit for an example of how this is
+            calculated (for HLS, it uses `Program Date-Time` tags, and for
+            WebRTC, it uses the latency from server to client at stream
+            startup)._
         endTime:
           type: number
-          description: End time of the clip in UNIX timestamp (in milliseconds)
+          description: >-
+            The end timestamp of the clip in Unix milliseconds. This is the
+            **server time** - not relative time to the start of the stream. _See
```
emranemran: Does this description make sense for the end time? I'm thinking server term terminology will be confusing for most users. Should we just leave out the server time bit from both startTime and endTime?

packages/api/src/schema/api-schema.yaml
emranemran: Yeah maybe we should clarify here that sessionId should be used to clip recordings otherwise it will clip the livestream?

packages/api/src/schema/api-schema.yaml
```diff
@@ -859,12 +883,18 @@ components:
           type: string
           example: >-
             https://livepeercdn.com/asset/ea03f37e-f861-4cdd-b495-0e60b6d753ad/index.m3u8
-          description: URL for HLS playback
+          description: >-
+            URL for HLS playback. **It is recommended to not use this URL**, and
+            instead use playback IDs with the Playback Info endpoint to retrieve
+            the playback URLs - this URL format is subject to change.
```
emranemran: An example of the URL would be useful here.

packages/api/src/schema/api-schema.yaml
0xcadams: Yeah, I understand what you mean - I mean, it _is_ the server time though, right? It seems like it's hard to explain. If we say Unix time, doesn't that mean we're constantly explaining the latency for HLS vs WebRTC?

I can remove that verbiage though. Maybe this is resolved with a better interface and not documentation.

0xcadams: I removed that verbiage and we can reconsider when we come back to clipping.

packages/api/src/schema/api-schema.yaml
0xcadams: Done

PR #2084, opened 2024-02-29
Stream pull startup improvements

mjh1:


packages/api/src/controllers/stream.ts
```diff
@@ -1093,11 +1093,14 @@ app.put(
       }
     }
 
-    const ingest = await getIngestBase(req);
-    await triggerCatalystPullStart(stream, getHLSPlaybackUrl(ingest, stream));
+    if (!stream.isActive) {
+      const ingest = await getIngestBase(req);
+      // TODO run in background
```
mjh1: @thomshutt @victorges not sure if it's worth it but maybe we should run the manifest call in the background in case the stream becomes active whilst we're still waiting for a response there..

thomshutt: Good idea. Can we also wait for other states besides Active? I think that as long as we're confident it's in a state that'll become active (e.g Pending) then that's probably okay

mjh1: Oh yeah good plan

mjh1: Ah no, just seen that the 'pending' label you see in studio dashboard comes after the stream is active so that doesn't help.

thomshutt (approved):

packages/api/src/controllers/stream.ts
```diff
@@ -1093,11 +1093,22 @@ app.put(
       }
     }
 
-    const ingest = await getIngestBase(req);
-    await triggerCatalystPullStart(stream, getHLSPlaybackUrl(ingest, stream));
+    if (!stream.isActive) {
```
victorges: Might need some logic in case the stream was just nuked above due to pull source changing. This isActive will certainly be true since it was fetched before the nuke above

victorges (approved):
LGTM

packages/api/src/controllers/stream.ts
mjh1: Ah yeah hadn't fully realised the significance, good catch

packages/api/src/controllers/stream.ts
victorges: yeah pending is just "active but waiting for stream health data"

PR #2085, opened 2024-02-29
Always log response body for pull streams

thomshutt:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -114,13 +114,19 @@ async function fireGateWebhook(
     console.log(
       `access-control: gate: webhook=${
         webhook.id
-      } statusCode=${statusCode} respSpanId=${resp.headers.get("X-Tlive-Spanid")} respBody=${Buffer.from(respBody).toString("base64")} duration=${process.hrtime(startTime)[1] / 1e6}ms`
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -106,14 +107,22 @@ async function fireGateWebhook(
...
-  console.log(
-    `access-control: gate: webhook=${
-      webhook.id
-    } statusCode=${statusCode} duration=${process.hrtime(startTime)[1] / 1e6}ms`
-  );
+  if ("pull" in content && content.pull) {
+    console.log(
+      `access-control: gate: webhook=${
+        webhook.id
+      } statusCode=${statusCode} respSpanId=${resp.headers.get("X-Tlive-Spanid")} respBody=${Buffer.from(respBody).toString("base64")} duration=${process.hrtime(startTime)[1] / 1e6}ms`
```
gioelecerati: I think this `resp` could be empty if the try catch fails, it's worth checking or do `resp?.`

gioelecerati (approved):
LGTM - tests are failing due to missing `yarn prettier`

packages/api/src/controllers/access-control.ts
thomshutt: Thanks, fixed both of those

PR #2086, opened 2024-03-01
Pull stream start improvements

mjh1:


thomshutt (approved):

packages/api/src/controllers/helpers.ts
```diff
@@ -661,12 +661,16 @@ export const triggerCatalystPullStart =
             timeout: PULL_START_TIMEOUT,
             maxRedirects: 10,
           });
-          const body = await res.text();
-          const isHlsErr = body.includes("#EXT-X-ERROR: Stream open failed");
+          const errHeader = res.headers.get("error");
```
victorges, line 664: ah nice

victorges (approved):
lgtm anyway

victorges (commented):
test

packages/api/src/controllers/stream.ts
```diff
@@ -1097,20 +1068,7 @@ app.put(
...
-          );
-        } catch (err) {
-          logger.error("Error triggering catalyst pull start err=", err);
-        }
-      });
-
-      if (waitActive === "true") {
-        stream = await pollWaitStreamActive(req, stream.id);
-      }
+      await triggerCatalystPullStart(stream, getHLSPlaybackUrl(ingest, stream));
```
victorges, line 1071: so in practice we always wait for the stream to start, is that right?

mjh1: yes if the manifest request has returned then the stream has started and ready for playback

PR #2087, opened 2024-03-04
Allow for 200 access denied for webhook access control

mjh1:
It doesn't look like there are any existing tests for webhook access control, I'll get some added as a follow up to this if that's ok. Tested manually for now.

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -83,6 +83,7 @@ async function fireGateWebhook(
   let respBody: string;
   let errorMessage: string;
   let statusCode: number;
+  let bodyStatusCode: number;
 
   try {
     resp = await fetchWithTimeoutAndRedirects(webhook.url, params);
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -132,6 +138,20 @@ async function fireGateWebhook(
   return statusCode;
 }
 
+function checkRespBody(resp: string): number | null {
+  if (!resp || resp == "") {
+    return;
+  }
+  try {
+    const body = JSON.parse(resp);
+    return body.ret;
```
thomshutt, line 146: I'm assuming Typescript is robust to this, but does it handle the auth server returning a blob of JSON without this field?

packages/api/src/controllers/access-control.ts
```diff
@@ -88,11 +88,21 @@ async function fireGateWebhook(
...
+      const respCode = checkRespBody(respBody);
+      if (respCode) {
+        statusCode = respCode;
+      }
+    }
+
+    if (statusCode < 200 || statusCode >= 300) {
       console.error(
-        `access-control: gate: webhook=${webhook.id} got response status != 2XX statusCode=${resp.status}`
+        `access-control: gate: webhook=${webhook.id} got response status != 2XX statusCode=${statusCode}`
```
thomshutt: It'll mess up your nice neat code, but it might be best to log the real response code and the body one, since they have different meanings / causes

packages/api/src/controllers/access-control.ts
mjh1: Yep, tested that, it returns undefined or null and the `if(respCode)` covers it

thomshutt (approved):

PR #2092, opened 2024-03-08
api: asset: delete & restore

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -1156,6 +1156,29 @@ app.put("/:id/restore", authorizer({}), async (req, res) => {
   res.end();
 });
 
+app.patch("/:id/deleted", authorizer({ anyAdmin: true }), async (req, res) => {
+  const { id } = req.params;
+  const asset = await db.asset.get(id);
+
+  if (!asset) {
+    throw new NotFoundError(`Asset not found`);
+  }
+
+  if (!(asset.status.phase === "deleting")) {
+    throw new BadRequestError(`Asset is not in a deleting phase`);
+  }
+
+  await db.asset.update(asset.id, {
+    status: {
+      phase: "deleted",
+      updatedAt: Date.now(),
+    },
+  });
+
+  res.status(204);
+  res.end();
+});
+
 app.delete("/", authorizer({ anyAdmin: true }), async (req, res) => {
   if (req.query.userId && req.query.userId !== req.user.id && !req.user.admin) {
     throw new ForbiddenError(`users may only delete their own assets`);
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -674,6 +674,16 @@ app.get("/", authorizer({}), async (req, res) => {
   query.push(
     sql`coalesce(asset.data->>'projectId', '') = ${req.project?.id || ""}`
   );
+  if (req.user.admin && deleting) {
+    const deletionThreshold = new Date(
+      Date.now() - DELETE_ASSET_DELAY
+    ).toISOString();
+
+    query.push(sql`asset.data->'status'->>'phase' = 'deleting'`);
+    query.push(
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -706,6 +719,14 @@ app.get("/", authorizer({}), async (req, res) => {
     res.links({ next: makeNextHREF(req, newCursor) });
   }
 
+  if (req.user.admin && deleting) {
+    output = output.filter((asset) => asset.status.phase == "deleting");
+    output = output.filter(
+      (asset) =>
+        asset.deletedAt && Date.now() > asset.deletedAt + DELETE_ASSET_DELAY
+    );
+  }
```
victorges: **Blocking:** We must filter assets on the SQL query instead, [like this](https://github.com/livepeer/studio/pull/2092/files#diff-ed7a14d016bf3f1e0d94b5823604fde2f1dba33ce1376c00c1237dfa3ed3775bL647), not filtering the DB response in memory.

packages/api/src/controllers/asset.ts
```diff
@@ -1089,11 +1110,64 @@ app.delete("/:id", authorizer({}), async (req, res) => {
...
+  if (asset.status.phase === "deleting" || asset.deleted) {
+    throw new BadRequestError(`Asset is already deleted`);
+  }
+
   await req.taskScheduler.deleteAsset(asset);
   res.status(204);
   res.end();
 });
 
+app.put("/:id/restore", authorizer({}), async (req, res) => {
```
victorges: I believe this should be a `POST` which AFAIK is the general fallback in REST for an "action" on a resource, which is what being triggered here.

packages/api/src/controllers/asset.ts
```diff
@@ -1089,11 +1110,64 @@ app.delete("/:id", authorizer({}), async (req, res) => {
...
+  const { id } = req.params;
+  const asset = await db.asset.get(id);
+
+  if (!asset) {
+    throw new NotFoundError(`Asset not found`);
```
victorges: I believe our errors are generally `lower case`, let's keep it consistent (other occurrences in this PR)

packages/api/src/task/scheduler.ts
```diff
@@ -539,10 +539,44 @@ export class TaskScheduler {
       if (typeof asset === "string") {
         asset = await db.asset.get(asset);
       }
+
+      let phase = asset.status?.phase;
+      // Prevent bump of updatedAt if phase isn't getting updated
+      let updatedAt = asset.status?.updatedAt;
+
+      if (phase == "ready") {
```
victorges: You gotta remember to use triple equals in JS, double equal can do weird stuff sometimes. As a rule of thumb, **always** use triple equals unless you have a reason not to.
```suggestion
      if (phase === "ready") {
```

packages/api/src/task/scheduler.ts
```diff
@@ -539,10 +539,44 @@ export class TaskScheduler {
...
+        // If the asset is ready, we need to schedule deletion
```
victorges, line 548: I wonder if we want to trigger this for failed assets too. Maybe there are left overs?

packages/api/src/task/scheduler.ts
```diff
@@ -539,10 +539,44 @@ export class TaskScheduler {
...
+    }
+  }
+
+  async restoreAsset(asset: string | Asset) {
+    try {
+      if (typeof asset === "string") {
+        asset = await db.asset.get(asset);
+      }
+
+      await this.updateAsset(asset, {
```
victorges, line 573: Will this send any webhooks as well?

victorges (changes requested):
Requesting changes because filtering assets in memory is a no go. The other comments are optional.

packages/api/src/task/scheduler.ts
gioelecerati: we could send an asset.restored event, but I don't think we need it
restoring a deleted asset is potentially a rare occasion

victorges (approved):
LGTM

PR #2095, opened 2024-03-11
Call stop sessions rather than nuke for stream pull

mjh1:


packages/api/src/controllers/stream.ts
```diff
@@ -1060,13 +1060,11 @@ app.put(
       // read from DB again to keep exactly what got saved
       stream = await db.stream.get(stream.id, { useReplica: false });
 
-      if (oldStream.pull?.source != stream.pull?.source) {
-        await triggerCatalystStreamNuke(req, stream.playbackId);
-        streamNuked = true;
-      }
+      await triggerCatalystStreamUpdated(req, stream.playbackId);
+      await triggerCatalystStreamStopSessions(req, stream.playbackId);
```
mjh1, line 1064: @thomshutt are you still feeling it makes sense to stop the sessions on every call to /pull? I'm thinking it would also be better this way to have the consistency, makes debugging easier.

thomshutt: Yep, I think so

thomshutt (approved):

PR #2102, opened 2024-03-13
projects: add projects and projectId scoping to assets/api-token cotrollers

emranemran:
* Reverting the reverted PR
* Added one commit on top to fix the issues with incorrect queries. Previously, records with a missing projectId was being selected. The updated query handles this case correctly.


victorges (approved):
Reviewed only the commit after the revert. LGTM

mjh1 (approved):
neat 👍

PR #2103, opened 2024-03-13
add projectId scoping to streams/sessions/webhook

emranemran:
* Adding projectId scoping for stream/session/webhook.ts. 
* TODO: add/fix tests

### DIFF
--- a/packages/api/src/controllers/session.ts
+++ b/packages/api/src/controllers/session.ts
@@ -151,8 +151,7 @@ app.get("/:id", authorizer({}), async (req, res) => {
   let session = await db.session.get(req.params.id);
   if (
     !session ||
-    (hasAccessToResource(req, session) &&
-      !req.user.admin &&
+    (!hasAccessToResource(req, session) &&
       !LVPR_SDK_EMAILS.includes(req.user.email))
   ) {
     // do not reveal that session exists
--- a/packages/api/src/controllers/session.ts
+++ b/packages/api/src/controllers/session.ts
@@ -2,7 +2,7 @@ import { Router, Request } from "express";
 import sql from "sql-template-strings";
 
 import { authorizer } from "../middleware";
-import { User } from "../schema/types";
+import { User, Project } from "../schema/types";
 import { db } from "../store";
 import { DBSession } from "../store/session-table";
 import { pathJoin } from "../controllers/helpers";
@@ -36,6 +36,7 @@ const fieldsMap: FieldsMap = {
   userId: `session.data->>'userId'`,
   "user.email": { val: `users.data->>'email'`, type: "full-text" },
   parentId: `session.data->>'parentId'`,
+  projectId: `session.data->>'projectId'`,
   record: { val: `session.data->'record'`, type: "boolean" },
   sourceSegments: `session.data->'sourceSegments'`,
   transcodedSegments: {
@@ -65,10 +66,14 @@ app.get("/", authorizer({}), async (req, res, next) => {
     limit = undefined;
   }
 
+  const query = parseFilters(fieldsMap, filters);
+
   if (!req.user.admin) {
     userId = req.user.id;
+    query.push(
+      sql`coalesce(session.data->>'projectId', '') = ${req.project?.id || ""}`
+    );
   }
-  const query = parseFilters(fieldsMap, filters);
   if (!all || all === "false" || !req.user.admin) {
     query.push(sql`(session.data->>'deleted')::boolean IS false`);
   }
@@ -146,7 +151,9 @@ app.get("/:id", authorizer({}), async (req, res) => {
   let session = await db.session.get(req.params.id);
   if (
     !session ||
-    ((session.userId !== req.user.id || session.deleted) &&
+    ((session.userId !== req.user.id ||
+      (session.projectId ?? "") !== (req.project?.id ?? "") ||
+      session.deleted) &&
       !req.user.admin &&
       !LVPR_SDK_EMAILS.includes(req.user.email))
   ) {
--- a/packages/api/src/controllers/session.ts
+++ b/packages/api/src/controllers/session.ts
@@ -1,7 +1,7 @@
 import { Router, Request } from "express";
 import sql from "sql-template-strings";
 
-import { authorizer } from "../middleware";
+import { authorizer, hasAccessToResource } from "../middleware";
 import { User, Project } from "../schema/types";
 import { db } from "../store";
 import { DBSession } from "../store/session-table";
@@ -151,9 +151,7 @@ app.get("/:id", authorizer({}), async (req, res) => {
   let session = await db.session.get(req.params.id);
   if (
     !session ||
-    ((session.userId !== req.user.id ||
-      (session.projectId ?? "") !== (req.project?.id ?? "") ||
-      session.deleted) &&
+    (hasAccessToResource(req, session) &&
       !req.user.admin &&
       !LVPR_SDK_EMAILS.includes(req.user.email))
   ) {
--- a/packages/api/src/controllers/signing-key.test.ts
+++ b/packages/api/src/controllers/signing-key.test.ts
@@ -99,9 +99,7 @@ describe("controllers/signing-key", () => {
     });
 
     it("should list all user signing keys", async () => {
-      const res = await client.get(
-        `/access-control/signing-key?projectId=${projectId}`
-      );
+      const res = await client.get(`/access-control/signing-key`);
       expect(res.status).toBe(200);
     });
 
--- a/packages/api/src/controllers/signing-key.test.ts
+++ b/packages/api/src/controllers/signing-key.test.ts
@@ -9,6 +9,7 @@ import { SigningKey, SigningKeyResponsePayload, User } from "../schema/types";
 import { WithID } from "../store/types";
 import jwt, { JsonWebTokenError, JwtPayload } from "jsonwebtoken";
 import { db } from "../store";
+import { createProject } from "../test-helpers";
 
 // includes auth file tests
 
@@ -48,6 +49,7 @@ describe("controllers/signing-key", () => {
     let samplePrivateKey: string;
     let decodedPublicKey: string;
     let otherPublicKey: string;
+    let projectId: string;
 
     beforeEach(async () => {
       ({ client, adminUser, adminToken, nonAdminUser, nonAdminToken } =
@@ -69,12 +71,18 @@ describe("controllers/signing-key", () => {
         otherSigningKey.publicKey,
         "base64"
       ).toString();
+
+      let project = await createProject(client);
+      projectId = project.id;
+      expect(projectId).toBeDefined();
     });
 
     it("should create a signing key and display the private key only on creation", async () => {
       const preCreationTime = Date.now();
-      let res = await client.post("/access-control/signing-key");
+      let res = await client.post("/access-control/signing-key", { projectId });
+      let resWithoutProject = await client.post("/access-control/signing-key");
       expect(res.status).toBe(201);
+      expect(resWithoutProject.status).toBe(201);
       const created = (await res.json()) as SigningKeyResponsePayload;
       expect(created).toMatchObject({
         id: expect.any(String),
@@ -91,7 +99,9 @@ describe("controllers/signing-key", () => {
     });
 
     it("should list all user signing keys", async () => {
-      const res = await client.get(`/access-control/signing-key`);
+      const res = await client.get(
+        `/access-control/signing-key?projectId=${projectId}`
+      );
       expect(res.status).toBe(200);
     });
 
--- a/packages/api/src/controllers/signing-key.test.ts
+++ b/packages/api/src/controllers/signing-key.test.ts
@@ -2,10 +2,16 @@ import serverPromise, { TestServer } from "../test-server";
 import {
   TestClient,
   clearDatabase,
+  createApiToken,
   setupUsers,
   verifyJwt,
 } from "../test-helpers";
-import { SigningKey, SigningKeyResponsePayload, User } from "../schema/types";
+import {
+  ApiToken,
+  SigningKey,
+  SigningKeyResponsePayload,
+  User,
+} from "../schema/types";
 import { WithID } from "../store/types";
 import jwt, { JsonWebTokenError, JwtPayload } from "jsonwebtoken";
 import { db } from "../store";
@@ -50,6 +56,7 @@ describe("controllers/signing-key", () => {
     let decodedPublicKey: string;
     let otherPublicKey: string;
     let projectId: string;
+    let apiKeyWithProject: WithID<ApiToken>;
 
     beforeEach(async () => {
       ({ client, adminUser, adminToken, nonAdminUser, nonAdminToken } =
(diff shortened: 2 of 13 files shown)


mjh1 (approved):

leszko (approved):
Would it maybe be possible to add some unit tests?

packages/api/src/controllers/session.ts
```diff
@@ -78,6 +79,9 @@ app.get("/", authorizer({}), async (req, res, next) => {
   if (parentId) {
     query.push(sql`session.data->>'parentId' = ${parentId}`);
   }
+  query.push(
+    sql`coalesce(session.data->>'projectId', '') = ${req.project?.id || ""}`
+  );
```
victorges: We probably don't want to include this when the caller is an admin and has provided the `all` qs (L73), or when they've provided an `userId` qs (some weird condition chain here, might wanna rewrite)

packages/api/src/controllers/session.ts
victorges, line 1: Hmm I find it weird that there are only 2 places in this file that needed project awareness? Is this the real life?

packages/api/src/controllers/stream.ts
```diff
@@ -388,6 +390,9 @@ app.get("/", authorizer({}), async (req, res) => {
   if (userId) {
     query.push(sql`stream.data->>'userId' = ${userId}`);
   }
+  query.push(
```
victorges: Similar for admin queries

packages/api/src/controllers/stream.ts
```diff
@@ -700,7 +708,8 @@ app.get("/:id", authorizer({}), async (req, res) => {
   let stream = await db.stream.get(req.params.id);
   if (
     !stream ||
-    ((stream.userId !== req.user.id || stream.deleted) && !req.user.admin)
+    ((stream.userId !== req.user.id || stream.deleted) && !req.user.admin) ||
+    stream.projectId !== req.project?.id
```
victorges: I think the `admin` checks should override the project Id checks, just like the override the userId check today. This applies to several other places in the PR. This would look like
```ts
((stream.userId !== req.user.id || stream.projectId !== req.project?.id || stream.deleted) && !req.user.admin)
```

packages/api/src/controllers/stream.ts
```diff
@@ -626,7 +632,8 @@ app.get("/sessions/:parentId", authorizer({}), async (req, res) => {
   if (
     !stream ||
     stream.deleted ||
-    (stream.userId !== req.user.id && !req.isUIAdmin)
+    (stream.userId !== req.user.id && !req.isUIAdmin) ||
+    stream.projectId !== req.project?.id
```
victorges: Aren't we missing some coalesce logic here? If the stream has an empty project id, we might compare `"" !== undefined` here which is true even though they should match. This is why I had suggested a similar logic to what we have in the SQL filters:
```suggestion
    stream.projectId ?? "" !== req.project?.id ?? ""
```
(applies to other similar checks in the PR)

Unless you think we will never have empty strings saved in the `projectId` in the DB, then we might be safe just comparing them raw everywhere.

packages/api/src/controllers/stream.ts
```diff
@@ -743,7 +752,8 @@ app.get("/playback/:playbackId", authorizer({}), async (req, res) => {
   });
   if (
     !stream ||
-    ((stream.userId !== req.user.id || stream.deleted) && !req.user.admin)
+    ((stream.userId !== req.user.id || stream.deleted) && !req.user.admin) ||
+    stream.projectId !== req.project?.id
```
victorges: I would looooove to have that helper that checks all this on a resource and we don't have to keep repeating it 😭

packages/api/src/controllers/stream.ts
```diff
@@ -1625,7 +1650,7 @@ app.patch(
 
     const exists = stream && !stream.deleted;
     const hasAccess = stream?.userId === req.user.id || req.isUIAdmin;
-    if (!exists || !hasAccess) {
+    if (!exists || !hasAccess || stream.projectId !== req.project?.id) {
```
victorges: **nit:** `hasAccess` should include the project check, by definition

packages/api/src/controllers/stream.ts
```diff
@@ -1836,7 +1864,12 @@ app.get("/:id/config", authorizer({ anyAdmin: true }), async (req, res) => {
   let stream = await db.stream.getByPlaybackId(id, {
     useReplica: false,
   });
-  if (!stream || stream.deleted || stream.suspended) {
+  if (
+    !stream ||
+    stream.deleted ||
+    stream.suspended ||
+    stream.projectId !== req.project?.id
```
victorges: Maybe a general concern you should have with these updates is regarding **admin** APIs. This is an example of such API, which is called internally (maybe by catalyst) with an admin API key which has cross-user access and thus should have cross-project access as well.

This means that we shouldn't have checks like this on most APIs that have
`authorizer({ anyAdmin: true })`
or
`authorizer({ admin: true })` (admin-restricted but only with a JWT (dashboard access))

There are also functions that have special logic in them for admin queries, which is the case of the List APIs where I made some comments about this too.

TLDR is: On this API you should have no `projectId` check. An admin API key should have access to any resource. There are likely other places where this is happening too, so you need to double check the conditions you change to see if they are breaking an admin API or an admin feature within an API.

victorges (changes requested):
Requesting changes due to the admin APIs/queries being potentially broken. Anything else is nit/optional.

packages/api/src/controllers/stream.ts
victorges: In general, I think the (naive) mindset should be: "if there's a userId check, there should be a projectId check":
- Here, there are no user ID checks, so there wasn't a need for a projectId check either.
- On places where we check userId but allow admins to override it, there should be a project Id check but admin check should also override it.
- On queries where we have conditionals on how to include the `userId` (normally around admin access), we should replicate those when including the `projectId` as well.

So it's like the project and user checks should happen on the same "logical level". I'm not sure if this will hold everywhere in the code tho, as I said it's like a naive mindset without actually checking all usages out there =P

packages/api/src/controllers/session.ts
emranemran: Yeah this file is actually pretty short. Most of the session logic happens in streams.ts actually

packages/api/src/controllers/session.ts
emranemran: Fixed

packages/api/src/controllers/stream.ts
emranemran: Fixed

packages/api/src/controllers/stream.ts
emranemran: Fixed

packages/api/src/controllers/stream.ts
emranemran: Fixed

packages/api/src/controllers/stream.ts
emranemran: Yeah i was trying to do it here but was too bogged down with addressing each usage and making sure it worked. I'll file a ticket as a follow up to try and make this a generic helper.

packages/api/src/controllers/stream.ts
emranemran: Thanks for these - helped a lot to clear through some confusion I had.

emranemran:
**Update:**
* Addressed all comments
* Fixed/added unit tests

packages/api/src/controllers/stream.test.ts
```diff
@@ -162,6 +170,32 @@ describe("controllers/stream", () => {
...
+      for (let i = 0; i < 5; i += 1) {
+        const document = {
+          id: uuid(),
+          kind: "stream",
+          deleted: i > 3 ? true : undefined,
+          projectId: i > 2 ? projectId : undefined,
+        } as DBStream;
+        await server.store.create(document);
+        const res = await client.get(
+          `/stream/${document.id}/?projectId=${projectId}`
```
victorges: Is the `projectId` required on a GET by ID here? Given this is only for an admin API, I'd say it should allow reading the object with just the ID

packages/api/src/controllers/stream.test.ts
```diff
@@ -1134,15 +1170,63 @@ describe("controllers/stream", () => {
...
         expect(res.status).toBe(200);
       }
+
+      // create a new project
+      client.jwtAuth = nonAdminToken;
+      projectId = await createProject(client);
+      expect(projectId).toBeDefined();
+
+      // then create a new api-key under that project
+      const allowedOrigins = ["http://localhost:3000"];
```
victorges: Is this used?

packages/api/src/controllers/stream.test.ts
```diff
@@ -1134,15 +1170,63 @@ describe("controllers/stream", () => {
...
+
+      // create streams with a projectId
+      for (let i = 0; i < 5; i += 1) {
+        const document = {
+          id: uuid(),
+          kind: "stream",
+          userId: nonAdminUser.id,
+          projectId: projectId,
+        };
+        await server.store.create(document);
```
victorges: Would be good to have some of these tests actually create objects through the API, not hardcoding them on the db. To see if `projectId` is added properly to the created objects on that flow as well (is it? 👀)

packages/api/src/controllers/stream.test.ts
```diff
@@ -1134,15 +1170,63 @@ describe("controllers/stream", () => {
...
     it("should get own streams", async () => {
       client.apiKey = nonAdminApiKey;
       let res = await client.get(`/stream/user/${nonAdminUser.id}`);
       expect(res.status).toBe(200);
       const streams = await res.json();
-      expect(streams.length).toEqual(3);
+      expect(streams.length).toEqual(8);
+      expect(streams[0].userId).toEqual(nonAdminUser.id);
+
+      // ensure project associated with a new api-key is not enforced for this endpoint
```
victorges: Not sure I understand this comment/what is being tested here. But if we're using an API key from a different project, shouldn't this API return no results? In the end game I don't think we should have any APIs that skip the project abstraction unless they're admin.

packages/api/src/controllers/stream.ts
```diff
@@ -1677,7 +1710,10 @@ app.patch(
     const stream = await db.stream.get(id);
 
     const exists = stream && !stream.deleted;
-    const hasAccess = stream?.userId === req.user.id || req.isUIAdmin;
+    const hasAccess =
+      stream?.userId === req.user.id ||
+      (stream?.projectId ?? "" !== req.project?.id ?? "") ||
+      req.isUIAdmin;
```
victorges: You're mixing up `!==` and `===` checks here. I guess it should be `===` and `&&`. (my helper function suggestion might help with these)

packages/api/src/controllers/stream.ts
```diff
@@ -1923,7 +1968,10 @@ app.patch("/:id/suspended", authorizer({}), async (req, res) => {
   const stream = await db.stream.get(id);
   if (
     !stream ||
-    (!req.user.admin && (stream.deleted || stream.userId !== req.user.id))
+    (!req.user.admin &&
+      (stream.deleted ||
+        stream.userId !== req.user.id ||
+        (stream.projectId ?? "") !== (req.project?.id ?? "")))
```
victorges: Ok, so I'm gonna bring up this idea one last time, which is to create a helper function that checks access to a resource, so we don't have to keep repeating this whole logical chain everywhere that accesses some object in the DB. 

So I've implemented the helper function here https://github.com/livepeer/studio/commit/cb6f2d7e7b3ec22e408f9961ce885ac41d31d22f and use it in a couple different places just to show how it could be used. The idea would be just replacing the other access checks, seen on this PR, with a call to that function. It should hopefully make it easier to migrate the other data types to the new project-aware authorization flow as well, in the future.

Notice that I also added this which I think you forgot, to make sure we don't spam the db with queries for project objects that never change (we have it similar for users and api keys): https://github.com/livepeer/studio/commit/cb6f2d7e7b3ec22e408f9961ce885ac41d31d22f#diff-ce0ba37b6f5d180196d67f9e4b868e51df33660eac5213b4372701773a7b046bR65
I'd say that line is a required change, if you don't adopt the helper functions it'd be good to copy at least that part over.

> Disclaimer: I'm also only insisting because I don't recall getting a reply from you on the other comments I made while I think it'd make the code much less repetitive and error-prone. I feel the suggestion could feel a little daunting without the TypeScript experience to refactor some stuff around as well, so I thought providing you with a commit head start could be helpful. I wrote that commit in 5 minutes and no problem if you prefer to stick with your current implementation tho.

victorges (approved):
LGTM with some minor changes (and a suggestion)

packages/api/src/controllers/session.ts
```diff
@@ -146,7 +151,9 @@ app.get("/:id", authorizer({}), async (req, res) => {
   let session = await db.session.get(req.params.id);
   if (
     !session ||
-    ((session.userId !== req.user.id || session.deleted) &&
+    ((session.userId !== req.user.id ||
+      (session.projectId ?? "") !== (req.project?.id ?? "") ||
```
victorges: Shouldn't this be using the helper?

packages/api/src/controllers/signing-key.test.ts
```diff
@@ -91,7 +99,9 @@ describe("controllers/signing-key", () => {
     });
 
     it("should list all user signing keys", async () => {
-      const res = await client.get(`/access-control/signing-key`);
+      const res = await client.get(
+        `/access-control/signing-key?projectId=${projectId}`
```
victorges: Is this query-string necessary? Thought it was auto detected from API key.

packages/api/src/controllers/stream.test.ts
victorges: Still holds. Not sure I get why it's needed here.

packages/api/src/controllers/stream.ts
```diff
@@ -583,21 +593,19 @@ app.get("/:parentId/sessions", authorizer({}), async (req, res) => {
...
   const query = [];
   query.push(sql`data->>'parentId' = ${stream.id}`);
   query.push(sql`(data->'lastSeen')::bigint > 0`);
   query.push(sql`(data->'sourceSegmentsDuration')::bigint > 0`);
   query.push(sql`data->>'partialSession' IS NULL`);
+  if (!req.isUIAdmin) {
+    query.push(
+      sql`coalesce(data->>'projectId', '') = ${req.project?.id || ""}`
+    );
+  }
```
victorges: I don't think this is required, there's no `userId` check here either right? Probably because child streams are necessarily from the same user (and also from the same project) as their parent.

I'd say we follow the rule of thumb I suggested and only add project checks on the same place where we have user checks. No need to filter by project here if we're not filtering by user.

If you prefer to keep this filter, WDYT of adding the corresponding user filter here as well? I think it's confusing to check only 1 of the 2, especially if only the project.

packages/api/src/controllers/stream.ts
```diff
@@ -1707,7 +1687,10 @@ app.post(
       return res.json({ errors: ["stream not found"] });
     }
 
-    if (stream.userId !== req.user.id) {
+    if (
```
victorges: Could use helper here (and remove above `if` as well)

packages/api/src/controllers/stream.ts
```diff
@@ -1751,7 +1734,10 @@ app.delete("/:id/multistream/:targetId", authorizer({}), async (req, res) => {
     return res.json({ errors: ["stream not found"] });
   }
 
-  if (stream.userId !== req.user.id) {
+  if (
```
victorges: Same for helper

packages/api/src/controllers/stream.ts
```diff
@@ -1790,7 +1776,7 @@ app.patch(
     const stream = await db.stream.get(id);
 
     const exists = stream && !stream.deleted;
-    const hasAccess = stream?.userId === req.user.id || req.user.admin;
+    const hasAccess = hasAccessToResource(req, stream, true);
     if (!exists || !hasAccess) {
```
victorges: You can also just use the full `req.checkResourceAccess` helper here. This was just a dummy example of when the `hasAccessToResource` was needed (but not where the logic actually needed a separate check).

packages/api/src/controllers/stream.ts
```diff
@@ -1870,7 +1856,11 @@ app.patch(
 app.patch("/:id/record", authorizer({}), async (req, res) => {
   const { id } = req.params;
   const stream = await db.stream.get(id);
-  if (!stream || stream.deleted) {
+  if (
+    !stream ||
+    stream.deleted ||
+    (stream.projectId ?? "") !== (req.project?.id ?? "")
+  ) {
```
victorges: Same for helper. Scary, were we not checking the `userId` here? 😳

packages/api/src/controllers/webhook.ts
```diff
@@ -270,7 +267,8 @@ app.delete("/", authorizer({}), async (req, res) => {
     const webhooks = await db.webhook.getMany(ids);
     if (
       webhooks.length !== ids.length ||
-      webhooks.some((s) => s.deleted || s.userId !== req.user.id)
+      webhooks.some((s) => s.deleted || s.userId !== req.user.id) ||
+      webhooks.some((s) => (s.projectId ?? "") !== (req.project?.id ?? ""))
```
victorges: Now **this** is one good place where you could use `hasAccessToResource` :)

victorges (approved):
LGTM overall, but I think there are a few other places that could use the new helpers.

There are also a couple comments from the past reviews that didn't get a reply nor fix. Could you make sure there's nothing important left behind?

PR #2104, opened 2024-03-14
Delay the stream pull if it has been recently terminated

mjh1:


packages/api/src/store/stream-table.ts
```diff
@@ -382,6 +382,7 @@ export default class StreamTable extends Table<DBStream> {
       transcodedSegmentsDuration: 0,
       sourceBytes: 0,
       transcodedBytes: 0,
+      lastTerminatedAt: null,
```
mjh1, line 385: @victorges the line below this: `...obj` will mean that any of these default values will be overwritten if `obj` has the params set right?

thomshutt (approved):

packages/api/src/store/stream-table.ts
victorges: Exactly

PR #2108, opened 2024-03-21
Implement locking stream pull to avoid multiple catalyst pulling the …

leszko:
…same stream

### DIFF
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -535,11 +535,11 @@ describe("controllers/stream", () => {
         const resLockPull = await client.post(`/stream/${stream.id}/lockPull`);
         expect(resLockPull.status).toBe(204);
 
-        // Check that the lockedAt is marked with the correct date
+        // Check that the pullLockedAt is marked with the correct date
         const res2 = await client.get(`/stream/${stream.id}`);
         expect(res2.status).toBe(200);
         const stream2 = await res2.json();
-        expect(stream2.pull.lockedAt).toBeGreaterThan(now);
+        expect(stream2.pullLockedAt).toBeGreaterThan(now);
       });
 
       it("should not lock pull for an active stream", async () => {
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -524,6 +524,76 @@ describe("controllers/stream", () => {
         expect(server.db.stream.addDefaultFields(document)).toEqual(stream);
       });
 
+      it("should lock pull for a new stream", async () => {
+        // Create stream pull
+        const now = Date.now();
+        const res = await client.put("/stream/pull", postMockPullStream);
+        expect(res.status).toBe(201);
+        const stream = await res.json();
+
+        // Request pull lock
+        const resLockPull = await client.post(`/stream/${stream.id}/lockPull`);
+        expect(resLockPull.status).toBe(204);
+
+        // Check that the lockedAt is marked with the correct date
+        const res2 = await client.get(`/stream/${stream.id}`);
+        expect(res2.status).toBe(200);
+        const stream2 = await res2.json();
+        expect(stream2.pull.lockedAt).toBeGreaterThan(now);
+      });
+
+      it("should not lock pull for an active stream", async () => {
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -1086,6 +1086,41 @@ app.put(
...
+  const stream = await db.stream.get(id, { useReplica: false });
+  if (!stream || (stream.deleted && !req.user.admin)) {
+    res.status(404);
+    return res.json({ errors: ["not found"] });
+  }
+  if (stream.isActive) {
+    return res.status(423);
+  }
+
+  const updateRes = await db.stream.update(
```
victorges, line 1107: WDYT of also adding a `WHERE` clause on this query instead, that disconsiders the stream if it has `isActive=true`? Given we don't have transactions on the DB access, this would guarantee we don't have race conditions here.

packages/api/src/schema/api-schema.yaml
```diff
@@ -479,6 +479,9 @@ components:
                     Longitude of the pull source in degrees. East is positive,
                     west is negative.
                   example: -104.988
+            lockedAt:
```
victorges: I think ideally this field should be hidden from responses. In other resources this would just mean setting it as a `writeOnly: true` here, but for streams we have a separate clean-up function that handles removing fields from some const in the code: https://github.com/livepeer/livepeer-com/blob/ff6c3a0453d2a0cce8f9ace8649c8c1bbfac7a7a/packages/api/src/store/stream-table.ts#L390

That function doesn't support nested properties today tho, so I'll leave it up to you if it's worth the work of making it work with this nested `pull.lockedAt` field.

packages/api/src/controllers/stream.test.ts
```diff
@@ -524,6 +524,76 @@ describe("controllers/stream", () => {
...
+        expect(res.status).toBe(201);
+        const stream = await res.json();
+
+        // Request pull lock
+        const resLockPull = await client.post(`/stream/${stream.id}/lockPull`);
+        expect(resLockPull.status).toBe(204);
+
+        // Requesting pull lock should fail, because pull has already locked b y another node (so it should be replicated instead of being pulled)
+        const resLockPull2 = await client.post(`/stream/${stream.id}/lockPull`);
+        expect(resLockPull2.status).toBe(423);
```
victorges: Could be fun to make these in parallel with a `Promise.all` to actually make sure there are no races.
e.g.
```suggestion
        const promises = [];
        for (int i = 0; i < 10; i++) {
          promises.push(client.post(`/stream/${stream.id}/lockPull`));
        }
        const resPulls = await Promise.all(promises);
        expect(resPulls.filter(r => r.status === 204).length).toBe(1);
        expect(resPulls.filter(r => r.status === 423).length).toBe(9);
```
(sounded too fun I had to try. not sure it even "compiles" tho lol)

victorges (approved):
LGTM

packages/api/src/controllers/stream.ts
mjh1: I was curious about the race condition too, of two catalyst calls acquiring the lock at the same time. But is that what we're guarding against with this:
```
sql`COALESCE((data->'pull'->>'lockedAt')::bigint,0) < ${
        Date.now() - leaseTimeout
      }`,
```
Does that ensure the update doesn't go through if there's already an active lock? Or can there still be a race condition because they can happen in parallel with no transactions?

I'm not sure checking `isActive` would help that much because there is a fairly big gap between acquiring the lock and the stream becoming active? I may be misunderstanding your idea Victor

leszko: I discussed it with Victor, it's fine as it is right now.

The JSON update is transactional, so it's ok.

packages/api/src/controllers/stream.test.ts
leszko: Good idea, updated.

packages/api/src/schema/api-schema.yaml
leszko: Good point. I've moved `lockedAt` to non-nested `pullLockedAt` and prevent returning it from the API response.

mjh1 (approved):

PR #2109, opened 2024-03-22
refactor: improve api schema

suhailkakar:

- fixes many issues and improves the api schema file 

- format yaml file 
- add operationId to all endpoints
- marked rooms api as deprected
- categories endpoints by tags
  - stream
  - asset
  - webhook
  - multistream
  - session
  - room
  - transcode
  - metrics  
  - playback  
  - access control  
- moved unused components to db-schema
- add examples to many components
- override name for sdk generation (so response method will always be `res.data` instead of `res.classes` or `res.object`
- override operation name for sdk generation (to make sdk methods a lot nices like `livepeer.asset.create` or `livepeer.stream.get`

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/compile-schemas.js
+++ b/packages/api/src/compile-schemas.js
@@ -76,7 +76,6 @@ const data = _.merge({}, apiData, dbData);
   | {
       type: "unverified";
       value: string;
-      [k: string]: unknown;
     }
   | string;`;
 
--- a/packages/api/src/compile-schemas.js
+++ b/packages/api/src/compile-schemas.js
@@ -76,6 +76,7 @@ const data = _.merge({}, apiData, dbData);
   | {
       type: "unverified";
       value: string;
+      [k: string]: unknown;
     }
   | string;`;
 
--- a/packages/api/src/compile-schemas.js
+++ b/packages/api/src/compile-schemas.js
@@ -55,7 +55,7 @@ const data = _.merge({}, apiData, dbData);
   const ajv = new Ajv({ sourceCode: true });
 
   const index = [];
-  const types = [];
+  let types = [];
 
   for (const [name, schema] of Object.entries(data.components.schemas)) {
     schema.title = name;
@@ -72,7 +72,17 @@ const data = _.merge({}, apiData, dbData);
   const indexPath = path.resolve(validatorDir, "index.js");
   write(indexPath, indexStr);
 
-  const typeStr = types.join("\n");
+  const typeDefinition = `export type InputCreatorId =
+  | {
+      type: "unverified";
+      value: string;
+    }
+  | string;`;
+
+  let typeStr = types.join("\n\n");
+  const cleanedTypeStr = typeStr.split(typeDefinition).join("");
+  typeStr = `${cleanedTypeStr.trim()}\n\n${typeDefinition}`;
+
   const typePath = path.resolve(schemaDir, "types.d.ts");
   write(typePath, typeStr);
 })().catch((err) => {
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -3369,6 +3369,7 @@ paths:
                 $ref: '#/components/schemas/error'
   /room:
     post:
+      deprecated: true
       operationId: createRoom
       summary: Create a room
       description: |
@@ -3388,6 +3389,7 @@ paths:
                 $ref: '#/components/schemas/error'
   '/room/{id}':
     get:
+      deprecated: true
       operationId: getRoom
       parameters:
         - name: id
@@ -3410,6 +3412,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/error'
     delete:
+      deprecated: true
       operationId: deleteRoom
       parameters:
         - name: id
@@ -3429,6 +3432,7 @@ paths:
                 $ref: '#/components/schemas/error'
   '/room/{id}/egress':
     post:
+      deprecated: true
       operationId: startRoomEgress
       requestBody:
         required: true
@@ -3458,6 +3462,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/error'
     delete:
+      deprecated: true
       operationId: stopRoomEgress
       summary: Stop room RTMP egress
       parameters:
@@ -3477,6 +3482,7 @@ paths:
                 $ref: '#/components/schemas/error'
   '/room/{id}/user':
     post:
+      deprecated: true
       operationId: createRoomUser
       requestBody:
         required: true
@@ -3515,6 +3521,7 @@ paths:
   '/room/{id}/user/{userId}':
     get:
       operationId: getRoomUser
+      deprecated: true
       parameters:
         - name: id
           in: path
@@ -3541,6 +3548,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/error'
     put:
+      deprecated: true
       operationId: updateRoomUser
       requestBody:
         required: true
@@ -3572,6 +3580,7 @@ paths:
                 $ref: '#/components/schemas/error'
     delete:
       operationId: deleteRoomUser
+      deprecated: true
       parameters:
         - name: id
           in: path
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -1951,19 +1951,28 @@ components:
                       type: string
                       description: IPFS CID of the exported data
                       example: Qmabc123xyz341
+
     input-creator-id:
-      type: object
-      additionalProperties: true
-      required:
-        - type
-        - value
-      properties:
-        type:
-          type: string
-        value:
-          type: string
+      oneOf:
+        - type: object
+          oneOf:
+            - type: object
+              additionalProperties: false
+              required:
+                - type
+                - value
+              properties:
+                type:
+                  type: string
+                  enum:
+                    - unverified
+                value:
+                  type: string
+
+        - type: string
           description: >-
-            Developer-managed ID of the user who created the resource.
+            Helper syntax to specify an unverified creator ID, fully managed by
+            the developer.
     creator-id:
       type: object
       oneOf:
@@ -1975,10 +1984,13 @@ components:
           properties:
             type:
               type: string
+              enum:
+                - unverified
             value:
               type: string
               description:
                 Developer-managed ID of the user who created the resource.
+
     export-task-params:
       description: Parameters for the export task
       oneOf:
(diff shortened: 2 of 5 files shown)


packages/api/src/schema/api-schema.yaml
```diff
@@ -1,19 +1,38 @@
...
+  - name: room
+    description: Operations related to rooms api
+  - name: transcode
+    description: Operations related to transcode api
+  - name: metrics
+    description: Operations related to metrics api
+  - name: playback
+    description: Operations related to playback api
+  - name: accessControl
+    description: Operations related to access contorl/signing keys api
```
hjpotter92: nit: typo `contorl`

suhailkakar: fixed :)

packages/api/src/schema/api-schema.yaml
```diff
@@ -768,8 +805,10 @@ components:
           example: de7818e7-610a-4057-8f6f-b785dc1e6f88
           description: Points to parent stream object
         record:
-          description: |
-            Whether the stream should be recorded. Uses default settings. For more
+          description: >
+            Whether the stream should be recorded. Uses default settings. For
+            more
+
```
0xcadams: I don't know if this is an issue, but just flagging that there's an extra newline here

packages/api/src/schema/api-schema.yaml
```diff
@@ -1957,18 +2036,11 @@ components:
           example: 1587667174725
         publicKey:
           type: string
+          example: "-----BEGIN PUBLIC KEY"
```
0xcadams: This may be misleading - this should be base64 encoded

packages/api/src/schema/api-schema.yaml
```diff
@@ -2511,17 +2591,24 @@ components:
...
+              example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
             failed:
               type: string
               description: ID of the last task to fail execution.
+              example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
 paths:
   /stream:
     post:
+      operationId: createStream
+      x-speakeasy-name-override: create
```
0xcadams, line 2691: Very cool

0xcadams (approved):
LGTM, some tiny nits

packages/api/src/schema/api-schema.yaml
```diff
@@ -167,17 +211,23 @@ components:
               - task.updated
               - task.completed
               - task.failed
+          example:
+            - stream.started
+            - stream.detection
```
victorges: The detection API is actually deprecated, let's give another example here
```suggestion
            - stream.idle
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -215,6 +265,13 @@ components:
                 Timestamp (in milliseconds) at which the webhook last was
                 triggered
               example: 1587667174725
+          example:
```
victorges: Do we need both an example for the full object and for each field?

packages/api/src/schema/api-schema.yaml
```diff
@@ -244,26 +303,33 @@ components:
...
               type: string
               description: HTTP request method
+              example: POST
             headers:
               type: object
               description: HTTP request headers
               additionalProperties:
                 type: string
+              example:
+                Authorization: Bearer 123
```
victorges: We don't send auth headers today so this may be confusing. Perhaps this?
```suggestion
                User-Agent: livepeer.studio
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1034,8 +1073,10 @@ components:
               description: Error message if the asset creation failed.
         name:
           type: string
-          description: |
-            The name of the asset. This is not necessarily the filename - it can be a
+          description: >
+            The name of the asset. This is not necessarily the filename - it can
+            be a
+
             custom name or title.
```
victorges: Potentially typo paragraph break here?
```suggestion
            be a custom name or title.
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -1169,21 +1210,17 @@ components:
...
       properties:
         name:
           type: string
-          description: |
-            The name of the asset. This is not necessarily the filename - it can be a
+          description: >
+            The name of the asset. This is not necessarily the filename - it can
+            be a
+
             custom name or title.
```
victorges: same
```suggestion
            be a custom name or title.
```

packages/api/src/schema/api-schema.yaml
victorges: Probably worth adding a `description` explaining that. @gioelecerati should know details on the format

packages/api/src/schema/api-schema.yaml
```diff
@@ -2182,15 +2221,19 @@ components:
         playbackId:
           type: string
           description: The playback ID associated with the metric.
+          example: 1bde4o2i6xycudoy
         creatorId:
           type: string
           description: The ID of the creator associated with the metric.
+          example: ucqw1bde4o2i6xycudoy
```
victorges: Maybe making this either an email, ETH address or UUID woul be clearer

packages/api/src/schema/api-schema.yaml
```diff
@@ -3578,7 +3856,11 @@ paths:
                 $ref: "#/components/schemas/error"
   /data/views/query:
     get:
+      operationId: getViewershipMetrics
+      x-speakeasy-name-override: getViewership
       summary: Query viewership metrics
+      tags:
+        - metrics
```
victorges, line 3945: What if it was more like this instead?
```suggestion
      operationId: getViewershipMetrics
      x-speakeasy-name-override: getMetrics
      summary: Query viewership metrics
      tags:
        - views
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -3799,17 +4093,21 @@ paths:
...
             application/json:
               schema:
                 $ref: "#/components/schemas/error"
-  "/data/usage/query":
+  /data/usage/query:
     get:
+      operationId: getUsageMetrics
+      x-speakeasy-name-override: getUsage
+      tags:
+        - metrics
```
victorges, line 4190: And here
```suggestion
      operationId: getUsageMetrics
      x-speakeasy-name-override: getMetrics
      tags:
        - usage
```

packages/api/src/schema/db-schema.yaml
```diff
@@ -1015,8 +1024,87 @@ components:
...
+          $ref: "#/components/schemas/webhook/properties/name"
+        url:
+          $ref: "#/components/schemas/webhook/properties/url"
+        events:
+          $ref: "#/components/schemas/webhook/properties/events"
+        sharedSecret:
+          $ref: "#/components/schemas/webhook/properties/sharedSecret"
+        streamId:
+          $ref: "#/components/schemas/webhook/properties/streamId"
     new-asset-from-url-payload:
```
victorges: I believe this is used on a public API, for uploading an asset with a URL. Is that endpoint not in our `endpoints:` on `api-schema.yaml`? I believe it's `/api/asset/upload/url`

victorges (approved):
Awesome!!

packages/api/src/schema/api-schema.yaml
suhailkakar: Agreed, I will remove it for now and will talk to Gioele about the correct format and example

packages/api/src/schema/api-schema.yaml
suhailkakar: Not really! I just removed the full object example

packages/api/src/schema/api-schema.yaml
suhailkakar: Love the idea! I will open a separate PR to update the tags :)

packages/api/src/schema/api-schema.yaml
suhailkakar: same, i will open a separate PR to update the tags

packages/api/src/schema/db-schema.yaml
suhailkakar: Thanks! confirmed with Chase and added it back to `api-schema.yaml`

packages/api/src/compile-schemas.js
```diff
@@ -72,7 +72,17 @@ const data = _.merge({}, apiData, dbData);
   const indexPath = path.resolve(validatorDir, "index.js");
   write(indexPath, indexStr);
 
-  const typeStr = types.join("\n");
+  const typeDefinition = `export type InputCreatorId =
```
victorges, line 75: So weird that this was necessary. Did you get to understand why it was being generated twice?

packages/api/src/schema/api-schema.yaml
```diff
@@ -1890,15 +1966,28 @@ components:
                     cid:
                       type: string
                       description: IPFS CID of the exported data
+                      example: Qmabc123xyz341
+
     input-creator-id:
       oneOf:
-        - $ref: "#/components/schemas/creator-id"
```
victorges, line 1895: Can we keep the inline `$ref` given the fix on `compile-schemas.js`? Or is this part of the fix?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2112,60 +2198,22 @@ components:
...
+          example: webhook
         webhookId:
           type: string
           description: ID of the webhook to use for playback policy
+          example: 1bde4o2i6xycudoy
         webhookContext:
           type: object
           description: User-defined webhook context
           additionalProperties: true
+          example: {}
```
victorges: Maybe add something in the example here for easier understanding. WDYT?
```suggestion
          example: {"streamerId":"my-custom-id"}
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2223,59 +2280,76 @@ components:
...
           type: string
           description: |
             The subdivision (e.g., state or province) where the viewer is
             located.
+          example: California
         timezone:
           type: string
           description: The timezone where the viewer is located.
+          example: America/Los_Angeles
         geohas:
```
victorges: Is this actually like this? Sounds like maybe a typo
```suggestion
        geohash:
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -2533,17 +2627,70 @@ components:
...
+        targetSegmentSizeSecs:
+          $ref: >-
+            #/components/schemas/new-asset-payload/properties/targetSegmentSizeSecs
+        objectStoreId:
+          type: string
+          description: Object store ID where the asset is stored
+          writeOnly: true
+          example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
+        catalystPipelineStrategy:
+          $ref: "#/components/schemas/task/properties/params/properties/upload/properties/catalystPipelineStrategy"
```
victorges, line 2685: I think these 2 fields should be moved to `db-schema` since they're somewhat internal

victorges (approved):
LGTM!

PR #2124, opened 2024-04-04
api/cannon: Re-fix local IP check with better error handling

victorges:
This brings back the fix from #2118, reverted by #2123, adding proper error checking
when a given domain has no IPs configured (likely for v4 or v6 only).

This also makes the resolution errors non-fatal, so we default to trust the domains
in case we have a problem with the local IP check. This is less secure but avoids 
surprises in case this breaks again. We should also add an alert when there are failures
in resolving the domain (from that log inside the `catch`).

- Re-apply the original commit
- Handle DNS errors when resolving v4/v6 (moved to helper func here)
- Make local IP check error non fatal

 - `yarn test`
 - ✅ call IPv4-only webhooks and make sure they work
 - ✅ call IPv6-only (if it exists) and make sure they work
 - ✅ call webhook with an IP in the host (private or public) and check it works/doesnt as expected
 - ✅ try localhost or private IP webhooks and make sure they don't work

Informal on discord.

**Checklist**


packages/api/src/webhooks/cannon.ts
```diff
@@ -464,6 +439,29 @@ export default class WebhookCannon {
...
+    }
+
+    const emptyIfNotFound = (err) => {
+      if ([dns.NODATA, dns.NOTFOUND, dns.BADFAMILY].includes(err.code)) {
+        return [] as string[];
+      }
+      throw err;
+    };
+
+    const urlObj = parseUrl(url);
```
victorges: gotta test what happens when the hostname is just an IP itself

emranemran:
LGTM but see discord thread on internal discussion.

packages/api/src/store/webhook-table.ts
```diff
@@ -49,12 +49,14 @@ export default class WebhookTable extends Table<DBWebhook> {
...
 
   async updateStatus(id: string, status: DBWebhook["status"]) {
+    const statusStr = JSON.stringify(status);
     const res = await this.db.query(
-      `UPDATE ${
-        this.name
-      } SET data = jsonb_set(data, '{status}', case when data->'status' is null then '{}' else data->'status' end || '${JSON.stringify(
-        status
-      )}') WHERE id = '${id}'`
+      sql``
```
leszko, line 54: Is this change related to the local IP change? Or is it just refactor? Both are fine, I'm just trying to connect the dots.

packages/api/src/webhooks/cannon.ts
```diff
@@ -464,6 +440,36 @@ export default class WebhookCannon {
     }
   }
 
+  private async checkIsLocalIp(url: string, isAdmin: boolean) {
```
leszko: I wonder if it's possible and if it makes sense to add some unit tests for this one. Wdyt?

leszko (approved):

packages/api/src/store/webhook-table.ts
victorges: It is not related, it is a fix to use proper SQL parameters in the query instead of raw string interpolation. As I was testing this there were some errors that contained `'` and the query failed, so we need to use proper `sql` strings here. I can move this to a separate PR if you think it's necessary.

packages/api/src/webhooks/cannon.ts
victorges: Yeah good call. Done!

PR #2126, opened 2024-04-04
Integrating June tracking tool

hthillman:
This tool allows us to track users through their onboarding journey.

**Testing instructions:**

**note: I've disabled the `identifyUser` step for staging, so you'll need to run locally - you can also use the preview from [3860bc8](https://github.com/livepeer/studio/pull/2126/commits/3860bc8c8a048d76ce38b2626c86722dce36865d) if you want but probably better to run locally **

We will need to check functionality of everywhere I've added June events.
- API key creation
- docs and feedback links in header
- all sidebar navlinks
- creation of multistream targets
- recording toggle
- render of stream health tab in stream detail page
- embed and copy links on stream detail page
- stream go live 
- stream key copy
- landing page billing cta
- billing page usage cta
- registration / account creation
- login flows

I've also ripped out the Ripe integration points for now, can revert if needed


packages/www/components/StreamDetails/MultistreamTargetsTable/index.tsx
```diff
@@ -75,7 +76,10 @@ const MultistreamTargetsTable = ({
     [state.tableId, stream, streamHealth, invalidateTargetId, targetRefs]
   );
 
-  const onCreateClick = saveDialogState.onOn;
+  const onCreateClick = () => {
+    June.track(events.stream.multistreamTarget);
+    return saveDialogState.onOn;
```
adamsoffer: `saveDialogState.onOn` function needs to be invoked `saveDialogState.onOn()`

packages/www/components/Header/index.tsx
```diff
@@ -105,7 +106,11 @@ const Header = ({ breadcrumbs = [] }) => {
               appearance: "none",
               border: "none",
             }}>
-            <Link href="https://docs.livepeer.studio" passHref legacyBehavior>
+            <Link
+              href="https://docs.livepeer.studio"
+              passHref
+              legacyBehavior
+              onClick={() => June.track(events.all.documentation)}>
```
adamsoffer: The legacy behavior of the next/link component we use requires `onClick` be set on the _child_ of next/link.

adamsoffer (changes requested):
Need to remove additional instances of Ripe:

https://github.com/livepeer/studio/blob/6432cc93e4f40c72baec0e0d7e820f13c22f684f/packages/www/pages/dashboard/index.tsx#L13

https://github.com/livepeer/studio/blob/6432cc93e4f40c72baec0e0d7e820f13c22f684f/packages/www/pages/dashboard/index.tsx#L15

https://github.com/livepeer/studio/blob/6432cc93e4f40c72baec0e0d7e820f13c22f684f/packages/www/pages/dashboard/index.tsx#L43

https://github.com/livepeer/studio/blob/6432cc93e4f40c72baec0e0d7e820f13c22f684f/packages/www/pages/index.tsx#L10



Also need to remove [ripe.js file](https://github.com/livepeer/studio/blob/6432cc93e4f40c72baec0e0d7e820f13c22f684f/packages/www/lib/ripe.ts):

packages/www/components/Header/index.tsx
hthillman: @adamsoffer I assume this is a universal change for this PR? There are a couple other `onClick`s like this

adamsoffer: Yeah anywhere you added an `onClick` to a `Link` component

adamsoffer:
When I navigate to a stream detail view and then refresh the page I get an error that `June` isn't available on `window`. @suhailkakar can you help debug this issue?

CleanShot 2024-04-10 at 16 35 41@2x

hthillman:
> When I navigate to a stream detail view and then refresh the page I get an error that `June` isn't available on `window`. @suhailkakar can you help debug this issue?
> 
> CleanShot 2024-04-10 at 16 35 41@2x

@adamsoffer I wonder if the `identify` call is in the wrong place, and `track` is not available until after identification. I thought I put it somewhere that would get initialized upfront every time but maybe there's a race condition

suhailkakar:
hey @hthillman 

- i've replaced the script and `window.analytics` object with the June next.js sdk[1] and a hook
- all the implementational will be same, but instead of importing June from `../lib/June`, we are using the June hook

```
import { useJune, events } from "hooks/use-june";
const June = useJune();
```
- i've tested and all the `tracks` events are being sent to June
image


cc: @adamsoffer 

[1] https://www.june.so/docs/tracking/web/react

suhailkakar:
Also, the big "file changes" are just types issue related to livepeer/design library which is also fixed

packages/www/hooks/use-june.tsx
```diff
@@ -0,0 +1,59 @@
...
+  [path: string]: string;
+}
+
+export function useJune() {
+  const [analytics, setAnalytics] = useState(undefined);
+
+  useEffect(() => {
+    const loadAnalytics = async () => {
+      let response = AnalyticsBrowser.load({
+        writeKey: "3VINzqYVjfOxFyIr",
```
suhailkakar, line 50: Not sure, if this needs to be added to `.env`. Please feel free to move it from here if it needs to kept private.

adamsoffer (approved):
LGTM thanks @suhailkakar !

PR #2127, opened 2024-04-09
Add pullRegion to streams to make the ingest node sticky

leszko:
This is the updated flow:
1. Fetch the region from Catalyst/Mist (effectively using MistUtilLoad to decide on region based on `stream.pull.location`)
2. Store this region in the stream object as `stream.pullRegion`
3. Use this region in catalyst-api (so if the it's being pulled by the wrong region, then it changes the region)

This way, we resolve region once in `/pull` and then re-use the same region for stream pulling. This should make some improvement in the ingest node, and what's even better, we can start logging that `(lat,lon)=>region`, and then have data to say if out method for choosing ingest node (`MistUtilLoad`) is correct.

One of the further idea would be to not use `MistUtilLoad`, but just always use the geographically closest node.

Related PRs:
- https://github.com/livepeer/go-api-client/pull/64
- https://github.com/livepeer/catalyst-api/pull/1215

### DIFF
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -728,7 +728,7 @@ describe("controllers/stream", () => {
           extractRegionFrom(
             "https://fra-staging-staging-catalyst-0.livepeer.monster:443/hls/video+not-used-playback/index.m3u8"
           )
-        ).toBe("fra");
+        ).toBe("fra-staging");
         expect(
           extractRegionFrom(
             "https://fra-staging-staging-catalyst-0.livepeer.monster:443/hls/video+other-playback/index.m3u8"
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -23,6 +23,7 @@ import {
 import serverPromise, { TestServer } from "../test-server";
 import { semaphore, sleep } from "../util";
 import { generateUniquePlaybackId } from "./generate-keys";
+import { extractRegionFrom } from "./stream";
 
 const uuidRegex = /[0-9a-f]+(-[0-9a-f]+){4}/;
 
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/stream.test.ts
```diff
@@ -712,6 +713,28 @@ describe("controllers/stream", () => {
         const document = await db.stream.get(stream.id);
         expect(db.stream.addDefaultFields(document)).toEqual(updatedStream);
       });
+      it("should extract region from redirected playback url", async () => {
```
emranemran, line 716: It might be good to have some flv urls in here to verify nothing breaks: 
e.g `/flv/<playback-id>`

packages/api/src/controllers/stream.ts
```diff
@@ -235,6 +236,39 @@ async function triggerManyIdleStreamsWebhook(ids: string[], queue: Queue) {
...
 
+async function resolvePullRegion(
+  stream: NewStreamPayload,
+  ingest: string
+): Promise<string> {
+  if (process.env.NODE_ENV === "test") {
+    return null;
+  }
+  const url = new URL(
+    pathJoin(ingest, `hls`, "not-used-playback", `index.m3u8`)
```
emranemran, line 247: don't we need to handle flv as well?

packages/api/src/controllers/stream.ts
```diff
@@ -235,6 +236,39 @@ async function triggerManyIdleStreamsWebhook(ids: string[], queue: Queue) {
...
+  );
+  const { lat, lon } = stream.pull?.location ?? {};
+  if (lat && lon) {
+    url.searchParams.set("lat", lat.toString());
+    url.searchParams.set("lon", lon.toString());
+  }
+  const playbackUrl = url.toString();
+  const response = await fetchWithTimeout(playbackUrl, { redirect: "manual" });
```
emranemran, line 256: nit: maybe a comment here to note that this is being sent to catalyst/MistUtilLoad

emranemran (approved):

packages/api/src/controllers/stream.ts
leszko: No, this call is not for the "any redirect", this is just to discover which should be the ingest node.

So, kind-of hack the system and use catalyst-api redirect balancing (effectively using `MistUtilLoad`) to divide on the ingest node.

packages/api/src/controllers/stream.test.ts
leszko: As pointed out below, we don't use it for `/flv`

packages/api/src/controllers/stream.ts
leszko: Added

PR #2137, opened 2024-04-18
Add additional logs to /pull

leszko:


gioelecerati (approved):
LGTM, I am not sure if starting the log line with / is having any impact with our metrics, it's probably better to be a lowercase normal character

PR #2138, opened 2024-04-18
Stick to the pull host while triggering the catalyst pull start

leszko:
Two changes:
1. While resolving pull region, follow many redirects => Needed after the change with the GeoIP redirection mechanism
2. Resolve also host and stick to that host => It will improve the number of timed out `/pull` requests, because when stream is pulled from a different region, but studio waits on another, then Mist Pull + Replication is to slow to satisfy `10s` timeout on `/pull`

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -653,7 +653,7 @@ export const triggerCatalystPullStart =
           url.searchParams.set("lon", lon.toString());
           playbackUrl = url.toString();
           console.log(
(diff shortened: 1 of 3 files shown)


emranemran:
LGTM. One question: Shouldn't `fetchWithTimeoutAndRedirects` have been called even when we had the single redirect from CDN?

emranemran (approved):

leszko:
> LGTM. One question: Shouldn't `fetchWithTimeoutAndRedirects` have been called even when we had the single redirect from CDN?

Now, we always use `fetchWithTimeoutAndRedirects`. Before, when we had 1 redirect, then I used `fetchWithTimeout` and checked the redirection URL, which worked as well.

PR #2144, opened 2024-04-22
api: Handle active cleanup from pull lock API

victorges:

This is (yet another) hotfix for the active cleanup logic, specifically on the lock pull flow.

We were sometimes not allowing a stream to be locked because it was marked as active,
even tho it was lost (old lastSeen).

This adds some logic in the pull lock flow to handle the `lastSeen` as well.

- Do active cleanup of streams on pull lock logic
- If cleanup was triggered, skip the isActive check on the actual lock query

`yarn test`

Fixes isActive issue on lock pull

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -1170,15 +1170,17 @@ app.post("/:id/lockPull", authorizer({ anyAdmin: true }), async (req, res) => {
     return res.json({ errors: ["not found"] });
   }
 
+  // We have an issue that some of the streams/sessions are not marked as inactive when they should be.
+  // This is a workaround to clean up the stream in the background
   const doingActiveCleanup = activeCleanupOne(
     req.config,
     stream,
     req.queue,
     await getIngestBase(req)
   );
+
   // the `isActive` field is only cleared later in background, so we ignore it
   // in the query below in case we triggered an active cleanup logic above.
-
   const leaseDeadline = Date.now() - leaseTimeout;
   const updateRes = await db.stream.update(
     [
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -1170,12 +1170,22 @@ app.post("/:id/lockPull", authorizer({ anyAdmin: true }), async (req, res) => {
...
+  const leaseDeadline = Date.now() - leaseTimeout;
   const updateRes = await db.stream.update(
     [
       sql`id = ${stream.id}`,
-      sql`(data->>'pullLockedBy' = ${host} OR (COALESCE((data->>'pullLockedAt')::bigint,0) < ${
-        Date.now() - leaseTimeout
-      } AND COALESCE((data->>'isActive')::boolean,FALSE) = FALSE))`,
+      doingActiveCleanup
+        ? sql`(data->>'pullLockedBy' = ${host} OR (COALESCE((data->>'pullLockedAt')::bigint,0) < ${leaseDeadline}))`
+        : sql`(data->>'pullLockedBy' = ${host} OR (COALESCE((data->>'pullLockedAt')::bigint,0) < ${leaseDeadline} AND COALESCE((data->>'isActive')::boolean,FALSE) = FALSE))`,
```
victorges, line 1190: @leszko I could also replicate the `lastSeen` logic here but this query would become even more complex, so I preferred to keep it "non-concurrent-safe" for now. Let me know if you think we need to go all the way here.

leszko: Yeah, I think it's fine as it is.

leszko (approved):

PR #2145, opened 2024-04-22
fix june trigger

suhailkakar:
Preivously june events used to trigger multiple times, this PR adds callback to make sure each event is being sent just once

- Updated the implementation and used usedCallback hook

- Manually though browser network tab and June dashboard

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/www/components/Sidebar/index.tsx
+++ b/packages/www/components/Sidebar/index.tsx
@@ -31,6 +31,7 @@ import Router from "next/router";
 import { RocketIcon, ChatBubbleIcon, LoopIcon } from "@radix-ui/react-icons";
 import Contact from "../Contact";
 import { useJune, events } from "hooks/use-june";
+import { useCallback, useEffect } from "react";
 
 export const NavLink = styled(A, {
   fontSize: 14,
@@ -73,9 +74,16 @@ export type SidebarId =
 
 const Sidebar = ({ id }: { id: SidebarId }) => {
   const { user, logout } = useApi();
-  const June = useJune();
 
-  June?.track(`sidebar ${id}`);
+  const june = useJune();
+
+  const trackSidebarEvent = useCallback(() => {
+    if (june) june.track(`sidebar ${id} new`);
+  }, [june]);
+
+  useEffect(() => {
+    trackSidebarEvent();
+  }, [trackSidebarEvent]);
 
   return (
     <Box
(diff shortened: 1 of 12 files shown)


packages/www/components/Sidebar/index.tsx
```diff
@@ -73,9 +74,16 @@ export type SidebarId =
...
+
   const June = useJune();
 
-  June?.track(`sidebar ${id}`);
+  const trackSidebarEvent = useCallback(() => {
+    if (June) June.track(`sidebar ${id}`);
+  }, [June]);
+
+  useEffect(() => {
+    trackSidebarEvent();
```
0xcadams: Could you explain what this useEffect is supposed to do? Is it on mount we fire this event?

0xcadams (approved):
LGTM just one question on the useEffect

packages/www/components/Sidebar/index.tsx
suhailkakar: Yeah! just want to make sure the event is fired once the component is mounted. Do you think we can update with a better approach? I don't like using useEffect here as well :)

0xcadams: What's being tracked here? A user is shown a sidebar option?

suhailkakar: Yes, the sidebar id in which the page is active, like `stream`, `asset`, `developers` is being tracked and sent to June

0xcadams: Okay maybe we use pageview events for this?
https://www.june.so/docs/quickstart/install

suhailkakar: yeah, i've checked that, but they don't have that method on their SDK. We might need to add their API request then. 
Should I do it @0xcadams?

0xcadams: Oh odd - maybe we just add a route change handler instead with the events (supported by their SDK)?

```
const router = useRouter(); 

  useEffect(() => {
    const handleRouteChange = (url, { shallow }) => {
      if (analytics) analytics.page(url);
    };

    router.events.on('routeChangeStart', handleRouteChange);

    return () => {
      router.events.off('routeChangeStart', handleRouteChange);
    };
  }, [analytics]);
  ```

suhailkakar: yep! sounds good! just updated it.

0xcadams: LGTM!

PR #2146, opened 2024-04-23
api: moved objectStoreId and catalystPipelineStrategy to db schema

suhailkakar:

Move the references of `objectStoreId` and `catalystPipelineStrategy` from `api-schema` and instead moving it to `db-schema`

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -1387,9 +1387,6 @@ components:
                       type: integer
     new-asset-from-url-payload:
       properties:
-        url:
-          type: string
-          description: URL of the asset to be imported
         objectStoreId:
           type: string
           description: Object store ID where the asset is stored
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/db-schema.yaml
```diff
@@ -1385,3 +1385,15 @@ components:
                       type: string
                     timestamp:
                       type: integer
+    new-asset-from-url-payload:
+      properties:
+        url:
+          type: string
+          description: URL of the asset to be imported
```
victorges: This doesn't need to be repeated here, the db-schema is deep merged!

victorges (approved):
LGTM

packages/api/src/schema/db-schema.yaml
suhailkakar: Cool!, we can remove the reference of `url` from here?

PR #2148, opened 2024-04-24
fix stream page ui issues 

suhailkakar:

**Screenshots (optional)**

**Checklist**


packages/www/components/StreamsTable/StreamFilter.tsx
```diff
@@ -196,8 +194,28 @@ const StreamFilter = ({ onDone, activeFilters }) => {
     }
   }, [activeFilters]);
 
+  const getDropdownMargin = (filterId) => {
```
adamsoffer, line 197: @suhailkakar we really should not need this switch logic. Dropdown components should automatically align to the parent element. We can't have custom logic depending the width of the dropdown element. That is not sustainable. I'm going to merge this PR but let's remove this logic in a follow up PR and use the component correctly.

adamsoffer (approved):

PR #2152, opened 2024-04-29
api: stream: test profiles for test creator ids

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -1038,6 +1038,13 @@ const pullStreamKeyAccessors: Record<string, string[]> = {
   "pull.source": ["pull", "source"],
 };
 
+const testCreatorIds: string[] = [
+  "73846_104901225_104901225",
+  "73846_116005487_116003843",
+  "73846_116003843_116003843",
+  "73846_115939837_115939837",
+];
+
 app.put(
   "/pull",
   authorizer({}),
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -1057,9 +1064,21 @@ app.put(
       });
     }
 
+    let profiles = req.config.defaultStreamProfiles;
```
victorges: Can you name this `defaultProfiles` (and change L1081 to `profiles: defaultProfiles`) so this is clearer? Otherwise the request will still override these and that can be misleading.

packages/api/src/controllers/stream.ts
```diff
@@ -1057,9 +1064,21 @@ app.put(
...
+
+    // Allow test creatorIds to set custom profiles
+    if (rawPayload.creatorId && typeof rawPayload.creatorId === "string") {
+      if (testCreatorIds.includes(rawPayload.creatorId)) {
+        logger.info(
+          `pull request creatorId=${rawPayload.creatorId} is a test creatorId`
+        );
+        profiles = rawPayload.profiles; // TODO: specify here the hardcoded profiles for testing
```
victorges: Only merge this after changing this to the actual set of profiles.

victorges (approved):

emranemran (approved):

PR #2155, opened 2024-04-30
api/stream: Add a couple more hacks for testing Trovo profiles

victorges:

- When isMobileflag is set to true in the pull API endpoint, all transcode profiles set by the end user should be overwritten by fps=0. The rest of the parameters can remain as-is. So for example, if fps=30 is set by the user, we should replace that profile with fps=0 and keep remaining parameters the same. This is a temporary workaround required to resolve stuttering issues seen with Trovo's mobile app ([PS-382](https://linear.app/livepeer/issue/PS-382/mobile-playback-stream-not-working-properly-when-qa-pushes-480p-30-fps))

For example, a /pull API request where the user set profiles like this:

```json
"profiles": [ { "name": "360p0", "fps": 30, "bitrate": 800000, "width": 640, "height": 360, "profile": "H264ConstrainedHigh" }, { "name": "480p0", "fps": 30, "bitrate": 1600000, "width": 854, "height": 480, "profile": "H264ConstrainedHigh" }, ] }
```
will become the following (only fps field overwritten and set to 0): 

```json
"profiles": [ { "name": "360p0", "fps": 0, "bitrate": 800000, "width": 640, "height": 360, "profile": "H264ConstrainedHigh" }, { "name": "480p0", "fps": 0, "bitrate": 1600000, "width": 854, "height": 480, "profile": "H264ConstrainedHigh" }, ] }
```

- Any request that has a width=853 and height=480, should be updated so that width=854. This is because 853 is not a standard dimension and it results in a vertical green line on the right side of the frame ([PS-381](https://linear.app/livepeer/issue/PS-381/mobile-playback-green-screen-when-streamer-starts-or-ends-pk-with))

The 2 things above.

`yarn test`

Implements PS-534

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -1,15 +1,12 @@
-import { Router, Request } from "express";
+import { Request, Router } from "express";
+import _ from "lodash";
 import { QueryResult } from "pg";
 import sql from "sql-template-strings";
 import { parse as parseUrl } from "url";
 import { v4 as uuid } from "uuid";
-import _ from "lodash";
 
 import logger from "../logger";
-import { authorizer } from "../middleware";
-import { validatePost } from "../middleware";
-import { geolocateMiddleware } from "../middleware";
-import { fetchWithTimeoutAndRedirects } from "../util";
+import { authorizer, geolocateMiddleware, validatePost } from "../middleware";
 import { CliArgs } from "../parse-cli";
 import {
   DetectionWebhookPayload,
@@ -20,7 +17,6 @@ import {
   User,
 } from "../schema/types";
 import { db } from "../store";
-import { DBSession } from "../store/session-table";
 import {
   BadRequestError,
   InternalServerError,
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -1,15 +1,12 @@
-import { Router, Request } from "express";
+import { Request, Router } from "express";
+import _ from "lodash";
```
gioelecerati, line 2: Probably imported by mistake?

gioelecerati (approved):

packages/api/src/controllers/stream.ts
gioelecerati: Ah, I see it was elsewhere

victorges: No it's used in this file
https://github.com/livepeer/livepeer-com/blob/2a52015e9c4a0a41d94b7fe5ba5996af9447b097/packages/api/src/controllers/stream.ts

pwilczynskiclearcode:
Deployed to prod a moment ago https://discord.com/channels/423160867534929930/1102968315137228900/1235138317050183721

PR #2156, opened 2024-04-30
refactor: improve api schema 

suhailkakar:

**Screenshots (optional)**

**Checklist**


packages/api/src/schema/api-schema.yaml
```diff
@@ -162,6 +162,20 @@ components:
             - VP8
             - VP9
           example: H.264
+    webhook-payload:
```
victorges: I think the previous name is more accurate, since `webhook-payload` adds no information to simply `webhook` which is the schema that already exists and is used for the create endpoint. If this is only used on the request to PATCH a webhook, so it should be named `webhook-patch-payload`.

UNLESS the create endpoint also uses the same payload, in which case we should just drop this and use `webhook` for both (or `webhook-payload` for both, if there are differences but these differences should also apply to the create endpoint).

packages/api/src/schema/api-schema.yaml
```diff
@@ -176,6 +190,11 @@ components:
         name:
           type: string
           example: test_webhook
+        kind:
+          type: string
+          example: webhook
```
victorges, line 181: If we don't add this on the API schema, will the SDK tests break? This is really legacy and I hope we get rid of them at some point. Users should not be relying on this.

packages/api/src/schema/api-schema.yaml
```diff
@@ -176,6 +190,11 @@ components:
...
+        userId:
+          type: string
```
victorges, line 185: This is also a field that doesn't help much users of the API, might only confuse them. They will always get their own `userId` which, as we just chatted on discord, is not something they should need to be aware about nor be passing around. Can we remove this or will the SDK tests break as well?

packages/api/src/schema/api-schema.yaml
```diff
@@ -217,7 +236,6 @@ components:
           example: "https://my-service.com/webhook"
         sharedSecret:
           type: string
-          writeOnly: true
```
victorges: Why removing this? This field should be write-only (it is (should be) on the implementation and should also be shown as write-only on docs)

packages/api/src/schema/api-schema.yaml
```diff
@@ -780,6 +802,12 @@ components:
           type: string
           readOnly: true
           example: de7818e7-610a-4057-8f6f-b785dc1e6f88
+        kind:
+          type: string
+          example: stream
+        userId:
+          type: string
+          example: 66E2161C-7670-4D05-B71D-DA2D6979556F
```
victorges: Same

packages/api/src/schema/api-schema.yaml
```diff
@@ -887,6 +913,9 @@ components:
           example: 09F8B46C-61A0-4254-9875-F71F4C605BC7
         name:
           type: string
+        userId:
+          type: string
+          example: 66E2161C-7670-4D05-B71D-DA2D6979556F
```
victorges: Same.

Won't keep repeating this, but ideally we don't document this `userId` anywhere, and shouldn't be returning it on any API. Will the SDK tests break if we return a couple fields that are not in the schema?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2220,12 +2255,6 @@ components:
       type: object
       description: |
         An individual metric about usage of a user.
-      required:
-        - UserID
-        - CreatorID
-        - DeliveryUsageMins
-        - TotalUsageMins
-        - StorageUsageMins
```
victorges, line 2252: Are these not always returned?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2929,7 +2958,7 @@ paths:
           content:
             application/json:
               schema:
-                type: array
+                type: object
```
victorges: Hmm this seems wrong. `type: object` wouldn't have an `items:` prop below. Maybe missing changing `items { $ref: ... }` to just `$ref` on line below?

packages/api/src/schema/api-schema.yaml
```diff
@@ -3039,6 +3068,12 @@ paths:
         To create a new webhook, you need to make an API call with the events
         you want to listen for and the URL that will be called when those events
         occur.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/webhook-payload"
```
victorges: This is not the payload used in the code: https://github.com/livepeer/livepeer-com/blob/d93c77ce4ea801b0a4a1f3443f8c5a45139b8b6e/packages/api/src/controllers/webhook.ts#L146

Maybe we don't need the 2 schemas? Perhaps just `webhook` could be used everywhere.

packages/api/src/schema/api-schema.yaml
```diff
@@ -4154,9 +4195,6 @@ paths:
                 description: |
                   A simplified metric object about aggregate viewership of an
                   asset. Either playbackId or dStorageUrl will be set.
-                required:
-                  - viewCount
-                  - playtimeMins
```
victorges, line 4183: Same, are these not always returned?

packages/api/src/schema/db-schema.yaml
```diff
@@ -587,12 +587,9 @@ components:
       table: webhook
       properties:
         kind:
-          type: string
-          example: webhook
           readOnly: true
```
victorges: This is not a custom field AFAIK. It exists in OpenAPI spec: https://swagger.io/docs/specification/data-models/data-types/#readonly-writeonly

If we're moving the field to public docs (api-schema), this prop should also go with it. It is part of the spec/definition.

Same for many other fields below, won't repeat the comment.

packages/www/pages/app/stream/[id].tsx
```diff
@@ -258,7 +258,10 @@ const ID = () => {
       } else if ("errors" in info) {
         throw new Error(info.errors.toString());
       }
+
+      // @ts-ignore
       setStream(info.stream);
+      // @ts-ignore
       setStreamOwner(info.user);
```
Comment (victorges) [inline packages/www/pages/app/stream/[id].tsx]: What's this about? 👀

victorges (approved):
Approved anyway but there are some minor comments.

packages/api/src/schema/api-schema.yaml
suhailkakar: yeah that is correct, create webhook also uses this payload, but we can't use webhook, since it has many other param which user can't/shouldn't update like userId

packages/api/src/schema/api-schema.yaml
suhailkakar: yeah, unfortunately both SDK and API should have exactly the same object fields, but we can mark it as deprecate from the SDK side if helpful.

packages/www/pages/app/stream/[id].tsx
Comment (suhailkakar) [reply packages/www/pages/app/stream/[id].tsx]: just testing the vercel deployment, will revert it back this one :)

packages/api/src/schema/api-schema.yaml
suhailkakar: As mentioned above, both SDK and API should have exactly the same responses. Since we have `userId` on rest api, we would also need for the SDK as well. If we plan to remove this, we should do the same for the rest api as well.

packages/api/src/schema/api-schema.yaml
suhailkakar: > As mentioned above, both SDK and API should have exactly the same responses. Since we have userId on rest api, we would also need for the SDK as well. If we plan to remove this, we should do the same for the rest api as well.

packages/api/src/schema/api-schema.yaml
suhailkakar: > As mentioned above, both SDK and API should have exactly the same responses. Since we have userId on rest api, we would also need for the SDK as well. If we plan to remove this, we should do the same for the rest api as well.

packages/api/src/schema/db-schema.yaml
suhailkakar: sounds good, will add readOnly/writeOnly to api-schema as well

packages/api/src/schema/api-schema.yaml
suhailkakar: I guess, we could just replace it with this:

```
application/json:
              schema:
                $ref: "#/components/schemas/multistream-target"
```

as the API returns object not array

packages/api/src/schema/api-schema.yaml
suhailkakar: Probably, it is removed by mistake, added it back :)

packages/api/src/schema/api-schema.yaml
suhailkakar: yes, but `webhook` also has many other param like `userId` which user can't send on request body. I will update the payload in code as well

packages/api/src/schema/api-schema.yaml
suhailkakar: the required items are just for response not request payload and sometimes, if there is data, the api returns and other times it doesn't, which is why i removed the required fields as it really doesn't have impact on user side validation

packages/api/src/schema/api-schema.yaml
suhailkakar: > the required items are just for response not request payload and sometimes, if there is data, the api returns and other times it doesn't, which is why i removed the required fields as it really doesn't have impact on user side validation

packages/api/src/schema/api-schema.yaml
victorges: If the field is `readOnly`, it cannot be sent on request payloads (correspondingly, `writeOnly` don't come in responses). That's the ideal way of setting up these schemas, but if it's not possible for some reason no problem, we can repeat the schema for request payload.

packages/api/src/schema/api-schema.yaml
victorges: Yeah that makes sense!

packages/api/src/schema/api-schema.yaml
victorges: Yeah definitely! Please tag all `kind` and `userId` fields as deprecated 🙏

packages/api/src/schema/api-schema.yaml
victorges: Ok, as long as we use it on the code as well (as you mentioned in the other comment you'll do), not only the schema, this is fine!

victorges (approved):
Was already approved but just confirming, LGTM

packages/api/src/schema/api-schema.yaml
```diff
@@ -548,18 +558,14 @@ components:
               example:
                 Authorization: "Bearer 123"
             isMobile:
-              type:
-                - boolean
```
0xcadams, line 548: @victorges this was invalid spec on the API schema and broke the SDK generation

packages/api/src/schema/api-schema.yaml
```diff
@@ -3466,6 +3464,26 @@ paths:
               $ref: "#/components/schemas/new-asset-from-url-payload"
       responses:
         "200":
+          description: Import in progress
```
victorges: Let's call it `Upload` for consistency
```suggestion
          description: Upload in progress
```

packages/api/src/schema/api-schema.yaml
```diff
@@ -3466,6 +3464,26 @@ paths:
...
+                required:
+                  - asset
+                  - task
+                properties:
+                  asset:
+                    $ref: "#/components/schemas/asset"
+                  task:
+                    type: object
+                    properties:
+                      id:
```
victorges, line 3494: Can also make this `required`

victorges (approved):
Reviewed the changes since last time. LGTM!

0xcadams (approved):

PR #2157, opened 2024-04-30
move auth routes to /dashboard

adamsoffer:
Moved the following routes inside /dashboard so that we can run a proxy to those routes when the site moves to webflow.

```
https://livepeer.studio/dashboard/register
https://livepeer.studio/dashboard/login
https://livepeer.studio/dashboard/forgot-password
https://livepeer.studio/dashboard/verify
https://livepeer.studio/dashboard/verify-new-email
```


iameli (changes requested):
Maybe let's add redirects for the old routes on this app too, so this can be deployed incrementally prior to the Webflow change?

adamsoffer:
> Maybe let's add redirects for the old routes on this app too, so this can be deployed incrementally prior to the Webflow change?

@iameli I put [these](https://github.com/livepeer/studio/pull/2157/files#diff-3f4e3e3df38174254eb3b3188c64083c4200c2b6b989854d2bc6045f883e3ff9L190) in place. Are there any others I missed?

iameli:
@adamsoffer Ah nice I missed that. Will redirects using that mechanism keep query strings in place? Because that'll be needed for email verification and stuff to work. Assuming it does, we're probably good

adamsoffer:
> @adamsoffer Ah nice I missed that. Will redirects using that mechanism keep query strings in place? Because that'll be needed for email verification and stuff to work. Assuming it does, we're probably good

@iameli just [made an update](https://github.com/livepeer/studio/pull/2157/commits/00dc5db81f47c5b10f857d565bcb7bfff0beeda2) to the redirects to ensure query strings are still in place.

iameli (approved):

PR #2162, opened 2024-05-02
api,www: Create `failed` state for recordings

victorges:

This is just to make a small improvement on the recordingStatus field of streams/sessions.

Previously we had a `none` state which was only set when we tried to create the recording but
the processing failed. This is pretty misleading as it looks like we never even tried creating one.

So I created a new `failed` state to be set on that case, and am setting `none` only when we
actually won't generate a recording, which is the case for stale sessions (exists today, will stop
happening with #2013).

- Add new `failed` status and set it when processing failed
- Process it on frontend showing "Failed" when that's the case

Simple hotfix, `yarn test` guarantees it's not breaking anything.

No.

**Checklist**


gioelecerati (approved):
lgtm

PR #2163, opened 2024-05-03
project: user: create default project id

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/api-token.js
+++ b/packages/api/src/controllers/api-token.js
@@ -27,7 +27,6 @@ async function enrichApiTokenProjectId(body, req, res) {
   return body;
 }
 
-// Apply the middleware globally
 app.use(mung.jsonAsync(enrichApiTokenProjectId));
 
 app.get("/:id", async (req, res) => {
--- a/packages/api/src/controllers/api-token.js
+++ b/packages/api/src/controllers/api-token.js
@@ -6,10 +6,29 @@ import { makeNextHREF, parseOrder, parseFilters } from "./helpers";
 import { authorizer, validatePost } from "../middleware";
 import { AuthPolicy } from "../middleware/authPolicy";
 import { db } from "../store";
+import mung from "express-mung";
 
 const app = Router();
 
 app.use(authorizer({ noApiToken: true }));
+
+async function enrichApiTokenProjectId(body, req, res) {
+  const enrichToken = (token) => {
+    if (!token.projectId || token.projectId === "") {
+      token.projectId = req.user.defaultProjectId;
+    }
+  };
+
+  if (Array.isArray(body)) {
+    body.forEach(enrichToken);
+  } else {
+    enrichToken(body);
+  }
+  return body;
+}
+
+// Apply the middleware globally
+app.use(mung.jsonAsync(enrichApiTokenProjectId));
 
 app.get("/:id", async (req, res) => {
   const { id } = req.params;
--- a/packages/api/src/controllers/api-token.js
+++ b/packages/api/src/controllers/api-token.js
@@ -2,7 +2,12 @@ import Router from "express/lib/router";
 import { v4 as uuid } from "uuid";
 import sql from "sql-template-strings";
 
-import { makeNextHREF, parseOrder, parseFilters } from "./helpers";
+import {
+  makeNextHREF,
+  parseOrder,
+  parseFilters,
+  addDefaultProjectId,
+} from "./helpers";
 import { authorizer, validatePost } from "../middleware";
 import { AuthPolicy } from "../middleware/authPolicy";
 import { db } from "../store";
@@ -12,22 +17,7 @@ const app = Router();
 
 app.use(authorizer({ noApiToken: true }));
 
-async function enrichApiTokenProjectId(body, req, res) {
-  const enrichToken = (token) => {
-    if (!token.projectId || token.projectId === "") {
-      token.projectId = req.user.defaultProjectId;
-    }
-  };
-
-  if (Array.isArray(body)) {
-    body.forEach(enrichToken);
-  } else {
-    enrichToken(body);
-  }
-  return body;
-}
-
-app.use(mung.jsonAsync(enrichApiTokenProjectId));
+app.use(mung.jsonAsync(addDefaultProjectId));
 
 app.get("/:id", async (req, res) => {
   const { id } = req.params;
--- a/packages/api/src/controllers/api-token.js
+++ b/packages/api/src/controllers/api-token.js
@@ -75,10 +75,6 @@ app.get("/", async (req, res) => {
 
   if (!userId) {
     const query = parseFilters(fieldsMap, filters);
-    query.push(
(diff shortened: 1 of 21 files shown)


packages/api/src/controllers/api-token.js
```diff
@@ -108,9 +112,13 @@ app.get("/", async (req, res) => {
   }
   const query = parseFilters(fieldsMap, filters);
   query.push(sql`api_token.data->>'userId' = ${userId}`);
-  query.push(
-    sql`coalesce(api_token.data->>'projectId', '') = ${req.project?.id || ""}`
-  );
+  if (!req.user.admin) {
```
leszko, line 115: So, what happens with the admin user? Won't it have the default project?

packages/api/src/controllers/asset.test.ts
```diff
@@ -221,6 +221,15 @@ describe("controllers/asset", () => {
...
     expect(await res.json()).toBeDefined(); //api-key be retrieve if adminApiKey is used..
+
+    res = await client.get(
+      `/asset?limit=10&allUsers=true&filters=[{"id":"playbackId","value":"${asset.playbackId}"}]`
+    );
+    expect(res.status).toBe(200);
+    let assets = await res.json();
+    expect(assets).toHaveLength(1);
+    expect(assets[0].projectId).toBe(projectId);
+    expect(assets[0].playbackId).toBe(asset.playbackId);
```
leszko, line 232: Thanks for this 🙏

packages/api/src/controllers/asset.test.ts
```diff
@@ -254,6 +263,25 @@ describe("controllers/asset", () => {
     expect(project.id).toBeDefined();
   });
 
+  it("should import asset when projectId is not passed and list with projectId", async () => {
```
leszko, line 266: Should we also add some tests for the admin user case? Like, how the assets/streams are listed? Do we have something similar to `allUsers=true`, something like `allProjects=true`?

packages/api/src/controllers/asset.ts
```diff
@@ -613,14 +615,24 @@ export async function toExternalAsset(
   return a;
 }
 
+app.use(mung.jsonAsync(addDefaultProjectId));
```
leszko, line 618: I like using this mung here, I think it simplifies the code a lot 🏅

packages/api/src/controllers/helpers.ts
```diff
@@ -734,3 +736,71 @@ export function mapInputCreatorId(inputId: InputCreatorId): CreatorId {
...
+
+export function getProjectId(req: Request): string {
+  let projectId = req.user.defaultProjectId ?? "";
+  if (req.project?.id) {
+    projectId = req.project.id;
+  }
+  return projectId;
+}
+
+export async function addDefaultProjectId(
```
leszko, line 748: Is it possible to unit test this function? Or is it already tested (indirectly) with some unit tests?

packages/api/src/controllers/project.ts
```diff
@@ -139,4 +139,47 @@ app.post("/", authorizer({}), async (req, res) => {
   res.json(project);
 });
 
+app.patch("/:id", authorizer({}), async (req, res) => {
+  const project = await db.project.get(req.params.id, {
```
leszko, line 137: Nit: Could you clean imports in this file? I see a lot of them are unused.

packages/api/src/controllers/project.ts
```diff
@@ -139,4 +139,47 @@ app.post("/", authorizer({}), async (req, res) => {
...
+
+  if (!project || project.deleted) {
+    throw new NotFoundError(`project not found`);
+  }
+
+  if (req.user.admin !== true && req.user.id !== project.userId) {
+    throw new ForbiddenError("user can only delete their own projects");
+  }
+
+  //await db.project.delete(req.params.id);
```
leszko: 1. Could you please remove the commented-out code? 
2. How do we plan to address removing projects? Do we want to delete all related resources, but it takes too long so we don't do it here? What's the plan around it?

packages/api/src/controllers/user.ts
```diff
@@ -1477,4 +1500,51 @@ app.post(
   }
 );
 
+app.post(
+  "/migrate/userDefaultProject",
```
leszko, line 1506: When do we plan to call it? Do we plan to do some global migration of the users? Is it something we'll do once and then remove this endpoint? 

I think you can add some comment to explain that.

packages/api/src/middleware/auth.ts
```diff
@@ -189,6 +190,10 @@ function authenticator(): RequestHandler {
     req.token = tokenObject;
     req.user = user;
 
+    if (!projectId) {
```
leszko, line 193: Could you add returning `x-livepeer-project-id` from the `/api/auth` endpoint?

leszko (changes requested):
Good work @gioelecerati 🎖️ I really like where it's heading, plus the `mung` injection. I think it simplified a lot of things.

Added some inline comments, other than that I tried the following:
| Test | Result |
| ---- | -----   |
| non-admin token `/stream/<id>` | 🆗 | 
| non-admin token `/stream` | 🚫 no `projectId` included |
| non-admin token `/asset/<id>` | 🆗 |
| non-admin token `/asset` | 🚫 no `projectId` included |
| admin token `/stream/<id>` | 🚫 no `projectId` included |
| admin token `/stream` | 🚫 no `projectId` included |
| admin token `/asset/<id>` | 🚫 no `projectId` included |
| admin token `/asset` | 🚫 no `projectId` included |

Could you please check? I think we should always include `projectId` in the resource. I checked it in the box.

packages/api/src/controllers/api-token.js
gioelecerati: The admin user can see all the keys unregarding of the projectId - for example on the /app/admin page

packages/api/src/controllers/asset.test.ts
gioelecerati: This is exactly an admin case where the list happens with an adminKey - we don't have `allProjects=true` the admin is seeing all the assets and streams either way - I can add some tests

packages/api/src/controllers/helpers.ts
gioelecerati: It's not tested directly but by all the other tests that are expecting a projectId back - but good idea, I add tests specific to this

packages/api/src/controllers/project.ts
gioelecerati: 1. Sure!
2. The plan is to wait for this https://github.com/livepeer/studio/pull/2174 and make the deletion a cron job that lists files in queue to get deleted and do it in batches. We don't want to just delete them here, because they could be thousands

packages/api/src/controllers/user.ts
gioelecerati: We plan to call it on production deploy - it's done once and then removed
Sure can add some comments

leszko (approved):
LGTM

PR #2166, opened 2024-05-04
api/stream: Make sure we use patched payload on /pull create stream

victorges:

There was a bug in #2155 which didn't use the patched profiles when the stream
was created. This is because the `handleCreateStream` function just got the raw
request and fetched the `req.body` again.

- Pass the patched payload to `handleCreateStream` so it gets the Trovo-hacked profiles

`yarn test`

Informal.

**Checklist**


gioelecerati (approved):
LGTM

PR #2167, opened 2024-05-06
api/cannon: Stop retrying invalid webhooks forever

victorges:

Turns out there were never any suspended users sending webhooks. Only webhooks
that never got acked from the queue after they got suspended.

When we got an invalid webhook for a deleted or suspended user or stream,
we threw an error from the fucntion that processes the events. Problem was the
caller of that function was retrying on any generic errors.

Made them throw an explicit `UnprocessableEntityError` instead so we drop
the messages accordingly.

- Throw `UnprocessableEntityError` from main process func
- Handle that error and make sure we `ack` messages from them
- Remove explicitly handling for `recording.waiting` events, we can just rethrow now

`yarn test`

Not sure, we talked about it tho and the webhook queue is never empty.

**Checklist**

### DIFF
--- a/packages/api/src/webhooks/cannon.ts
+++ b/packages/api/src/webhooks/cannon.ts
@@ -101,7 +101,9 @@ export default class WebhookCannon {
     try {
       ack = await this.processWebhookEvent(event);
     } catch (err) {
-      ack = isRuntimeError(err);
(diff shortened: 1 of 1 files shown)


gioelecerati (approved):
LGTM

PR #2168, opened 2024-05-06
Reuse catalyst stream pull node if pull is locked

leszko:
This is a fix for the following scenario that happens from time to time on prod:
1. We get `/pull` request which is successful
2. A few seconds after (without getting `/terminate`) we receive another `/pull` request for the same stream
3. If it's redirected to a different catalyst node, then the pull fails, because we lock stream pull to one node

The workaround for this is to check if the given stream is currently locked (read: it's being pulled atm) and if yes, then trigger pulling from the same catalyst node. If we redirect to the same node, then Mist handles this.

Sample logs: https://eu-metrics-monitoring.livepeer.live/grafana/goto/wDCuqDYSg?orgId=1

fix https://linear.app/livepeer/issue/PS-518/more-pull-failuress


packages/api/src/controllers/stream.ts
```diff
@@ -233,6 +233,28 @@ async function triggerManyIdleStreamsWebhook(ids: string[], queue: Queue) {
...
 }
 
+export function resolvePullUrlFromExistingStreams(
+  existingStreams: DBStream[]
+): { pullUrl: string; pullRegion: string } {
+  if (existingStreams.length !== 1) {
+    return null;
+  }
+  const stream = existingStreams[0];
+  const lockLeaseTimeout = 60 * 1000; // 1 min
```
victorges: Should we hardcode this value on the lock API as well?

victorges (approved):
LGTM

Such race condition. I'd imagine that the second `/pull` request would simply be redirected to the node currently pulling it. Does this only happen when the second `/pull` arrives after the stream started but before the load balancer gets its state to redirect to the right node?

emranemran (approved):

leszko:
> Such race condition. I'd imagine that the second `/pull` request would simply be redirected to the node currently pulling it. Does this only happen when the second `/pull` arrives after the stream started but before the load balancer gets its state to redirect to the right node?

Yeah, from our perspective it happens, when Load Balancer redirects to a different node AND the stream is not available (so, Mist load balancer does not see it active). I believe that for Trovo it happens when a streamer starts streaming, stops streaming, and then suddenly starts again (maybe some retry from OBS). So, Trovo does not send `/terminate` properly.

packages/api/src/controllers/stream.ts
leszko: Good idea. Updated.

packages/api/src/controllers/stream.ts
```diff
@@ -233,6 +235,35 @@ async function triggerManyIdleStreamsWebhook(ids: string[], queue: Queue) {
...
+  }
+  const stream = existingStreams[0];
+  if (
+    stream.pullRegion &&
+    stream.pullLockedBy &&
+    ((stream.pullLockedAt &&
+      stream.pullLockedAt > Date.now() - DEFAULT_PULL_LOCK_LEASE_TIMEOUT) ||
+      (stream.lastSeen &&
+        stream.lastSeen > Date.now() - PULL_REUSE_SAME_NODE_TIMEOUT))
+  ) {
```
victorges: Could be more readable if we break these up in intermediate bools wdyt?
```suggestion
  const hasPullInfo = stream.pullRegion && stream.pullLockedBy;
  const pullLockedRecently = stream.pullLockedAt && stream.pullLockedAt > Date.now() - DEFAULT_PULL_LOCK_LEASE_TIMEOUT;
  const stoppedRecently = stream.lastSeen && stream.lastSeen > Date.now() - PULL_REUSE_SAME_NODE_TIMEOUT;
  if (hasPullInfo && (pullLockedRecently || stoppedRecently)) {
```

packages/api/src/controllers/stream.ts
```diff
@@ -233,6 +235,35 @@ async function triggerManyIdleStreamsWebhook(ids: string[], queue: Queue) {
...
+    return null;
+  }
+  const stream = existingStreams[0];
+  if (
+    stream.pullRegion &&
+    stream.pullLockedBy &&
+    ((stream.pullLockedAt &&
+      stream.pullLockedAt > Date.now() - DEFAULT_PULL_LOCK_LEASE_TIMEOUT) ||
+      (stream.lastSeen &&
+        stream.lastSeen > Date.now() - PULL_REUSE_SAME_NODE_TIMEOUT))
```
victorges: I wonder if we need to check if the stream already `isActive` as well? Or do we already check this somewhere else? Or do we _not_ want to trust that given the issues we were just discussing, that some of these streams are left for clean-up cause the `/setactive` doesn't work?

victorges (approved):
LGTM either way

packages/api/src/controllers/stream.ts
leszko: Yeah, I think that checking the `lastSeen` already covers it. And yeah, we don't tryst `isActive` 🙃

emranemran:
How do we test this to ensure no regressions?

leszko:
> How do we test this to ensure no regressions?

I think the same way as with other changes => Test pull stream in staging (with the stream stopped and started). And then monitor the prod. Any other thoughts how we could test it?

PR #2173, opened 2024-05-09
api/stream: Only clear fps if isMobile === 2

victorges:
Implements [PS-569](https://linear.app/livepeer/issue/PS-569/studio-change-ismobile-from-bool-to-int)


gioelecerati (approved):
lgtm

PR #2174, opened 2024-05-09
api: Create entrypoints for long-running jobs

victorges:

This is to implement a long-due change, to start doing long-running jobs as a 
separate execution instead of calling an API to do it. The idea is that these jobs
will run directly in our Kubernetes clusters in a necessarily isolated environment,
not making calls to the API and potentially competing for resources.

There are 3 jobs being created here:
- `active-cleanup`: Runs the logic to clean-up active streams that are lost. This was
just created in #2013 and is currently running on a GitHub action.
- `update-usage`: Runs the logic to update the usage tables every hour. This is crucial
for our user billing workflow as well. Currently it's also running on a GitHub action that
always fails because it doesn't complete under 1 minute.
- `create-db-tables`: This will simply create new `DB` clients to create all tables and
indexes. The intention is to run this as an `initContainer` of the API, so we don't need to
do it on the main container startup instead.

- Create CLI args to say which job should run (and some for custom args for each job)
- Call that separate job func from the startup
- Implement each of the jobs as separate functions on new `./jobs` dir
- Call the jobs from the APIs to avoid repetition (having the API might still be useful to run on demand)

For now mostly made sure it still compiles after the refactors. Will run each job separately to make sure it works.

Implement ENG-2034

**Checklist**

### DIFF
--- a/packages/api/src/cli.ts
+++ b/packages/api/src/cli.ts
@@ -4,6 +4,7 @@ import makeStreamInfoSvc from "./app/stream-info/stream-info-app";
 import "./dotenv";
 import makeApp from "./index";
 import runActiveCleanup from "./jobs/active-cleanup";
+import createDbTables from "./jobs/create-db-tables";
 import parseCli from "./parse-cli";
 
 if (require.main === module) {
@@ -12,6 +13,8 @@ if (require.main === module) {
     makeStreamInfoSvc(args);
   } else if (args.job === "active-cleanup") {
     runActiveCleanup(args);
+  } else if (args.job === "create-db-tables") {
+    createDbTables(args);
   } else {
     makeApp(args);
   }
--- a/packages/api/src/cli.ts
+++ b/packages/api/src/cli.ts
@@ -1,14 +1,17 @@
 #!/usr/bin/env node
 
+import makeStreamInfoSvc from "./app/stream-info/stream-info-app";
 import "./dotenv";
 import makeApp from "./index";
-import makeStreamInfoSvc from "./app/stream-info/stream-info-app";
+import runActiveCleanup from "./jobs/active-cleanup";
 import parseCli from "./parse-cli";
 
 if (require.main === module) {
   const args = parseCli();
   if (args.streamInfoService) {
     makeStreamInfoSvc(args);
+  } else if (args.job === "active-cleanup") {
+    runActiveCleanup(args);
   } else {
     makeApp(args);
   }
--- a/packages/api/src/cli.ts
+++ b/packages/api/src/cli.ts
@@ -3,21 +3,15 @@
 import makeStreamInfoSvc from "./app/stream-info/stream-info-app";
 import "./dotenv";
 import makeApp from "./index";
-import activeCleanup from "./jobs/active-cleanup";
-import createDbTables from "./jobs/create-db-tables";
-import updateUsage from "./jobs/update-usage";
+import { runJob } from "./jobs";
 import parseCli from "./parse-cli";
 
 if (require.main === module) {
   const args = parseCli();
   if (args.streamInfoService) {
     makeStreamInfoSvc(args);
-  } else if (args.job === "active-cleanup") {
-    activeCleanup(args);
-  } else if (args.job === "create-db-tables") {
-    createDbTables(args);
-  } else if (args.job === "update-usage") {
-    updateUsage(args);
+  } else if (args.job) {
+    runJob(args.job, args);
   } else {
     makeApp(args);
   }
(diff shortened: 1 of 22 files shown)


packages/api/src/cli.ts
```diff
@@ -1,14 +1,23 @@
...
 import parseCli from "./parse-cli";
 
 if (require.main === module) {
   const args = parseCli();
   if (args.streamInfoService) {
     makeStreamInfoSvc(args);
+  } else if (args.job === "active-cleanup") {
+    activeCleanup(args);
+  } else if (args.job === "create-db-tables") {
+    createDbTables(args);
```
leszko: I think I don't understand the idea of `createDbTables()`. Why do we want to run a separate `initContainer` to create DB tables?

leszko (commented):
Looks fine, added one inline comment. Other than that:

1. Could you try to add some unit testing? 
2. I see you created CronJobs as thick jobs actually doing the work. How would you compare it to the solution in which we would have thin CronJobs only doing the API calls to Studio API and Studio API doing the work? Not saying that your approach is worse. I just had in my mind that the k8 CronJob will only make API calls, but I see they'll execute the whole logic.

packages/api/src/cli.ts
victorges: By running an `initContainer` we can have a more controlled environment for creating the tables and indexes, which has been taking a long time these days. I think the main advantages are:
- Simpler healthchecks for the main container. After https://github.com/livepeer/livepeer-infra/pull/2097 we changed the `startupProbe` config to allow the API to take up to 2 minutes to start with a granularity (check period) of 10 seconds. With the `initContainer` we will be done with the startup as soon as it's done and have a more reasonable/faster startup delay for the API, maybe sth like up to 20s checking ever 2-5s or sth like that.
- Quicker turnaround in case of container crashes. I think this is the biggest benefit. Today, if an API container crashes for any reason, it will start again and try to do an expensive operation which is creating tables and indexes on the DB all over again. This often causes crash loops when there is some network issue going on, since the chance of getting more errors on that startup is increased. With an `initContainer` (and `create-db-tables` disabled on main container), the main API container will startup really fast, being only a matter of connecting to DB and Rabbit again.

victorges:
> Could you try to add some unit testing?

Yeah sure. The active-cleanup already has tests on the `stream.test` api, but I can make a test on the jobs just to make sure they run well as well. I'll see what I can do for the other jobs (index should be easier, but not sure how to test the update-usage, I think there are no tests for it).

> I see you created CronJobs as thick jobs actually doing the work. How would you compare it to the solution in which we would have thin CronJobs only doing the API calls to Studio API and Studio API doing the work? Not saying that your approach is worse. I just had in my mind that the k8 CronJob will only make API calls, but I see they'll execute the whole logic.

Some benefits from the top of my head:
- No long running connections. Currently the `update-usage` job fails because it runs for longer than 60s. This is only because it goes through the CDN which enforces that timeout, but we could still have connections dropping internally in the cluster if they're running for too long.
- Further isolated resources from the APIs. Even though I already created the separate jobs DB pool, this pool is still used for other background operations like tasks and webhooks handling. These jobs also require some CPU time which would compete with other operations in the API. By moving to jobs, these are also isolated and the jobs won't compete for connections or CPU time with other API operations (will still share the DB itself tho, obv)
- Easier monitoring. The jobs will get [their own `app`s on Loki](https://eu-metrics-monitoring.livepeer.live/grafana/goto/ByJkQ-YSg?orgId=1) from which we could check their logs in isolation. We could also turn on verbose logging for jobs so we can see every query that they run on the DB for example, which is really useful for monitoring, debugging and accountability of these jobs over time. (still need to make these configurable, but I enabled verbose logging when I was testing active-cleanup on staging and it was super useful)

WDYT?

victorges:
Forgot to tag you on above msg cc @leszko

leszko:
> WDYT?

I think it makes sense :)

packages/api/src/cli.ts
leszko: The drawback is that now studio api app will be less "self-contained". You'll need to know that before using it you need to run another up to set up the DB. 

Saying that, if the current approach caused us issues, then I'm with having this as an `initContainer`.

victorges: My idea is to keep the index creation on by default!
https://github.com/livepeer/livepeer-com/blob/9ffcbf58e551cd0354a89bf54d78010a4a8b7d1c/packages/api/src/parse-cli.ts#L156-L161

So if you just run the API, it will create the indexes as usual. Only if you want to run the separate "create indexes" step that you would:
- disable `postgres-create-tables` on the main container
- define the `initContainer` (or whatever else on your stack) to create the indexes instead

victorges: Ohh just another another idea too: In the future, maybe we can even make these not an `initContainer` but a step in the `livepeer-infra` deployment itself. So we run only once when deploying a new API version and all the API pods can start immediately without spending dozens of seconds creating indexes.

leszko (approved):

PR #2176, opened 2024-05-14
api: access-control: allowedOrigins

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.test.ts
+++ b/packages/api/src/controllers/access-control.test.ts
@@ -243,22 +243,19 @@ describe("controllers/access-control", () => {
     });
 
     it("should not allow playback if origin is not in playback.allowedOrigins", async () => {
-      const webhook = await db.webhook.create({
-        id: uuid(),
-        name: "test",
-        url: `http://localhost:3004/api/access-control/webhook-test`,
-        events: ["playback.accessControl"],
-      });
-      gatedAsset.playbackPolicy.webhookId = webhook.id;
-      gatedAsset.playbackPolicy.type = "webhook";
+      gatedAsset.playbackPolicy.type = "jwt";
       gatedAsset.playbackPolicy.allowedOrigins = ["http://localhost:3000"];
       await db.asset.update(gatedAsset.id, {
         playbackPolicy: gatedAsset.playbackPolicy,
       });
+      let asset = await db.asset.get(gatedAsset.id);
+      expect(asset.playbackPolicy.allowedOrigins).toEqual([
+        "http://localhost:3000",
+      ]);
       const res3 = await client.post("/access-control/gate", {
         stream: `video+${gatedAsset.playbackId}`,
-        type: "accessKey",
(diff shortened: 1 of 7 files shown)


packages/api/src/controllers/access-control.test.ts
```diff
@@ -242,6 +242,48 @@ describe("controllers/access-control", () => {
...
+        stream: `video+${gatedAsset.playbackId}`,
+        type: "jwt",
+        pub: "notExistingPubKey",
+        webhookPayload: {
+          headers: {
+            origin: "http://localhost:3000",
+          },
+        },
+      });
+      expect(res4.status).toBe(403);
```
victorges, line 280: Could you add one example that passes all checks as well?

packages/api/src/controllers/access-control.ts
```diff
@@ -242,6 +243,24 @@ app.post(
       return res.end();
     }
 
+    let origin: string = null;
+    if (req.body?.webhookPayload?.headers) {
+      origin = req.body?.webhookPayload?.headers["origin"];
+    }
```
victorges: nit: could do in one line if you want
```suggestion
    const origin = req.body?.webhookPayload?.headers?.origin;
```

packages/api/src/controllers/asset.ts
```diff
@@ -287,6 +287,24 @@ async function validateAssetPlaybackPolicy(
...
     }
+    const allowedOrigins = playbackPolicy?.allowedOrigins;
+    if (allowedOrigins) {
+      if (allowedOrigins.length > 0) {
+        const allowedOriginsValid = allowedOrigins.every((origin) => {
+          const url = new URL(origin);
+          return (
+            ["http:", "https:"].includes(url.protocol) &&
+            url.hostname &&
+            url.port
```
victorges: Not sure if necessary (or even correct), but should we check that there is no path or the path is just `/`? I'm not sure if there could be a path tho, gotta read spec

packages/api/src/controllers/asset.ts
```diff
@@ -287,6 +287,24 @@ async function validateAssetPlaybackPolicy(
         `webhook ${playbackPolicy.webhookId} not found`
       );
     }
+    const allowedOrigins = playbackPolicy?.allowedOrigins;
+    if (allowedOrigins) {
+      if (allowedOrigins.length > 0) {
+        const allowedOriginsValid = allowedOrigins.every((origin) => {
+          const url = new URL(origin);
```
victorges: You might need to try catch this to throw a friendly error to the user. Or does it never explode here? Maybe this present is very lenient to invalid input, gpt told me that once lol

victorges (approved):
LGTM

packages/api/src/controllers/asset.ts
gioelecerati: They should never put the path, even with the final slash, I will add a check

PR #2178, opened 2024-05-14
fix: change API schema to allow null for transcode profile

0xcadams:


0xcadams:
closing in favor of https://github.com/livepeer/studio/pull/2156

PR #2179, opened 2024-05-15
fix: change to nullable create/patch stream

0xcadams:


victorges (approved):
LGTM for the schema part

mjh1 (approved):

PR #2180, opened 2024-05-17
www: multistream: support question mark in stream key

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
+++ b/packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
@@ -69,15 +69,6 @@ const addQuery = (
 
 const composeIngestUrl = (ingestUrlStr: string, streamKeyStr: string) => {
   const ingestUrl = parseUrlSafe(ingestUrlStr);
-  if (streamKeyStr) {
-    if (streamKeyStr.includes("?")) {
-      streamKeyStr += "?";
-    }
-  } else {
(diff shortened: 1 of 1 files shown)


packages/www/components/StreamDetails/MultistreamTargetsTable/SaveTargetDialog.tsx
```diff
@@ -76,6 +76,9 @@ const composeIngestUrl = (ingestUrlStr: string, streamKeyStr: string) => {
   switch (ingestUrl.protocol) {
     case "rtmp:":
     case "rtmps:":
+      if (streamKeyStr.includes("?") && !streamKeyStr.endsWith("?")) {
```
victorges, line 79: Can you add a comment explaining why this is necessary?

victorges: Also, why shouldn't we add the `?` if the `streamKey` ends in a `?`? Don't we still want to keep it there so we'd add another question mark?

In the end, maybe the logic should be to always add a question mark? It will be removed anyway cause it's always the last, be it the only one or not. Unless I misunderstood how mist fumbles with it

victorges (approved):
LGTM, but as we chatted offline now, I'd rather have this logic be inside catalyst which is our interface to Mist. We shouldn't be saving these dirty URLs in our DB.

gioelecerati:
Closed in favor of catalyst-api fix

PR #2181, opened 2024-05-20
api/task: Prevent tasks from suspended users from running

victorges:
In order to fix the `sao` crisis, I wanna clear at least all the tasks from `record-tester` that are
in the queue, which turns out get executed even if the user is suspended.


gioelecerati (approved):
LGTM, do we want to check also user.disabled here?

PR #2183, opened 2024-05-21
access-control: use referer on redirects

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -243,9 +243,15 @@ app.post(
       return res.end();
     }
 
-    const origin = req.body?.webhookPayload?.headers?.Origin;
+    let origin = req.body?.webhookPayload?.headers?.Origin;
 
     if (origin) {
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -243,9 +243,15 @@ app.post(
       return res.end();
     }
 
-    const origin = req.body?.webhookPayload?.headers?.Origin;
+    let origin = req.body?.webhookPayload?.headers?.Origin;
 
     if (origin) {
```
victorges, line 248: Shouldn't we try fetching from the Referrer if origin is false here too?

victorges (approved):
LGTM

packages/api/src/controllers/access-control.ts
gioelecerati: No! Only when the origin is set as a string `null`, because that's what the browser does during redirects. If it's false then it's not coming in entirely

PR #2197, opened 2024-06-03
api: stream: add cache to /stream/playback

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -840,7 +840,7 @@ app.get("/:id", authorizer({}), async (req, res) => {
   res.json(db.stream.addDefaultFields(stream));
 });
 
-// returns stream by steamKey
+// returns stream by playbackId
 app.get("/playback/:playbackId", authorizer({}), async (req, res) => {
   const stream = await cache.getOrSet(
     `strm-stream-by-playback-${req.params.playbackId}`,
@@ -872,7 +872,7 @@ app.get("/playback/:playbackId", authorizer({}), async (req, res) => {
   );
 });
 
-// returns stream by steamKey
+// returns stream by streamKey
 app.get("/key/:streamKey", authorizer({}), async (req, res) => {
   const useReplica = req.query.main !== "true";
   const [docs] = await db.stream.find(
--- a/packages/api/src/controllers/stream.ts
+++ b/packages/api/src/controllers/stream.ts
@@ -54,6 +54,7 @@ import {
 } from "./helpers";
 import { toExternalSession } from "./session";
 import wowzaHydrate from "./wowza-hydrate";
+import { cache } from "../store/cache";
 
 type Profile = DBStream["profiles"][number];
 type MultistreamOptions = DBStream["multistream"];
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/stream.ts
```diff
@@ -841,12 +842,21 @@ app.get("/:id", authorizer({}), async (req, res) => {
 
 // returns stream by steamKey
```
victorges: mind fixing this comment while you're here 😆 
```suggestion
// returns stream by playbackId
```

packages/api/src/controllers/stream.ts
```diff
@@ -841,12 +842,21 @@ app.get("/:id", authorizer({}), async (req, res) => {
...
+      const {
+        data: [stream],
+      } = await req.store.queryObjects<DBStream>({
+        kind: "stream",
+        query: { playbackId: req.params.playbackId },
+      });
+
+      return stream;
+    },
+    5
```
victorges, line 857: Is this the TTL? If so, any reason to have it so short? Is it gonna have any gain at all?

AFAIK this API is not widely used, and the cache is local to the API instance. So a cache of 5s would hardly ever get any hits.

I'd go with 60s at least.

victorges (approved):
LGTM

packages/api/src/controllers/stream.ts
gioelecerati: The issue is explained here https://discord.com/channels/423160867534929930/1102968315137228900/1246573851358265384

It's true that it's not widely used, but go-api-client is pointing to that via `GetStreamByPlaybackId` and this endpoint is getting flooded by calls due to a `lockPull` bug that Rafal is fixing.
We're receiving around 4k req/s to this endpoint during the fails, so this is to mitigate the amount of queries going into postgres. We probably want to call the /stream endpoint instead from go-api-client, but that's without cache too.

5s because the active state is needed, but we can increase, maybe to 10 or 15, I think it's a @leszko call depending on how much the state of the stream needs to be updated.

victorges: Thanks for explaining. This would have been super helpful context to add on the PR description. It also stays in the git history if anyone asks why it was done. Please try to add some helpful descriptions next times instead of just a title.

5s sounds fine if it's only coming in surges.

packages/api/src/controllers/stream.ts
```diff
@@ -839,14 +840,23 @@ app.get("/:id", authorizer({}), async (req, res) => {
   res.json(db.stream.addDefaultFields(stream));
 });
 
-// returns stream by steamKey
+// returns stream by playbackId
 app.get("/playback/:playbackId", authorizer({}), async (req, res) => {
```
leszko, line 844: I'm a little lost. I thought that the issue was that we were flooding with `/lockPull` requests, so why to add the cache to `/playback/<playbackId>`? 

It the DB peak was related to the `GET /playback/<playbackId>`?

gioelecerati: The flood I am seeing is to that endpoint, caused by catalyst-api using the method `GetStreamByPlaybackId` from go-api-client

leszko: Ok 👌

PR #2201, opened 2024-06-11
db: index: asset phase, deleted, deletedAt

gioelecerati:

Adding indexes for
- asset.deleted
- asset.deletedAt
- asset.status.phase

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -957,7 +957,6 @@ components:
         deleted:
           type: boolean
           description: Set to true when the asset is deleted
-          index: true
         sourcePlaybackReady:
           type: boolean
           description: Whether source playback is ready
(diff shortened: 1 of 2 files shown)


leszko (approved):
LGTM 👍 

One thing to maybe double check (google in the Internet 🙃 ) is if applying indexes to existing DB with 3M entries is ok, in other words if applying indexes will not be a very slow opretion. Though, I guess for 3M entries, it should be fine.

packages/api/src/schema/db-schema.yaml
```diff
@@ -1018,6 +1019,39 @@ components:
...
+              type: number
+              description:
+                Timestamp (in milliseconds) at which the asset was last updated
+              example: 1587667174725
+            progress:
+              type: number
+              description: Current progress of the task creating this asset.
+            errorMessage:
+              type: string
+              description: Error message if the asset creation failed.
```
victorges: We should not repeat the whole object, but just add the fields that are missing (`index`). The anti-pattern has spread in this file so it's hard to not follow it, but any time you see repetition it can be removed!
```suggestion
        status:
          properties:
            phase:
              index: true
```

packages/api/src/schema/db-schema.yaml
```diff
@@ -1018,6 +1019,39 @@ components:
           description:
             Timestamp (in milliseconds) at which the asset was deleted
           example: 1587667174725
+          index: true
```
victorges: I don't think we need all these indexes, let's chat about it to minimize how many we create here.

packages/api/src/schema/db-schema.yaml
```diff
@@ -957,6 +957,7 @@ components:
         deleted:
           type: boolean
           description: Set to true when the asset is deleted
+          index: true
```
victorges: Same, not sure if this is necessary.

victorges (approved):
LGTM

victorges:
> One thing to maybe double check (google in the Internet 🙃 ) is if applying indexes to existing DB with 3M entries is ok, in other words if applying indexes will not be a very slow opretion. Though, I guess for 3M entries, it should be fine.

Only saw this now. FTR we create indexes using the CONCURRENTLY directive which doesn't lock the table while doing it. It has been historically fine running it on the prod DB, only delayed startup a bit sometimes (not anymore with the initContainer only on `mad` region 😄)

PR #2202, opened 2024-06-11
api/stripe: Avoid sleeping too much when reporting usage

victorges:
Instead of adding additional 100ms delay, let's only make sure that the whole function executes for at least 100ms. This might make these jobs much quicker.


gioelecerati (approved):
lgtm

PR #2203, opened 2024-06-11
api/jobs/update-usage: Process users in parallel and allow cancellation

victorges:
This improves the `update-usage` job by processing users in parallel (up to 10 at a time).

The job is currently taking ~70 minutes to run so I hope this would bring it under 10min.

- Respect SIGTERM from the job runtime
- Paralelize user usage reporting
- Fix console.log logs to logger

`yarn test`

Fixes issue noticed after jobs reply, not tracked to Linear.

**Checklist**


gioelecerati (approved):
Ooooh thank you for doing this :heart: lgtm - let's make sure that everything is getting reported correctly in staging and in dev stripe

PR #2204, opened 2024-06-11
fix: update old workspace deps

0xcadams:
This updates old dependencies in the root of the workspace.

I think there was something to do with Box for the workspace build across the entire project. If this breaks, I think Box needs to update its scripts to include a call to build. This should not happen on each `yarn install`


victorges (approved):
neat!

PR #2222, opened 2024-06-18
Fix thumbs for direct uploads

mjh1:


mjh1:
I'm thinking we may want to add coverage for this in MediaGuard? Unless @victorges you think it should be easy to add a unit test here, I couldn't see an existing one I could piggy back on.

packages/api/src/controllers/asset.ts
```diff
@@ -980,6 +980,10 @@ app.post(
           url: downloadUrl,
           catalystPipelineStrategy: catalystPipelineStrategy(req),
           encryption,
+          thumbnails: !(await isExperimentSubject(
+            "vod-thumbs-off",
+            req.user?.id,
+          )),
```
victorges, line 986: If you want to have this logic for all assets, might wanna check usages of this `createTask` function. I know there's at least recordings tasks which are created from `cannon.ts`

victorges (approved):
LGTM

victorges:
re tests: I think we are fine without a test here, it's a really simple logic. I also understood that MediaGuard was supposed to test only the basic stuff for a continuous test,  while we test all features like this from E2E tests instead. So maybe those would be better? (probably on `catalyst` with the box?)

PR #2223, opened 2024-06-18
Add viewerLimit to Gate API

leszko:
Related to: https://www.notion.so/livepeer/Cap-50k-Fishtank-viewers-a828ef8d037a4f27b9a6769d91c50b23

### DIFF
--- a/packages/api/src/controllers/access-control.ts
+++ b/packages/api/src/controllers/access-control.ts
@@ -36,6 +36,8 @@ const app = Router();
 type GateConfig = {
   refresh_interval: number;
   rate_limit: number;
+  viewer_limit_per_user: number;
+  user_id: string;
 };
 
 async function fireGateWebhook(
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -36,6 +36,8 @@ const app = Router();
 type GateConfig = {
   refresh_interval: number;
   rate_limit: number;
+  viewer_limit_per_user: number;
```
victorges: nit: WDYT of viewer_limit_for_user or user_viewer_limit? I was slightly confused with "per" for a second as it sounded like there would be multiple users in the stream (e.g. "playback users", even tho it wouldn't make sense).

victorges (approved):
LGTM with a small, optional, suggestion on the field name for catalyst

thomshutt (approved):

mjh1 (approved):

packages/api/src/controllers/access-control.ts
leszko: Updated

PR #2228, opened 2024-06-26
api: migration: flag inactive and fake users

gioelecerati:

- update the migration endpoints for projects to also flag inactive and fake users as `toBeDeleted`, looking forward to user cleanup
- allow a dry run for the endpoint

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -1536,7 +1536,7 @@ app.post(
     const OLD_USER_CUTOFF = Date.now() - ms("12 months");
 
     for (const user of users) {
-      if (isFakeEmail(user.email)) {
+      if (isFakeEmail(user.email) && !user.admin) {
         if (actuallyMigrate) {
           await db.user.update(user.id, { toBeDeleted: true });
         }
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -1516,8 +1516,9 @@ app.post(
   "/migrate/userDefaultProject",
   authorizer({ anyAdmin: true }),
   async (req, res) => {
-    // parse limit from querystring
+    // parse limit and actually_migrate from querystring
     const limit = parseInt(req.query.limit?.toString() || "100");
+    const actuallyMigrate = req.query.actually_migrate === "true";
 
     const [users] = await db.user.find(
       [
@@ -1536,7 +1537,9 @@ app.post(
 
     for (const user of users) {
       if (isFakeEmail(user.email)) {
-        await db.user.update(user.id, { toBeDeleted: true });
+        if (actuallyMigrate) {
+          await db.user.update(user.id, { toBeDeleted: true });
+        }
         results.flagged.push({ id: user.id, email: user.email });
         continue;
       }
@@ -1553,7 +1556,9 @@ app.post(
           tokens.every((token) => token.lastSeen < OLD_USER_CUTOFF);
 
         if (allTokensOld) {
-          await db.user.update(user.id, { toBeDeleted: true });
+          if (actuallyMigrate) {
+            await db.user.update(user.id, { toBeDeleted: true });
+          }
           results.flagged.push({ id: user.id, email: user.email });
           continue;
         }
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -1525,28 +1528,62 @@ app.post(
...
-    const results = [];
+    const results = {
+      migrated: [],
+      flagged: [],
+    };
+
+    const OLD_USER_CUTOFF = Date.now() - ms("12 months");
 
     for (const user of users) {
+      if (isFakeEmail(user.email) && !user.admin) {
```
victorges: I think we're better off using the existing Sendgrid email validation than just a lib that checks if the string is a valid email: https://github.com/livepeer/livepeer-com/blob/bd525fc10ed62abba126471a35d191c7b8d23237/packages/api/src/controllers/helpers.ts#L435

It will also use the data that Sendgrid itself has.

victorges: We should also check the `user.emailValid` flag here. If the user has already verified their email we shouldn't trust neither `fakefilter` nor Sendgrid haha.

packages/api/src/controllers/user.ts
```diff
@@ -1525,28 +1528,62 @@ app.post(
       { limit },
     );
 
-    const results = [];
+    const results = {
+      migrated: [],
+      flagged: [],
+    };
+
+    const OLD_USER_CUTOFF = Date.now() - ms("12 months");
```
victorges: Can you make this a constant in the file, not in the HTTP handler, so we're not constantly calculating this date? Also, can you make this a well-defined fixed date instead of a computation that depends on `Date.now()`? e.g. Just pick the timestamp corresponding to 2023/06/26 and hardcode that in the constant.

packages/api/src/controllers/user.ts
```diff
@@ -1525,28 +1528,62 @@ app.post(
...
+        user.lastSeen < OLD_USER_CUTOFF &&
+        user.lastStreamedAt < OLD_USER_CUTOFF;
+
+      if (isOldUser && isInactive) {
+        const [tokens] = await db.apiToken.find({ userId: user.id });
+        const allTokensOld =
+          tokens.length === 0 ||
+          tokens.every((token) => token.lastSeen < OLD_USER_CUTOFF);
+
+        if (allTokensOld) {
```
victorges: **nit:** I think this whole logic could be made clearer by having a single condition like 
```
shouldDeleteUser = !user.admin && (isFakeEmail(user.email) || (isOldUser && isInactive && allTokensOld))
```
WDYT?

packages/api/src/controllers/user.ts
```diff
@@ -1509,13 +1510,15 @@ app.post(
...
+// This is also flagging users to be deleted for cleanup
 // To call once and then we can remove it
 app.post(
   "/migrate/userDefaultProject",
   authorizer({ anyAdmin: true }),
   async (req, res) => {
-    // parse limit from querystring
+    // parse limit and actually_migrate from querystring
     const limit = parseInt(req.query.limit?.toString() || "100");
+    const actuallyMigrate = req.query.actually_migrate === "true";
```
victorges, line 1524: Nice I like having the dry run approach

victorges (approved):
LGTM either way

packages/api/src/controllers/user.ts
victorges: Actually, making this a fixed value is way more important than moving to a constant in the file. That's a required change otherwise the behavior of this migration API is not deterministic.

All other comments optional.

packages/api/src/controllers/user.ts
gioelecerati: Yeah, but fakefilter is also checking for temporary emails! For example, generated from temp-email.org
They may be verified, but they are disposable so the user won't receive more mails and it's certainly a fake one. wdyt?

packages/api/src/controllers/user.ts
gioelecerati: Sure, makes sense, harcoding the millis unix timestamp for that date

packages/api/src/controllers/user.ts
gioelecerati: Followed your suggestion! should be much clearer now

packages/api/src/controllers/user.ts
victorges: Hmmm got it. Yeah that's sketchy then. WDYT of having both Sendgrid and fakefilter then, but only logging (or flagging differently) if we see a user that is verified but didn't pass in one of those checks?

gioelecerati: Cool, added the sendgrid check and the flag in the user object

PR #2234, opened 2024-06-28
api: projects: enable back projectId filter & remove migration api

gioelecerati:

- Enable back projectId filtering in check resource access
- Remove migration API

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/user.ts
```diff
@@ -1510,86 +1510,4 @@ app.post(
   },
 );
 
-const OLD_USER_CUTOFF = 1656312726000; // 27 June 2022
```
victorges, line 1513: Oh good practice. We should probably remove some other migration APIs that never got cleaned up :disappear:

victorges (approved):
neat

PR #2237, opened 2024-06-28
Add "x-livepeer-is-project-default" to /auth

leszko:
related to https://linear.app/livepeer/issue/ENG-1817/livepeer-data-update-for-projectid-handling

### DIFF
--- a/packages/api/src/controllers/auth.ts
+++ b/packages/api/src/controllers/auth.ts
@@ -70,6 +70,10 @@ app.all(
     res.header("x-livepeer-user-id", userId);
     res.header("x-livepeer-is-caller-admin", req.user.admin ? "true" : "false");
     res.header("x-livepeer-project-id", req.project?.id || "");
+    res.header(
+      "x-livepeer-is-project-default",
+      (req.project?.id == req.user?.defaultProjectId).toString(),
+    );
 
     res.status(204).end();
   },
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/auth.ts
```diff
@@ -72,7 +72,7 @@ app.all(
     res.header("x-livepeer-project-id", req.project?.id || "");
     res.header(
       "x-livepeer-is-project-default",
-      (req.project?.id == req.user?.defaultProjectId).toString(),
```
victorges: Was this not working the same way? 0o

packages/api/src/controllers/auth.ts
```diff
@@ -72,7 +72,7 @@ app.all(
...
+      req.project?.id == req.user?.defaultProjectId ? "true" : "false",
```
victorges, line 75: Given we have the `?` in the `project`, which implies that it could be undefined, WDYT of being resilient to that?

```suggestion
      !req.project?.id || req.project.id == req.user.defaultProjectId ? "true" : "false",
```
(also `req.user` is **never** undefined if there's an auth middleware)

On the bigger picture, all the `?` on projects solved the case when there was no default project for a user, which is not possible anymore AFAIK? Either way, I think we should keep the code consistent and we have a lot of other places that are considering the `req.project` nullable so might as well do it here. Later we can do a little clean-up removing all this "optional" logic which can get confusing for readers in the future.

victorges (approved):
LGTM with an optional nit

leszko:
Closing this PR, sincce this change won't be needed (Evan migrated all data at rest, so we don't need to know the default project id).

PR #2238, opened 2024-07-01
www: billing: pricing update

gioelecerati:

- Hacker plan become `Sandbox` plan
- Scale plan is removed from the plan selection
- Frontend pricing update + wording update
- Usage report is changing for the growth plan:
 Instead of calculating the amount X that exceeded the limit set for each product item, we are now calculating the amount of overage above a minimum spend. After 100$ of spend, we start reporting the correct product item usage to stripe  
- Storage price updated for all products  

**Screenshots (optional)**

**Checklist**


packages/api/src/config.js
```diff
@@ -114,11 +114,13 @@ export const products = {
...
         description: "Storage (minutes)",
-        price: 0.0035,
+        price: 0.0015,
         limit: 10_000,
       },
     ],
-    monthlyPrice: 100,
+    monthlyPrice: 1, // test value for staging
+    minimumSpend: true,
+    concurrent_viewers: 50_000,
```
victorges: Can we use camel 🐫 for some consistent casing here? This is the only snake 🐍 
```suggestion
    concurrentViewers: 50_000,
```

packages/api/src/config.js
```diff
@@ -114,11 +114,13 @@ export const products = {
       {
         name: "Storage",
         description: "Storage (minutes)",
-        price: 0.0035,
+        price: 0.0015,
         limit: 10_000,
       },
     ],
-    monthlyPrice: 100,
+    monthlyPrice: 1, // test value for staging
```
victorges: Should this be removed before merging? Not sure what it means.

packages/api/src/controllers/stripe.test.ts
victorges, line 1: Nice! Thanks for writing these

packages/api/src/controllers/stripe.test.ts
```diff
@@ -0,0 +1,158 @@
...
+          {
+            name: "Storage",
+            description: "Storage (minutes)",
+            price: 0.0015,
+            limit: 10_000,
+          },
+        ],
+      };
+
+      let billingUsage = {
```
victorges, line 47: Can you add a little comment before each of these explaining what you are testing? I don't get why everyting is 0 here for example. I'll exemplify with some hypothesis I have for this one:

```suggestion
      // no charges when all usages are below the minimum spend
      let billingUsage = {
```
(but then no idea why the one below also has everything 0. need to clarify on each comment)

Another way of organizing this would be making each of these a separate `it()` test and you describe what you are testing in the test name instead of a comment (`product` can be a static const on top of file or in the `describe` scope). That would also be easier to inspect and debug when running the tests, but just some comments would already be enough here.

packages/api/src/controllers/stripe.ts
```diff
@@ -214,6 +225,70 @@ const sendUsageRecordToStripe = async (
   );
 };
 
+export const calculateOverageOnMinimumSpend = async (product, billingUsage) => {
+  const minimumSpend = product.monthlyPrice;
```
victorges, line 229: Did you mean `const minimumSpend = product.minimumSpend`? Otherwise please either rename the variable or add a comment to explain the concept change here (FWIW the caller doesn't check `monthlyPrice`)

packages/api/src/controllers/stripe.ts
```diff
@@ -214,6 +225,70 @@ const sendUsageRecordToStripe = async (
...
+
+  let storageUsage = billingUsage.storageUsage;
+  let deliveryUsage = billingUsage.deliveryUsage;
+  let totalUsage = billingUsage.totalUsage;
```
victorges: **ts-nit:** deconstruction is helpful in such cases
```suggestion
  let { storageUsage, deliveryUsage, totalUsage } = billingUsage;
```

packages/api/src/controllers/stripe.ts
```diff
@@ -214,6 +225,70 @@ const sendUsageRecordToStripe = async (
...
+
+  let prices: any = {};
+
+  if (product?.usage) {
+    product.usage.forEach((item) => {
+      if (item.name.toLowerCase() === "transcoding")
+        prices.transcoding = item.price;
+      if (item.name.toLowerCase() === "delivery") prices.streaming = item.price;
+      if (item.name.toLowerCase() === "storage") prices.storage = item.price;
```
victorges: **nit:** Please avoid blocks of code without curly braces as those are pretty error-prone as the code evolves, especially if multi-line. 
```suggestion
      if (item.name.toLowerCase() === "transcoding") {
        prices.transcoding = item.price;
      } else if (item.name.toLowerCase() === "delivery") {
        prices.streaming = item.price;
      } else if (item.name.toLowerCase() === "storage") {
        prices.storage = item.price;
      }
```

(it's also easier to read IMO, with less stuff happening on each line)

packages/api/src/controllers/stripe.ts
```diff
@@ -214,6 +225,70 @@ const sendUsageRecordToStripe = async (
...
+        transcoding: prices.transcoding * totalUsage,
+      };
+
+      const totalSpent =
+        totalSpentInCategories.storage +
+        totalSpentInCategories.delivery +
+        totalSpentInCategories.transcoding;
+
+      if (totalSpent > 0) {
+        const overageRatio = remainingOverage / totalSpent;
```
victorges: This seems like some pretty complex logic. Is it described anywhere? The description you gave on the PR doesn't seem to describe this stuff like "overage ratio". Maybe worth a comment in this function declaration explaining how it works wdyt?

packages/api/src/controllers/stripe.ts
```diff
@@ -214,6 +225,70 @@ const sendUsageRecordToStripe = async (
...
+          (totalSpentInCategories.delivery * overageRatio) / prices.streaming;
+        overageUsage.TotalUsageMins =
+          (totalSpentInCategories.transcoding * overageRatio) /
+          prices.transcoding;
+      }
+    }
+
+    return overageUsage;
+  }
+  return {
```
victorges: **nit:** A general tip to keep less indentation in the code is also to just invert some conditions.

e.g. in this case instead of 
```
if (total > minimum) {
  /* a long branch of code */
}
// POV: You're reading the code linearly... "where was I again?" 😆 
return 0
```
it could be
```
if (total <= minimum) {
  return 0
}
/* the long branch of code, with 1 level less indentation 😌  */
```

packages/www/components/Plans/index.tsx
```diff
@@ -224,6 +224,50 @@ const Plans = ({ dashboard = false, stripeProductId }: PlanProps) => {
                   fontSize: "$3",
                 }}
               />
+              <Item
+                displayCheck={false}
+                title={
+                  <Tooltip
+                    multiline
+                    content="Max amount of concurrent viewers.">
```
victorges, line 232: Neat!

packages/www/components/Plans/index.tsx
```diff
@@ -294,7 +338,25 @@ const Plans = ({ dashboard = false, stripeProductId }: PlanProps) => {
...
+                    <span>5,000 minutes / month</span>
+                  </Flex>
+                }
+                css={{ borderBottom: 0 }}
+              />
+              <Item
+                displayCheck={false}
+                title={
+                  <Flex css={{ justifyContent: "center", width: "100%" }}>
+                    <span>Up to 30</span>
```
victorges, line 350: Missing unit & time range?

packages/www/components/Plans/index.tsx
```diff
@@ -594,33 +484,18 @@ const Plans = ({ dashboard = false, stripeProductId }: PlanProps) => {
...
               <Item
-                css={{ borderBottom: 0, fontSize: "11px" }}
+                css={{
+                  borderBottom: 0,
+                  fontSize: "11px",
+                  justifyContent: "center",
+                }}
                 displayCheck={false}
-                title={"* Pay as you go past allotted minutes"}
+                title={"* (we round up if below $100)"}
```
victorges, line 498: I think this could be rephrased more clearly. Rounding sounds like you pay $100 if you spent $99.01.

Maybe?
```suggestion
                title={"* Monthly pay of $100 or more depending on your usage"}
```
(the parenthesis seem redundant too, this is already a `*` remark)

packages/www/components/Plans/index.tsx
```diff
@@ -672,28 +547,54 @@ const Plans = ({ dashboard = false, stripeProductId }: PlanProps) => {
...
+                    <span>Unlimited</span>
+                  </Flex>
+                }
+                css={{ borderBottom: 0 }}
+              />
+              <Item
+                displayCheck={false}
+                title={
+                  <Flex css={{ justifyContent: "center", width: "100%" }}>
+                    <span>$2.5k minimum spend</span>
```
victorges, line 584: Same `*` explaining minimum spend?

packages/api/src/config.js
victorges: Finished the PR and came back here. Shouldn't this be enforced on the gate API? Also shouldn't there be a `concurrentViewers: 10` somewhere for the free plan?

victorges (approved):
LGTM

packages/api/src/config.js
gioelecerati: Right, will do camel!
Yeah that's enforced by the gate API - but this is just for the frontend display on the plan page. All of these products should end up in a `product` table in the future with billing v2, and both the frontend and the api will fetch from this field

packages/api/src/controllers/stripe.ts
gioelecerati: minimumSpend is a boolean and monthlyPrice is the actual minimumSpend per month. Agree that I could give a better naming, will give a think

packages/www/components/Plans/index.tsx
gioelecerati: It's exactly how it is! from now on, if you spend 1 or 99$ you still pay 100$ monthly at least for the `Growth` plan. The phrasing in the frontend was given to me by hunter from the livepeer.studio pricing page that is coming

packages/www/components/Plans/index.tsx
gioelecerati: On the livepeer.studio pricing page the only * it's with annual commit, will ask hunter

packages/api/src/controllers/stripe.test.ts
gioelecerati: Added some comments for each step. Won't put it right now in separate `it()`, I would like to write all tests for billing here in billing v2 - so I will refactor it

packages/www/components/Plans/index.tsx
gioelecerati: What are you referring to? This is the row `Concurrent viewers` which this plan (`Sandbox`) has `Up to 30`
Again, exactly the same as the new livepeer.studio pricing page

packages/api/src/controllers/stripe.ts
gioelecerati: I added several comments across the steps!
This is the part where I mostly need the feedback, so I add some description here:

Stripe does not have a built-in way to implement minimum spend for metered usage, so if I report 30$ of usage to it on a product with a 100$ fixed monthly price, it will produce a 130$ invoice. Before these changes, we had fixed amount of minutes included in the plan, so whenever one of the items (transcoding, storage or delivery) had an overage, we would have reported it.

Instead, now, we don't have a fixed amount of minutes, we do have instead a fixed amount of dollars included in the plan, with a minimum monthly spend. So, if you use 1$ or 99$ or anything under 100$ across all three product items, you will receive an invoice of 100$.

This has an implication: 
Whenever the 100$ is exceeded, we need to calculate the amount of minutes in all three product items to subtract the included amount. To do this (with this proposed change) we take the total overage (i.e. 3$ out of a 103$ total usage) and distribute it across the items that have more than 0$ usage.

Stripe is going to get reported usage rows for 3$ in this example

PR #2239, opened 2024-07-01
api: projects: fetch user if missing subitem

gioelecerati:

When the document is missing the subitem `user`, fetch it based on the document.userId

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -772,9 +772,9 @@ export async function addDefaultProjectId(
   };
 
   const enrichResponseWithUserProjectId = async (document) => {
-    if ("id" in document && "userId" in document && "user" in document) {
-      if (!document.projectId || document.projectId === "") {
-        if (document.user?.defaultProjectId) {
+    if ("id" in document && "userId" in document) {
+      if (!document.projectId) {
+        if (document.user) {
           document.projectId = document.user.defaultProjectId;
         } else {
           let user = await db.user.get(document.userId, { useCache: true });
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -777,7 +777,10 @@ export async function addDefaultProjectId(
         if (document.user) {
           document.projectId = document.user.defaultProjectId;
         } else {
-          let user = await db.user.get(document.userId, { useCache: true });
+          let user =
+            document.userId === req.user.id
+              ? req.user
+              : await db.user.get(document.userId, { useCache: true });
           if (user.defaultProjectId) {
             document.projectId = user.defaultProjectId;
           }
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -814,7 +814,7 @@ export async function addDefaultProjectId(
   };
 
   if (Array.isArray(clonedBody)) {
-    await Promise.all(clonedBody.map(processItem));
+    await Promise.all(clonedBody.map((item) => processItem(item)));
   } else if (typeof clonedBody === "object" && clonedBody !== null) {
     await processItem(clonedBody);
   }
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
     }
   };
 
-  const enrichResponseWithUserProjectId = (document) => {
+  const enrichResponseWithUserProjectId = async (document) => {
     if ("id" in document && "userId" in document && "user" in document) {
-      if (
-        (!document.projectId || document.projectId === "") &&
-        document.user?.defaultProjectId
-      ) {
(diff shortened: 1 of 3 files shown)


leszko (approved):
LGTM. It'd be nice if you could add a unit test.

packages/api/src/controllers/helpers.ts
```diff
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
...
 
-  const enrichResponseWithUserProjectId = (document) => {
+  const enrichResponseWithUserProjectId = async (document) => {
     if ("id" in document && "userId" in document && "user" in document) {
-      if (
-        (!document.projectId || document.projectId === "") &&
-        document.user?.defaultProjectId
-      ) {
-        document.projectId = document.user.defaultProjectId;
+      if (!document.projectId || document.projectId === "") {
```
victorges: **[js-wat](https://www.destroyallsoftware.com/talks/wat)-protip**: empty string is already falsy! same behavior:
```suggestion
      if (!document.projectId) {
```

packages/api/src/controllers/helpers.ts
```diff
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
...
+        if (document.user?.defaultProjectId) {
+          document.projectId = document.user.defaultProjectId;
```
victorges: We shouldn't be fetching the user from the DB again if the user is already embedded in the object.
```suggestion
        if (document.user) {
          document.projectId = document.user.defaultProjectId;
```

packages/api/src/controllers/helpers.ts
```diff
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
     }
   };
 
-  const enrichResponseWithUserProjectId = (document) => {
+  const enrichResponseWithUserProjectId = async (document) => {
     if ("id" in document && "userId" in document && "user" in document) {
```
victorges: This condition needs to be updated.

packages/api/src/controllers/helpers.ts
```diff
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
...
-      if (
-        (!document.projectId || document.projectId === "") &&
-        document.user?.defaultProjectId
-      ) {
-        document.projectId = document.user.defaultProjectId;
+      if (!document.projectId || document.projectId === "") {
+        if (document.user?.defaultProjectId) {
+          document.projectId = document.user.defaultProjectId;
+        } else {
+          let user = await db.user.get(document.userId, { useCache: true });
```
victorges: We shouldn't be fetching the user if they are already the same as the caller.
```suggestion
          let user = document.userId === req.user.id ? req.user : await db.user.get(document.userId, { useCache: true });
```
(good call on caching)

packages/api/src/controllers/helpers.ts
```diff
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
...
+          if (user.defaultProjectId) {
+            document.projectId = user.defaultProjectId;
+          }
```
victorges, line 786: It's not really a problem if `defaultProjectId` is not set, right? We'd just be setting the `projectId` to empty or undefined, which it actually is anyway (if it has anything else it's a mistake!)

Not a problem checking it either, just feels like unnecessary branches to be reasoning about in the code, apart from the required ones.

```suggestion
          document.projectId = user.defaultProjectId;
```

packages/api/src/controllers/helpers.ts
```diff
@@ -771,45 +771,49 @@ export async function addDefaultProjectId(
...
 
-      Object.values(item).forEach((subItem) => {
-        if (typeof subItem === "object" && subItem !== null) {
-          if (req.user.admin) {
-            enrichResponseWithUserProjectId(subItem);
-          } else {
-            enrichResponse(subItem);
+      await Promise.all(
+        Object.values(item).map(async (subItem) => {
+          if (typeof subItem === "object" && subItem !== null) {
```
victorges, line 804: Remind me why do we need this recursion again? Do we have project-objects nested inside project-objects as well?

packages/api/src/controllers/helpers.ts
gioelecerati: Am I tripping? I already see the `async` in the changes

packages/api/src/controllers/helpers.ts
gioelecerati: Wasn't it for the cases like the asset creation, which returns {asset:{},task:{}}?

packages/api/src/controllers/helpers.ts
gioelecerati: Ah I see what you're referring to, fixed it

packages/api/src/controllers/helpers.ts
victorges: Ahh yeah right!

victorges (approved):

PR #2240, opened 2024-07-02
ui: various fixes and enhancements

suhailkakar:

- fix: 404 error on stream delete (ENG-2148)
- fix: display error message for invalid payment card (ENG-2149)
- ui: add 30 seconds time limit to password reset request

**Screenshots (optional)**

**Checklist**


packages/api/src/controllers/user.test.ts
```diff
@@ -363,51 +363,57 @@ describe("controllers/user", () => {
...
-      expect(resp.errors[0]).toBe("user not found");
-
-      // should return 422 when extra property added to request
-      req = await client.post(`/user/password/reset-token`, {
-        email: "noemail@gmail.com",
-        password: "a".repeat(64),
-      });
-      expect(req.status).toBe(422);
+      // should return 201 when user email is valid
+      if (user.emailValid) {
```
leszko: Why do we need this "if" here. It's a unit test, so we should know at this point if the email is valid or not.

packages/www/components/Sidebar/index.tsx
```diff
@@ -93,6 +95,25 @@ const Sidebar = ({ id }: { id: SidebarId }) => {
     };
   }, [June]);
 
+  const changePassword = async (e) => {
```
leszko, line 230: Sorry to be picky, but the UX of changing password is not great. Here's what happened to me:
1. I click "change password"
2. Nothing happens, so I click again
3. I see "Please wait 29 seconds before sending another email."

One quick improvement would be to have `toast("Password reset link sent to your email.");` before `const res = await makePasswordResetToken(user.email);`. And then if there is an error, we just toast the error message on top of that.

leszko (approved):
Thanks for the PR @suhailkakar 

Added some comments, other than that LGTM

packages/api/src/controllers/user.test.ts
suhailkakar: Oh, yes,  good point!  I didn't bring any change in `api` so, i wasn't sure. Will remove the if condition

packages/www/components/Sidebar/index.tsx
suhailkakar: No problem, and sounds good! will update it :)

packages/api/src/controllers/user.test.ts
suhailkakar: i've talked with Victor, i will remove it for now and will confirm with Hunter about its logic first :)

packages/www/components/Sidebar/index.tsx
```diff
@@ -149,13 +169,18 @@ const Sidebar = ({ id }: { id: SidebarId }) => {
                 }}>
                 Billing
               </DropdownMenuItem>
+              <DropdownMenuItem
+                key="change-password-dropdown-item"
+                onSelect={changePassword}>
+                Change Password
```
0xcadams:
```suggestion
                Change password
```

packages/www/lib/utils/can-send-email.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ * Function to check if a user can request reset password or verfication email
+ */
+
+export function canSendEmail(emailType: "resetPassword" | "verifyEmail"): {
```
0xcadams, line 5: I don't love that the frontend is responsible for doing rate limiting for reset password. This should be a responsibility of the backend, not the frontend. A user can easily clear their local storage.

0xcadams (changes requested):

packages/www/lib/utils/can-send-email.ts
suhailkakar: As much as i agree with you, it might be a little more work on backend compared to frontend. 

From Amrit, 
> usually, emails should get queued and processed independently. when clicking the button -> queue under the <user_id>.<action_hash> unique id, so that if same action is done multiple times (with same inputs), it wont trigger multiple emails.

So i think for now, we can keep it on frontend, just to avoid situation like this[1] but in future we can implement it directly in the backend. 

[1]:

image

packages/www/components/Sidebar/index.tsx
```diff
@@ -223,6 +227,24 @@ const Sidebar = ({ id }: { id: SidebarId }) => {
...
+      openSnackbar(
+        `Please wait ${response.waitTime} seconds before sending another email.`,
+      );
+      return;
+    }
+    openSnackbar("Password reset link sent to your email.");
+    const res = await makePasswordResetToken(user.email);
+    if (res.errors) {
+      res.errors.forEach((error) => {
+        openSnackbar(error);
```
0xcadams: We should only show one error here - not all errors included

packages/www/lib/utils/can-send-email.ts
0xcadams: OK, that's fine for now

0xcadams (commented):
small nits but LGTM

suhailkakar:
@0xcadams please let me know, when we can merge this

0xcadams (approved):

PR #2241, opened 2024-07-04
Add api schema for Realtime Viewership API

leszko:
Related to creating documentation for Realtime Viewership API: https://linear.app/livepeer/issue/ENG-1719/documentation

### DIFF
--- a/packages/api/src/schema/api-schema.yaml
+++ b/packages/api/src/schema/api-schema.yaml
@@ -2391,10 +2391,42 @@ components:
           type: number
           description: Total minutes of storage usage.
           example: 100
+    realtime-viewership-metric:
+      type: object
+      description: |
+        An individual metric about realtime viewership of a stream/asset.
+      required:
+        - viewCount
+        - errorRate
+      properties:
+        playbackId:
+          type: string
+          description: The playback ID associated with the metric.
+          example: 1bde4o2i6xycudoy
+        device:
+          type: string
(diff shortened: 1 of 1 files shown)


leszko:
@suhailkakar how to actually check if it works well with Speakeasy?

suhailkakar:
Once the PR gets merged, a automatic pr would be opened on the SDK repository which looks like this: https://github.com/livepeer/livepeer-js/pull/18

Inside of that, we have tests to make sure that all API responses is valid

packages/api/src/schema/api-schema.yaml
```diff
@@ -2391,10 +2391,42 @@ components:
           type: number
           description: Total minutes of storage usage.
           example: 100
+    realtime-viewership-metric:
+      type: object
+      description: |
+        An individual metric about realtime viewership of a stream/asset.
+      required:
+        - viewCount
+        - errorRate
```
suhailkakar, line 2400: Shouldn't the `playbackId` here be required too, because it is the only filter to query results. Otherwise LGTM ✅

suhailkakar (approved):

packages/api/src/schema/api-schema.yaml
leszko: If the query is done against the whole account (project id / user id), then we don't include `playbackId` in the response.

PR #2242, opened 2024-07-08
api: projects: long running job cleanup (draft)

gioelecerati:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/project.ts
+++ b/packages/api/src/controllers/project.ts
@@ -3,6 +3,7 @@ import { authorizer } from "../middleware";
 import { db, jobsDb } from "../store";
 import { v4 as uuid } from "uuid";
 import {
+  deleteAllOwnedObjects,
   makeNextHREF,
   parseFilters,
   parseOrder,
@@ -216,7 +217,7 @@ export function triggerCleanUpProjectsJob(
 
   const jobPromise = Promise.resolve().then(async () => {
     try {
-      await Promise.all(projects.map((s) => cleanUpProject(jobsDb, s, req)));
+      await Promise.all(projects.map((s) => cleanUpProject(s, req)));
     } catch (err) {
       const ids = projects.map((s) => s.id);
       logger.error(`Error cleaning up projectId=${ids} err=`, err);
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/project.ts
```diff
@@ -177,4 +181,115 @@ app.delete("/:id", authorizer({}), async (req, res) => {
   res.end();
 });
 
+app.post(
+  "/job/projects-cleanup",
+  authorizer({ anyAdmin: true }),
```
victorges, line 187: Btw if you're creating this only because the other jobs have sth like it, it's unnecessary! The only reason they're like that is tech debt: we already had such APIs that were called from GitHub actions.

The cleanest way for a new job IMO will be to implement everything on the job. Then if later we want the API to trigger it manually, we can make an API that calls the job function (not the other way around).

Your call though, especially if you're already too far into having this API that it would be a big refactor to get rid of it now.

packages/api/src/controllers/project.ts
```diff
@@ -177,4 +181,115 @@ app.delete("/:id", authorizer({}), async (req, res) => {
...
+
+  let [sessions] = await db.session.find({
+    filters: {
+      projectId: project.id,
+      deleted: false,
+    },
+  });
+
+  for (const asset of assets) {
+    await req.taskScheduler.deleteAsset(asset.id);
```
victorges: Make sure this doesn't send webhooks I guess? Or do we want webhooks?

packages/api/src/controllers/project.ts
```diff
@@ -177,4 +181,115 @@ app.delete("/:id", authorizer({}), async (req, res) => {
...
+  }
+
+  for (const stream of streams) {
+    await db.stream.update(stream.id, {
```
victorges: FWIW there's a `markDeleted` function on the tables which is generally used for the soft deletion

packages/api/src/controllers/project.ts
```diff
@@ -177,4 +181,115 @@ app.delete("/:id", authorizer({}), async (req, res) => {
...
+      deleted: true,
+    });
+  }
+
+  for (const signingKey of signingKeys) {
+    await db.signingKey.update(signingKey.id, {
+      deleted: true,
+      disabled: true,
```
victorges: Do we need to disable them as well? Does the code that check `disabled` not check `deleted`? If so, that's the thing that needs to be fixed instead imo.

My reasoning is that if we ever want to "undelete" a project then we would need to figure out which signing keys were already disabled and which ones we disabled automatically here.

packages/api/src/jobs/projects-cleanup.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        type: "url" as const,
+        url: "someSource",
+      },
+    };
+  };
+
+  it("it should deleted related assets and streams", async () => {
+    const triggerSpy = jest
+      .spyOn(projectsController, "triggerCleanUpProjectsJob")
+      .mockImplementation(() => [[], Promise.resolve()]);
```
victorges, line 85: Same regarding the "inverted" nature of the current jobs implementation. Let's make sure the "real" tests are implemented here, which check the effects of the deletion job. The API that should be the "side-kick" which calls the job, but the core logic of the job should be the job itself

packages/api/src/jobs/projects-cleanup.ts
```diff
@@ -0,0 +1,35 @@
...
+    throw new Error("ingest not configured");
+  }
+  const { jobsDb } =
+    clients ?? (await initClients(config, "projects-cleanup-job"));
+  const { projectsCleanupLimit: limit, ingest } = config;
+
+  let [projects] = await jobsDb.stream.find([sql`data->>'deleted' = 'true'`], {
+    limit,
+    order: "data->>'lastSeen' DESC",
+  });
```
victorges, line 26: We probably need sth like a "`deleting` phase" on the projects too, otherwise this will keep listing deleted files that have already been cleaned-up ad infinitum. Maybe a simple `cleanedUp` boolean in the project that gets set by this job?

packages/api/src/jobs/projects-cleanup.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (!config.ingest?.length) {
+    throw new Error("ingest not configured");
+  }
+  const { jobsDb } =
+    clients ?? (await initClients(config, "projects-cleanup-job"));
+  const { projectsCleanupLimit: limit, ingest } = config;
+
+  let [projects] = await jobsDb.stream.find([sql`data->>'deleted' = 'true'`], {
+    limit,
+    order: "data->>'lastSeen' DESC",
```
victorges, line 25: Does this exist? I thought it was an user/api-key/stream thing only

victorges (commented):
did a skim pre-review 👀

PR #2243, opened 2024-07-08
api: access-control: viewers limit

gioelecerati:

- Remove rate limit
- Apply viewer limit to all users 
- If user has a specific limit set, use it
- If user has a product with a limit, use it
- If user has neither, use default

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/config.js
+++ b/packages/api/src/config.js
@@ -25,6 +25,7 @@ export const products = {
       },
     ],
     monthlyPrice: 0,
+    viewerLimit: 30,
   },
   prod_1: {
     deprecated: true,
@@ -52,7 +53,7 @@ export const products = {
       },
     ],
     monthlyPrice: 0,
-    concurrentViewers: 50_000,
+    viewerLimit: 50_000,
   },
   prod_2: {
     deprecated: true,
@@ -94,6 +95,7 @@ export const products = {
       },
     ],
     monthlyPrice: 0,
+    viewerLimit: 30,
   },
   growth_1: {
     order: 4,
@@ -121,7 +123,7 @@ export const products = {
     ],
     monthlyPrice: 100,
     payAsYouGo: true,
-    concurrentViewers: 50_000,
+    viewerLimit: 50_000,
   },
   scale_1: {
     order: 5,
@@ -149,7 +151,7 @@ export const products = {
     ],
     monthlyPrice: 500,
     payAsYouGo: true,
-    concurrentViewers: 50_000,
+    viewerLimit: 50_000,
   },
   pay_as_you_go_1: {
     order: 6,
@@ -223,7 +225,7 @@ export const products = {
       },
     ],
     monthlyPrice: 0,
-    concurrentViewers: 50_000,
+    viewerLimit: 30,
   },
   prod_O9XtHhI6rbTT1B: {
     order: 9,
@@ -251,7 +253,7 @@ export const products = {
     ],
     monthlyPrice: 100,
     payAsYouGo: true,
-    concurrentViewers: 50_000,
+    viewerLimit: 50_000,
   },
   prod_O9XtcfOSMjSD5L: {
     order: 10,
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/access-control.ts
```diff
@@ -30,12 +31,11 @@ import { DBStream } from "../store/stream-table";
 import { WithID } from "../store/types";
 
 const WEBHOOK_TIMEOUT = 30 * 1000;
-const MAX_ALLOWED_VIEWERS_FOR_HACKER_TIER_PER_NODE = 5;
+const DEFAULT_MAX_CONCURRENT_VIEWERS = 50_000;
```
victorges: Shouldn't this and some of the plan limits on `config.js` be `10`?

packages/api/src/controllers/access-control.ts
```diff
@@ -225,17 +225,20 @@ app.post(
...
+    if (userProduct.concurrentViewers) {
+      concurrentViewers = userProduct.concurrentViewers;
     }
 
-    if (user.createdAt > HACKER_DISABLE_CUTOFF_DATE) {
-      if (isFreeTierUser(user)) {
-        config.rate_limit = MAX_ALLOWED_VIEWERS_FOR_HACKER_TIER_PER_NODE;
-      }
+    if (user.viewerLimit) {
+      concurrentViewers = user.viewerLimit;
```
victorges: Now that I see this, since we already have the `viewerLimit` field on users WDYT of using the same name on the plans? Feels unnecessary to use a different name, even if the existing one is not ideal, and make it look like they mean different things (especially since we don't usually have a lot of documentation for internal stuff)

victorges (approved):

packages/api/src/controllers/access-control.ts
gioelecerati: The new `Sandbox` limit (free tier) is 30 - all other plans are 50k. I added a default here in case for deprecated or not tracked plans (i.e. TheLRadio)

packages/api/src/controllers/access-control.ts
gioelecerati: Make sense! Changing everything to `viewerLimit`

packages/api/src/config.js
```diff
@@ -25,6 +25,7 @@ export const products = {
       },
     ],
     monthlyPrice: 0,
+    viewerLimit: 30,
```
leszko, line 28: What are these config changes?

packages/api/src/controllers/access-control.ts
```diff
@@ -225,17 +225,22 @@ app.post(
...
-    if (user.viewerLimit) {
-      config.user_viewer_limit = user.viewerLimit;
-      config.user_id = user.id;
-    }
+    let viewerLimit = DEFAULT_MAX_CONCURRENT_VIEWERS;
 
-    if (user.createdAt > HACKER_DISABLE_CUTOFF_DATE) {
-      if (isFreeTierUser(user)) {
-        config.rate_limit = MAX_ALLOWED_VIEWERS_FOR_HACKER_TIER_PER_NODE;
+    if (user.stripeProductId) {
```
leszko, line 230: What is stripe product id?

leszko (approved):
LGTM. Added 2 comments, but it's mostly to improve my understanding. The PR is good.

packages/api/src/config.js
gioelecerati: It's the viewer limit specified per product. In the future (billing v2) it will be a column in the `product` table, with a default viewerLimit specified on a specific product

packages/api/src/controllers/access-control.ts
gioelecerati: It's the stripe `ID` of the product (plan) of the user. 
In this case either `Sandbox` (30 viewers as specified in config.js), `Growth` (50_000 viewers) and all the other untracked products (`Enterprise`, `Lotr`, ecc.).
If the user has no viewerLimit set in his object, it will fall back to product[user.productId].viewersLimit - if also that is missing, it will use the default const

PR #2245, opened 2024-07-08
api/api-token: Avoid read-after-write inconsistency on creation

victorges:
When we crete the API token for some reason we read it back instead of trusting the DB transaction worked. The problem is that we read from the DB replica by default and there might be some replication lag.

Fix it by flagging not to read from the replica instead. I could remove that logic altogether but I want to make as less intrusive of a fix as possible.


gioelecerati (approved):
LGTM

PR #2246, opened 2024-07-08
api: Create generative AI APIs using AI subnet

victorges:

This creates the initial versions of the AI Generate APIs, based on the [design doc](https://www.notion.so/livepeer/Studio-Generative-AI-27ec0fb899c64341957ee56617f66767).

This initial version implements basically the proxy to an internal AI Gateway service, thus
also implementing the exact same interface as the gateway.

The main complexity of this was adding support to the multipart requests, currently used by
the AI Gateway interface for these APIs. This required upgrading the major version of the
`ajv` library and some of its related packages, as well as adding a bit of code to handle
multipart form validation and error handling.

Note: haven't added the actual API reference on purpose (`paths:` section on API schema), given we're
not yet sure how this new API is going to be advertised. Right now it is under an experiment so no end-users
will be able to use by themselves.

- Add schemas for the new APIs inputs
- Implement the JSON handler for `/generate/text-to-image` API
- Implement the multipart handler for other `/generate` APIs
- Create a couple tests for the API proxy behavior

- `yarn test`
- Put on staging and called the APIs

Implements ENG-2181

**Checklist**


packages/api/src/parse-cli.ts
```diff
@@ -216,6 +216,12 @@ export default function parseCli(argv?: string | readonly string[]) {
         type: "string",
         default: "https://{{ip}}:8935",
       },
+      "ai-gateway-url": {
+        describe:
+          "base URL of the AI Gateway to call for generative AI requests",
+        type: "string",
+        default: "http://staging-livepeer-ai-gateway/", // TODO: Remove this default, staging testing only
```
leszko: 1. What's this TODO saying? Why don't we remove this default right now if we don't want to have it?
2. Do I understand correctly that currently this feature won't work with Livepeer in Box (and it won't be simple to implement it in the box, because Livepeer AI requires GPUs)?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2854,7 +2856,137 @@ components:
...
+      required:
+        - prompt
+      properties:
+        prompt:
+          type: string
+        model_id:
+          type: string
+          default: SG161222/RealVisXL_V4.0_Lightning
+          enum:
+            - SG161222/RealVisXL_V4.0_Lightning
```
leszko, line 2873: Do I understand correctly that currently, we accept only 2 models and that these 2 specific models are always available in Os? 

Btw. what happens if the model is not available in the specific O?

packages/api/src/middleware/validators.ts
```diff
@@ -0,0 +1,62 @@
...
+      res.status(422);
+      return res.json({
+        errors: validate.errors.map((err) => JSON.stringify(err)),
+      });
+    }
+    next();
+  };
+};
+
+export const validateFormData = (name: string): RequestHandler => {
```
leszko, line 22: Would it make sense to add a unit test for this? Or is it maybe already tested inside `generate.test.ts`?

leszko (approved):
Looks good. Added some minor comments. I have some other questions comments (about billing and monitoring) but I'll open them in Discord.

packages/api/src/controllers/generate.test.ts
```diff
@@ -0,0 +1,166 @@
...
+    });
+    client.apiKey = null;
+    client.jwtAuth = nonAdminToken;
+
+    aiGatewayCalls = {};
+  });
+
+  describe("API proxies", () => {
+    it("should call the AI Gateway for generate API /text-to-image", async () => {
+      const res = await client.post("/beta/generate/text-to-image", {
```
gioelecerati, line 97: How expensive is this when it runs on every github actions during tests? This and all the tests after this in sequence

packages/api/src/schema/api-schema.yaml
```diff
@@ -120,7 +120,7 @@ components:
         height:
           type: integer
           minimum: 128
-          expample: 720
+          example: 720
```
gioelecerati, line 123: A typo lurking for 3 months!

packages/api/src/schema/api-schema.yaml
```diff
@@ -2854,7 +2856,137 @@ components:
         targetSegmentSizeSecs:
           $ref: >-
             #/components/schemas/new-asset-payload/properties/targetSegmentSizeSecs
-
+    # AI Generate payloads. Keep in mind that these use snake_case instead of camelCase since
+    # they implement the same interface as the AI Gateway Livepeer node.
+    text-to-image-payload:
+      type: object
+      required:
+        - prompt
```
gioelecerati, line 2864: Any plan to add a check in the validator for the prompt for potentially harmful content?

gioelecerati (approved):

gioelecerati (commented):
LGTM, a couple tests are failing

packages/api/src/middleware/validators.ts
```diff
@@ -0,0 +1,62 @@
...
+        allFields[key] = fval;
+      } else if (value === "true" || value === "false") {
+        allFields[key] = value === "true";
+      } else {
+        allFields[key] = value;
+      }
+    }
+    if (Array.isArray(files)) {
+      for (const file of files) {
+        // we don't need to validate the file contents, just that it is there
```
gioelecerati, line 49: Don't we need to validate the mime type of the files to ensure they're either `image/png` or `image/jpeg`?

packages/api/src/parse-cli.ts
victorges: This is only from the commit https://github.com/livepeer/studio/pull/2246/commits/802d297f060db3713392c72496787059c8a541ae which will be reverted before merging this. It was used for testing this branch in staging without needing to modify the deployment envs on livepeer-infra

victorges: > Do I understand correctly that currently this feature won't work with Livepeer in Box (and it won't be simple to implement it in the box, because Livepeer AI requires GPUs)?

Yeah that's correct, so on the box one will just get that error that the AI Gateway is not configured. We need to figure out how to add it E2E in the box, but requiring a 16GB NVIDIA GPU is probably not an option and CPU inference is too slow to be minimally useful.

Maybe the best bet would be requiring you to configure a wallet and run an AI Gateway connected to mainnet, but we can figure that out later especially if/when we get some AI on the video pipeline critical path.

packages/api/src/middleware/validators.ts
victorges: Yeah the easier way to test will be with a real API, with a schema. I can add some test cases on the `generate.test.ts` that checks these validations are working properly.

packages/api/src/controllers/generate.test.ts
victorges: We actually don't run the inference on tests! These APIs will hit the local server initialized on the `beforeAll` which is only dummy, returns a 200

packages/api/src/schema/api-schema.yaml
victorges: Hmm not yet! I wonder if we should ever do that though. What we currently have is the `safety_check` which I believe checks if the output is NSFW (not sure if it returns an error or just flags the response `nsfw` field)

packages/api/src/middleware/validators.ts
victorges: I left this validation for the AI Gateway itself! We will proxy any 4xx we get back from them

packages/api/src/parse-cli.ts
leszko: Yeah, we can wait with the box implementation.

In general:
- If we believe the box will be used not only for dev, then we can have a URL to AI Gateway
- If we believe box is just for dev env, then we can create an AI Gateway stub

packages/api/src/schema/api-schema.yaml
victorges: @leszko Forgot to reply this. Yeah right now we accept only these 2 models (or others in other APIs) which are the ones recommended by the AI Network team on the docs. e.g. https://docs.livepeer.org/ai/pipelines/text-to-image#models

Other models (called "on-demand") are still supported, but they might have less support from Orchestrators so I opted to keep them out on this first version. 

Currently, an O can keep only 1 model (from any pipeline) warmed up, as it has to keep an `ai-worker` container running with that model loaded in the GPU. When the model is not available (warm) in the O, it will just kill that running container and start another `ai-worker` configured with the requested model. The model is usually available on the disk if the operator previously configured it (see [this getting started](https://docs.livepeer.org/ai/orchestrators/models-download)), otherwise the `ai-worker` will download the model dynamically from [hugging face](https://huggingface.co/).

This means that asking for an uncommon model can make the request _really_ slow. Just loading the model on the GPU already takes a couple dozen seconds, and if the O has to download it as well it could take even minutes. So that's why I opted to limit the models that can be asked here (but there's still a risk no O has it warm and it takes longer than usual).

I believe this whole flow is being improved on by the AI network team, like with the "remote worker" architecture (similar to splitting O+T nodes). Maybe @rickstaa could correct if I said anything wrong or provide more info here 😄

PR #2252, opened 2024-07-11
ui: add basic color definitions

suhailkakar:

add basic color definitions for tailwind css

**Screenshots (optional)**

**Checklist**


packages/www/css/tailwind.css
```diff
@@ -1,3 +1,26 @@
...
+
+  /* 
+  The class name here is temporory. Since old class names are still in use. We can replace this with .dark once we are ready to switch to the new class names.
+  */
+  .dark-theme-green {
+    --color-border: 240 5% 16%;
+    --color-background: 0 0% 9%;
+    --color-foreground: 0 0% 100%;
+    --color-muted: 249 4 64;
+    --color-surface: 240° 4% 9%;
```
0xcadams: Is this part of the shadcn or separate?

https://ui.shadcn.com/docs/theming

Also, there's a degree symbol here and below

0xcadams (changes requested):

packages/www/css/tailwind.css
suhailkakar: > Is this part of the shadcn or separate?

There are just css variable from Studio's color scheme. And yeah we will use them in shadcn components. But we don't need to the same exact variable written in shadcn docs, like foreground and etc..

0xcadams (approved):

PR #2263, opened 2024-07-12
ui: reskin dashboard

suhailkakar:

Updated the colors in Studio dashboard. 

NOTE: I've updated only the background or foreground colors not the components. Migrating components will take more time so until they are completed, we can use the already added tailwind colors 

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/www/components/AssetDetails/AssetOverviewTab.tsx
+++ b/packages/www/components/AssetDetails/AssetOverviewTab.tsx
@@ -135,8 +135,9 @@ const AssetOverviewTab = ({
           </Text>
         </Box>
         <div
-        className="bg-background text-foreground p-4 rounded-lg border" 
-        size="2" css={{ borderRadius: "$2" }}>
+          className="bg-background text-foreground p-4 rounded-lg border"
+          size="2"
+          css={{ borderRadius: "$2" }}>
           <Flex css={{ justifyContent: "space-between" }}>
             <Flex align="center">
               <StyledIpfsIcon />
--- a/packages/www/components/AssetDetails/AssetOverviewTab.tsx
+++ b/packages/www/components/AssetDetails/AssetOverviewTab.tsx
@@ -136,7 +136,6 @@ const AssetOverviewTab = ({
         </Box>
         <div
           className="bg-background text-foreground p-4 rounded-lg border"
-          size="2"
           css={{ borderRadius: "$2" }}>
           <Flex css={{ justifyContent: "space-between" }}>
             <Flex align="center">
--- a/packages/www/components/AssetDetails/AssetOverviewTab.tsx
+++ b/packages/www/components/AssetDetails/AssetOverviewTab.tsx
@@ -134,7 +134,9 @@ const AssetOverviewTab = ({
             may also archive them using the following storage providers.
           </Text>
         </Box>
-        <Promo size="2" css={{ borderRadius: "$2" }}>
+        <div
+        className="bg-background text-foreground p-4 rounded-lg border" 
+        size="2" css={{ borderRadius: "$2" }}>
           <Flex css={{ justifyContent: "space-between" }}>
             <Flex align="center">
               <StyledIpfsIcon />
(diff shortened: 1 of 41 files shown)


packages/www/components/AssetDetails/AssetOverviewTab.tsx
```diff
@@ -134,7 +134,9 @@ const AssetOverviewTab = ({
             may also archive them using the following storage providers.
           </Text>
         </Box>
-        <Promo size="2" css={{ borderRadius: "$2" }}>
+        <div
+          className="bg-background text-foreground p-4 rounded-lg border"
+          css={{ borderRadius: "$2" }}>
```
0xcadams, line 139: When we move to tailwind, can we remove all of the previous CSS? In this, they are competing for priority (they are both applying border radius here).

packages/www/components/AssetsTable/CreateAssetDialog/index.tsx
```diff
@@ -221,6 +222,7 @@ const CreateAssetDialog = ({
             <Button
               css={{ display: "flex", ai: "center" }}
```
0xcadams, line 223: Same here - if we touch styling of a component, we should move it entirely to Tailwind.

packages/www/components/FeaturesModel/index.tsx
```diff
@@ -45,7 +45,9 @@ export default function FeaturesModel() {
 
   return (
     <AlertDialog open={shouldShowFeature}>
-      <AlertDialogContent css={{ p: "0", borderRadius: 0 }}>
+      <AlertDialogContent
+        className="bg-surface"
```
0xcadams, line 49: This should be styled at the component level, not at every Alert Dialog level. We don't want to have to apply this class to *every* dialog, the underlying dialog should be responsible for that.

packages/www/components/GettingStarted/index.tsx
```diff
@@ -140,7 +141,7 @@ curl --request POST \
           </Heading>
         </Flex>
         <Link href={appendProjectId("/settings")} passHref legacyBehavior>
-          <Button as="a" size="2">
+          <Button className="bg-accent" as="a" size="2">
```
0xcadams, line 144: Same thing here, we should have a handful of consistent Button styles, and then use that to re-style the dash.

0xcadams (commented):
Generally let's sync on how we want to go about the redesign. I'd like to add the shadcn components (some `components` folder with Button, Dialog, etc), get those styled according to our style guide, and *then* move to using those in the dashboard. Right now this is top-down, and it makes more sense to go bottom-up

suhailkakar:
Thanks Chase, sharing the same reply from discord:

__

Like we talked earlier, we're planning to switch all our components from @livepeer/design to shadcn and tailwindcss and realized it’s  will take some time, so we’re gonna tackle it step by step instead of in one big go.

To make sure our new stuff matches the Studio dashboard, we need to start by just changing the colors. This way, even if we roll out small updates one by one, they’ll still fit with the dashboard theme. I know the Alert background component should ideally be updated on component level, but since we don’t have the updated component yet, we need to update the colors first. This means we’ll replace the components gradually. We can’t do it all at once, so we’ll continue by adding the new component with updated colors and remove the current components along with their styles. 

Hope this clears things up! Let me know if you want to chat more about this. We can also hop on a call if that helps.

packages/www/components/AssetDetails/AssetOverviewTab.tsx
suhailkakar: Yep, but we will create this component from scratch with shadcn, so it doesn't make sense to style all of it here again

packages/www/components/AssetsTable/CreateAssetDialog/index.tsx
suhailkakar: Yes, but again, we will create all of them from scratch, and once we do all of them should have styling with tailwind

packages/www/components/FeaturesModel/index.tsx
suhailkakar: Same here... once we create the Alert from scratch, all the styling will be inside the component level

packages/www/components/GettingStarted/index.tsx
suhailkakar: Yep, with `shadcn` we can do that, but as we are still didn't migrated all component, we can keep them as it is and just change color

PR #2265, opened 2024-07-15
www: fix upcoming invoice & update pricing lookup key

gioelecerati:

- Fix upcoming invoice to reflect the recent pricing changes
- The table is displaying limits for Sandbox plan
- The table is now displaying the correct overage spend calculation for the `Growth` plan, considering the minimum spend

Also, this updates the storage lookup key to the newly created product. For tax information info, we cannot update the price of the previous product, so a new one got created. This ensures that newly created users will have the right product assigned, and usage records are reported to the right item price

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stripe.ts
+++ b/packages/api/src/controllers/stripe.ts
@@ -193,26 +193,26 @@ const sendUsageRecordToStripe = async (
   // Invoice items based on overusage
   await Promise.all(
     products[user.stripeProductId].usage.map(async (product) => {
-      if (product.name === "Transcoding") {
-        await stripe.subscriptionItems.createUsageRecord(
-          subscriptionItemsByLookupKey["transcoding_usage"],
-          {
-            quantity: parseInt(overUsage.TotalUsageMins.toFixed(0)),
-            timestamp: Math.floor(new Date().getTime() / 1000),
-            action: "set",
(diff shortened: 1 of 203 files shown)


packages/api/src/controllers/stripe.ts
```diff
@@ -209,19 +209,32 @@ const sendUsageRecordToStripe = async (
...
+            subscriptionItemsByLookupKey["tstorage_usage2"],
+            {
+              quantity: parseInt(overUsage.StorageUsageMins.toFixed(0)),
+              timestamp: Math.floor(new Date().getTime() / 1000),
+              action: "set",
+            },
+          );
+        } catch (e) {
+          // Fallback to old storage lookup key if new one fails.
+          // To avoid breaking the usage reporting during migration
```
victorges: Will there be an exception if the other lookup key doesn't exist? afaik the lookup will just return `undefined`

victorges (approved):
`yarn prettier` before merging

packages/api/src/controllers/stripe.ts
gioelecerati: Yeah you're right, no exception, just a bad status code from stripe. I think we can just migrate without this try catch then

PR #2268, opened 2024-07-17
api: task: add projectId to tasks

gioelecerati:

- Add projectId to tasks. The project is either the projectId of the input asset, the output asset or the defaultprojectId is neither is specified or existing

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -1091,6 +1091,8 @@ components:
           index: true
         outputAssetId:
           index: true
+        projectId:
+          index: true
         requesterId:
           type: string
           description: ID of the requester hash(IP + SALT + PlaybackId)
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -350,14 +350,14 @@ export class TaskScheduler {
     user: User,
     inputAsset?: Asset,
     outputAsset?: Asset,
-    userId?: string,
     requesterId?: string,
   ) {
-    const projectId = inputAsset?.projectId || user?.defaultProjectId;
+    const projectId =
+      inputAsset?.projectId || outputAsset.projectId || user?.defaultProjectId;
     const task = await this.createTask(
       type,
       params,
-      userId,
+      user.id,
       projectId,
       inputAsset,
       outputAsset,
@@ -371,8 +371,8 @@ export class TaskScheduler {
   async createTask(
     type: Task["type"],
     params: Task["params"],
-    userId?: string,
-    projectId?: string,
+    userId: string,
+    projectId: string,
     inputAsset?: Asset,
     outputAsset?: Asset,
     requesterId?: string,
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -383,18 +383,24 @@ export class TaskScheduler {
     userId?: string,
     requesterId?: string,
   ) {
+    const uId = inputAsset?.userId || outputAsset?.userId || userId;
+    const user = await db.user.get(uId, { useCache: true });
     const task = await db.task.create({
       id: uuid(),
       createdAt: Date.now(),
       type: type,
       outputAssetId: outputAsset?.id,
       inputAssetId: inputAsset?.id,
-      userId: inputAsset?.userId || outputAsset?.userId || userId,
+      userId: uId,
       params,
       status: {
         phase: "pending",
         updatedAt: Date.now(),
       },
+      projectId:
+        inputAsset?.projectId ||
+        outputAsset?.projectId ||
+        user.defaultProjectId,
       requesterId,
     });
     await this.queue.publishWebhook("events.task.spawned", {
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -361,11 +361,11 @@ export class TaskScheduler {
     const task = await this.createTask(
       type,
       params,
+      projectId,
       inputAsset,
       outputAsset,
       userId,
       requesterId,
-      projectId,
     );
 
     if (user?.disabled) {
@@ -379,11 +379,11 @@ export class TaskScheduler {
   async createTask(
     type: Task["type"],
     params: Task["params"],
+    projectId: string,
     inputAsset?: Asset,
     outputAsset?: Asset,
     userId?: string,
     requesterId?: string,
-    projectId?: string,
   ) {
     const uId = inputAsset?.userId || outputAsset?.userId || userId;
     const user = await db.user.get(uId, { useCache: true });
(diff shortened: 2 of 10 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -378,18 +378,24 @@ export class TaskScheduler {
     userId?: string,
     requesterId?: string,
   ) {
+    const uId = inputAsset?.userId || outputAsset?.userId || userId;
+    const user = await db.user.get(uId, { useCache: true });
```
victorges: I think the caller of this function always has the `User` object available on `req.user`. Shouldn't we just pass that down in the args instead of risking another DB round-trip here? Or even, they can just send the `req.project`, or even send _only_ the `req.project.id` and avoid any repetition of the default project ID logic here.

packages/api/src/schema/db-schema.yaml
```diff
@@ -1085,6 +1085,8 @@ components:
           index: true
         outputAssetId:
           index: true
+        projectId:
+          index: true
```
victorges, line 1095: don't we need to modify all the task APIs to filter by the context project ID as well?

victorges (commented):
need clarification on the task APIs behavior

packages/api/src/task/scheduler.ts
```diff
@@ -352,23 +352,24 @@ export class TaskScheduler {
     userId?: string,
     requesterId?: string,
   ) {
+    let uId = inputAsset?.userId || outputAsset?.userId || userId;
+    let user: User;
+    if (uId) {
+      user = await db.user.get(uId);
+    }
+    const projectId = inputAsset?.projectId || user?.defaultProjectId;
```
victorges: Shouldn't the caller already send the user object and/or projectId? Haven't checked it deeply in the code but I feel like it will always be available (after all, the caller sends a `userId` here)

The `projectId` is less of a priority, but we should really avoid doing an extraneous DB query whenever possible (for the user). Every query is more DB load and more possibility of failure.

packages/api/src/task/scheduler.ts
```diff
@@ -352,23 +352,24 @@ export class TaskScheduler {
...
 
-    let uId = inputAsset?.userId || outputAsset?.userId || userId;
-
-    if (uId) {
-      const user = await db.user.get(uId);
-
-      if (user?.disabled) {
-        throw new Error("user is disabled");
-      }
+    if (user?.disabled) {
```
victorges: I don't think we should ever be executing this logic without a user.

packages/api/src/task/scheduler.ts
```diff
@@ -378,23 +379,27 @@ export class TaskScheduler {
   async createTask(
     type: Task["type"],
     params: Task["params"],
+    projectId: string,
```
victorges, line 375: Similarly, we should never have a `projectId` but not have a `userId`. Can you make this interface clearer?

packages/api/src/task/scheduler.ts
```diff
@@ -378,23 +379,27 @@ export class TaskScheduler {
...
     inputAsset?: Asset,
     outputAsset?: Asset,
     userId?: string,
     requesterId?: string,
   ) {
+    const uId = inputAsset?.userId || outputAsset?.userId || userId;
+    const user = await db.user.get(uId, { useCache: true });
```
victorges: Should not be re-fetching the user everywhere. Can you just pass the object around?

victorges (approved):
LGTM but I think this is overcomplicating things grabbing the `userId` from several different places and re-fetching it from the DB all the time. I think it could be greatly simplified by not making arguments optional and just passing the user object around. The tasks logic should _never_ execute without a user.

packages/api/src/task/scheduler.ts
```diff
@@ -347,54 +347,50 @@ export class TaskScheduler {
   async createAndScheduleTask(
     type: Task["type"],
     params: Task["params"],
+    user: User,
     inputAsset?: Asset,
     outputAsset?: Asset,
     userId?: string,
     requesterId?: string,
   ) {
+    const projectId = inputAsset?.projectId || user?.defaultProjectId;
```
victorges: Shouldn't this come from the request API key instead? i.e. `req.project?.id || user.defaultProjectId` (user is never null, should not `?`)

packages/api/src/task/scheduler.ts
```diff
@@ -347,54 +347,50 @@ export class TaskScheduler {
   async createAndScheduleTask(
     type: Task["type"],
     params: Task["params"],
+    user: User,
     inputAsset?: Asset,
     outputAsset?: Asset,
     userId?: string,
```
victorges: This arg is redundant now 👀

packages/api/src/task/scheduler.ts
```diff
@@ -347,54 +347,50 @@ export class TaskScheduler {
...
-
     await this.scheduleTask(task);
     return task;
   }
 
   async createTask(
     type: Task["type"],
     params: Task["params"],
+    userId?: string,
+    projectId?: string,
```
victorges: Should these really be optional? I think we always need a user and project.

packages/api/src/task/scheduler.ts
```diff
@@ -347,54 +347,50 @@ export class TaskScheduler {
...
     inputAsset?: Asset,
     outputAsset?: Asset,
-    userId?: string,
     requesterId?: string,
   ) {
+    const uId = inputAsset?.userId || outputAsset?.userId || userId;
```
victorges: `userId` arg should be supreme, no need to grab it from other places. In fact, if any of these ids mismatch it should be a fatal error (no need to check tho, I think we can just assume they are the same)

packages/api/src/webhooks/cannon.ts
```diff
@@ -603,6 +603,8 @@ export default class WebhookCannon {
         this.secondaryRecordObjectStoreId,
       );
 
+      const user = await db.user.get(session.userId);
```
victorges: Maybe we should move this fetch https://github.com/livepeer/livepeer-com/blob/c87b3a1176de9ffca972c6ff6750149ec70a0659/packages/api/src/webhooks/cannon.ts#L173 to the beginning of the function and then pass the `user` around?

Otherwise if that's too big of a change for now, we can probably add `useCache` to both of those fetches (that and this) to hopefully avoid a db round trip

packages/api/src/task/scheduler.ts
gioelecerati: Makes sense, but in the `clip` case the owner is different than the requester - I could pass req.project.id everywhere but there, but I instead removed userId and I am passing the user object, passing the owner during the clip gen, how does it look?

gioelecerati: Seeing a bunch of test failure after addressing the comments tho, looking into it

PR #2270, opened 2024-07-18
ui: add asset status on assets page

suhailkakar:

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/www/layouts/assetDetail.tsx
+++ b/packages/www/layouts/assetDetail.tsx
@@ -64,7 +64,7 @@ const AssetDetail = ({
         await refetchAsset();
       }
     },
-    [asset, patchAsset, refetchAsset]
+    [asset, patchAsset, refetchAsset],
   );
 
   useEffect(() => {
--- a/packages/www/layouts/assetDetail.tsx
+++ b/packages/www/layouts/assetDetail.tsx
@@ -18,6 +18,8 @@ import {
   useEffect,
   useState,
 } from "react";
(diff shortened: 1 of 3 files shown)


packages/www/layouts/assetDetail.tsx
```diff
@@ -86,17 +88,44 @@ const AssetDetail = ({
...
+          ) : asset?.status?.phase === "processing" ||
+            asset?.status?.phase === "uploading" ? (
+            <Alert variant="warning">
+              <Clock10 size={18} />
+              <AlertTitle>
+                <span className="capitalize">{asset.status.phase}</span> the
+                asset
+                {asset.status.progress
+                  ? ` (${asset.status.progress * 100}%)`
+                  : ""}
```
victorges, line 113: I guess this will only be the case when the asset is uploading. If it's easy enough, could show the upload progress here as well which should be available on the frontend. Nevermind if not easy tho

victorges (approved):

packages/www/layouts/assetDetail.tsx
suhailkakar: ha, yeah! sounds good!

I will add that as well :)

suhailkakar: Just catching up with the PRs :)

i've checked it and for now, it might be a little complicated because when the file is uploading, we don't have their asset Id, we could filter the `getFilteredFileUploads` by name, but that is not really ideal solution. 

I'm going to merge this PR, but once we completed the asset redesign, I will make sure to include it as well :)

PR #2273, opened 2024-07-23
api: Log AI generate requests in DB

victorges:

This implements the logging of all AI Generate requests both in the DB and in Prometheus.

- create new schema for the log resource
- implement logic in API handler to save entry on db
- implement logic in API handler to send Prometheus metric with detailed latency 

`yarn test`

Implements ENG-2182

**Checklist**


packages/api/src/controllers/generate.ts
```diff
@@ -39,58 +167,49 @@ function registerGenerateHandler(
...
-      }
-      if (response.status >= 500) {
-        return res.status(500).json({ errors: [`Failed to generate ${name}`] });
+      if (gatewayRes.status >= 500) {
+        // We hide internal server error details from the user.
+        return res.status(500).json({ errors: [`Failed to generate ${type}`] });
       }
 
-      res.status(response.status).json(body);
+      // forward content metadata headers
```
leszko, line 197: Do I understand correctly that this is not related to any logging, it's just an improvement to proxy back to the user the headers from the response?

packages/api/src/schema/api-schema.yaml
```diff
@@ -2858,6 +2858,72 @@ components:
             #/components/schemas/new-asset-payload/properties/targetSegmentSizeSecs
     # AI Generate payloads. Keep in mind that these use snake_case instead of camelCase since
     # they implement the same interface as the AI Gateway Livepeer node.
+    ai-generate-log:
```
leszko: Shouldn't this schema be in `db-schema.yaml`? I guess it's related to our (internal) DB not to the API exposed to the users. Wdyt?

packages/api/src/controllers/generate.ts
```diff
@@ -39,58 +167,49 @@ function registerGenerateHandler(
...
-          form.append(file.fieldname, file.buffer, {
-            filename: file.originalname,
-            contentType: file.mimetype,
-            knownLength: file.size,
-          });
+        response = await gatewayRes.buffer();
+      } catch (err) {
+        error = err;
+      } finally {
+        logAiGenerateRequest(
```
leszko, line 177: What happens if this throws an exception?

packages/api/src/controllers/generate.ts
```diff
@@ -1,33 +1,161 @@
...
+const aiGenerateDurationMetric = new promclient.Histogram({
+  name: "livepeer_api_ai_generate_duration",
+  buckets: [100, 500, 1000, 2500, 5000, 10000, 30000, 60000],
+  help: "Duration in milliseconds of the AI generation request",
+  labelNames: ["type", "complexity"], // see aiGenerateMetricComplexity
+});
+
+// We divide obervations in "complexity" buckets. The complexity is only relevant on image generation pipelines that
+// have num images and/or inference steps params. The complexity is based on the product of them.
+function aiGenerateMetricComplexity(request: AiGenerateLog["request"]) {
```
leszko: What is this complexity? 🤔 Is is something well-known in the AI world or did we define it (just to bucket somehow the metrics in Prometheus?

packages/api/src/controllers/generate.ts
```diff
@@ -1,33 +1,161 @@
...
+    form.append(key, value);
+  }
+  for (const file of req.files) {
+    form.append(file.fieldname, file.buffer, {
+      filename: file.originalname,
+      contentType: file.mimetype,
+      knownLength: file.size,
+    });
+    // We save only the hash of the file on the `payload` since that will be saved in the DB.
+    // TODO: Make this save the contents of the file in some OS and only reference them here.
```
leszko, line 67: Why would we ever save the payload file content?

packages/api/src/controllers/generate.test.ts
```diff
@@ -190,4 +224,113 @@ describe("controllers/generate", () => {
...
+
+      const res = await client.post("/beta/generate/text-to-image", {
+        prompt: "a man in a suit and tie",
+      });
+      expect(res.status).toBe(500);
+      expect(await res.json()).toEqual({
+        errors: ["Failed to generate text-to-image"], // 5xx errors get hidden from end users
+      });
+
+      await sleep(100);
```
leszko, line 288: The question I always ask, won't it make the tests flaky? Don't we need to wait in the loop with timeout?

leszko (approved):
Added some comments. Other than that LGTM

gioelecerati (approved):

packages/api/src/controllers/generate.ts
victorges: Actually this was necessary because I started treating the response as a buffer instead of always a `json()`. This was required just to get to the guarantee we don't miss information (like response content) when the AI Gateway gives a bad response that is not a JSON (like a literal `500 Internal Server Error`, could be some proxy, idk)

So previously, this code would call sth like `res.json()` that would headers appropriately. Now that we proxy the response verbatim with `buffer()` and `send()`, we need to proxy the content headers too.

packages/api/src/controllers/generate.ts
victorges: We have a general exception handler in the API, so it becomes a `500` [here](https://github.com/livepeer/livepeer-com/blob/a9979b7bffa8e1852aa191efe70a7a894cc8c733/packages/api/src/middleware/errorHandler.ts#L23)

packages/api/src/controllers/generate.ts
victorges: It's actually something I made up. I wanted to get some use out of this prometheus metric apart from the ai logs table itself, so thought of buckets it'd be useful to break by and use to understand behavior on prometheus:
- one bucket is the histogram itself, useful to see latency percentiles (relevant for our arch where models can be not warm sometimes)
- another bucket is this "complexity", so we can see how the latency scales linearly (or not) with these params, and monitor the request pattern, see if it changes etc

WDYT? I can drop this complexity label if you think it's unnecessary/should do all analysis from db

packages/api/src/controllers/generate.ts
victorges: More to debug/reproduce the AI generations if we get weird reports for example.

The log of the input/outputs could also be useful to train/fine tune other AI models (e.g. if users gave feedback to those generations). So if we ever get to that point could want to do this. We could consider starting to store them early tho, to collect data ahead of time.

packages/api/src/controllers/generate.test.ts
victorges: I think it is theoretically possible for this test to fail yeah. But we really only need this because we save the log with a `setImmediate` instead of synchronously. `setImediate` executes as soon as possible on the node event loop. Then it's going to save on the DB on localhost, which also shouldn't take more than a couple `ms`.

I could make this into a loop tho if you think there's still a risk.

packages/api/src/controllers/generate.ts
leszko: TBH I believe we don't need it. And "complexity" brings some complexity 🙃 

The thing is that that Prometheus metrics is mostly about monitoring and I believe we don't need the complexity here. To analyze the data, I guess we'll anyway go to DB.

packages/api/src/controllers/generate.test.ts
leszko: Not sure to be honest. I'm somewhere on a fence, because:
- on one hand => I think we should use the pattern of retries everywhere, because if we have a lot of these, then at some point we'll start having flaky test (and it's a pain, we already have them in go-livepeer, catalyst-api, and catalyst)
- on the other hand => we have these in many places, so we can leave it for now and take the reactive approach, so leave it as it as and maybe it won't happen in practice

victorges: OK, let's leave as is and consider some general refactor later. Can probably create some kind of helper like `waitUntil` with a check period/timeout, and replace every usage of `sleep` on tests with that.

packages/api/src/controllers/generate.ts
victorges: That makes sense. Replaced with a status code label instead.

PR #2274, opened 2024-07-24
api/generate: Implement rate limiting based on req logs

victorges:

This creates the capability of rate-limiting API calls to the `/generate` endpoints.
It uses the request log built on #2273 to track the rate of requests and limits based
on a CLI argument.

We can later build support for per billing plan limits depending on the user, but I left
that for the future.

- Create new rate limit logic on generate API through a middleware
- Add tests to the rate limiter logic

`yarn test`

Implements ENG-2180

**Checklist**

### DIFF
--- a/packages/api/src/controllers/generate.ts
+++ b/packages/api/src/controllers/generate.ts
@@ -223,7 +223,7 @@ function registerGenerateHandler(
         }
       }
 
-      if (gatewayRes.status >= 500) {
+      if (!req.user.admin && gatewayRes.status >= 500) {
         // We hide internal server error details from the user.
         return res.status(500).json({ errors: [`Failed to generate ${type}`] });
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/generate.ts
```diff
@@ -15,6 +16,7 @@ import { experimentSubjectsOnly } from "./experiment";
 import { pathJoin2 } from "./helpers";
 
 const AI_GATEWAY_TIMEOUT = 10 * 60 * 1000; // 10 minutes
+const MINUTE_MS = 60 * 1000;
```
leszko: Maybe call it with something meaningful, like `RATE_LIMIT_TIMEOUT`?

leszko (approved):

PR #2280, opened 2024-08-02
api: webhook: improve asset webhooks behavior

gioelecerati:

There is currently some confusion in the way that the asset webhooks work.
This is to address the issue, making these changes to how webhooks behave

- Make sure the payload sent on asset.deleted is from before any internal updates made to the asset.
- Stop sending asset.updated for deleted assets.
- Move all assets to deleting when deleted, not only ready
- Stop updating assets deleting phase if there are further updates

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -602,7 +602,7 @@ export class TaskScheduler {
       );
     }
     const event = updates.deleted ? "asset.deleted" : "asset.updated";
-    const publishWebhook = async (asset: Asset) => {
+    const publishWebhook = async (asset: Asset, event: EventKey) => {
       const snapshot = await toExternalAsset(asset, this.config, true);
       const timestamp = Date.now();
       await this.queue.publishWebhook(`events.${event}`, {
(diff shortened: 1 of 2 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -639,14 +644,14 @@ export class TaskScheduler {
       await this.queue.publishWebhook(routingKey, {
```
victorges: Should this use the `publishWebhook` helper too?

packages/api/src/task/scheduler.ts
```diff
@@ -608,29 +601,41 @@ export class TaskScheduler {
...
+            id: asset.id,
+            snapshot,
+          },
+        },
+      });
+    };
+
+    if (event === "asset.deleted") {
+      await publishWebhook(asset);
+    }
```
victorges: WDYT of sending this after the `db.asset.update` below (but before we update the `asset` var)? Otherwise, we may send the event then the update fails

victorges (approved):
Great!

packages/api/src/task/scheduler.ts
gioelecerati: Make sense! doing it

packages/api/src/task/scheduler.ts
gioelecerati: Done! Could you take a look to the routingKeys :pray: ? @victorges  just a double check, as I needed to move things around

PR #2281, opened 2024-08-03
api: user: mark user as deleted

gioelecerati:

Mark the user as deleted instead of hard deleting the row from the db

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -296,7 +296,11 @@ app.delete("/:id", authorizer({ anyAdmin: true }), async (req, res) => {
     res.status(404);
     return res.json({ errors: ["Account not found"] });
   }
-  await db.user.delete(id);
+
+  await db.user.update(id, {
+    deleted: true,
+    deletedAt: Date.now(),
+  });
 
   // TODO: remove all streams owned by user
   // TODO: remove all assets owned by user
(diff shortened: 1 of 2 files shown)


thomshutt (approved):

packages/api/src/controllers/user.ts
```diff
@@ -296,7 +296,11 @@ app.delete("/:id", authorizer({ anyAdmin: true }), async (req, res) => {
     res.status(404);
     return res.json({ errors: ["Account not found"] });
   }
-  await db.user.delete(id);
+
+  await db.user.update(id, {
+    deleted: true,
+    deletedAt: Date.now(),
+  });
```
victorges, line 303: I don't think it is that simple. We have several places that read the user object which don't check either of these fields. I believe we should either keep it as a hard-delete or do this big refactor fixing these places.

But IMO as an after-thought, we probably want to just keep the hard delete. This is only done in case of GDPR requests, in which case we do want to completely delete any internal records right? For any other cases we have both the `suspended` and `disabled` fields to use.

victorges (changes requested):

packages/api/src/controllers/user.ts
gioelecerati: Yeah, it was opened in a rush, but let's close it and I focus on making a proper change to deleted all the streams + assets + sessions when a user is hard deleted. 

We may also consider to delete all the contents but keep the user row redacting all the personal info to avoid userIds pointing to non exsiting users, wdyt? Is it GDPR compliant?

PR #2282, opened 2024-08-06
Pass through UTM params on registration

thomshutt:
This will allow marketing to better track where registrations are coming from. The parameters should already be picked up and passed to Hubspot.

It also moves Register to using a different Hubspot form ID to Login, to allow us to differentiate the flows better.


0xcadams (approved):
LGTM

gioelecerati (approved):
LGTM - haven't checked the frontend in the box, but in the vercel preview everything looks fine

PR #2283, opened 2024-08-08
api: scheduler: improve fetch of task during process task event

gioelecerati:

- Use `get` instead of `find` when fetching a task incoming in a process task event
- Do not use replica for it, to improve consistency
- Add a second retry after 500ms in case the task is missing

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -99,12 +99,20 @@ export class TaskScheduler {
   }
 
   async processTaskEvent(event: messages.TaskResult): Promise<boolean> {
-    const tasks = await db.task.find({ id: event.task.id });
-    if (!tasks?.length || !tasks[0].length) {
-      console.log(`task event process error: task ${event.task.id} not found`);
-      return true;
+    let task = await db.task.get(event.task.id, { useReplica: false });
+    if (!task) {
+      console.log(
+        `task event process error: task ${event.task.id} not found, retrying after 500ms`,
+      );
+      await new Promise((resolve) => setTimeout(resolve, 500));
+      task = await db.task.get(event.task.id);
+      if (!task) {
+        console.log(
+          `task event process error: task ${event.task.id} not found in process task event`,
+        );
+        return true;
+      }
     }
-    const task = tasks[0][0];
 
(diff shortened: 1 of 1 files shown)


packages/api/src/task/scheduler.ts
```diff
@@ -99,12 +99,20 @@ export class TaskScheduler {
   }
 
   async processTaskEvent(event: messages.TaskResult): Promise<boolean> {
-    const tasks = await db.task.find({ id: event.task.id });
-    if (!tasks?.length || !tasks[0].length) {
-      console.log(`task event process error: task ${event.task.id} not found`);
-      return true;
+    let task = await db.task.get(event.task.id, { useReplica: false });
+    if (!task) {
```
victorges: check `task.deleted` too

packages/api/src/task/scheduler.ts
```diff
@@ -99,12 +99,20 @@ export class TaskScheduler {
...
+      console.log(
+        `task event process error: task ${event.task.id} not found, retrying after 500ms`,
+      );
+      await new Promise((resolve) => setTimeout(resolve, 500));
```
victorges: I don't think this is necessary if we are already not using the replica. It is not possible to have inconsistencies when reading the primary, so if the task doesn't exist it won't be created later either. So this will overcomplicate the code unnecessarily IMO

packages/api/src/task/scheduler.ts
```diff
@@ -211,12 +219,20 @@ export class TaskScheduler {
...
```
victorges: Same for checking `deleted` and not having this `sleep`.

victorges (approved):
LGTM either way, but we should avoid having the sleep-retry logic which makes the code more complex but is not necessary

packages/api/src/task/scheduler.ts
gioelecerati: added the check!

packages/api/src/task/scheduler.ts
gioelecerati: removed all the sleeps

victorges (approved):

PR #2286, opened 2024-08-14
Don't use db replica for asset get when trying to update asset

mjh1:
This seems to have caused a vod problem where we failed to read an asset from the db when processing the task complete event

### DIFF
--- a/packages/api/src/task/scheduler.ts
+++ b/packages/api/src/task/scheduler.ts
@@ -592,7 +592,7 @@ export class TaskScheduler {
     filters?: { allowedPhases: Array<Asset["status"]["phase"]> },
   ) {
     if (typeof asset === "string") {
-      asset = await db.asset.get(asset);
+      asset = await db.asset.get(asset, { useReplica: false });
     }
     const phaseChanged =
       updates.status && asset.status.phase !== updates.status.phase;


thomshutt (approved):

packages/api/src/task/scheduler.ts
```diff
@@ -592,7 +592,7 @@ export class TaskScheduler {
     filters?: { allowedPhases: Array<Asset["status"]["phase"]> },
   ) {
     if (typeof asset === "string") {
-      asset = await db.asset.get(asset);
+      asset = await db.asset.get(asset, { useReplica: false });
```
victorges, line 595: Reminder that what we really need to investigate is how the replication lag is getting so high. It is not normal for the db to be so out of date that _minutes_ after creating the asset/task it is still not in the replicas.

That said, if we're going the "drop the replica" approach, we should probably update every call to the DB in this file to use the primary instead. Can just search for `db.` for all calls to the `db` and add the replica option

victorges (approved):

mjh1:
Merging this despite the test failure since this is happening in master branch too: https://github.com/livepeer/studio/actions/runs/10390427712/job/28854769615
Not ideal but we can fix that separately.

PR #2288, opened 2024-08-16
Don't use recording fallback logic if no secondary exists

mjh1:
This was needed for the new recording fallback logic to work, if we had to use the fallback location to write the manifest file to then we don't want studio to try to send the fallback location itself. We want studio to continue to send the primary location and let catalyst's new fallback logic to handle it.


victorges (approved):
LGTM

PR #2291, opened 2024-08-16
api: webhook: by project

gioelecerati:

- Send webhooks only to webhook created on related projectId (i.e. asset.created on P1 will be sent only to webhooks existing in P1, not P2)

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/webhooks/cannon.test.ts
+++ b/packages/api/src/webhooks/cannon.test.ts
@@ -392,35 +392,25 @@ describe("webhook cannon", () => {
       expect(calledFlags).toEqual([true, true]);
     });
 
-    it("should receive events for related project and not for unrelated", async () => {
-      let res = await client.post(`/webhook?projectId=${nonAdminProject.id}`, {
-        ...mockWebhook,
-        name: "test-related",
-      });
-      expect(res.status).toBe(201);
-
+    it("should not receive events for unrelated project", async () => {
       const differentProject = await db.project.create({
         name: "different project",
         id: uuid(),
         userId: nonAdminUser.id,
       });
 
-      res = await client.post(`/webhook?projectId=${differentProject.id}`, {
+      let res = await client.post(`/webhook?projectId=${differentProject.id}`, {
         ...mockWebhook,
         name: "test-unrelated",
       });
 
       expect(res.status).toBe(201);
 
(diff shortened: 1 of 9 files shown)


packages/api/src/webhooks/cannon.ts
```diff
@@ -123,7 +123,7 @@ export default class WebhookCannon {
   }
 
   async processWebhookEvent(msg: messages.WebhookEvent): Promise<boolean> {
-    const { event, streamId, sessionId, userId } = msg;
+    const { event, streamId, sessionId, userId, projectId } = msg;
```
victorges, line 127: Can you make the `projectId` field non-optional on this `WebhookEvent` type just to have the type safety that we will never send events without a project ID? It is definitely not an optional field, just like the `userId`.

victorges: In fact, I tried doing that and **several** places broke (most tests, but some actual runtime code [1]). So please do that, fix the errors, and let's keep in mind to use typing in our favor to catch thing like this.

[1] e.g. no `projectId` passed [here](https://github.com/livepeer/livepeer-com/blob/d983f4573ee3dd47534142eee0430a50551fe6ab/packages/api/src/task/scheduler.ts#L511) :)

victorges (changes requested):

victorges (commented):
Can you also add a proper title/description to the PR and make sure you squash the commits when merging? Ideally the commits should have proper titles/descriptions themselves, but no need to worry to change the ones already there.

packages/api/src/webhooks/cannon.test.ts
```diff
@@ -327,21 +385,69 @@ describe("webhook cannon", () => {
...
         userId: nonAdminUser.id,
-        projectId: nonAdminUser.defaultProjectId,
+        projectId: nonAdminProject.id,
       });
 
       await Promise.all(sems.map((s) => s.wait(3000)));
       expect(calledFlags).toEqual([true, true]);
     });
 
-    it("should send multiple events to same webhook", async () => {
```
victorges, line 319: Could you keep these tests separate? It gets hard to understand/debug if the test case is too long and from the description it seems like they are testing different things.

victorges (approved):
LGTM!

PR #2293, opened 2024-08-18
feat: new usage charts

suhailkakar:

**Screenshots (optional)**

**Checklist**


0xcadams:
It looks like prettier is failing on this

suhailkakar:
Let's wait for the redesign pr (#2298) to get merged and then we can merge this so both are same design and colors

packages/www/components/ui/chart.tsx
```diff
@@ -0,0 +1,370 @@
...
+      labelKey,
+    ]);
+
+    if (!active || !payload?.length) {
+      return null;
+    }
+
+    const nestLabel = payload.length === 1 && indicator !== "dot";
+    const formatValue = (value) => {
+      if (value >= 1000) {
```
louisbrown0212, line 176: I think it might be better to update it with `value >= 100`.
so that it can show `0.9k` instead of `952` for example.

PR #2295, opened 2024-08-21
api/schema: Make AI schema automatic

victorges:
In order to add the AI APIs into our schema I decided to just make this process automatic.
Created a `pull-ai-schema` script that pulls the schema from the AI Gateway source of
truth ([ai-worker](https://github.com/livepeer/ai-worker/blob/main/runner/gateway.openapi.yaml)) and downloads it to the repository after some changes (like the path prefix).

This will avoid manual work in future updates to this schema, by hopefully being able to
simply run the `pull-ai-schema` script and have things up to date.

- Create `pull-ai-schema` script
- Implement patches to the schema on that script
- Implement support for separate AI Schema file on `compile-schemas`
- Add tests for new pipeline coming in (segment-anything-2)

- `yarn pull-ai-schema`
- `yarn compile-schemas`
- `yarn test`
- Generate Speakeasy SDK using the 2 schema files (need to update their config)

Implements ENG-2251

**Checklist**


packages/api/src/schema/ai-api-schema.yaml
```diff
@@ -0,0 +1,869 @@
...
+          description: The generated images.
+      type: object
+      required:
+        - images
+      title: VideoResponse
+      description: Response model for image generation.
+    chunk:
+      properties:
+        timestamp:
+          items: {}
```
gioelecerati, line 844: Not sure I ever encountered this type, is `{}` just defining a generic object? could it be defined specifically as an array of numbers with [number]?

packages/api/src/schema/compile-schemas.js
```diff
@@ -24,23 +24,47 @@ const write = (dir, data) => {
   console.log(`wrote ${dir}`);
 };
 
-const schemaDir = path.resolve(__dirname, "schema");
+// Remove the title from the schema to avoid conflicts with the TypeScript type name
+function removeAllTitles(schema) {
```
gioelecerati, line 28: Just a thought, any risk of this affecting generating tools like documentation tools or sdk generating tools, or they are going to be merged anyway? If any of them require `title`

gioelecerati (approved):

packages/api/src/schema/ai-api-schema.yaml
victorges: This is defined in `ai-worker` repository, we only clone it here. Any more detailed specification should go there.

packages/api/src/schema/compile-schemas.js
victorges: This will not affect SDK/docs, since it's only used in build time to generate that combined schema that we use in our code! The SDK/docs will use the raw `api-schema.yaml` and `ai-api-schema.yaml` files.

PR #2296, opened 2024-08-23
Reduce timeout for auto stream start on pull API

mjh1:
The stream pull API was proving too slow so I'm reducing the auto-start timeout significantly, the processing for that request should still continue even when dropping it from the studio side (needs more testing). We then need to catch these timeout errors in a try catch and not propagate these to the user. So this auto start becomes more of a best effort process which I think is acceptable since even if something goes wrong the stream startup will happen when viewers start making requests.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -21,7 +21,7 @@ import { v4 as uuid } from "uuid";
 
 const ITERATIONS = 10000;
 const PAYMENT_FAILED_TIMEFRAME = 3 * 24 * 60 * 60 * 1000;
-const PULL_START_TIMEOUT = 60 * 1000;
+const PULL_START_TIMEOUT = 3 * 1000;
 
 const crypto = new Crypto();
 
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/helpers.ts
```diff
@@ -21,7 +21,7 @@ import { v4 as uuid } from "uuid";
 
 const ITERATIONS = 10000;
 const PAYMENT_FAILED_TIMEFRAME = 3 * 24 * 60 * 60 * 1000;
-const PULL_START_TIMEOUT = 60 * 1000;
```
victorges, line 24: FYI just in case the problem with these were the `nginx` timeouts, I've just [increased those from 60s to 300s](https://github.com/livepeer/livepeer-infra/pull/2298) for the API, so the request wont' get cut anymore.

packages/api/src/controllers/helpers.ts
```diff
@@ -669,34 +669,40 @@ export const triggerCatalystPullStart =
           );
         }
 
-        const deadline = Date.now() + 2 * PULL_START_TIMEOUT;
+        const deadline = Date.now() + PULL_START_TIMEOUT;
```
victorges, line 672: Doesn't this mean we will only really try once, for 3s?

packages/api/src/controllers/helpers.ts
```diff
@@ -669,34 +669,40 @@ export const triggerCatalystPullStart =
...
+            logger.warn(
+              `error while trying to trigger catalyst pull for stream=${
+                stream.id
+              } playbackUrl=${playbackUrl} error=${e}`,
+            );
+          }
           await sleep(250);
         }
-
-        throw new Error(`failed to trigger catalyst pull`);
```
victorges, line 699: So this means that if we can't get the stream pull to start we will return a success for the user as if it worked? Or is this handled somewhere else? Seems "optimistically misleading" to return a success if we didn't get a successful response to the playback URL.

victorges (approved):
LGTM either way but have some questions

packages/api/src/controllers/helpers.ts
mjh1: Yeah it's not a great solution, I'm just trying to do whatever I can to reduce the response time, if they take over 5 seconds then the customer will time them out and fallback to their provider. But I think maybe I have a better solution now, will open a different PR.

PR #2297, opened 2024-08-23
Buffer webhook status updates

mjh1:
We were finding the database was getting overloaded with these webhook status updates so we should buffer them up like we do with the lastSeen field.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/middleware/tracking.ts
+++ b/packages/api/src/middleware/tracking.ts
@@ -2,6 +2,7 @@ import sql from "sql-template-strings";
 import { ApiToken } from "../schema/types";
 import { db } from "../store";
 import Table from "../store/table";
+import { DBWebhook } from "../store/webhook-table";
 
 const flushDelay = 60 * 1000; // 60s
 
@@ -12,7 +13,7 @@ type PendingLastSeen = {
 };
 
 class Tracker {
-  pendingUpdates: Map<string, PendingLastSeen> = new Map();
+  pendingLastSeenUpdates: Map<string, PendingLastSeen> = new Map();
 
   recordUser(userId: string) {
     this.recordLastSeen(db.user, userId);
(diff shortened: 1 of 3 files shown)


packages/api/src/middleware/tracking.ts
```diff
@@ -60,13 +61,40 @@ class Tracker {
     }
   }
 
-  async flushAll() {
-    const all = this.pendingUpdates;
-    this.pendingUpdates = new Map();
```
mjh1, line 65: @victorges it seemed wrong that we were re-initialising the map here because then we try to read from it in `flushLastSeen` in the loop right?

packages/api/src/middleware/tracking.ts
```diff
@@ -60,13 +61,40 @@ class Tracker {
...
+    const alreadyScheduled = this.pendingWebhookStatusUpdates.has(key);
+    this.pendingWebhookStatusUpdates.set(key, status);
+    if (alreadyScheduled) return;
+
+    setTimeout(() => this.flushWebhookStatus(key), flushDelay);
+  }
 
-    for (const key of all.keys()) {
+  private async flushWebhookStatus(id: string) {
+    const status = this.pendingWebhookStatusUpdates.get(id) ?? {};
```
victorges: We should probably not have this `?? {}` default, otherwise line 78 will never be true and we might update the webhook with an empty status.
```suggestion
    const status = this.pendingWebhookStatusUpdates.get(id);
```

victorges (approved):
Neat!

So I also understand that the webhook response table is not a problem right now, right? Since it's append only and there's no index contention I guess. The problem is only with updating the same webhook object over and over

mjh1:
> Neat!
> 
> So I also understand that the webhook response table is not a problem right now, right? Since it's append only and there's no index contention I guess. The problem is only with updating the same webhook object over and over

That's right I think yeah, we were only seeing update contention

PR #2299, opened 2024-08-27
Add index on stream.multistream.targets

leszko:
Lack of this index caused issues with toggling multistream targets when user has many streams.

Fix https://linear.app/livepeer/issue/PS-793/august-22-2024-1142-am-stream-id-b924ca03-71fb-4af8-b526-087e2285576a#comment-7457b434

### DIFF
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -745,21 +745,10 @@ components:
           type: string
           example: D8321C3E-B29C-45EB-A1BB-A623D8BE0F65
         multistream:
-          type: object
-          additionalProperties: false
           properties:
             targets:
-              type: array
               index: true
               indexType: gin
-              example:
-                - id: PUSH123
-                  profile: 720p
-              description: |
-                References to targets where this stream will be simultaneously
-                streamed to
-              items:
-                $ref: "#/components/schemas/target"
         presets:
           type: array
           items:
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -744,6 +744,22 @@ components:
         objectStoreId:
           type: string
           example: D8321C3E-B29C-45EB-A1BB-A623D8BE0F65
+        multistream:
+          type: object
+          additionalProperties: false
+          properties:
+            targets:
+              type: array
+              index: true
+              indexType: gin
+              example:
+                - id: PUSH123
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/db-schema.yaml
```diff
@@ -744,6 +744,22 @@ components:
...
           type: string
           example: D8321C3E-B29C-45EB-A1BB-A623D8BE0F65
+        multistream:
+          type: object
+          additionalProperties: false
+          properties:
+            targets:
+              type: array
+              index: true
+              indexType: gin
```
victorges, line 751: FTR as we chatted offline: these are the only fields that we need to have in this `db-schema` file, while all the rest can be kept in the `api-schema` (i.e. no changes there). These files are deep merged so the fields added here simply override the fields in the `api-schema`.

That also means that there's no problem if we repeat the whole field definition to add the index fields here, but I generally try to avoid the redundancy and add here only the fields that can't go on the OpenAPI schema (our custom stuff, like indexes). This pattern has not been followed in this entire file tho

There's a `README` in this directory as well with some more info.

packages/api/src/store/table.ts
```diff
@@ -371,15 +371,6 @@ export default class Table<T extends DBObject> {
 
   // on startup: auto-create indices if they don't exist
   async ensureIndex(propName: string, prop: FieldSpec, parents: string[] = []) {
-    if (
-      process.env.NODE_ENV !== "test" &&
-      ["stream", "session"].includes(this.name)
```
victorges, line 376: Btw if you want to be extra sure that the indexes are being created normally you can also create a test [like this](https://github.com/livepeer/livepeer-com/blob/4e24f8e656e204089b0d0057428820045c162c89/packages/api/src/store/asset-table.test.ts#L8) for the `stream-table`

victorges (approved):
LGTM either way! But we should keep the `api-schema` unchanged to keep the field in our docs

packages/api/src/store/table.ts
leszko: Ok. I'll probably not add the test, because to check it would not only need to check if the index exists, but also its content. So, probably not worth adding it.

packages/api/src/schema/db-schema.yaml
leszko: Updated in c89f4293808695dca7edca506a785f68f9da51f3

PR #2306, opened 2024-09-05
Use stream metadata request to detect when stream is ready

mjh1:

Rather than wait for the playback call to finish we can improve the speed of this API by using the metadata call instead. We still make the playback call to ensure ingest ends up in the right region but the metadata endpoint will return a meta object once the stream is ready for playback, which only seems to take a second or two.

This was tested in staging: [Logs from a test in staging](https://eu-metrics-monitoring.livepeer.monster/grafana/explore?schemaVersion=1&panes=%7B%22bcz%22:%7B%22datasource%22:%22P8E80F9AEF21F6940%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P8E80F9AEF21F6940%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bapp%3D%5C%22staging-livepeer-api%5C%22%7D%20%7C~%20%5C%22pull%20start%7CgetCatalystStreamMetadata%20failed%7Cfailed%20to%20trigger%20catalyst%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22to%22:%221726141145714%22,%22from%22:%221726137545714%22%7D%7D%7D)

Screenshot 2024-09-12 at 13 20 30

https://linear.app/livepeer/issue/PS-778
https://linear.app/livepeer/issue/PS-572

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/helpers.ts
+++ b/packages/api/src/controllers/helpers.ts
@@ -656,6 +656,7 @@ export const triggerCatalystPullStart =
   process.env.NODE_ENV === "test"
     ? async () => {} // noop in case of tests
     : async (stream: DBStream, playbackUrl: string) => {
+        console.log("pull start for " + playbackUrl);
         const { lat, lon } = stream.pull?.location ?? {};
         if (lat && lon) {
           // Set the lat/lon qs to override the observed "client location" and
@@ -682,6 +683,7 @@ export const triggerCatalystPullStart =
             errHeader != "" &&
             !errHeader.includes("not allowed to view this stream");
           if (res.ok && !isHlsErr) {
+            console.log("pull start finished for " + playbackUrl);
             return;
           }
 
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/helpers.ts
```diff
@@ -699,6 +699,33 @@ export const triggerCatalystPullStart =
...
+    : async (url: string) => {
+        const deadline = Date.now() + 2 * PULL_START_TIMEOUT;
+        while (Date.now() < deadline) {
+          const res = await fetchWithTimeoutAndRedirects(url, {
+            method: "GET",
+            timeout: PULL_START_TIMEOUT,
+            maxRedirects: 10,
+          });
+          const body = await res.text();
+          if (res.ok && body.includes(`"meta":`)) {
```
leszko, line 718: Does the stream always include `meta: ` or can there be a case in which the playback works but the `meta: ` is not returned?

I'm trying to think if there is any possibility of breaking of what we currently have.

leszko (approved):
Nice improvement @mjh1 

Added one comment, other than that, it looks fine.

packages/api/src/controllers/helpers.ts
mjh1: That came from Jaron, so hopefully we're good:
> If there's a meta member, the stream is active

packages/api/src/controllers/helpers.ts
```diff
@@ -699,6 +699,33 @@ export const triggerCatalystPullStart =
         throw new Error(`failed to trigger catalyst pull`);
       };
 
+export const getCatalystStreamMetadata =
```
victorges: super-nit: wdyt of:
- Renaming this to `waitCatalystStreamReady` since this function doesn't just make a single request
- Moving the `metadataUrl` construction to this function instead, since it is the one that has the logic on how the metadata works. It could receive the `stream` object instead of the `url` and simplify usage

victorges (approved):
LGTM with some optional nit

packages/api/src/controllers/helpers.ts
mjh1: Oh yep that sounds good

packages/api/src/controllers/helpers.ts
mjh1: Jaron confirmed that `meta` is only missing in these cases:
> Yeah. It's only missing while the stream is offline or in the middle of booting/shutting down

So we just need to double check we don't see a dip in the success rate of the API after rolling out.

PR #2313, opened 2024-09-18
Add surname to registration form

thomshutt:
At the request of Marketing, this PR adds an explicit surname field to the registration form.

### DIFF
--- a/packages/www/components/Register/index.tsx
+++ b/packages/www/components/Register/index.tsx
@@ -12,6 +12,7 @@ const Register = ({ id, buttonText, onSubmit, loading, errors }) => {
   const router = useRouter();
   const [organization, setOrganization] = useState("");
   const [firstName, setFirstName] = useState("");
+  const [surname, setSurname] = useState("");
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
   const [utmCampaign, setUtmCampaign] = useState("");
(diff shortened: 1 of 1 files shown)


0xcadams:
I think we'll actually need to do something with this value here. Where do they want it to go? Hubspot?

thomshutt:
@0xcadams Yep, Hubspot - from the last time I used this page, I thought all values from the form got passed through to Hubspot by default

packages/www/components/Register/index.tsx
```diff
@@ -123,6 +124,20 @@ const Register = ({ id, buttonText, onSubmit, loading, errors }) => {
             onChange={(e) => setFirstName(e.target.value)}
           />
 
+          <TextField
```
victorges, line 128: Does this already go magically on the request sent to backend? I think it's called `lastName` on the API https://github.com/livepeer/livepeer-com/blob/4e24f8e656e204089b0d0057428820045c162c89/packages/api/src/controllers/user.ts#L314

victorges (approved):

packages/www/components/Register/index.tsx
thomshutt: @victorges good call, you were right - I updated it (it's `lastName` in Hubspot too) and tested locally the the last name makes it into the DB _and_ Hubspot

PR #2314, opened 2024-09-18
Add a file extension to download url

mjh1:
This relies on a CDN rule to re-route the request to the standard `/video` path. I'll probably put it behind an experiment to begin with and also probably should have the mapping of format to file extension in catalyst.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/asset.ts
+++ b/packages/api/src/controllers/asset.ts
@@ -418,7 +418,12 @@ function getDownloadUrl(
     os.id !== vodObjectStoreId ? os.publicUrl : pathJoin(ingest, "asset");
   const source = asset.files?.find((f) => f.type === "source_file");
   if (source) {
-    return pathJoin(base, asset.playbackId, source.path);
+    return pathJoin(
+      base,
+      asset.playbackId,
+      source.path,
+      "download." + asset.videoSpec.format,
+    );
   }
   return pathJoin(base, asset.playbackId, "video");
 }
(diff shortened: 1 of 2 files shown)


packages/api/src/controllers/asset.ts
```diff
@@ -418,7 +418,23 @@ function getDownloadUrl(
...
+      source.path != "video" ||
+      !formatToFileExtension.has(asset.videoSpec.format)
+    ) {
+      return pathJoin(base, asset.playbackId, source.path);
+    }
+    return pathJoin(
+      base,
+      asset.playbackId,
+      source.path,
+      "download." + formatToFileExtension.get(asset.videoSpec.format),
```
victorges, line 439: wdyt of naming the file with the asset name instead?
```suggestion
      encodeURIComponent(`${asset.name}.${formatToFileExtension.get(asset.videoSpec.format)}`),
```

victorges (approved):

packages/api/src/controllers/asset.ts
mjh1: Only issue with that is that the name for direct uploads will contain the original file extension and so they'll end up with double file extensions :( I like the idea though, it is annoying when you end up downloading files all with the same name.
Screenshot 2024-09-19 at 12 21 13

mjh1: could just strip anything after the final `.`? Or just use everything before the first `.`

mjh1:
@thomshutt @victorges I'm also tempted to put this behind an experiment for now... although the old URLs will still work I feel like we probably need to announce somewhere we're making this change to the API

thomshutt:
@mjh1 Let's not do an experiment, but you can work with @Shih-Yu on messaging and giving enough notice

PR #2317, opened 2024-09-27
Fix segment anything default model ID

mjh1:
Model ID doesn't need trailing `:`, this was causing 503s when trying to use the segment anything api.


thomshutt (approved):

victorges:
oops. i blame claude

PR #2318, opened 2024-10-01
ai: generate: remove endpoint from experiment & remove beta from path

gioelecerati:

- Remove `generate` from experiment
- Remove `beta` from path
- Make sure that the endpoint still works with the deprecated `beta`, to be removed on a second time for a full phase out
- Modified the tests to test both the new and the `beta` paths

**Screenshots (optional)**

**Checklist**

- [ X] I have read the **CONTRIBUTING** document.
- [ X] My change requires a change to the documentation.
- [ X] I have added tests to cover my changes.

### DIFF
--- a/packages/api/src/controllers/generate.test.ts
+++ b/packages/api/src/controllers/generate.test.ts
@@ -71,6 +71,16 @@ afterEach(async () => {
   await clearDatabase(server);
 });
 
+const testBothRoutes = (testFn) => {
+  describe("generate route", () => {
+    testFn("/generate");
+  });
+
+  describe("beta generate route", () => {
+    testFn("/beta/generate");
+  });
+};
+
 describe("controllers/generate", () => {
   let client: TestClient;
   let adminUser: User;
@@ -145,9 +155,9 @@ describe("controllers/generate", () => {
     return form;
   };
 
-  describe("API proxies", () => {
-    it("should call the AI Gateway for generate API /audio-to-text", async () => {
-      const res = await client.fetch("/beta/generate/audio-to-text", {
+  testBothRoutes((basePath) => {
+    it(`should call the AI Gateway for ${basePath}/audio-to-text`, async () => {
+      const res = await client.fetch(`${basePath}/audio-to-text`, {
         method: "POST",
         body: buildMultipartBody(
           {},
@@ -162,8 +172,8 @@ describe("controllers/generate", () => {
       expect(aiGatewayCalls).toEqual({ "audio-to-text": 1 });
     });
 
-    it("should call the AI Gateway for generate API /text-to-image", async () => {
-      const res = await client.post("/beta/generate/text-to-image", {
+    it(`should call the AI Gateway for ${basePath}/text-to-image`, async () => {
+      const res = await client.post(`${basePath}/text-to-image`, {
         prompt: "a man in a suit and tie",
       });
       expect(res.status).toBe(200);
@@ -174,8 +184,8 @@ describe("controllers/generate", () => {
       expect(aiGatewayCalls).toEqual({ "text-to-image": 1 });
     });
 
-    it("should call the AI Gateway for generate API /image-to-image", async () => {
-      const res = await client.fetch("/beta/generate/image-to-image", {
+    it(`should call the AI Gateway for ${basePath}/image-to-image`, async () => {
+      const res = await client.fetch(`${basePath}/image-to-image`, {
         method: "POST",
         body: buildMultipartBody({
           prompt: "replace the suit with a bathing suit",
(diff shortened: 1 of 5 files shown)


packages/api/src/controllers/generate.test.ts
```diff
@@ -71,6 +71,16 @@ afterEach(async () => {
...
 
+const testBothRoutes = (testFn) => {
+  describe("generate route", () => {
+    testFn("/generate");
+  });
+
+  describe("beta generate route", () => {
+    testFn("/beta/generate");
+  });
+};
```
victorges, line 82: Keep in mind we need to update CloudFlare routes to also skip the regular worker on the new route.

packages/api/src/controllers/generate.test.ts
```diff
@@ -260,7 +270,7 @@ describe("controllers/generate", () => {
 
     for (const [title, input, error] of testCases) {
       it(title, async () => {
-        const res = await client.fetch("/beta/generate/image-to-image", {
+        const res = await client.fetch("/generate/image-to-image", {
```
victorges, line 273: Why are these (everything below) not using the base path?

packages/api/src/controllers/generate.ts
```diff
@@ -35,7 +35,13 @@ const aiGenerateDurationMetric = new promclient.Histogram({
...
 const app = Router();
 
-app.use(experimentSubjectsOnly("ai-generate"));
+// TODO: Remove beta paths middleware
+app.use((req, res, next) => {
+  if (req.path.startsWith("/beta/generate")) {
+    req.url = req.url.replace("/beta/generate", "/generate");
+  }
+  next();
+});
```
victorges: Not sure why do we need this. Aren't the route already registered on both paths?

packages/api/src/controllers/generate.ts
```diff
@@ -35,7 +35,13 @@ const aiGenerateDurationMetric = new promclient.Histogram({
 
 const app = Router();
 
-app.use(experimentSubjectsOnly("ai-generate"));
```
victorges, line 38: Yeah this is fine to remove .

packages/api/src/schema/pull-ai-schema.js
```diff
@@ -60,10 +60,20 @@ const downloadAiSchema = async () => {
...
 
+  // TODO: Remove beta paths
+  const newPaths = {};
+  Object.entries(schema.paths).forEach(([path, value]) => {
+    const generatePath = `/api/generate${path}`;
+    const betaGeneratePath = `/api/beta/generate${path}`;
+    newPaths[generatePath] = value;
+    newPaths[betaGeneratePath] = value;
+  });
+  schema.paths = newPaths;
```
victorges: I don't think we should have both in the API schema. It should be just the new one. Meaning we only change L76 (prev L66) in this file.

Also, it seems like you updated this script but didn't run it? The current changes on `ai-api-schema` do not seem to correspond to the changes you made in this file. Meaning, in the current code, the `ai-api-schema` should have both the `/beta/` and non`-/beta/` APIs. But instead, it looks like you manually updated the APIs to strip the `/beta/`?

FTR that file should NEVER be updated manually. More info in the schemas README.

gioelecerati: Oh, right, I updated it manually when it should have been via script, removing the old path here and fixing the schema

packages/api/src/controllers/generate.test.ts
gioelecerati: I wanted to just test the new path working, considering we will remove the `beta` endpoint at some point, but yeah I can add it to all of them

packages/api/src/controllers/generate.ts
gioelecerati: Right, this is not needed

packages/api/src/controllers/generate.test.ts
victorges: Yeah if we have some of the tests testing both, let's be consistent. I'm fine with just testing the new one as well and leaving the old untested. But we can do that when we remove the `/beta`/ ones

packages/api/src/schema/ai-api-schema.yaml
```diff
@@ -368,6 +368,65 @@ paths:
               schema:
                 $ref: '#/components/schemas/studio-api-error'
       x-speakeasy-name-override: segmentAnything2
+  /api/generate/llm:
```
victorges, line 371: Keep in mind you might have conflict with @mjh1's #2319, but it should be solved by simply running the `pull-ai-schema` script again

victorges (approved):
LGTM

PR #2319, opened 2024-10-02
Add AI LLM endpoint

mjh1:

Adding a new AI endpoint for LLM.

Testing in staging and unit testing.

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/generate.test.ts
+++ b/packages/api/src/controllers/generate.test.ts
@@ -227,6 +227,19 @@ describe("controllers/generate", () => {
       });
       expect(aiGatewayCalls).toEqual({ "segment-anything-2": 1 });
     });
+
+    it("should call the AI Gateway for generate API /llm", async () => {
+      const res = await client.fetch("/beta/generate/llm", {
+        method: "POST",
+        body: buildMultipartBody({}),
+      });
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/generate.test.ts
```diff
@@ -237,6 +243,19 @@ describe("controllers/generate", () => {
       });
       expect(aiGatewayCalls).toEqual({ "segment-anything-2": 1 });
     });
+
+    it("should call the AI Gateway for generate API /llm", async () => {
+      const res = await client.fetch("/beta/generate/llm", {
+        method: "POST",
+        body: buildForm({ prompt: "foo" }),
```
victorges, line 250: lol so the LLM API is a multipart form even tho there's no files in the payload? or is it multi-modal and _could_ receive files? if it could, I'd say we test with a dummy file here as well just in case

victorges (approved):
LGTM

packages/api/src/controllers/generate.test.ts
mjh1: yeah.... it did strike me as odd too but wasn't sure. there aren't any file fields in the spec 😬 Maybe we're futureproofing..?

PR #2323, opened 2024-10-17
api: stream: allow name update

gioelecerati:

- Allow `name` as field in `PATCH` stream api

**Screenshots (optional)**

**Checklist**

### DIFF
--- a/packages/api/src/controllers/stream.test.ts
+++ b/packages/api/src/controllers/stream.test.ts
@@ -1240,6 +1240,13 @@ describe("controllers/stream", () => {
         });
       });
 
+      it("should allow patch of name", async () => {
+        const res = await client.patch(patchPath, {
+          name: "new name",
+        });
+        expect(res.status).toBe(204);
+      });
+
       it("should allow patch of playbackPolicy", async () => {
         const res = await client.patch(patchPath, {
           playbackPolicy: {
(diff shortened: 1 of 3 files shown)


gioelecerati:
@victorges do you know why the stream name was immutable till now?
There was also a specific test making sure it was immutable (`the stream name is immutable`) - is a name update going to affect anything?

victorges:
@gioelecerati oh wow super weird that there's even a test. I have no idea 😅 can't think of anything that would break but now it's better to test it anyway. Maybe on the very beginning it was used for configuring streams on mist (or even wowza lol)

gioelecerati:
> @gioelecerati oh wow super weird that there's even a test. I have no idea 😅 can't think of anything that would break but now it's better to test it anyway. Maybe on the very beginning it was used for configuring streams on mist (or even wowza lol)

Coold :D was added here https://github.com/livepeer/studio/pull/430 and I think it was just added without any specific reason, but let's test throughly either way!

packages/api/src/controllers/stream.test.ts
```diff
@@ -1240,6 +1240,13 @@ describe("controllers/stream", () => {
         });
       });
 
+      it("should allow patch of name", async () => {
+        const res = await client.patch(patchPath, {
+          name: "new name",
+        });
+        expect(res.status).toBe(204);
```
victorges, line 1247: You should check if the database object is actually updated

victorges (approved):
neat

victorges:
@gioelecerati lol got it. that PR has nothing to do with stream name then, probably just a confusing string i used to test the `additionalProperties` validation.

PR #2325, opened 2024-11-07
Put hack in place for StreamETH to do 20 concurrent clips instead of 5

thomshutt:
We don't have time / inclination to do the work to make this more generic

### DIFF
--- a/packages/api/src/controllers/clip.ts
+++ b/packages/api/src/controllers/clip.ts
@@ -121,7 +121,13 @@ app.post(
       5,
     );
 
-    if (processingClips.length >= MAX_PROCESSING_CLIPS && !clippingUser.admin) {
+    // Hack to let StreamETH have a higher clipping limit without having to do the work to put this on the User object
+    let maxProcessingClips = MAX_PROCESSING_CLIPS;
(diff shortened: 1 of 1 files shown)


packages/api/src/controllers/clip.ts
```diff
@@ -121,7 +121,13 @@ app.post(
       5,
     );
 
-    if (processingClips.length >= MAX_PROCESSING_CLIPS && !clippingUser.admin) {
+    // Hack to let StreamETH have a higher clipping limit without having to do the work to put this on the User object
+    let maxProcessingClips = MAX_PROCESSING_CLIPS;
+    if (req.user.id == "05f04681-b600-4a9b-81ea-1f56e7871f04") {
```
gioelecerati, line 126: WDYT of checking with `owner.id` instead, so it's applied to the owner of the content itself and not the user owning the api key doing the request? I am fine with it either way as only admins are allowed to do cross-clips, for now

gioelecerati (approved):

packages/api/src/controllers/clip.ts
thomshutt: Yeah let's leave it for now then - we can hopefully roll this back after Devcon anyway

PR #2332, opened 2025-04-07
rafal/fix-slow-multistream-delete

leszko:
We have a problem with PATCH/DELETE multistream target. The issue is that when user has a lot of streams, then this becomes a very DB-heavy request.

The proposed workaround is to skip notifying Catalyst about the multistream update in case of not passing by the user a helper `streamId` or `playbackId` as query param. So it will work right now this way:
1. If user does not change anything in their API usage, then the Catalyst will not be notified about the multistream delete or patch => Then their multistream will be refreshed only after 30s (instead of instant refresh)
2. If user updates their API call to include `streamId` or `playbackId` in the query params, then it will work as before.
   - Now: `DELETE /api/multistream/target/fabb8085-5d71-4807-919e-abd7410f051b`
   - After: `DELETE /api/multistream/target/fabb8085-5d71-4807-919e-abd7410f051b?streamId=abcdef`

The root cause is our studio DB architecture and the design decision that 1 multistream can be related to multiple streams. This with comparison with storing everything as JSON and multiple target in a single stream creates the issue: we can only use GIN index, which happens to be too slow with the number of streams users have.

### DIFF
--- a/packages/api/src/controllers/multistream.ts
+++ b/packages/api/src/controllers/multistream.ts
@@ -14,6 +14,7 @@ import {
 } from "./helpers";
 import { db } from "../store";
 import { FindOptions, FindQuery, WithID } from "../store/types";
+import logger from "../logger";
 import {
   MultistreamTarget,
   MultistreamTargetPatchPayload,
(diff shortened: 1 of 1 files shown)


hjpotter92 (approved):

thomshutt (approved):

leszko:
Closing this one, since @victorges found a better solution here: https://github.com/livepeer/studio/pull/2333

PR #2335, opened 2025-06-10
Introduce retries for queue message sending

mjh1:
We have been seeing timeouts with pushing messages to queues so add some retries


j0sh (approved):
the next step would be a local queue to push to a remote queue

it's ~~turtles~~ queues all the way down

victorges (approved):
LGTM

PR #2348, opened 2026-02-24
Add reencode flag

mjh1:
Give the option of reencoding segments in transcode API

### DIFF
--- a/packages/api/src/schema/db-schema.yaml
+++ b/packages/api/src/schema/db-schema.yaml
@@ -1185,6 +1185,9 @@ components:
                     - background_mist
                     - fallback_external
                     - external
+                reencode:
+                  type: boolean
+                  default: false
         deleted:
           type: boolean
           description: Set to true when the task is deleted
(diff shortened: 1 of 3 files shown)


packages/api/src/schema/db-schema.yaml
```diff
@@ -1185,6 +1185,9 @@ components:
                     - background_mist
                     - fallback_external
                     - external
+                reencode:
```
gioelecerati, line 1188: Any reason why this change is happening in the db schema and not the api schema?

gioelecerati (approved):
LGTM - just a question posted

packages/api/src/schema/db-schema.yaml
mjh1: Just because I wanted to keep it as a hidden flag like we have with `catalystPipelineStrategy`, users shouldn't normally be using it but I want to try it out to see if it fixes a customer issue.

